PR #8, opened 2016-10-26
Mqtt getting started guide

2m:


mqtt/src/test/scala/akka/stream/contrib/mqtt/MqttSourceSpec.scala
```diff
@@ -54,6 +53,46 @@ class MqttSourceSpec extends WordSpec with Matchers with ScalaFutures {
...
+      import f._
+
+      //#create-settings
+      val settings = MqttSourceSettings(
+        MqttConnectionSettings(
+          "tcp://localhost:1883",
+          "test-client",
+          new MemoryPersistence
+        ),
+        Map("topic1" -> 0, "topic2" -> 0)
```
johanandren: what are these numbers?

johanandren: "These zeroes needs to be here, trust me"

mqtt/src/main/scala/Mqtt.scala
```diff
@@ -11,6 +11,19 @@ import org.eclipse.paho.client.mqttv3.{ MqttMessage => PahoMqttMessage, _ }
...
 
+/**
+ * Quality of Service constants as defined in
+ * https://www.eclipse.org/paho/files/mqttdoc/Cclient/qos.html
+ */
+object QoS {
+  def atMostOnce = 0
+  def atLeastOnce = 1
+  def exactlyOnce = 2
+}
```
johanandren: Better. But I would have liked a typed encoding even better. As in sealed trait + three objects (with a byte value method each for the internal representation).

2m:
Added typed `MqttQoS`. Its a sealed abstract *class* because only then one can access a method defined in `MqttQoS` objecy from Java like: `MqttQoS.atMostOnce();`.

mqtt/src/test/scala/akka/stream/contrib/mqtt/MqttSourceSpec.scala
```diff
@@ -54,7 +53,47 @@ class MqttSourceSpec extends WordSpec with Matchers with ScalaFutures {
...
+      import f._
+
+      //#create-settings
+      val settings = MqttSourceSettings(
+        MqttConnectionSettings(
+          "tcp://localhost:1883",
+          "test-client",
+          new MemoryPersistence
+        ),
+        Map("topic1" -> MqttQoS.AtMostOnce, "topic2" -> MqttQoS.AtMostOnce)
```
johanandren: 100 times much nicer, thanks!

PR #13, opened 2016-11-01
Add some more to the documentation

patriknw:


docs/src/main/paradox/external-connectors.md
```diff
@@ -0,0 +1,28 @@
+# External Connectors
+
+Connectors hat are maintained by other projects.
```
johanandren: hat -> that

docs/src/main/paradox/index.md
```diff
@@ -1,13 +1,16 @@
 # Alpakka
 
-Welcome to the home of the Alpakka innitiative, which harbours various Akka Streams connectors. Here you can find documentation of the connectors that are part of this project as well as links to connectors that are maintained by other projects.
+Welcome to the home of the Alpakka innitiative, which harbours various Akka Streams components for integration use cases. Here you can find documentation of the components that are part of this project as well as links to components that are maintained by other projects.
```
johanandren: We shouldn't use "connectors" because it will also contain transformation stages?

patriknw: I see Alpakka as containing connectors, integration patterns, and data transformations.
That's why I tried to find a more general word here

harksin: Why not an exhaustive list of component type? it's bits longer, but clearer

patriknw: @harksin You mean something like:

Welcome to the home of the Alpakka innitiative, which harbours various Akka Streams connectors, integration patters, and data transformations for integration use cases.

PR #18, opened 2016-11-03
journal-writer

dnvriend:
A journal-writer, a flow and sink component for Alpakka that can be handy for the following use cases:

- migrating events from one journal to another
- transforming event serialization strategy from one journal to another
- bulk loading events from a big flat file (eg. CSV or JSON) into a journal.

### DIFF
--- a/journal-writer/src/main/scala/akka/persistence/journal/writer/WriteJournalAdapterCameo.scala
+++ b/journal-writer/src/main/scala/akka/persistence/journal/writer/WriteJournalAdapterCameo.scala
@@ -1,0 +1,39 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.persistence.journal.writer
+
+import akka.actor.{ Actor, ActorRef }
+import akka.persistence.JournalProtocol._
+import akka.persistence.query.EventEnvelope
+import akka.stream.alpakka.persistence.writer._
+
+import scala.collection.immutable.Seq
+
+object WriteJournalAdapterCameo {
+  def writeMessages(xs: Seq[EventEnvelope], writerJournalAdapter: ActorRef): WriteMessages =
+    WriteMessages(toAtomicWrite(xs), writerJournalAdapter, 1)
+}
+
+class WriteJournalAdapterCameo(writePlugin: ActorRef, replyTo: ActorRef, messages: Seq[EventEnvelope]) extends Actor {
+  import WriteJournalAdapterCameo._
+  override def preStart(): Unit = {
+    if (messages.isEmpty)
+      replyWithSuccess(replyTo)
+    else
+      writePlugin ! writeMessages(messages, self)
+  }
+
+  override def receive: Receive = {
+    case msg: WriteMessageSuccess =>
+      replyWithSuccess(replyTo)
+    case WriteMessagesSuccessful =>
+      replyWithSuccess(replyTo)
+    case WriteMessagesFailed(cause) =>
+      replyWithFailure(replyTo, cause)
+    case WriteMessageRejected(_, cause, _) =>
+      replyWithFailure(replyTo, cause)
+    case WriteMessageFailure(_, cause, _) =>
+      replyWithFailure(replyTo, cause)
+  }
+}
--- a/journal-writer/src/main/scala/akka/stream/alpakka/persistence/writer/JournalWriter.scala
+++ b/journal-writer/src/main/scala/akka/stream/alpakka/persistence/writer/JournalWriter.scala
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.alpakka.persistence.writer
+
+import akka.actor.{ Actor, ActorRef, ActorSystem, Props }
+import akka.pattern.ask
+import akka.persistence.Persistence
+import akka.persistence.journal.writer.WriteJournalAdapterCameo
+import akka.persistence.query.{ EventEnvelope, EventEnvelope2 }
+import akka.stream.scaladsl._
+import akka.util.Timeout
+import akka.{ Done, NotUsed }
+
+import scala.collection.immutable._
+import scala.concurrent.duration._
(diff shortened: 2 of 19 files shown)


johanandren:
Fails on `com.github.dnvriend#akka-persistence-inmemory_2.12.0-RC2;1.3.13: not found`, not published to maven central?

ktoso:
Should target 2.12 stable anyway :) (changed)

journal-writer/src/main/scala/akka/persistence/journal/writer/WriteJournalAdapterCameo.scala
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.persistence.journal.writer
```
patriknw: this must be in the `akka.persistence` package because it is using internal API?

journal-writer/src/main/scala/akka/stream/alpakka/persistence/writer/package.scala
```diff
@@ -0,0 +1,55 @@
...
+
+package object writer {
+  val unsupported: Throwable = new IllegalArgumentException("JournalWriter only accepts EventEnvelope, immutable.Seq[EventEnvelope], EventEnvelope2 and immutable.Seq[EventEnvelope2]")
+
+  def replyWithFailure(replyTo: ActorRef, cause: Throwable = unsupported)(implicit ctx: ActorContext, self: ActorRef): Unit = {
+    replyTo ! akka.actor.Status.Failure(cause)
+    ctx.stop(self)
+  }
+
+  def replyWithSuccess(replyTo: ActorRef, msg: String = "ack")(implicit ctx: ActorContext, self: ActorRef): Unit = {
```
patriknw: I think this is too much hidden magic, and it's only used in one Actor

journal-writer/src/main/scala/akka/stream/alpakka/persistence/writer/JournalWriter.scala
```diff
@@ -0,0 +1,50 @@
...
+import akka.util.Timeout
+import akka.{ Done, NotUsed }
+
+import scala.collection.immutable._
+import scala.concurrent.duration._
+import scala.concurrent.{ ExecutionContext, Future }
+import scala.reflect.ClassTag
+
+class WriteJournalAdapter(writePlugin: ActorRef) extends Actor {
+  def cameo(replyTo: ActorRef, writePlugin: ActorRef, messages: Seq[EventEnvelope]): ActorRef =
```
patriknw: I'm not sure I see the value of the `WriteJournalAdapterCameo` actor

journal-writer/src/main/scala/akka/persistence/journal/writer/WriteJournalAdapterCameo.scala
dnvriend: Only because akka.persistence.JournalProtocol is private ie. 'internal API'. It uses Jamie Allen's Cameo pattern which is handy for multiple things, as it turns out, also a bridge between packages.

journal-writer/src/main/scala/akka/stream/alpakka/persistence/writer/JournalWriter.scala
dnvriend: Could you please elaborate. What cost do you see using the cameo pattern, and if so, what alternatives do you see?

journal-writer/src/main/scala/akka/stream/alpakka/persistence/writer/package.scala
dnvriend: package objects are visible to all classes in the package, there is nothing hidden about that. When using a structure where each package contains a single component like the JournalWriter, all functionality in the package object belongs to one component with one goal, writing to the journal. 

Given the component-per-package structure, the value I see in using package objects for methods and functions is that it makes classes more readable, focussing on the core functionality and not distracting you with details. When you want to see the details, look at the contents of the method. This lets you focus on the essentials. 

Also the code uses an architectural evident style in the naming conventions where each method speaks for himself. Think of it as zooming out on Google Maps, looking at cities (classes) and zooming in seeing roads and houses (clicking on methods).

patriknw: I would have found it easier to read if it was a plain private method in the actor. Passing around ActorContext is also a bad smell, since it's not thread safe.

journal-writer/src/main/scala/akka/stream/alpakka/persistence/writer/JournalWriter.scala
patriknw: As far as I can see it's stateless and doesn't do anything that can't be done in the adapter actor, but perhaps the mapping of the replies is the motivation for it.

patriknw:
@dnvriend Do you have any plans of completing this, or shall we close this PR for now?

journal-writer/src/main/scala/akka/stream/alpakka/persistence/writer/package.scala
drewhk: I agree with @patriknw here. I understand each person have a different style that he likes but I would like to see these refactored simply because it is of a different style than is used elsewhere in this project. We are much more lenient in Alpakka than in Akka because it is more community driven but I think this is just too different from other subprojects here.

journal-writer/src/test/scala/akka/stream/alpakka/persistence/writer/TestSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+      tp.expectTerminated(actor)
+    }
+  }
+
+  implicit class SourceOps[A](that: Source[A, _])(implicit system: ActorSystem) {
+    def withTestProbe(f: TestSubscriber.Probe[A] => Unit): Unit =
+      f(that.runWith(TestSink.probe(system)))
+  }
+
+  implicit class PimpedFuture[T](self: Future[T]) {
```
drewhk: please rename this to ExtendedFuture or something else.

Issue #15, opened 2016-11-02
scaladsl vs javadsl


patriknw:
We should use the convention with `scaladsl` and `javadsl` packages to make it consistent with other things (e.g. Akka Streams). We should do this before the 0.1 release so that we have good examples.

PR #19, opened 2016-11-03
Introduced javadsl and scaladsl packages for language specific parts

johanandren:
Refs #15


cassandra/src/test/scala/akka/stream/alpakka/cassandra/scaladsl/CassandraSourceSpec.scala
```diff
@@ -106,7 +107,7 @@ class CassandraSourceSpec extends WordSpec with ScalaFutures with BeforeAndAfter
     "stream the result of Cassandra statement that results in no data" in {
       val stmt = new SimpleStatement("SELECT * FROM akka_stream_scala_test.test")
 
-      val rows = CassandraSource(stmt)
+      val rows = CassandraSourceStage(stmt)
```
2m: This and above renames are probably collateral damage. Should stay `CassandraSource`. This is what fails PR validation.

cassandra/src/test/scala/akka/stream/alpakka/cassandra/scaladsl/README.txt
```diff
@@ -0,0 +1,7 @@
+All the tests must be run with a local Cassandra running on default port 9042.
+
+When using docker, a temporary server can by started by running:
+
+```
+docker run --rm -p 9042:9042 cassandra:3
+```
```
2m: We have this information in the paradox docs and in scaladoc of the spec classes. I think we can remove these `txt` files.

johanandren:
I think we can keep them common and together with the shared model as long as they do not require huge differences.

cassandra/src/test/scala/akka/stream/alpakka/cassandra/scaladsl/README.txt
2m: I agree. One might not be coming from the docs. What is more, this information is only relevant when someone tries to run the tests. In a light of this realization I think it would be best to have this information printed out to the console in `beforeAll` methods of the spec (and java test).

This might be an unorthodox suggestion, be we should not be reserved when it comes to communication through sbt console. :) P.S. [Scala has it](https://github.com/scala/scala/blob/2.12.x/build.sbt#L859).

johanandren: What about just in a scaladoc comment up top in those tests, that should cover it?

johanandren: Oh, it already has, alrightey then removing the readme

2m:
Updated LGTM

johanandren:
I'm going to merge this, want to get "final" paths to the classes for updating the blog entries.

PR #22, opened 2016-11-05
A DigestCalculator for the Alpakka Project

dnvriend:


digest/src/main/scala/akka/stream/alpakka/digest/DigestCalculator.scala
```diff
@@ -0,0 +1,59 @@
...
+import scala.concurrent.Future
+import scala.util.Success
+
+/**
+ * The DigestCalculator transforms/digests a stream of akka.util.ByteString to a
+ * DigestResult according to a given Algorithm
+ */
+object DigestCalculator {
+
+  def flow(algorithm: Algorithm): Flow[ByteString, DigestResult, NotUsed] =
```
patriknw: we decided to use scaladsl/javadsl packages, as in https://github.com/akka/alpakka/pull/19

digest/src/main/scala/akka/stream/alpakka/digest/DigestCalculator.scala
```diff
@@ -0,0 +1,59 @@
...
+    setHandler(in, new InHandler {
+      override def onPush(): Unit = {
+        val chunk = grab(in)
+        digest.update(chunk.toArray)
+        pull(in)
+      }
+
+      override def onUpstreamFinish(): Unit = {
+        val messageDigest: Array[Byte] = digest.digest()
+        emit(out, DigestResult(ByteString(messageDigest), digestToHexString(messageDigest), Success(Done)))
```
patriknw: when will it be Failure?

dnvriend: Good question! I believe it cannot happen at all. All other exceptions will cause the stream to fail. The interface I copied of `akka.stream.IOResult`.  Maybe it needs another interface?

digest/src/main/scala/akka/stream/alpakka/digest/DigestCalculator.scala
dnvriend: Java... alright, let me look into that!

digest/src/main/scala/akka/stream/alpakka/digest/DigestCalculator.scala
patriknw: ok, then you can remove the `status` from the `DigestResult`

digest/src/main/scala/akka/stream/alpakka/digest/DigestCalculator.scala
```diff
@@ -0,0 +1,59 @@
...
+
+private[digest] class DigestCalculator(algorithm: Algorithm) extends GraphStage[FlowShape[ByteString, DigestResult]] {
+  val in: Inlet[ByteString] = Inlet("DigestCalculator.in")
+  val out: Outlet[DigestResult] = Outlet("DigestCalculator.out")
+  override val shape: FlowShape[ByteString, DigestResult] = FlowShape.of(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+    val digest = java.security.MessageDigest.getInstance(algorithm.toString)
+
+    def digestToHexString(messageDigest: Array[Byte]): String =
```
patriknw: perhaps this should be a `def` in DigestResult instead, so you don't have to pay for the hex string unless you need it

drewhk:
LGTM

patriknw:
@dnvriend Do you have any plans of completing this, or shall we close this PR for now?

PR #23, opened 2016-11-07
Aggregate file sub-project

2m:


build.sbt
```diff
@@ -31,6 +24,14 @@ lazy val cassandra = project
     Dependencies.Cassandra
   )
 
+lazy val files = project
```
johanandren: isn't `files` vs `file` worse than the full name? (that's why I choose that over `files`)

2m: The full name really stood out from the others. For example the output of `projects` command in sbt was:
```
> projects
[info] In file:/home/martynas/projects/alpakka/
[info]   * alpakka
[info]     akka-stream-alpakka-file
[info]     amqp
[info]     cassandra
[info]     docs
[info]     mqtt
```

What if we rename it to `files` everywhere?

patriknw: the name in the build doesn't matter, suggestion: use `fileIO` in the build

Issue #25, opened 2016-11-09
ReactiveSocket connector


patriknw:
[ReactiveSocket](http://reactivesocket.io/) is interesting, especially for integration between microservices (implemented with different technology stacks).

[reactivesocket-java](https://github.com/ReactiveSocket/reactivesocket-java) API is based on Reactive Streams Publisher and Subscriber and could therefore be used directly with Akka Streams. 

However, it might be beneficial to provide a more idiomatic API on top of `reactivesocket-java`.

The `reactivesocket-java` API is: https://github.com/ReactiveSocket/reactivesocket-java/blob/0.5.x/reactivesocket-core/src/main/java/io/reactivesocket/ReactiveSocket.java
(note that `requestSubscription` is being removed)

The same interface is used on both client and server side.

A Scala API wrapper might look like:

```scala
def fireAndForget(payload: Payload): Future[Done]
def requestResponse(payload: Payload): Future[Payload]
def requestStream(payload: Payload): Source[Payload]
def requestChannel(payloads: Source[Payload]): Source[Payload]
```

patriknw:
So far, I'm not convinced that the value of such a wrapper API would add much value compared to using the `reactivesocket-java` API directly with Akka Streams Publisher/Subscriber conversions. I'll dig deeper.

johanandren:
After looking at it a first time I think that on the server side we could provide a nice `SocketAcceptor` integration, something closer to how the Tcp streams api looks server side perhaps, client side maybe not so much need.

maciekciolek:
It seems that first attempts of implementing this were stopped. I am interested to provide this kind of alpakka implementation - do you have any new thoughts how should it look like or would be great to make some initial proposal by me?

patriknw:
@maciekciolek That would be awesome. It would be great if you could start by sketching out the API and we can review that as the first step.
Take a look at the prototype at https://github.com/akka/alpakka/pull/26/files

maciekciolek:
@patriknw I went deeply into #26, #69 prototypes and the implementation of ReactiveSocket. I can see two ways how Alpakka can support ReactiveSocket:
1. We can implement just Akka Streams wrapper (abstract class with method overrides) which will allow developers to use ReactiveSocket with Akka Streams easily but it will be just raw wrapper.
2. We can do it like akka-http - I can imagine approach as follow:
```scala
object ReactiveSockets {

  def bindAndHandle(
[...]

PR #26, opened 2016-11-09
ReactiveSocket build, and HelloWorld

patriknw:


reactivesocket/src/main/scala/akka/stream/alpakka/reactivesocket/scaladsl/HelloWorld.scala
```diff
@@ -0,0 +1,73 @@
...
+import io.reactivesocket.client.SetupProvider._
+import akka.actor.ActorSystem
+import akka.stream.ActorMaterializer
+import akka.stream.scaladsl.Source
+import akka.stream.scaladsl.Sink
+import io.reactivesocket.util.PayloadImpl
+import scala.concurrent.Future
+import io.reactivesocket.frame.ByteBufferUtil
+
+object HelloWorld {
```
patriknw: This corresponds to https://github.com/ReactiveSocket/reactivesocket-java/blob/0.5.x/reactivesocket-examples/src/main/java/io/reactivesocket/examples/transport/tcp/requestresponse/HelloWorldClient.java

patriknw:
@2m @johanandren just pushed a naive example for discussion

reactivesocket/src/main/scala/akka/stream/alpakka/reactivesocket/scaladsl/HelloWorld.scala
```diff
@@ -0,0 +1,73 @@
...
+    println(s"started server: $serverAddress") // FIXME
+
+    val clientPublisher = ReactiveSocketClient.create(TcpTransportClient.create(serverAddress),
+      keepAlive(never()).disableLease()).connect()
+
+    Source.fromPublisher(clientPublisher).runWith(Sink.head).foreach { socket =>
+      println(s"# got socket") // FIXME
+      val response: Future[Payload] =
+        Source.fromPublisher(socket.requestResponse(new PayloadImpl("Hello"))).runWith(Sink.head)
+
```
patriknw: we could hide a lot of this Publisher-to-Future juggling

reactivesocket/src/main/scala/akka/stream/alpakka/reactivesocket/scaladsl/HelloWorld.scala
```diff
@@ -0,0 +1,73 @@
...
+      response.onFailure {
+        case e => println(e.getMessage)
+      }
+
+      response.map(_.getData)
+        .map(x => { println("got response"); x }) // FIXME remove
```
patriknw: strangely, the sample doesn't work, never got this, but the server got the request

reactivesocket/src/main/scala/akka/stream/alpakka/reactivesocket/scaladsl/HelloWorld.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  def socketAcceptor: SocketAcceptor = new SocketAcceptor {
+    override def accept(setup: ConnectionSetupPayload, sendingSocket: ReactiveSocket): LeaseEnforcingSocket =
+      new DisabledLeaseAcceptingSocket(new AbstractReactiveSocket {
+        override def requestResponse(p: Payload): Publisher[Payload] = {
+          println(s"# got request") // FIXME
+          Source.single(p).runWith(Sink.asPublisher(fanout = false))
+        }
+      })
+  }
```
johanandren: So if we provide an api it should be something like `(ConnectionSetupPayload, ReactiveSocket) => Payload => Payload` here, and possibly also `Payload => Payload` and `Payload => Future[Payload]`?

patriknw:
Published a suggestion of a simplified client api: 03e8823

reactivesocket/src/main/scala/akka/stream/alpakka/reactivesocket/scaladsl/HelloWorld.scala
patriknw: I think the final return type on the server side should be `Future[Payload]`. I wonder why they didn't use `CompletionStage`? I should ask that.

viktorklang:
Because Publisher can be lazy?

-- 
Cheers,
√

patriknw:
ok, that might be a reason. Do you think that would be needed in for example:
```
def requestResponse(payload: Payload): Future[Payload]
```

patriknw:
Refs https://github.com/akka/alpakka/issues/25

patriknw:
I didn't find an obvious way to make the `SocketAcceptor` more scala idiomatic and still retain the full power of returning different lease implementations. Tried an adapter approach, with convenience factory for the disabled lease: 2fec76c

reactivesocket/src/main/scala/akka/stream/alpakka/reactivesocket/scaladsl/HelloWorld.scala
```diff
@@ -0,0 +1,163 @@
...
+        .map(ByteBufferUtil.toUtf8String)
+        .map(println)
+        .andThen { case _ => socket.close() }
+    }
+
+  }
+
+}
+
+trait ReactiveSocketScaladsl {
```
patriknw: Scaladsl should of course not be in the names later, just prototyping the api here

reactivesocket/src/main/scala/akka/stream/alpakka/reactivesocket/scaladsl/HelloWorld.scala
```diff
@@ -0,0 +1,163 @@
...
+
+  def fireAndForget(payload: Payload): Future[Done]
+
+  def requestResponse(payload: Payload): Future[Payload]
```
johanandren: The original API allows for reacting on the `ConnectionSetupPayload` which may contain mime type and metadata for example to choose how to deal with the connection, I'm thinking auth for example, shouldn't we make that available through the dsl? (So as I noted before, something like `ConnectionStuffs => Payload => Future[Payload]`)

reactivesocket/src/main/scala/akka/stream/alpakka/reactivesocket/scaladsl/HelloWorld.scala
johanandren: Same with my parallel implementation.

reactivesocket/src/main/scala/akka/stream/alpakka/reactivesocket/scaladsl/HelloWorld.scala
johanandren: No, I'm just too tired to even write comments. :/

reactivesocket/src/main/scala/akka/stream/alpakka/reactivesocket/scaladsl/HelloWorld.scala
```diff
@@ -0,0 +1,167 @@
...
+
+  def requestChannel(payloads: Source[Payload, NotUsed]): Source[Payload, NotUsed] = ???
+
+  def close(): Unit = ???
+
+}
+
+object ServerSocketAcceptorAdapter {
+
+  def disabledLease(handler: ConnectionSetupPayload => ReactiveSocketApi)(implicit materializer: Materializer): SocketAcceptor =
```
patriknw: @johanandren You are right that the `ConnectionSetupPayload` should be available, so I added it like this.
Note that one server can handle different interaction types, e.g. fireAndForget and requestResponse, so I don't see how we could do that with plain functions `ConnectionSetupPayload` => `Payload` => `Future[Payload]`.

That is why I defined `AbstractReactiveSocketApi` (similar to `AbstractReactiveSocket`) so that only methods of interest have to be implemented.

patriknw: I think it will be very common to use more than one interaction type in a microservice.
E.g. `sendTweet` (fire-n-forget), `getTweetDetails` (request-response), `getTweetStream` (request-stream).

If we dislike the `Abstract` pattern we could create a `Builder` dsl instead.

```scala
ReactiveSocketApi
  .withFireAndForget(...)
  .withRequestResponse(...)
```

2m:
Another Reactive Socket API sketch: https://github.com/akka/alpakka/pull/69

patriknw:
Refs #25

reactivesocket/src/main/scala/akka/stream/alpakka/reactivesocket/scaladsl/HelloWorld.scala
```diff
@@ -0,0 +1,167 @@
...
+
+trait ReactiveSocketApi {
+
+  def fireAndForget(payload: Payload): Future[Done]
+
+  def requestResponse(payload: Payload): Future[Payload]
+
+  def requestStream(payload: Payload): Source[Payload, NotUsed]
+
+  def requestChannel(payloads: Source[Payload, NotUsed]): Source[Payload, NotUsed]
```
patriknw: when we discussed this a suggestion was to use `Flow` instead of `Source`

patriknw:
Closing this for now. We intend to pick it up again later.

PR #29, opened 2016-11-10
Use common bintray package

2m:
So all of the subprojects can be released at once, and new subprojects do not need maintenance when being released the first time (like requesting new bintray package to be included into JCenter).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,8 @@ script:
 deploy:
   provider: script
   script:
-    - sbt -J-XX:ReservedCodeCacheSize=128m +publish bintraySyncMavenCentral
+    # do the maven central sync only once, as all of the artifacts are under the same bintray package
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -27,7 +27,8 @@ script:
 deploy:
   provider: script
   script:
-    - sbt -J-XX:ReservedCodeCacheSize=128m +publish bintraySyncMavenCentral
+    # do the maven central sync only once, as all of the artifacts are under the same bintray package
+    - sbt -J-XX:ReservedCodeCacheSize=128m +publish amqp/bintraySyncMavenCentral
```
patriknw: why amqp here?

johanandren: so this is only to publish amqp?

johanandren: just like it said in the comment, sorry I didn't read

patriknw: ah, reading comments is over rated ;-)

PR #31, opened 2016-11-11
Publish unidoc

2m:
So Lightbend Tech Hub can consume the scaladoc jar and server Alpakka Api docs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,4 +86,4 @@ Example:
 
 1. Create a [new release](https://github.com/akka/alpakka/releases/new) with the next tag version (e.g. `v0.3`), title and release decsription including notable changes mentioning external contributors.
 2. Travis CI will start a [CI build](https://travis-ci.org/akka/alpakka/builds) for the new tag and publish artifacts to Bintray and will sync them to Maven Central.
-3. Checkout the newly created tag and run `sbt "deployRsync repo.akka.io"` to deploy API and reference documentation.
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -86,4 +86,4 @@ Example:
 
 1. Create a [new release](https://github.com/akka/alpakka/releases/new) with the next tag version (e.g. `v0.3`), title and release decsription including notable changes mentioning external contributors.
 2. Travis CI will start a [CI build](https://travis-ci.org/akka/alpakka/builds) for the new tag and publish artifacts to Bintray and will sync them to Maven Central.
-3. Checkout the newly created tag and run `sbt "deployRsync repo.akka.io"` to deploy API and reference documentation.
+3. Update the configuration of [Lightbend Tech Hub](https://github.com/typesafehub/tech-hub/blob/master/docsites.sbt#L70) to include the latest version.
```
patriknw: Does this mean that we have to manually create a PR to techhub for each new release?

build.sbt
```diff
@@ -34,7 +29,7 @@ lazy val files = project
 
 lazy val mqtt = project
   .in(file("mqtt"))
-  .enablePlugins()
+  .enablePlugins(AutomateHeaderPlugin)
```
johanandren: Shouldn't this be enabled in the common default config rather?

2m: It would be great if we could do it. But autoplugins can not enable other autoplugins.

johanandren: Ah, didn't remember we did the defaults using an auto plugin, thanks for clarifying.

PR #45, opened 2016-11-12
Caffeine Flow initial commit.

cheleb:
Provide reconcilation flow backed by caffeine.

Documentation will follow :)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,7 @@
 lazy val alpakka = project
   .in(file("."))
   .enablePlugins(PublishUnidoc)
-  .aggregate(amqp, awslambda, cassandra, csv, dynamodb, files, ftp, googleCloudPubSub, hbase, jms, mqtt, s3, simpleCodecs, sqs, sse)
+  .aggregate(amqp, awslambda, caffeine, cassandra, csv, dynamodb, files, ftp, googleCloudPubSub, hbase, jms, mqtt, s3, simpleCodecs, sqs, sse)
 
 lazy val amqp = project
   .enablePlugins(AutomateHeaderPlugin)
@@ -15,6 +15,13 @@ lazy val awslambda = project
   .settings(
     name := "akka-stream-alpakka-awslambda",
     Dependencies.AwsLambda
+  )
+
+lazy val caffeine = project
+  .enablePlugins(AutomateHeaderPlugin)
+  .settings(
+    name := "akka-stream-alpakka-caffeine",
+    Dependencies.Caffeine
   )
 
 lazy val cassandra = project
--- a/docs/src/main/paradox/caffeine.md
+++ b/docs/src/main/paradox/caffeine.md
@@ -1,0 +1,75 @@
+# Caffeine Flow
+
+The [Caffeine](https://github.com/ben-manes/caffeine) flow provides Akka Stream flow, to reconcile data in memory.
+
+
+Let imagine 3 event types with a correlation id:
+ 
+scala
+:   @@snip (../../../../caffeine/src/test/scala/akka/stream/alpakka/caffeine/scaladsl/CaffeineFlowSpec.scala) { #events }
+
+And an aggregate class:
+
+scala
+:   @@snip (../../../../caffeine/src/test/scala/akka/stream/alpakka/caffeine/scaladsl/CaffeineFlowSpec.scala) { #aggregated-event }
+
+
+This 3 different events type can enter in the flow. This Caffeine Flow will try to gather those events on their correlation id.
+
+An aggregation class will be emitted:
+
+* as completed, if all event for a given id entered the flow during an interval of time.
+* as expired, after the timeouts.
+
+ 
+## Artifacts
+
+sbt
+:   @@@vars
+    ```scala
+    libraryDependencies += "com.typesafe.akka" %% "akka-stream-alpakka-caffeine" % "$version$"
+    ```
+    @@@
+
+Maven
+:   @@@vars
+    ```xml
+    <dependency>
+      <groupId>com.typesafe.akka</groupId>
+      <artifactId>akka-stream-alpakka-caffeine_$scala.binaryVersion$</artifactId>
+      <version>$version$</version>
+    </dependency>
+    ```
+    @@@
+
+Gradle
+:   @@@vars
+    ```gradle
+    dependencies {
+      compile group: "com.typesafe.akka", name: "akka-stream-alpakka-caffeine_$scala.binaryVersion$", version: "$version$"
+    }
+    ```
+    @@@
+
+## Usage
+
+Reconcilation logic must be provided under the form of an Aggregator.
+
+
+scala
+:   @@snip (../../../../caffeine/src/test/scala/akka/stream/alpakka/caffeine/scaladsl/CaffeineFlowSpec.scala) { #aggregator }
+
+
+
+2 flow stages are provided
+
+### CaffeineFanOut2Stage
+
+2 outputs that emit reconciled and expired events.
+
+### CaffeineFlowStage
+
+1 OutLet that emit an Either
+
+* Right containing reconciled events.
+* Left containing expired events
--- a/caffeine/src/main/scala/akka/stream/alpakka/caffeine/Aggregator.scala
+++ b/caffeine/src/main/scala/akka/stream/alpakka/caffeine/Aggregator.scala
@@ -1,0 +1,71 @@
+/*
+ * Copyright (C) 2016-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.alpakka.caffeine
+
+import scala.concurrent.duration.FiniteDuration
+
+/**
+ * Aggregation contract to:
+ * <ul>
+ * <li>Initializes a new aggregate: when a element with a new correlation id (cache key) hits the flow</li>
+ * <li>Extracts the correlation id from incoming/outgoing element</li>
+ * <li>Combines an incoming element with current aggregate</li>
+ * <li>Decides if aggregate is complete</li>
+ * <li>Defines the retention time (TTL) of element in the cache</li>
+ * </ul>
+ *
+ * @tparam K key in the cache
+ * @tparam V value in the cache
+ * @tparam R aggregation type
+ */
+trait Aggregator[K, V, R] {
+
+  /**
+   * Initiate a fresh aggregate.
+   *
+   * @param key a new correlation id
+   * @return a fresh aggregate with the key, containing not event.
+   */
+  def newAggregate(key: K): R
+
+  /**
+   * Extract the key from Event (in).
+   *
+   * @param v event to extract the key from.
+   * @return the key from an event.
+   */
+  def inKey(v: V): K
+
+  /**
+   * Extract the key from Aggregate (out) event.
+   *
+   * @param r aggregate to extract the key from.
+   * @return the key from aggregate
+   */
+  def outKey(r: R): K
+
+  /**
+   * Combine event and aggregate.
+   *
+   * @param r previous aggregate
+   * @param v new event
+   * @return aggregate r + v
+   */
+  def combine(r: R, v: V): R
+
+  /**
+   * Tells when aggregate in complete, thus can be emitted.
+   *
+   * @param r an aggregate
+   * @return true when aggregation is complete
+   */
+  def isComplete(r: R): Boolean
+
+  /**
+   * TTL (after write)
+   *
+   * @return the TTL for event in cache.
+   */
+  def expiration: FiniteDuration
+}
--- a/caffeine/src/main/scala/akka/stream/alpakka/caffeine/internal/CaffeineFanOut2Stage.scala
+++ b/caffeine/src/main/scala/akka/stream/alpakka/caffeine/internal/CaffeineFanOut2Stage.scala
@@ -1,0 +1,61 @@
+/*
+ * Copyright (C) 2016-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.alpakka.caffeine.internal
+
+import akka.stream.alpakka.caffeine.Aggregator
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
+import akka.stream.{FanOutShape2, _}
+
+import scala.language.postfixOps
+
+class CaffeineFanOut2Stage[K, V, R](aggregator: Aggregator[K, V, R]) extends GraphStage[FanOutShape2[V, R, R]] {
+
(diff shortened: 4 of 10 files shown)


docs/src/main/paradox/caffeine.md
```diff
@@ -0,0 +1,52 @@
+# Caffeine Flow
+
+The [Caffeine](https://github.com/ben-manes/caffeine) flow provides Akka Stream flow, to reconcile data in memory.
+
+
+Let A(1), A(2), A(3), ...., B(1), C(1) event to enter the flow, this flow will emit (A(1), B(1), C(1)) on reconciled OutLet and (A(2), A(3)) on Expired OutLet
```
patriknw: Could you elaborate what A, B and C are?

patriknw: I would not understand when I should use this component based on above description. I'm not familiar with Caffeine.

caffeine/src/main/scala/akka/caffeine/internal/CaffeineCache.scala
```diff
@@ -0,0 +1,23 @@
...
+package akka.caffeine.internal
+
+import akka.actor.{ Actor, ActorRef }
+import akka.actor.Actor.Receive
+import akka.stream.stage.GraphStageLogic.StageActor
+
+/**
+ * Created by olivier.nouguier@gmail.com on 04/11/2016.
+ */
+class CaffeineCache extends Actor {
```
patriknw: not used?

caffeine/src/main/scala/akka/caffeine/internal/CaffeineGraphLogic.scala
```diff
@@ -0,0 +1,99 @@
...
+
+    val old = cache.getIfPresent(key)
+
+    val r = aggregator.combine(old.getOrElse(aggregator.newAggregate(key)), a)
+
+    if (aggregator.isComplete(r)) {
+      if (old.isDefined)
+        cache.invalidate(key)
+      else
+        println("WTF !")
```
patriknw: no println in production code

cheleb:
Done changes requested.
Is the README clear enough now ? 
Is it possible/recommended to provide schema ?
Thx for the review.

build.sbt
```diff
@@ -10,6 +10,14 @@ lazy val amqp = project
     Dependencies.Amqp
   )
 
+lazy val caffeine = project
+  .in(file("caffeine"))
```
hseeberger: Not needed, please delete.

caffeine/src/main/scala/akka/caffeine/Aggregator.scala
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.caffeine
+
+import scala.concurrent.duration.FiniteDuration
+
+/**
+ * Aggregation contract
```
hseeberger: OK, an `Aggregator` does aggregation. I would have concluded that from the class name. Could you please provide a more descriptive comment here?

caffeine/src/main/scala/akka/caffeine/Aggregator.scala
```diff
@@ -0,0 +1,58 @@
...
+ * @tparam K key in the cache
+ * @tparam V value in the cache
+ * @tparam R aggregation type
+ */
+trait Aggregator[K, V, R] {
+
+  /**
+   * Initiate a fresh aggregate.
+   *
+   * @param key
```
hseeberger: Either provide nonempty comments or get rid of these (possibly autogenerated) tags.

caffeine/src/main/scala/akka/caffeine/internal/CaffeineFanOut2Stage.scala
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.caffeine.internal
+
+import akka.caffeine.Aggregator
+import akka.stream.{ FanOutShape2, _ }
+import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
+
+import scala.language.postfixOps
```
hseeberger: @patriknw Do we want to allow postfix ops? If so, maybe define the switch in the project settings instead, i.e. "globally".

caffeine/src/main/scala/akka/caffeine/internal/CaffeineFanOut2Stage.scala
```diff
@@ -0,0 +1,61 @@
...
+
+class CaffeineFanOut2Stage[K, V, R](aggregator: Aggregator[K, V, R]) extends GraphStage[FanOutShape2[V, R, R]] {
+
+  private val in = Inlet[V]("in")
+  private val out = Outlet[(R)]("out")
+  private val exp = Outlet[(R)]("exp")
+
+  override def shape: FanOutShape2[V, R, R] = new FanOutShape2(in, out, exp)
```
hseeberger: Type annotation not needed for overridden member.

caffeine/src/main/scala/akka/caffeine/internal/CaffeineFanOut2Stage.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new CaffeineGraphLogic[K, V, R](shape, aggregator, in) with InHandler {
+
+      protected def pumpComplete(): Boolean =
```
hseeberger: Please add `override` here, that helps understanding!

caffeine/src/main/scala/akka/caffeine/internal/CaffeineFanOut2Stage.scala
```diff
@@ -0,0 +1,61 @@
...
+        if (isAvailable(out)) {
```
hseeberger: Consider changing into `isAvailable(out) && ...` instead of if-else.

caffeine/src/main/scala/akka/caffeine/internal/CaffeineFanOut2Stage.scala
```diff
@@ -0,0 +1,61 @@
...
+          if (completeQueue.isEmpty) {
+            if (!isClosed(in) && !hasBeenPulled(in))
```
hseeberger: This can also be simplified using boolean operators.

caffeine/src/main/scala/akka/caffeine/internal/CaffeineFlowStage.scala
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.caffeine.internal
+
+import akka.caffeine.Aggregator
+import akka.stream.{ FanOutShape2, _ }
```
hseeberger: No wildcard imports at top-level, please.

caffeine/src/main/scala/akka/caffeine/internal/CaffeineFlowStage.scala
```diff
@@ -0,0 +1,58 @@
...
+import scala.annotation.tailrec
+import scala.collection.mutable
+import scala.language.postfixOps
+
+class CaffeineFlowStage[K, V, R](aggregator: Aggregator[K, V, R]) extends GraphStage[FlowShape[V, Either[R, R]]] {
+
+  private val in = Inlet[V]("in")
+  private val out = Outlet[Either[R, R]]("out")
+
+  override def shape: FlowShape[V, Either[R, R]] = new FlowShape(in, out)
```
hseeberger: s/def/val

caffeine/src/main/scala/akka/caffeine/internal/CaffeineFlowStage.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new CaffeineGraphLogic[K, V, R](shape, aggregator, in) with InHandler {
+
+      override final def pumpComplete(): Boolean =
+        if (isAvailable(out)) {
```
hseeberger: Same boolean stuff like above.

caffeine/src/main/scala/akka/caffeine/internal/CaffeineGraphLogic.scala
```diff
@@ -0,0 +1,98 @@
...
+    extends GraphStageLogic(shape)
+    with InHandler {
+
+  val completeQueue = mutable.Queue.empty[R]
+
+  val expiredQueue = mutable.Queue.empty[R]
+
+  var cache: Cache[K, R] = _
+
+  var onAir = mutable.Set[K]()
```
hseeberger: Ah, if it's a mutable collection, it can and should be a val, right?

caffeine/src/main/scala/akka/caffeine/internal/CaffeineGraphLogic.scala
```diff
@@ -0,0 +1,98 @@
...
+
+    } else {
+      if (old.isEmpty)
+        onAir.add(key)
+      cache.put(key, r)
+    }
+    pump()
+  }
+
+  var closing = false
```
hseeberger: Move up to val/var section.

caffeine/src/main/scala/akka/caffeine/internal/CaffeineGraphLogic.scala
```diff
@@ -0,0 +1,98 @@
...
+import com.github.blemale.scaffeine.{ Cache, Scaffeine }
+
+import scala.annotation.tailrec
+import scala.collection.mutable
+
+abstract class CaffeineGraphLogic[K, V, R](shape: Shape, aggregator: Aggregator[K, V, R], in: Inlet[V])
+    extends GraphStageLogic(shape)
+    with InHandler {
+
+  val completeQueue = mutable.Queue.empty[R]
```
hseeberger: Use `private` modifier for vals/vars.

caffeine/src/test/scala/akka/caffeine/CaffeineTestSuite.scala
```diff
@@ -0,0 +1,125 @@
...
+import scala.concurrent.duration.{ Duration, DurationInt }
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.language.postfixOps
+
+sealed trait Event {
+  def id: String
+}
+
+case class EventA(id: String, data: String) extends Event
+
```
hseeberger: Delete blank lines.

cheleb: AFAIU formatting is forced ...with scalafmt

johanandren:
If I look at caffeine it's just a cache, so these are in fact not at all "Caffeine Flows" but a much more specific use case and that Caffeine is used is pretty much an implementation detail as far as I can see?

Additionally, when I try to read up on "data reconciliation" (on [wikipedia](https://en.wikipedia.org/wiki/Data_validation_and_reconciliation#Data_reconciliation) for example) I cannot figure out how this does that or find the terminology used here.

Maybe if you could shortly describe why and in what scenario this is useful? (I'm not quite convinced this fits in Alpakka)

cheleb:
https://en.wikipedia.org/wiki/Reconciliation_(accounting) is what I mean by
reconciliation, the terminology is used in our domain (at least in my
company :) )

You are right that caffeine is just an implementation detail, then ?

Our use case in advertising with an auction system:

*BidRequest*: We are receiving BidRequest from SSP Company like google,
this BidRequest contains a lot of information (user agent, "referer",
domain, IP address, geoloc etc etc ).
*BidResponse*: Some time (up to few second) we are replying with
BidResponse with a price, advertising content and some meta
*WinNotice*: Sometime (up to few minutes or never) we are winning the
auction, and are notified with an WinNotice even with winPrice et some
other meta.
*Impression*: Sometime (up to few minutes or never) our banner, video (...)
is displayed to the user, then we receive an Impression event if the
network is up
*Click*: The banner is clicked
Etc etc (many other event days after).

All those event are pushed in a kafka cluster in many many topics.

Now, what we need is to reconciled all those event, because for example,
when a *Click* event arrives, we need, the information of the *BidRequest.*
All related event shares a correlation id (uuid), but they occurs in very
variable temporality. Validation rules (hence *reconciliation*) occurs with
all information aggregated in the cache.

We need (near) real time reconciliation, and that is the purpose on this
flow. *BidRequest, **BidResponse, **WinNotice etc *enter the flow and
*ReconciledWin,
ReconciledClick *are emitted on OutLet "complete" when they are complete
and on OutLet expired when caffeine (cache eviction policies) reject data
because of TTL expiration  or capacity limit.

HIH



Le ven. 18 nov. 2016 à 20:42, Johan Andrén <notifications@github.com> a
écrit :

johanandren:
Thanks for the very detailed explanation @cheleb!

cheleb:
You welcome, does it make more sense now ? Do you think "reconciliation" is
a real poor term in this situation ?

PS: I'm french so ... don't hesitate :)


Le sam. 19 nov. 2016 à 14:54, Johan Andrén <notifications@github.com> a
écrit :

johanandren:
My current thought is that it isn't really a connector but would be better described as a pattern or even a specific use case. I'm not quite sure how/where this fits in Alpakka, we do have integration patterns as one topic in the docs, maybe it could fit there. Let's see what Patrik and Heiko thinks about this.

As the term seems to be overloaded I think it is important to describe which of the meanings it is about and how it is useful.

patriknw:
We have a section called Integration Patterns in the docs. Such a pattern can be backed released code, or just be documentation.

caffeine/src/main/scala/akka/caffeine/internal/CaffeineFanOut2Stage.scala
patriknw: I don't think that should be used more than necessary, i.e. only when it adds some substantial readability value.

patriknw:
@cheleb What is the status here. Shall we move it to the Patterns section of the docs? The code can still be in same place.

cheleb:
Yes I will try to do this asap.


Le dim. 5 févr. 2017 à 22:33, Patrik Nordwall <notifications@github.com> a
écrit :

raboof:
@cheleb I see you've since moved the documentation to the Patterns section of the documentation - sorry, we seem to have missed that. Would you be interested in fixing the last merge conflict and getting this ball rolling again?

ennru:

Issue #30, opened 2016-11-11
Move S3 configuration to conf file


agolubev:
Continuation of #1 ticket

johanandren:
Support both loading from file and programatic config would be the nicest, IMO.

2m:
Fixed in #24

PR #24, opened 2016-11-08
Initial import of https://github.com/bluelabsio/s3-stream

jypma:
Including chunk-to-disk support and minor repackaging. 

PR'ing this to get things rolling with https://github.com/akka/alpakka/issues/1, hopefully with @joearasin 's blessing :-)

**TODO**
- [x] Remove scala 2.12 support from this module, or wait till there's `akka-http` for scala 2.12

**LATER**
- [ ] Write tests against WireMock (but that may require implementing proxy support in akka-http, https://github.com/akka/akka-http/issues/115 )
- [ ] Have `S3Location` take `Uri` instead of `String`, since currently the signature fails if the string contains characters that need to be URL-encoded.

PR #54, opened 2016-11-14
Adding config file and config class

agolubev:
PR is for #30 however it was branched from  #24


s3/src/main/scala/akka/stream/alpakka/s3/auth/Signer.scala
```diff
@@ -26,29 +25,29 @@ object Signer {
...
   }
 
   def stringToSign(algorithm: String, signingKey: SigningKey, requestDate: ZonedDateTime, canonicalRequest: CanonicalRequest): String = {
     val digest = MessageDigest.getInstance("SHA-256")
-    val hashedRequest = Utils.encodeHex(digest.digest(canonicalRequest.canonicalString.getBytes()))
-    s"$algorithm\n${requestDate.format(dateFormatter)}\n${signingKey.scope.scopeString}\n$hashedRequest"
+    val hashedRequest = encodeHex(digest.digest(canonicalRequest.canonicalString.getBytes()))
+    val date = requestDate.format(dateFormatter)
+    val scope = signingKey.scope.scopeString
+    s"$algorithm\n$date\n$scope\n$hashedRequest"
```
hseeberger: Ah, that's so much better ;-)

2m:
This has been merged as a part of #24

PR #56, opened 2016-11-14
Cassandra sink

stephennancekivell:
Hey, should we provide a Cassandra sink like this.

It seems pretty simple, but I think this is what a user would want.

What do you think ?

```
+def apply[T](parallelism: Int, statement: PreparedStatement, statementBinder: (T, PreparedStatement) => BoundStatement)(implicit session: Session, ex: ExecutionContext): Sink[T, Future[Done]] =
+    Flow[T].mapAsyncUnordered(parallelism)(t ⇒
+      session.executeAsync(statementBinder(t, statement)).toFuture()
+    ).toMat(Sink.ignore)(Keep.right)
+}
```


cassandra/src/main/scala/akka/stream/alpakka/cassandra/GuavaFutureOpts.scala
```diff
@@ -0,0 +1,25 @@
...
+    }
+    Futures.addCallback(guavaFut, callback)
+    p.future
+  }
+}
+
+object GuavaFutureOptsImplicits {
+  implicit def toGuavaFutureOpts[A](guavaFut: ListenableFuture[A]): GuavaFutureOpts[A] =
+    new GuavaFutureOpts[A](guavaFut)
+}
```
patriknw: use `implicit class ... extends AnyVal` instead, like this: https://github.com/akka/akka-persistence-cassandra/blob/master/src/main/scala/akka/persistence/cassandra/package.scala#L22

cassandra/src/test/scala/akka/stream/alpakka/cassandra/scaladsl/CassandraSourceSpec.scala
```diff
@@ -5,13 +5,16 @@ package akka.stream.alpakka.cassandra.scaladsl
...
 import akka.stream.ActorMaterializer
-import akka.stream.scaladsl.Sink
+import akka.stream.scaladsl.{ Sink, Source }
 import com.datastax.driver.core.{ Cluster, SimpleStatement }
 import org.scalatest._
 import org.scalatest.concurrent.ScalaFutures
 
 import scala.concurrent._
 import scala.concurrent.duration._
+import scala.concurrent.ExecutionContext.Implicits.global
```
patriknw: Avoid this EC, also in tests, unless there is a good reason. Use `import system.dispatcher` inside the test instead.

cassandra/src/test/scala/akka/stream/alpakka/cassandra/scaladsl/CassandraSourceSpec.scala
```diff
@@ -5,13 +5,16 @@ package akka.stream.alpakka.cassandra.scaladsl
...
+
+import scala.collection.convert.decorateAsScala._
```
patriknw: I think it's the same but I prefer `scala.collection.JavaConverters._`

stephennancekivell:
Hi @patriknw, I've implemented your feedback, thanks for that.

Im not sure how to do the java dsl at the moment and it might be a few weeks before I have time to look at it. I agree it will be good for this project.

So, I dont know if you want to merge this in without java or not. Up to you.

Cheers,

cassandra/src/main/scala/akka/stream/alpakka/cassandra/scaladsl/CassandraSink.scala
```diff
@@ -0,0 +1,18 @@
...
+import akka.Done
+import akka.stream.scaladsl.{ Flow, Keep, Sink }
+import com.datastax.driver.core.{ BoundStatement, PreparedStatement, Session }
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+import akka.stream.alpakka.cassandra.cassandra._
+
+object CassandraSink {
+  def apply[T](parallelism: Int, statement: PreparedStatement, statementBinder: (T, PreparedStatement) => BoundStatement)(implicit session: Session, ex: ExecutionContext): Sink[T, Future[Done]] =
```
patriknw: some hints for the javadsl
* statementBinder: BiFunction[T, PreparedStatement, BoundStatement]
* delegate to `scaladsl.CassandraSink.apply`
* `mapMaterializedValue` and `scala.compat.java8.FutureConverters._` to convert the `Future` to `CompletionStage`
* Sink.asJava to convert to `akka.stream.javadsl.Sink`

patriknw:
@stephennancekivell We would like to see it completed with javadsl and some small usage example in the [docs](https://github.com/akka/alpakka/blob/master/docs/src/main/paradox/cassandra.md) before merging.

stephennancekivell:
Awesome, Thanks for the tips. Will do.

stephennancekivell:
Hi @patriknw I've completed the Java DSL and the logging. Do you or anyone else have any feedback?

cassandra/src/main/scala/akka/stream/alpakka/cassandra/javadsl/CassandraSink.scala
```diff
@@ -0,0 +1,29 @@
...
+
+object CassandraSink {
+
+  def create[T](parallelism: Int,
+                statement: PreparedStatement,
+                statementBinder: BiFunction[T, PreparedStatement, BoundStatement],
+                session: Session): Sink[T, CompletionStage[Done]] = {
+    val sink = ScalaCSink.apply[T](parallelism, statement, (t, p) => statementBinder.apply(t, p))(
+      session,
+      scala.concurrent.ExecutionContext.global)
```
johanandren: Take the EC as a parameter to `create` instead, just like the Scala api does (if done like this the user has no control).

cassandra/src/main/scala/akka/stream/alpakka/cassandra/package.scala
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.alpakka.cassandra
+
+import com.google.common.util.concurrent.{ FutureCallback, Futures, ListenableFuture }
+
+import scala.concurrent.{ Future, Promise }
+
+package object cassandra {
```
johanandren: `private[cassandra]`

cassandra/src/main/scala/akka/stream/alpakka/cassandra/package.scala
```diff
@@ -0,0 +1,22 @@
...
+  implicit class GuavaFutureOpts[A](val guavaFut: ListenableFuture[A]) extends AnyVal {
```
johanandren: `final`?

cassandra/src/main/scala/akka/stream/alpakka/cassandra/CassandraSourceStage.scala
```diff
@@ -6,11 +6,12 @@ package akka.stream.alpakka.cassandra
...
 import com.datastax.driver.core.{ ResultSet, Row, Session, Statement }
-import com.google.common.util.concurrent.{ FutureCallback, Futures, ListenableFuture }
 
-import scala.concurrent.{ Future, Promise }
+import scala.concurrent.Future
 import scala.util.{ Failure, Success, Try }
 
+import akka.stream.alpakka.cassandra.cassandra._
+
 class CassandraSourceStage(futStmt: Future[Statement], session: Session) extends GraphStage[SourceShape[Row]] {
```
johanandren: `final class` (please boyscout this for us)

cassandra/src/main/scala/akka/stream/alpakka/cassandra/package.scala
stephennancekivell: Hi Johan, where do you want me to put that. I cant get it to work. Its a compiler error on the object and cant be used on the implicit class.  ???

johanandren: Ah, you cannot put access restrictions on the package object itself, so then it needs to go on the implicit decorator, if kept public the risk is that library consumers start using it and then it cannot be moved/changed in the future, and this library is not here to provide guice-conversions.

stephennancekivell: Good point. Thanks.

I had two issues,
* I had to change the object name from `cassandara`, I think the `private[cassandra]` declaration was referencing the object instead of the package, which stopped it working.
* Then I had to add a implicit def to convert to the opts. It seems like the implicit class wasnt getting picked up with the private.

I that right,

cassandra/src/main/scala/akka/stream/alpakka/cassandra/package.scala
```diff
@@ -7,8 +7,8 @@ import com.google.common.util.concurrent.{ FutureCallback, Futures, ListenableFu
 
 import scala.concurrent.{ Future, Promise }
 
-package object cassandra {
-  implicit final class GuavaFutureOpts[A](val guavaFut: ListenableFuture[A]) extends AnyVal {
+package object GuavaFutureOpts {
```
johanandren: With this name it is no longer a package object, just a normal object, so remove that and make it package private for cassandra with `private[cassandra]` which will cover the class inside it as well. You will have to import the implicit class into the scope where it is used though, so that will be an `import GuavaFutureOpts._` in the classes you call `asScala`

stephennancekivell: Right, thanks. However i still need the `implicit def` the `implicit class` alone doesnt seem to get picked up...

cassandra/src/main/scala/akka/stream/alpakka/cassandra/GuavaFutureOpts.scala
```diff
@@ -20,6 +20,6 @@ package object GuavaFutureOpts {
     }
   }
 
-  private[cassandra] implicit def toGuavaFutureOpts[A](guavaFut: ListenableFuture[A]): GuavaFutureOpts[A] =
+  implicit def toGuavaFutureOpts[A](guavaFut: ListenableFuture[A]): GuavaFutureOpts[A] =
```
johanandren: if you make the class implicit you do not need this either

stephennancekivell:
Amqp tests are broken there, TimeoutException.

stephennancekivell:
@johanandren I've just pushed a commit with the changes like you suggested. However now Its a compiler error. So what do I do here. It seemed ok to me with the def. Is there a better way.

https://travis-ci.org/akka/alpakka/jobs/177938167#L773

johanandren:
If you look at the docs for implicit classes (http://docs.scala-lang.org/overviews/core/implicit-classes.html) you can see that an implicit class cannot have the same name as something else that is already in scope, therefore `GuavaFutureOpts` inside the object `GuavaFutureOpts` which is in the same package as where it is used will not work.

Rename the object or the implicit class and that will solve the problem.

stephennancekivell:
Thanks @patriknw

Thanks @johanandren I hadn't come across that rule before.

stephennancekivell:
hey @patriknw @johanandren,

Can this be merged in ?

Thanks

patriknw:
Refs https://github.com/akka/alpakka/issues/57

patriknw:
Excellent contribution @stephennancekivell !

Issue #44, opened 2016-11-12
AWS SQS Connector


dpfeiffer:
Although there is an implementation of an SQS Source in https://github.com/MfgLabs/commons-aws, I can start one which does not need external dependencies (except aws-java-sdk-sqs) and is based on `GraphStage` and not `ActorPublisher`.

Issue #55, opened 2016-11-14
FAILED: AmqpConnectorsTest


jypma:
[TimeoutException](https://travis-ci.org/akka/alpakka/jobs/175643872#L728) on `AmqpConnectorsTest`, but only for Scala 2.12. Timing issues? That timeout already is at 10 seconds...
```
[error] Test akka.stream.alpakka.amqp.javadsl.AmqpConnectorsTest.publishFanoutAndConsume failed: java.util.concurrent.TimeoutException: null, took 10.326 sec

[error]     at java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1771)

[error]     at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1915)

[error]     at scala.concurrent.java8.FuturesConvertersImpl$CF.super$get(FutureConvertersImpl.scala:84)

[error]     at scala.concurrent.java8.FuturesConvertersImpl$CF.$anonfun$get$2(FutureConvertersImpl.scala:84)

[error]     at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:53)

[error]     at scala.concurrent.package$.blocking(package.scala:123)

[error]     at scala.concurrent.java8.FuturesConvertersImpl$CF.get(FutureConvertersImpl.scala:84)

[error]     at akka.stream.alpakka.amqp.javadsl.AmqpConnectorsTest.publishFanoutAndConsume(AmqpConnectorsTest.java:142)
```

stephennancekivell:
I've noticed this as well, can we simply increase the timeout ?

2m:
Looks like a duplicate of https://github.com/akka/alpakka/issues/20

PR #59, opened 2016-11-14
Amazon SQS Source

dpfeiffer:
Hi, this is the SqsSource implementation that depends only on the aws-java-sdk-sqs and no other dependencies (issue #44). 

The source does not automatically delete messages from a queue (not even configurable) because usually you will ensure that the message was successfully processed before removing it from the queue. Either way, it would be only one line to explicitly remove a message from a SQS queue.

The SqsSourceStage gets the `AmazonSQSAsyncClient` as an explicit argument and does not create its own for exact the same reason and if it would create an own one, the user had to create a second client to remove the message.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ before_install:
   - docker run -p 9042:9042 -d cassandra:3
   - docker pull rabbitmq:3
   - docker run -p 5672:5672 -d rabbitmq:3
+  - docker pull expert360/elasticmq
+  - docker run -p 9324:9324 -d expert360/elasticmq
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,8 @@ before_install:
   - docker run -p 9042:9042 -d cassandra:3
   - docker pull rabbitmq:3
   - docker run -p 5672:5672 -d rabbitmq:3
-  - docker pull expert360/elasticmq
-  - docker run -p 9324:9324 -d expert360/elasticmq
+  - docker pull expert360/elasticmq:0.8.12
+  - docker run -p 9324:9324 -d expert360/elasticmq:0.8.12
 
 cache:
   directories:
--- a/docs/src/main/paradox/sqs.md
+++ b/docs/src/main/paradox/sqs.md
@@ -68,7 +68,9 @@ SQS queues never finished because there is no direct way to determine the end of
 The code in this guide is part of runnable tests of this project. You are welcome to edit the code and run it in sbt.
 
 > The test code uses [ElasticMQ](https://github.com/adamw/elasticmq) as queuing service which serves an AWS SQS 
-> compatible API.
+> compatible API.  You can start one quickly using docker:
+>
+> `docker run -p 9324:9324 -d expert360/elasticmq`
 
 Scala
 :   ```
--- a/docs/src/main/paradox/sqs.md
+++ b/docs/src/main/paradox/sqs.md
@@ -1,6 +1,7 @@
 # AWS SQS Connector
 
 The AWS SQS connector allows to stream SQS `Message` from a AWS SQS queue.
+For more information about AWS SQS please visit the [official docmentation](https://aws.amazon.com/documentation/sqs/).
 
 ## Artifacts
 
--- a/docs/src/main/paradox/sqs.md
+++ b/docs/src/main/paradox/sqs.md
@@ -1,0 +1,83 @@
+# AWS SQS Connector
+
+The AWS SQS connector allows to stream SQS `Message` from a AWS SQS queue.
+
+## Artifacts
+
+sbt
+:   @@@vars
+    ```scala
+    libraryDependencies += "com.typesafe.akka" %% "akka-stream-alpakka-sqs" % "$version$"
+    ```
+    @@@
+
+Maven
+:   @@@vars
+    ```xml
+    <dependency>
+      <groupId>com.typesafe.akka</groupId>
+      <artifactId>akka-stream-alpakka-sqs_$scala.binaryVersion$</artifactId>
+      <version>$version$</version>
+    </dependency>
+    ```
+    @@@
+
+Gradle
+:   @@@vars
+    ```gradle
+    dependencies {
+      compile group: "com.typesafe.akka", name: "akka-stream-alpakka-sqs_$scala.binaryVersion$", version: "$version$"
+    }
+    ```
+    @@@
+
+## Usage
+
+Sources provided by this connector need a prepared `AmazonSQSAsyncClient` to load messages from a queue. 
+
+Scala
+: @@snip (../../../../sqs/src/test/scala/akka/stream/alpakka/sqs/scaladsl/SqsSourceSpec.scala) { #init-client }
+
+Java
+: @@snip (../../../../sqs/src/test/java/akka/stream/alpakka/sqs/javadsl/SqsSourceTest.java) { #init-client }
+
+We will also need an @scaladoc[ActorSystem](akka.actor.ActorSystem) and an @scaladoc[ActorMaterializer](akka.stream.ActorMaterializer).
+
+Scala
+: @@snip (../../../../sqs/src/test/scala/akka/stream/alpakka/sqs/scaladsl/SqsSourceSpec.scala) { #init-mat }
+
+Java
+: @@snip (../../../../sqs/src/test/java/akka/stream/alpakka/sqs/javadsl/SqsSourceTest.java) { #init-mat }
+
+This is all preparation that we are going to need.
+
+Now we can stream AWS Java SDK SQS `Message` objects from any SQS queue where we have access to by providing the queue URL to the
+@scaladoc[SqsSource](akka.stream.alpakka.sqs.scaladsl.SqsSource$) factory.
+
+Scala
+: @@snip (../../../../sqs/src/test/scala/akka/stream/alpakka/sqs/scaladsl/SqsSourceSpec.scala) { #run }
+
+Java
+: @@snip (../../../../sqs/src/test/java/akka/stream/alpakka/sqs/javadsl/SqsSourceTest.java) { #run }
+
+As you have seen we consume the stream for 100 milliseconds. The reason for this is, that reading messages from
+SQS queues never finished because there is no direct way to determine the end of a queue.
+
+### Running the example code
+
+The code in this guide is part of runnable tests of this project. You are welcome to edit the code and run it in sbt.
+
+> The test code uses [ElasticMQ](https://github.com/adamw/elasticmq) as queuing service which serves an AWS SQS 
+> compatible API.
+
+Scala
+:   ```
+    sbt
+    > sqs/testOnly *.SqsSourceSpec
+    ```
+
+Java
+:   ```
+    sbt
+    > sqs/testOnly *.SqsSourceTest
+    ```
--- a/docs/src/main/paradox/sqs.md
+++ b/docs/src/main/paradox/sqs.md
@@ -8,7 +8,7 @@ For more information about AWS SQS please visit the [official docmentation](http
 sbt
 :   @@@vars
     ```scala
-    libraryDependencies += "com.typesafe.akka" %% "akka-stream-alpakka-sqs" % "$version$"
+    libraryDependencies += "com.lightbend.akka" %% "akka-stream-alpakka-sqs" % "$version$"
     ```
     @@@
 
@@ -16,7 +16,7 @@ Maven
 :   @@@vars
     ```xml
     <dependency>
-      <groupId>com.typesafe.akka</groupId>
+      <groupId>com.lightbend.akka</groupId>
       <artifactId>akka-stream-alpakka-sqs_$scala.binaryVersion$</artifactId>
       <version>$version$</version>
     </dependency>
@@ -27,7 +27,7 @@ Gradle
 :   @@@vars
     ```gradle
     dependencies {
-      compile group: "com.typesafe.akka", name: "akka-stream-alpakka-sqs_$scala.binaryVersion$", version: "$version$"
+      compile group: "com.lightbend.akka", name: "akka-stream-alpakka-sqs_$scala.binaryVersion$", version: "$version$"
     }
     ```
     @@@
@@ -61,8 +61,8 @@ Scala
 Java
 : @@snip (../../../../sqs/src/test/java/akka/stream/alpakka/sqs/javadsl/SqsSourceTest.java) { #run }
 
-As you have seen we consume the stream for 100 milliseconds. The reason for this is, that reading messages from
-SQS queues never finished because there is no direct way to determine the end of a queue.
+As you have seen we take the first 100 elements from the stream. The reason for this is, that reading messages from
+SQS queues never finishes because there is no direct way to determine the end of a queue.
 
 ### Running the example code
 
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -14,9 +14,9 @@ object Dependencies {
       "junit"              % "junit"               % "4.12"        % Test  // Eclipse Public License 1.0
     )
   )
-  
+
   val AkkaHttpVersion = "10.0.0-RC2"
-  
+
   val S3 = Seq(
     libraryDependencies ++= Seq(
       "com.typesafe.akka" %% "akka-http"     % AkkaHttpVersion,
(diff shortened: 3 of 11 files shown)


hseeberger:
@dpfeiffer Have you seen that CI fails for both Scala 2.11 and 2.12 (for different reasons). The latter is due to elasticmq-rest-sqs not being available for 2.12 => if you run `+test` or at least `+test:compile` locally this would have become clear earlier.

project/Dependencies.scala
```diff
@@ -40,4 +40,12 @@ object Dependencies {
       "com.google.jimfs"  %  "jimfs"               % "1.1"  % Test  // ApacheV2
     )
   )
+
+  val Sqs = Seq(
+    libraryDependencies ++= Seq(
+      "com.amazonaws"   % "aws-java-sdk-sqs"    % "1.11.51",          // ApacheV2
+      "org.elasticmq"  %% "elasticmq-rest-sqs"  % "0.10.1"    % Test  // ApacheV2
```
hseeberger: Please align first operator chars ;-)

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -0,0 +1,77 @@
...
+ */
+package akka.stream.alpakka.sqs
+
+import akka.stream.stage.{ GraphStage, GraphStageLogic, OutHandler }
+import akka.stream.{ Attributes, Outlet, SourceShape }
+import com.amazonaws.handlers.AsyncHandler
+import com.amazonaws.services.sqs.AmazonSQSAsyncClient
+import com.amazonaws.services.sqs.model.{ Message, ReceiveMessageRequest, ReceiveMessageResult }
+
+import scala.collection.JavaConversions._
```
hseeberger: Please use `JavaConverters`, because `JavaConversions` are deprecated in 2.12. If you build locally against 2.12, too, you'd notice.

sqs/src/test/scala/akka/stream/alpakka/sqs/scaladsl/SqsSourceSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+import com.amazonaws.services.sqs.AmazonSQSAsyncClient
+import org.elasticmq.rest.sqs.{SQSRestServer, SQSRestServerBuilder}
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.{BeforeAndAfterAll, Matchers, WordSpec}
+
+import scala.concurrent.duration._
+import scala.concurrent.{Await, Future}
+import scala.util.{Random, Try}
+
+class SqsSourceSpec extends WordSpec with BeforeAndAfterAll with ScalaFutures with Matchers {
```
hseeberger: Instead of using `WordSpec` with `Await.ready/result` I suggest you switch over to `AsyncWordSpec` which has been introduced in ScalaTest 3.

sqs/src/test/scala/akka/stream/alpakka/sqs/scaladsl/SqsSourceSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+    //This test has to block for a seconds because it is not possible to go below 1 second with the wait time
+    "continue streaming if receives an empty response" in {
+
+      val queue = randomQueueUrl()
+
+      val f = SqsSource(queue, SqsSourceSettings(1.second, 100))
+        .takeWithin(1100.millis)
+        .runWith(Sink.seq)
+
+      Thread.sleep(1000)
```
hseeberger: Ouch! Get rid of that, please.

sqs/src/test/scala/akka/stream/alpakka/sqs/scaladsl/SqsSourceSpec.scala
dpfeiffer: Thank you didn't know that.

project/Dependencies.scala
hseeberger: Oops, my bad. I didn't know that the authors have invented a new individual style. Keep it, I might try to change all of them later, if I have too much time ;-)

sqs/src/test/scala/akka/stream/alpakka/sqs/scaladsl/SqsSourceSpec.scala
hseeberger: Sure thing. For some inspiration you could take a look at https://github.com/hseeberger/akka-sse/blob/master/akka-sse/src/test/scala/de/heikoseeberger/akkasse/client/ServerSentEventParserSpec.scala#L50 (but don't tell me that I should use `Sink.seq` ;-).

dpfeiffer:
@hseeberger What do you think will be the best solution to make it run in 2.12. I could include a elastic docker container into the build so i don't need to depend on elasticmq. I open an issue on elasticmq which depends on akka-http so it could take a while until elasticmq is compiled for 2.12.

hseeberger:
I'm not sure, if this is possible, but you might want to try to set the `crossScalaVersions` for this project to "2.11.8" only.

.travis.yml
```diff
@@ -13,6 +13,11 @@ before_cache:
   - find $HOME/.ivy2 -name "ivydata-*.properties" -print -delete
   - find $HOME/.sbt  -name "*.lock"               -print -delete
 
+before_install:
+  - docker pull expert360/elasticmq
+  - docker run -p 9324:9324 -d expert360/elasticmq
+  - docker ps -a
```
hseeberger: Why that?

dpfeiffer: Wanted to make sure that it really started. I can also remove it.

dpfeiffer:
The build now fails because of this: https://github.com/travis-ci/travis-ci/issues/5502

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+      }
+
+      setHandler(out, new OutHandler {
+        override def onPull(): Unit = {
+          if (buffer.nonEmpty) {
+            push(out, buffer.dequeue())
+          } else {
+            receiveMessages()
```
agolubev: can we  receive messages to buffer asynchronously while downstream is pulling them?

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -0,0 +1,77 @@
...
+        })
+      }
+
+      def handleFailure(ex: Exception): Unit = {
+        failStage(ex)
+      }
+
+      def handleSuccess(result: ReceiveMessageResult): Unit = {
+
+        buffer.enqueue(result.getMessages.asScala.reverse: _*)
```
agolubev: there is no logic to prevent putting in buffer more than maxBufferSize

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -0,0 +1,77 @@
...
+import com.amazonaws.services.sqs.AmazonSQSAsyncClient
+import com.amazonaws.services.sqs.model.{ Message, ReceiveMessageRequest, ReceiveMessageResult }
+
+import scala.collection.JavaConverters._
+import scala.collection.mutable
+import scala.concurrent.duration.FiniteDuration
+
+case class SqsSourceSettings(longPollingDuration: FiniteDuration, maxBufferSize: Int)
+
+class SqsSourceStage(queueUrl: String, settings: SqsSourceSettings, sqsClient: AmazonSQSAsyncClient) extends GraphStage[SourceShape[Message]] {
```
agolubev: both classes should be final

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -0,0 +1,77 @@
...
+
+      private val buffer = mutable.Queue[Message]()
+
+      private val successCallback = getAsyncCallback[ReceiveMessageResult](handleSuccess)
+      private val failureCallback = getAsyncCallback[Exception](handleFailure)
+
+      def receiveMessages(): Unit = {
+
+        val request = new ReceiveMessageRequest(queueUrl)
+          .withMaxNumberOfMessages(10)
```
agolubev: perhaps this can be a value from settings

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -0,0 +1,77 @@
...
+          .withWaitTimeSeconds(settings.longPollingDuration.toSeconds.toInt)
+
+        sqsClient.receiveMessageAsync(request, new AsyncHandler[ReceiveMessageRequest, ReceiveMessageResult] {
```
agolubev: need to mention somewhere that this will init thread pool of 50 threads

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -0,0 +1,77 @@
...
+        if (buffer.nonEmpty && isAvailable(out)) {
+          push(out, buffer.dequeue())
+        }
+
+      }
+
+      setHandler(out, new OutHandler {
+        override def onPull(): Unit = {
+          if (buffer.nonEmpty) {
+            push(out, buffer.dequeue())
```
agolubev: maybe `emit` can be used here?

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
dpfeiffer: The threadpool get's configured on the cient itself not on the receive method. You are right that maybe users won't change the threadpool size. The only thin that could be changed in my opinion is to create an own AmazonSqsAsyncClient and configure it. (pushing the thread count as a setting into the Settings case class)

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
dpfeiffer: Why? As far as I know emit makes then sense when you don't want to check for the demand of downstream as it emits as soon as downstream demand is there. Within `onPull` i don't thin that makes sense because the downstream demand is clear at that point.

dpfeiffer:
@agolubev I implemented the requesting in `onPull that way that as soon as `maxBatchSize` elements are free in the buffer (when pulling) `maxBatchSize` elements are requested. The solution is a tradeoff between throughput and number of requests (as you get billed for them)

docs/src/main/paradox/sqs.md
```diff
@@ -0,0 +1,85 @@
+# AWS SQS Connector
+
+The AWS SQS connector allows to stream SQS `Message` from a AWS SQS queue.
```
patriknw: link to AWS SQS documentation?

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -0,0 +1,84 @@
...
+
+final class SqsSourceStage(queueUrl: String, settings: SqsSourceSettings, sqsClient: AmazonSQSAsyncClient) extends GraphStage[SourceShape[Message]] {
+
+  val out: Outlet[Message] = Outlet("SqsSource.out")
+  override val shape: SourceShape[Message] = SourceShape(out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = {
+    new GraphStageLogic(shape) {
+
+      private val buffer = mutable.Queue[Message]()
```
patriknw: perhaps use `java.util.ArrayDeque` for better performance?

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -0,0 +1,84 @@
...
+
+      def handleSuccess(result: ReceiveMessageResult): Unit = {
+
+        buffer.enqueue(result.getMessages.asScala.reverse: _*)
+
+        if (buffer.nonEmpty && isAvailable(out)) {
+          push(out, buffer.dequeue())
+        }
+
+        if (buffer.size < settings.maxBufferSize - settings.maxBatchSize) {
```
patriknw: should we simplify it by only using one `maxBufferSize` setting and trigger receiveMessages when `maxBufferSize / 2`?

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
patriknw: I think the idea is that `maxBatchSize < maxBufferSize`, but see my comment below

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
dpfeiffer: I am not sure about that, it also depends on the implementation of the `OutHandler`. Right now it's implemented like that, to ensure that `OutHandler.onPull` doesn't request as long as `handleSuccess` would do another request (without tracking the current requests). 

Request until buffer full (if consumer slower than producer) -> Don't request until exactly `maxBatchSize` is free in the buffer -> Request again (from `OutHandler.onPull` which will force `handleSuccess` to do more requests if the buffer has still space for `maxBatch` elements.

Request until buffer full (if consumer slower than producer) -> Don't request until exactly buffer is half empty -> Request again (from `OutHandler.onPull` which will force `handleSuccess` to do more requests if the buffer is still not full.

patriknw: ok, please add a `require` in the settings that `maxBatchSize <= maxBufferSize` (if that is not already done)

dpfeiffer:
@patriknw pushed your requested changes right now. Waiting for #65 to be merged to make the tests green.

2m:
@dpfeiffer thanks for the #65. After merging it, this one needs an update in the `.travis.yml` file.

dpfeiffer:
@2m Everything looks good now except that the amqp java test now fails with a timeout.

johanandren:
I think the failure is the same as: #55

docs/src/main/paradox/sqs.md
```diff
@@ -0,0 +1,86 @@
...
+For more information about AWS SQS please visit the [official docmentation](https://aws.amazon.com/documentation/sqs/).
+
+## Artifacts
+
+sbt
+:   @@@vars
+    ```scala
+    libraryDependencies += "com.typesafe.akka" %% "akka-stream-alpakka-sqs" % "$version$"
```
2m: s/typesafe/lightbend

docs/src/main/paradox/sqs.md
```diff
@@ -0,0 +1,86 @@
...
+    ```
+    @@@
+
+Maven
+:   @@@vars
+    ```xml
+    <dependency>
+      <groupId>com.typesafe.akka</groupId>
```
2m: s/typesafe/lightbend

docs/src/main/paradox/sqs.md
```diff
@@ -0,0 +1,86 @@
...
+      <version>$version$</version>
+    </dependency>
+    ```
+    @@@
+
+Gradle
+:   @@@vars
+    ```gradle
+    dependencies {
+      compile group: "com.typesafe.akka", name: "akka-stream-alpakka-sqs_$scala.binaryVersion$", version: "$version$"
```
2m: s/typesafe/lightbend

docs/src/main/paradox/sqs.md
```diff
@@ -0,0 +1,86 @@
...
+Now we can stream AWS Java SDK SQS `Message` objects from any SQS queue where we have access to by providing the queue URL to the
+@scaladoc[SqsSource](akka.stream.alpakka.sqs.scaladsl.SqsSource$) factory.
+
+Scala
+: @@snip (../../../../sqs/src/test/scala/akka/stream/alpakka/sqs/scaladsl/SqsSourceSpec.scala) { #run }
+
+Java
+: @@snip (../../../../sqs/src/test/java/akka/stream/alpakka/sqs/javadsl/SqsSourceTest.java) { #run }
+
+As you have seen we consume the stream for 100 milliseconds. The reason for this is, that reading messages from
```
2m: The example code has `take(100)` which is takes 100 stream elements.

docs/src/main/paradox/sqs.md
```diff
@@ -0,0 +1,86 @@
...
+SQS queues never finished because there is no direct way to determine the end of a queue.
```
2m: s/finished/finishes

patriknw:
@dpfeiffer looks like this needs another rebase, and then we can merge it.

dpfeiffer:
@patriknw done

patriknw:
Refs https://github.com/akka/alpakka/issues/44

patriknw:
Excellent, thanks

PR #65, opened 2016-11-16
Start services as docker containers

### LINKED ISSUE #62: Using Docker service on Travis fails Cassandra connector build
I think it is related to that: travis-ci/travis-ci#5502

I tried to use docker service to run elasticmq as it is not available for Scala 2.12 yet. Even if elasticmq will be available soon as they are just waiting for akka-http, I think that other connectors could need docker as well.

dpfeiffer:
- closes #62 
- Provide docker-compose.yml to start up services locally


hseeberger:
LGTM

ktoso:
Seems good for now, I wonder how many of such travis will survive at the same time, but we'll worry about that once we hit that limit.
LGTM

docker-compose.yml
```diff
@@ -0,0 +1,10 @@
+version: '2'
+services:
+  cassandra:
+    image: cassandra:3
+    ports:
+      - "9042:9042"
+  rabbitmq:
+    image: rabbitmq:3
+    ports:
+      - "5672:5672"
```
2m: Can you also tell how to use this? Preferably as a comment in the file.

2m: I think it makes sense. Created https://github.com/akka/alpakka/issues/75

s12v:
@ktoso, maybe it will be possible to create N builds, 1 per project(xM scala versions) and start only relevant containers

PR #69, opened 2016-11-17
Another stab at Reactive Socket API

2m:
RFC

The client tries to keep the same API feeling as other Source/Flow factories.

Server looks similar to Tcp/Http stream API


reactivesocket/src/test/scala/akka/stream/alpakka/reactivesocket/scaladsl/ReactiveSocketSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+    "support aeron transport" in {
+
+      import io.reactivesocket.aeron.internal._
+      import io.reactivesocket.aeron.internal.reactivestreams._
+      import io.reactivesocket.aeron.server.AeronTransportServer
+      import io.reactivesocket.aeron.client.AeronTransportClient
+
+      // server
+
+      // aeron config, which should be hidden from the user
```
patriknw: We should hide the reactive socket library, we should provide a more convenient api for integration with Akka things. I don't think we need to provide anything for the transport creation.

reactivesocket/src/test/scala/akka/stream/alpakka/reactivesocket/scaladsl/ReactiveSocketSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+  "reactive socket API" should {
+
+    "support server and client" in {
+
+      import io.reactivesocket.transport.tcp.client.TcpTransportClient
+      import io.reactivesocket.transport.tcp.server.TcpTransportServer
+
+      val binding = ReactiveSocketServer(TcpTransportServer.create())
+          .toMat(Sink.foreach {
+            _.flow.runWith(Source.repeat(new PayloadImpl("Pong")), Sink.foreach(println))
```
patriknw: How would the different interaction types (fire-forget, request-response, ...) be represented?
Compare with https://github.com/akka/alpakka/pull/26/files#diff-35147bb7a6ac0eb12c60538ecb260128R45

2m: All of the interaction types can be generalized as a flow on the server side. I think it only matters to distinguish between these interaction types on the initiating side - the client.

johanandren: Fire and forget can't be generalized to a `Flow`?

patriknw: I don't think that is true, you can almost think of them as method types (GET, POST, ...) in http. The server need to know the type to have full information about the request.

I think we should embrace the reactive socket specification instead of hiding/changing it. The different interaction types are defined in the spec for a reason (we like them or not, doesn't matter).

reactivesocket/src/test/scala/akka/stream/alpakka/reactivesocket/scaladsl/ReactiveSocketSpec.scala
patriknw: ouch, missed **NOT**

johanandren: now it became so much more clear :D thanks

reactivesocket/src/test/scala/akka/stream/alpakka/reactivesocket/scaladsl/ReactiveSocketSpec.scala
patriknw: The alternative is to encode it in the elements, different types of `IncomingConnection`, but I couldn't see a way to describe in the types that for example request-response should only return one element.

I also like the idea of that client and server interfaces are the same.

patriknw:
Refs #25

patriknw:
Closing this for now. We intend to pick it up again later.

PR #70, opened 2016-11-18
Link to release docs

2m:


README.md
```diff
@@ -8,7 +8,7 @@ This project provides a home to Akka Streams connectors to various technologies,
 Documentation
 -------------
 
-Latest [documentation](http://developer.lightbend.com/docs/alpakka/latest/).
+See [Releases](https://github.com/akka/alpakka/releases) page for the reference and API documentation.
```
patriknw: Hm, I'm not sure. A direct link to latest released documentation is much better. I didn't know that `latest` was master. If there is no support for such `current` symlink in techhub we should open an issue, and until solved we need to update this link manually.

That said, we can add more links here, one to the the Releases and another to the API docs.

johanandren: Oh, I'm to tired to review stuff. I my brain auto-corrected this to a different release path on techhub... I agree with Patrik.

2m: I like this, because it allows us to have one less step when cutting a release. And releases page has some useful information.

johanandren: I wish there was a `current` in techhub. Could we open a ticket for that?

patriknw: Until there is a `current` we must think about users, and not only what is least work for us. This is currently the landing page and there should be quick links to the documentation.

2m: Fair enough. Will change this to point to the current version.

Issue #35, opened 2016-11-11
HBase connector


cheleb:
Hi
  I can provide HBase (sink) connector.

  Any interest ?

patriknw:
that is useful, please go ahead

PR #73, opened 2016-11-18
Initial import of HBase connector.

cheleb:


cheleb:
Hum, work on my machine ... ideas someone ?

cheleb:
Done, I have encountered some issues with resolvers. At home I'm using [coursier](https://github.com/alexarchambault/coursier). Coursier and Ivy (original?) resolver are not resolving the same way some transitive dependencies (hbase-common was not included with original resolver): Resulting class not found during compilation ... 

AFAIU it was a "maven issue" ... as this dependency appears twice in the pom.xml (as a "jar" and "test-jar".

(sorry for the CI usage, but it was long for me to understand this issue).

To "fix" this I had to add the "hbase-common" dependency.
 
IMHO this issue should be documented in "contribute.md"

hbase/src/main/scala/akka/hbase/internal/HBaseFlowStage.scala
```diff
@@ -0,0 +1,107 @@
...
+
+      setHandler(in, new InHandler {
+        override def onPush() = {
+          val msg = grab(in)
+          val r = Promise[A]
+
+          awaitingConfirmation.incrementAndGet()
+
+          Future {
+            table.put(settings.converter.put(msg))
```
johanandren: This is a blocking operation, what dispatcher is it run on? Should be the `akka.stream.blocking-io-dispatcher` or a custom blocking dispatcher for this stage

hbase/src/test/scala/akka/hbase/Person.scala
```diff
@@ -0,0 +1,6 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.hbase
+
+case class Person(id: Int, name: String)
```
johanandren: just here for the test, so put it in the same file as the test

project/Dependencies.scala
```diff
@@ -27,6 +27,14 @@ object Dependencies {
     )
   )
 
+  val HBase = Seq(
+    libraryDependencies ++= Seq(
+      "org.apache.hbase" % "hbase-client" % "1.2.3",
+      "org.apache.hbase" % "hbase-common" % "1.2.3",
+      "org.apache.hadoop" % "hadoop-common" % "2.7.3"
```
johanandren: The client doesn't pull these two dependencies in even though it needs them?

hbase/src/main/scala/akka/hbase/HBaseConverter.scala
```diff
@@ -0,0 +1,20 @@
...
+ */
+package akka.hbase
+
+import org.apache.hadoop.hbase.client.Put
+import org.apache.hadoop.hbase.util.Bytes
+
+import scala.language.implicitConversions
+
+trait HBaseConverter[T] {
+  def put(t: T): Put
```
johanandren: `convert` would be a better name as it doesn't really put anything

hbase/src/main/scala/akka/hbase/HBaseConverter.scala
```diff
@@ -0,0 +1,20 @@
...
+}
+
+object HBaseConverter {
+  object Implicits {
+    implicit def toBytes(string: String): Array[Byte] = Bytes.toBytes(string)
```
johanandren: I don't like this, implicit conversions are magic, I think this belongs in the user code, because it depends on how they want to transform strings to bytes (what character encoding they want to use for example).

project/Dependencies.scala
cheleb: Yep AFAIU, sbt default resolver (ivy?) fails with transient dependencies here ...

cheleb:
Applied simple changes, according to review of  @johanandren (thanks a lot).
Will check others asap (future, thread pool related).

docs/src/main/paradox/hbase.md
```diff
@@ -0,0 +1,48 @@
...
+
+###Flow usage 
+
+@@snip (../../../../hbase/src/test/scala/akka/stream/alpakka/hbase/HBaseStageTestSuite.scala) { #flow }
+
+###Sink usage
+
+@@snip (../../../../hbase/src/test/scala/akka/stream/alpakka/hbase/HBaseStageTestSuite.scala) { #sink }
+
+##java TBD
```
johanandren: Java samples should preferably be for each step as in the existing connectors, rather than as a separate section.

hbase/src/main/scala/akka/stream/alpakka/hbase/HBaseConverter.scala
```diff
@@ -0,0 +1,20 @@
...
+package akka.stream.alpakka.hbase
+
+import org.apache.hadoop.hbase.client.Put
+import org.apache.hadoop.hbase.util.Bytes
+
+import scala.language.implicitConversions
+
+trait HBaseConverter[T] {
+  def put(t: T): Put
+}
```
johanandren: As this is just a single function, couldn't this just be a `T => Put` parameter then, rather than a specific type?

hbase/src/main/scala/akka/stream/alpakka/hbase/HBaseConverter.scala
```diff
@@ -0,0 +1,20 @@
...
+
+object HBaseConverter {
+  object Implicits {
+    implicit def toBytes(string: String): Array[Byte] = Bytes.toBytes(string)
+  }
+
+}
```
johanandren: Still not quite sure why this is needed, could you elaborate?

hbase/src/main/scala/akka/stream/alpakka/hbase/HBaseSettings.scala
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.alpakka.hbase
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.hbase.{ HBaseConfiguration, TableName }
+
+case class HBaseSettings(conf: Configuration = HBaseConfiguration.create(), parallelism: Int, dispatcher: String = "") {
```
johanandren: `final` for all types that are not intended to be subtyped.

Add an object for each settings type with a `create` method mirroring the generated apply-methods (using Java types for collections, options etc) for the Java API

hbase/src/main/scala/akka/stream/alpakka/hbase/HBaseSettings.scala
```diff
@@ -0,0 +1,24 @@
...
+                   columnFamilies = columnFamilies,
+                   converter = converter,
+                   parallelism = parallelism,
+                   dispatcher = dispatcher)
+}
+
+case class HTableSettings[T](conf: Configuration,
+                             tableName: TableName,
+                             columnFamilies: Seq[String],
+                             converter: HBaseConverter[T],
```
johanandren: here is where it could just be `T => Put`

hbase/src/main/scala/akka/stream/alpakka/hbase/internal/HBaseFlowStage.scala
```diff
@@ -0,0 +1,107 @@
...
+  private val in = Inlet[A]("messages")
+  private val out = Outlet[Future[A]]("result")
+
+  override val shape = FlowShape(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = {
+
+    implicit val connection = connect(settings.conf)
+
+    val table: Table = getOrCreateTable(settings.tableName, settings.columnFamilies)
```
johanandren: both these two does blocking network logic, now it happens during materialization stopping the graph from complete materialization until the blocking call is completed, in general we rather try to isolate such logic inside of the graph stage logic and let the materialization complete and have the logic backpressure until a connection has been made. The blocking logic needs to be isolated on a separate dispatcher for this or it will starve the default dispatcher. (I think I mentioned this in a previous review)

hbase/src/main/scala/akka/stream/alpakka/hbase/internal/HBaseFlowStage.scala
```diff
@@ -0,0 +1,107 @@
...
+      override def postStop() = {
+        log.debug("Stage completed")
+        try {
+          table.close()
+          connection.close()
+          log.debug("Producer closed")
+        } catch {
+          case NonFatal(ex) => log.error("Problem occurred during producer close", ex)
+        }
+        super.postStop()
```
johanandren: Post stop is empty in `GraphStageLogic`, so no need to call super here

hbase/src/main/scala/akka/stream/alpakka/hbase/internal/HBaseFlowStage.scala
```diff
@@ -0,0 +1,107 @@
...
+
+      val checkForCompletionCB = getAsyncCallback[Unit] { _ =>
+        checkForCompletion()
+      }
+
+      override def postStop() = {
+        log.debug("Stage completed")
+        try {
+          table.close()
+          connection.close()
```
johanandren: shouldn't these be separate tries, if table.close throws this will leave the connection unclosed?

hbase/src/main/scala/akka/stream/alpakka/hbase/internal/HBaseFlowStage.scala
```diff
@@ -0,0 +1,107 @@
...
+
+      setHandler(in, new InHandler {
+        override def onPush() = {
+          val msg = grab(in)
+          val r = Promise[A]
+
+          awaitingConfirmation.incrementAndGet()
+
+          Future {
+            table.put(settings.converter.put(msg))
```
johanandren: This is a blocking network op, it should run on a dedicated dispatcher, not the global default one as done now.

hbase/src/main/scala/akka/stream/alpakka/hbase/internal/HBaseFlowStage.scala
```diff
@@ -0,0 +1,107 @@
...
+          }.map { r =>
+            if (awaitingConfirmation.decrementAndGet() == 0 && inIsClosed)
+              checkForCompletion()
```
johanandren: this looks at stage mutable state from a different thread so it is not safe, use the `asyncCallback` api to communicate with the stage from another thread instead, you can read about it in the docs here: http://doc.akka.io/docs/akka/2.4.14/scala/stream/stream-customize.html#using-asynchronous-side-channels

hbase/src/main/scala/akka/stream/alpakka/hbase/internal/HBaseFlowStage.scala
```diff
@@ -0,0 +1,107 @@
...
+import akka.stream.{ Attributes, FlowShape, Inlet, Outlet }
+import org.apache.hadoop.hbase.client.Table
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.{ Future, Promise }
+import scala.util.control.NonFatal
+import scala.util.{ Failure, Success, Try }
+
+private[hbase] class HBaseFlowStage[A](settings: HTableSettings[A])
+    extends GraphStage[FlowShape[A, Future[A]]]
```
johanandren: This signature is weird, it should be `FlowShape[A, A]` or else you are actually killing backpressure here, as you can emit any number of futures that has not yet completed.

cheleb: Hum, definitively I took "reactive kafka" as example, but in reactive kafka "sender" has a buffer and blocks when buffer is full !
Sorry about that.

patriknw: it's used with `mapAsync` and the parallelism is the limiting factor, so I think it's fine

hbase/src/main/scala/akka/stream/alpakka/hbase/internal/HBaseFlowStage.scala
patriknw: There is no guarantee that the `table.put` will be performed in the same order as the stream elements since this is executed in a Future. That is probably not acceptable.

That means that it's probably a lot easier to do the blocking operations directly in the stage and run the stage with the `IODispatcher`.

hbase/src/main/scala/akka/stream/alpakka/hbase/internal/HBaseFlowStage.scala
johanandren: Ah, I didn't think about the fact that this is internal api and that the public api is the factory, that indeed does `mapAsync`. I agree with Patrik, this is fine!

patriknw: but I don't think spawning futures is fine, so then it will be `[A, A]`

docs/src/main/paradox/hbase.md
cheleb: I have started to write java stuff, but now that converter is a function, how can I convert a java function to a scala one ?

hbase/src/main/scala/akka/stream/alpakka/hbase/internal/HBaseFlowStage.scala
cheleb: If I run the stage in IODispatcher and move the blocking operattions in the GraphStageLogic it is ok? Or a have to run this init even more async ?

hbase/src/main/scala/akka/stream/alpakka/hbase/internal/HBaseFlowStage.scala
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        override def onUpstreamFailure(ex: Throwable) = {
+          completionState = Some(Failure(ex))
+          checkForCompletion()
+        }
+
+      })
+
+      def checkForCompletion() =
```
johanandren: The original behaviour of GraphStageLogic is to complete `onUpstreamFinish` and to fail `onUpstreamFailure` so this is now just a more convoluted way of doing the same, I think you can remove this method, the `completionState` and the overrides.

hbase/src/main/scala/akka/stream/alpakka/hbase/internal/HBaseFlowStage.scala
```diff
@@ -0,0 +1,87 @@
...
+
+      implicit val connection = connect(settings.conf)
+
+      val table: Table = getOrCreateTable(settings.tableName, settings.columnFamilies)
+
+      var completionState: Option[Try[Unit]] = None
+
+      setHandler(out, new OutHandler {
+        override def onPull() =
+          tryPull(in)
```
johanandren: As we do not do any detaching in this stage now, it can just do a pull here (we will never get two pulls in a row)

cheleb:
Done, I've squashed the review commits.
Thanks for your patience.

hbase/src/main/scala/akka/stream/alpakka/hbase/internal/HBaseCapabilities.scala
```diff
@@ -0,0 +1,32 @@
...
+      else {
+        val tableDescriptor: HTableDescriptor = new HTableDescriptor(tableName)
+        columnFamilies.foreach { cf =>
+          tableDescriptor.addFamily(new HColumnDescriptor(cf))
+        }
+        admin.createTable(tableDescriptor)
+        //log.info(s"Table $tableName created with cfs: $columnFamilies.")
+        connection.getTable(tableName)
+      }
+    admin.close()
```
patriknw: should this be done in a `finally` (to do it in case of exceptions also)?

hbase/src/main/scala/akka/stream/alpakka/hbase/HBaseSettings.scala
```diff
@@ -0,0 +1,13 @@
...
+ */
+package akka.stream.alpakka.hbase
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.hbase.TableName
+import org.apache.hadoop.hbase.client.Put
+
+final case class HTableSettings[T](conf: Configuration,
+                                   tableName: TableName,
+                                   columnFamilies: Seq[String],
```
patriknw: `Seq` can be mutable, so we prefer to use:

```scala

columnFamilies: immutable.Seq[String]
```

hbase/src/main/scala/akka/stream/alpakka/hbase/HBaseSettings.scala
```diff
@@ -0,0 +1,13 @@
...
+                                   converter: T => Put)
```
patriknw: `HTableSettings` is used as method parameter in `javadsl` and the scala function is not java friendly in Scala 2.11

I suggest that you add a java friendly factory method in the companion of `HTableSettings`, or use two different `scaladsl.HTableSettings` and `javadsl.HTableSettings`

hbase/src/main/scala/akka/stream/alpakka/hbase/internal/HBaseFlowStage.scala
```diff
@@ -0,0 +1,65 @@
...
+
+      })
+
+      override def postStop() = {
+        log.debug("Stage completed")
+        try {
+          table.close()
+          log.debug("table closed")
+        } catch {
+          case NonFatal(ex) => log.error("Problem occurred during producer table close", ex)
```
patriknw: the `ex` should be the first parameter

hbase/src/main/scala/akka/stream/alpakka/hbase/internal/HBaseFlowStage.scala
```diff
@@ -0,0 +1,65 @@
...
+        }
+        try {
+          connection.close()
+          log.debug("connection closed")
+        } catch {
+          case NonFatal(ex) => log.error("Problem occurred during producer connection close", ex)
```
patriknw: the `ex` should be the first parameter

hbase/src/main/scala/akka/stream/alpakka/hbase/internal/HBaseFlowStage.scala
```diff
@@ -0,0 +1,65 @@
...
+  override protected def initialAttributes: Attributes =
+    Attributes.name("HBaseFLow").and(ActorAttributes.dispatcher("akka.stream.default-blocking-io-dispatcher"))
+
+  private val in = Inlet[A]("messages")
+  private val out = Outlet[A]("result")
+
+  override val shape = FlowShape(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) with StageLogging {
```
patriknw: when using `StageLogging` you will get a slightly better log source name if you

```scala
override protected def logSource = classOf[HBaseFlowStage]
```

cheleb:
AFAIU I had to:
* handle timeout of connexion to hbase cluster
* use lazy (on table) to be able to use logger from StageLogging.

patriknw:
@johanandren I think this is ready for merge. Do you want to take a final look?

johanandren:
Refs #35

PR #76, opened 2016-11-21
Update to stable sbt-dynver

2m:


project/plugins.sbt
```diff
@@ -1,6 +1,6 @@
 addSbtPlugin("de.heikoseeberger"                 % "sbt-header"       % "1.6.0")
 addSbtPlugin("com.geirsson"                      % "sbt-scalafmt"     % "0.4.10")
-addSbtPlugin("com.dwijnand"                      % "sbt-dynver"       % "1.0.0+15-3655d014")
+addSbtPlugin("com.dwijnand"                      % "sbt-dynver"       % "1.1.0")
```
patriknw: does that mean that we can remove the 2m resolver?

patriknw:
Lgtm

Issue #48, opened 2016-11-13
JMS connector


johanandren:
Likely not the most exciting connector to write, but I think it would be interesting to a lot of Java people needing to integrate JEE things.

myfear:
Anybody who want's to contribute here can use my WildFly Swarm based example as an endpoint test. https://github.com/myfear/wfs-jms

PR #80, opened 2016-11-23
Add JMS connectors query #48

larousso:


jms/src/main/scala/akka/stream/alpakka/jms/javadsl/JmsSink.scala
```diff
@@ -0,0 +1,20 @@
...
+
+import akka.NotUsed
+import akka.stream.alpakka.jms.{ JmsSinkSettings, JmsSinkStage }
+
+object JmsSink {
+
+  /**
+   * Java API: Creates an [[JmsSink]]
+   *
+   * @param jmsSinkSettings the jms connexion settings
```
johanandren: `connexion` -> `connection`, actually I think the type convey this information already, so skip the `@param` and `@return` annotations here, they do not  add any information that isn't already in the signature. Same for the other factory methods below.

jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsConnectorsTest.java
```diff
@@ -0,0 +1,242 @@
...
+
+            CompletionStage<List<String>> result = JmsSource
+                    .textSource(JmsSourceSettings
+                            .create(connectionFactory)
+                            .withQueue("test")
+                            .withBufferSize(1)
+                    )
+                    .mapAsync(1, e ->
+                            after(FiniteDuration.create(1, TimeUnit.SECONDS), system.scheduler(), system.dispatcher(), CompletableFuture.completedFuture(e))
+                    )
```
johanandren: would maybe be more clear using `.throttle` instead

jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsConnectorsTest.java
```diff
@@ -0,0 +1,242 @@
...
+                    .take(in.size()).runWith(Sink.seq(), materializer);
+
+            assertEquals(in, result.toCompletableFuture().get(15, TimeUnit.SECONDS));
```
johanandren: I don't think we need to try 15 of them to verify that backpressure works, already after 2 elements we can be sure of that? (tests that take long time are nice to avoid)

jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsConnectorsTest.java
```diff
@@ -0,0 +1,242 @@
...
+        });
+    }
+
+    @Test
+    public void deconnexionShouldFail() throws Exception {
```
johanandren: `deconnexion` -> `disconnection`

johanandren: Also, the test cases that are not part of the docs are fine to just have in one version, java or scala, no need to run the exact same test from both languages.

jms/src/test/scala/akka/stream/alpakka/jms/JmsSpec.scala
```diff
@@ -0,0 +1,48 @@
...
+    with Matchers
+    with BeforeAndAfterAll
+    with BeforeAndAfterEach
+    with ScalaFutures {
+
+  implicit val system = ActorSystem(this.getClass.getSimpleName)
+  implicit val materializer = ActorMaterializer()
+
+  override protected def afterAll(): Unit =
+    system.terminate()
```
johanandren: `TestKit.shutdown(system)` will terminate and block with a timeout, use that instead

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
```diff
@@ -0,0 +1,129 @@
...
+
+  def fail = getAsyncCallback[JMSException](e => failStage(e))
+
+  def openQueue(): Unit = {
+    val factory: ConnectionFactory = jmsSettings.connectionFactory
+    try {
+      jmsConnection = Option(jmsSettings.credentials match {
+        case Some(Credentials(username, password)) => factory.createConnection(username, password)
+        case _ => factory.createConnection()
+      })
```
larousso: How do you get the right dispatcher inside the graphstage ?

johanandren: It's a little bit messy as there is no predefined API way to do it, but you can get it from the underlying:

```
val blockingDispatcher = materializer match {
        case m: ActorMaterializer => m.system.dispatchers.lookup("akka.stream.default-blocking-io-dispatcher")
        case x => throw new IllegalArgumentException(s"Stage only works with the ActorMaterializer, was: $x")
      }
```

Just running the entire stage on the blocking dispatcher is easier and is done by overriding `initialAttributes` returning `ActorAttributes.Dispatcher("akka.stream.default-blocking-io-dispatcher")` on the `GraphStage`.

johanandren: If the entire stage runs in the blocking dispatcher it is easier to write as it can block whenever it wants without problems, it will however always introduce an asynchronous boundary, as the other stages in streams it is used in will not run on that same dispatcher (so same as `.async`), if just some parts of the internal uses future/async-callback you will have the boundary sort of inside the stage.

What is best is different from case to case, but as you already deal with a separate thread doing interacting with JMS and then handing over the elements to the stage and vice versa I think just running the small parts on a suitable dispatcher is a good solution.

johanandren:
Also, needs a rebase now when other things got merged, should be painless.

jms/src/main/scala/akka/stream/alpakka/jms/JmsSinkStage.scala
```diff
@@ -0,0 +1,51 @@
...
+          initProducer.invoke(producer)
+        }
+
+      setHandler(in, new InHandler {
+        override def onPush(): Unit = {
+          val elem = grab(in)
+          Future {
+            val textMessage: TextMessage = jmsSession.session.createTextMessage(elem)
+            jmsProducer.send(textMessage)
+          }.foreach(_ => askNext.invoke(Done))
```
larousso: I'm not sure of what is the good behavior. If an error occurs should I ignore it ? 

Like this : 

```
 Future {
            val textMessage: TextMessage = jmsSession.session.createTextMessage(elem)
            jmsProducer.send(textMessage)
          }.onComplete {
            case Success(_) => askNext.invoke(Done)
            case Failure(e) =>
              log.error(e, "Error sending message")
              askNext.invoke(Done)
          }
```

I will push the rest of the PR, sorry I didn't merge all I've done.

johanandren: I think you should fail the stage just if sending a message throws, like you do in the `JmsSourceStage`, so using a callback but passing the exception back and then `failStage` with it.

build.sbt
```diff
@@ -64,6 +58,6 @@ lazy val docs = project
       "extref.java-api.base_url" -> "https://docs.oracle.com/javase/8/docs/api/index.html?%s.html",
       "extref.paho-api.base_url" -> "https://www.eclipse.org/paho/files/javadoc/index.html?%s.html",
       "scaladoc.akka.base_url" -> s"http://doc.akka.io/api/akka/${Dependencies.AkkaVersion}",
-      "scaladoc.akka.stream.alpakka.base_url" -> s"http://developer.lightbend.com/docs/api/alpakka/${version.value}"
+      "scaladoc.akka.stream.alpakka.base_url" -> s"http://doc.akka.io/api/alpakka/${version.value}"
```
patriknw: why this change, the right url is http://developer.lightbend.com/docs/api/alpakka/0.2/

docs/src/main/paradox/connectors.md
```diff
@@ -5,8 +5,10 @@
 * [AMQP Connector](amqp.md)
 * [Cassandra Connector](cassandra.md)
 * [File Connectors](file.md)
+* [JMS Connectors](jms.md)
 * [MQTT Connector](mqtt.md)
 * [FTP Connector](ftp.md)
+* [JMS Connectors](jms.md)
```
patriknw: included twice

docs/src/main/paradox/jms.md
```diff
@@ -0,0 +1,123 @@
...
+First define a jms `ConnectionFactory` depending on the implementation you're using. Here we're using Active MQ. 
+
+Scala
+: @@snip (../../../../jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsConnectorsSpec.scala) { #connection-factory }
+
+Java
+: @@snip (../../../../jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsConnectorsTest.java) { #connection-factory }
+
+
+Create a sink, that accepts and forwards @scaladoc[String](java.lang.String)s to the JMS server.
```
patriknw: linking to String looks overkill, will it even work?

I'd use

```
`String`
```

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
```diff
@@ -0,0 +1,87 @@
...
+      })
+      connection.start()
+      val session = connection.createSession(false, jms.Session.CLIENT_ACKNOWLEDGE)
+      val dest = jmsSettings.destination match {
+        case Some(Queue(name)) => session.createQueue(name)
+        case Some(Topic(name)) => session.createTopic(name)
+        case _ => throw new IllegalArgumentException("Destination is missing")
+      }
+      onSession.invoke(JmsSession(connection, session, dest))
+    }
```
patriknw: perhaps use `.onFailure` in case an exception is thrown inside that Future

jms/src/main/scala/akka/stream/alpakka/jms/JmsSinkStage.scala
```diff
@@ -0,0 +1,57 @@
...
+
+      override private[jms] def onSessionOpened(): Unit =
+        jmsSession.createProducer().foreach { producer =>
+          initProducer.invoke(producer)
+        }
+
+      setHandler(in, new InHandler {
+        override def onPush(): Unit = {
+          val elem = grab(in)
+          Future {
```
patriknw: The alternative would be to run the stage with the `default-blocking-io-dispatcher` and do the blocking calls directly in the stage. Can `send` block for a long time, making the stage unresponsive (to cancelation) for too long, or what is the motivation for delegating the blocking to a Future?

Opinions, @johanandren @drewhk ?

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  override def shape: SourceShape[Message] = SourceShape[Message](out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) with JmsConnector with StageLogging {
+
+      override private[jms] def jmsSettings = settings
+
+      private val bufferSize = settings.bufferSize
+      private val queue = mutable.Queue[Message]()
```
patriknw: just to be nice to GC (and in case the stage is still referenced from somewhere) you could clear the queue in `postStop`

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
```diff
@@ -0,0 +1,82 @@
...
+      }
+
+      override private[jms] def onSessionOpened(): Unit =
+        jmsSession.createConsumer().onComplete {
+          case Success(consumer) =>
+            consumer.setMessageListener(new MessageListener {
+              override def onMessage(message: Message): Unit = {
+                backpressure.acquire()
+                try {
+                  message.acknowledge()
```
patriknw: Here is another typical example where we could use something like the `Committable` in akka-stream-kafka. That is for a later PR. Makes it much more complicated, but very useful. I have created issue https://github.com/akka/alpakka/issues/104 for this.

project/Dependencies.scala
```diff
@@ -52,16 +41,11 @@ object Dependencies {
...
     libraryDependencies ++= Seq(
-      "commons-net"           % "commons-net"          % "3.5",             // ApacheV2
-      "com.jcraft"            % "jsch"                 % "0.1.54",          // BSD-style
-      "org.apache.ftpserver"  % "ftpserver-core"       % "1.0.6"    % Test, // ApacheV2
-      "org.apache.sshd"       % "sshd-core"            % "1.3.0"    % Test, // ApacheV2
-      "com.google.jimfs"      % "jimfs"                % "1.1"      % Test, // ApacheV2
-      "org.slf4j"             % "slf4j-api"            % "1.7.21"   % Test, // MIT
-      "ch.qos.logback"        % "logback-classic"      % "1.1.7"    % Test, // Eclipse Public License 1.0
-      "ch.qos.logback"        % "logback-core"         % "1.1.7"    % Test  // Eclipse Public License 1.0
+      "javax.jms" % "javax.jms-api" % "2.0.1", // CDDL + GPLv2
```
patriknw: We are against anything GPL (yeah, I know this is standard JEE API) and the safe bet would be to define this as "provided" and document that user must add the dependency. Spring is also doing that: https://mvnrepository.com/artifact/org.springframework/spring-jms/4.3.4.RELEASE

build.sbt
larousso: Error during merge I will fix it.

docs/src/main/paradox/connectors.md
larousso: Error during merge I will fix it.

docs/src/main/paradox/jms.md
larousso: Ok

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
larousso: The `message.acknowledge()` must be called in the thread where onMessage is called. It make this really hard to implement with commitable ?

jms/src/main/scala/akka/stream/alpakka/jms/JmsSinkStage.scala
larousso: In this case the motivation is just to have something coherent between source and sink. 
I think I have a couple of option here : 
- I run sink in the `default-blocking-io-dispatcher` but opening connection, session ... stay in a future to keep the code mutualized
- I open connection, session ... in a blocking way in sink and in a future in source. 
- I open connection, session ... in a blocking way and I run both sink and source in the `default-blocking-io-dispatcher` 

Delegating to a future force the user to use `ActorMaterializer`, may be it's not so good ?

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
```diff
@@ -0,0 +1,87 @@
...
+                    handleError.invoke(e)
+                }
+              }
+            })
+          case Failure(e) =>
+            settings.destination match {
+              case Some(Queue(name)) =>
+                log.error(e, "Error creating consumer on queue {}", name)
+              case Some(Topic(name)) =>
+                log.error(e, "Error creating consumer on topic {}", name)
```
johanandren: put this information in a new exception together with the original exception and then just pass that to failStage, streams will log that exception so no need to log it manually here.

jms/src/main/scala/akka/stream/alpakka/jms/JmsSinkStage.scala
johanandren: There actually seems to be a more async version `MessageProducer.send(Destination, Message, CompletionListener)` – maybe that could be used instead?

larousso: I've tried to implements and I realize that activemq was not implementing JMS 2. 
I have the feeling that the JMS 2 is not implemented a lot.
I could use the async way to send message with the risk to make the sink not usable for most of the users or downgrade JMS to 1.1 and keep the blocking way to send messages. 

What do you think ?

patriknw: @larousso Thanks for checking. We should absolutely have support for JMS 1.1. We could add another variant (probably another stage) that makes use of JMS 2, but that can be done later in a separate PR, I think.

larousso: The job is already done on my local branch. I've downgrade to JMS 1.1, the source is using future for establishing a connexion and the rest is done using the jms asnyc listener. 
On the sink, all is done in a blocking way using the `default-blocking-io-dispatcher` as attribute like you have suggested before. 
I will push that tonight .

patriknw: Excellent, looking forward to have JMS support in Alpakka. Will absolutely release as soon it is completed.

johanandren:
Seems to fail with a timeout now, maybe up the test timeouts a bit?

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  private def onSession =
+    getAsyncCallback[JmsSession](session => {
+      jmsSession = session
+      onSessionOpened()
+    })
+
+  private[jms] def initSessionAsync() = {
+    ec = materializer match {
+      case m: ActorMaterializer => m.system.dispatchers.lookup("akka.stream.default-blocking-io-dispatcher")
```
patriknw: One important nitpick; this should be taken from the Attributes so that it can be overridden.

In `createLogic` you can grab the dispatcher id with something like `inheritedAttributes.get[ActorAttributes.Dispatcher](Dispatcher("akka.stream.default-blocking-io-dispatcher"))`

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
```diff
@@ -0,0 +1,85 @@
...
+        override def onPull(): Unit =
+          if (queue.nonEmpty) {
+            pushMessage(queue.dequeue())
+          }
+      })
+
+      override def postStop(): Unit = {
+        queue.clear()
+        Option(jmsSession).foreach(_.closeSessionAsync().onFailure {
+          case e => fail.invoke(e)
```
patriknw: There is no point in calling a callback here, since the stage is already stopped. There is not much that can be done about close failures, so it's fine to catch and ignore

jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsConnectorsTest.java
```diff
@@ -0,0 +1,182 @@
...
+        JavaTestKit.shutdownActorSystem(system);
+    }
+
+    private final Random randomPort  = new Random();
+
+    private void withServer(ConsumerChecked<Context> test) throws Exception {
+        BrokerService broker = new BrokerService();
+        broker.setPersistent(false);
+        String host = "localhost";
+        Integer port = randomPort.nextInt(500) + 6500;
```
patriknw: use `akka.testkit.SocketUtil.temporaryServerAddress` instead

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
drewhk: +1

patriknw:
@larousso Thanks for incorporating the feedback and sorry for delay. Now it just need a fix for the merge conflict, and then I will merge and release.

patriknw:
Refs #48

larousso:
Thanks for your patience!

patriknw:
Thank you for an excellent PR of a very important piece of Alpakka. I'll release it today.

PR #89, opened 2016-11-26
=s3 initial add of WireMock tests #86

agolubev:
- localhost certificates for https added
- S3 exception added
- some refactoring
- temporary added proxy mechanism while it's not in akka-http

Ref #86


agolubev:
Basically WIrMock can be started as Docker service so let me know if should look into this.

s3/src/main/scala/akka/stream/alpakka/s3/S3Settings.scala
```diff
@@ -26,15 +31,19 @@ object S3Settings {
   /**
    * Create [[S3Settings]] from a Config subsection.
    */
-  def apply(config: Config): S3Settings =
+  def apply(config: Config): S3Settings = {
+    val proxyConfig = if (config.hasPath("proxy")) config.getConfig("proxy") else ConfigFactory.empty()
```
patriknw: The problem with not defining "all" properties in `reference.conf` is that it's difficult for the user to know about their existence. Yeah, it could be documented, but I think it's nice to have them in reference.conf, even if it means having a special value for undefined.

could we use something like:
```
proxy {
  # Hostname of the proxy. If undefined ("") proxy is not enabled.
  host = ""
  port = 8000
}

**Comment by patriknw** [comment]
It would be nice to start with the documentation for the S3 connector, so that it could be incrementally updated with features like this. I created issue https://github.com/akka/alpakka/issues/103

**Comment by agolubev** [comment]
MqttSourceSpec was failed.
PR is ready for review.

**Comment by agolubev** [comment]
Will start with document as soon as have some API refactoring(remove AWSCredntials and add S3 as extension)

**Comment by agolubev** [comment]
Rebased to fix merge conflicts

**Comment by agolubev** [comment]
build failed reason
```
[info] MqttSourceSpec:
[info] mqtt source
[info] - should receive a message from a topic (704 milliseconds)
[info] - should receive messages from multiple topics *** FAILED ***
```

patriknw:
I have started a new travis build

patriknw:
@agolubev tests green, this is ready to merged, right?

agolubev:
@patriknw yeah - it's ready. Can do sqaush if you want.

patriknw:
thanks, squashed with UI

PR #91, opened 2016-11-28
format s3 code

patriknw:


s3/src/test/scala/akka/stream/alpakka/s3/auth/SigningKeySpec.scala
```diff
@@ -16,12 +16,45 @@ class SigningKeySpec extends FlatSpec with Matchers {
...
+                                      26,
+                                      94,
+                                      134,
+                                      218,
+                                      110,
+                                      211,
+                                      193,
+                                      84,
+                                      164,
+                                      185).map(_.toByte)
```
patriknw: this looks too bad, strange that it breaks that even if I write it as short lines:

```scala
    val expected: Array[Byte] = Array(
      196, 175, 177, 204, 87, 113, 216, 113, 118, 58, 57, 62, 68, 183, 
      3, 87, 27, 85, 204, 40, 66, 77, 26, 94, 134, 218, 110, 211, 193, 
      84, 164, 185).map(_.toByte)
```

TimothyKlim: The only way to avoid it is to use `format: OFF/ON`:
```scala
// format: OFF
val expected: Array[Byte] = Array(
      196, 175, 177, 204, 87, 113, 216, 113, 118, 58, 57, 62, 68, 183, 
      3, 87, 27, 85, 204, 40, 66, 77, 26, 94, 134, 218, 110, 211, 193, 
      84, 164, 185).map(_.toByte)
// format: ON
```

patriknw: ok, I added that

Issue #95, opened 2016-11-29
AMQP Connect Should Support Different SSL protocols


irishshagua:
The default Rabbit MQ Connection Factory protocol for the version of the rabbit client in the dependencies (3.6.1) is TLSv1. This is not advised as per the [RabbitMQ site](https://www.rabbitmq.com/ssl.html) and has been updated in their 4.x series as per [here](https://github.com/rabbitmq/rabbitmq-java-client/issues/139)

The current AMQP Config doesn't offer the ability to select an alternate protocol version but looks like it could be added with a very minor change. I'm not sure if ye have plans to upgrade the rabbit client version (I havent actually used their 4.x client at all and am not sure how battle tested it is yet) so I'm happy to throw in a PR to add the extra AMQP Config item if people agree with the issue.

PR #96, opened 2016-11-29
Add AMQP Config for SSL protocol

irishshagua:
Add SSL protocol version override to the AmqpConnectionDetails case class. This is for issue https://github.com/akka/alpakka/issues/95


patriknw:
Thanks @irishshagua

patriknw:
Refs https://github.com/akka/alpakka/issues/95

PR #106, opened 2016-12-02
[fix] Same order of arguments for s3 multipartUpload

TimothyKlim:
In previous PR I forget to fix S3Client arguments for same consistency.


ktoso:
LGTM :-) 

(Why "fastfix" :-) Fix is a fix)

PR #108, opened 2016-12-02
Camel DSL for Akka Streams

### LINKED ISSUE #74: Generic Camel endpoint connector
Provide a generic [Camel endpoint](http://camel.apache.org/components.html) connector that is compatible with the back-pressure model of Reactive Streams. Camel endpoints that are compatible with back-pressure are, for example, polling consumers, batch consumers and (async) producers. Not compatible are consumers that push messages downstream independent of demand, for example. A generic connector would make a large number of Camel endpoints re-usable for Akka Streams (and will hopefully increase the adoption of Alpakka 😃)

I've done some experimental work into this direction with an [FS2 - Camel integration](https://github.com/krasserm/streamz#combinators-for-apache-camel) in the Streamz project (although the consumer side doesn't support back-pressure yet). This integration internally uses [akka-camel](http://doc.akka.io/docs/akka/current/scala/camel.html). Using the [FS2 - Akka Streams interface](https://github.com/krasserm/streamz#combinators-for-akka-stream) one can already use Camel endpoints with Akka Streams in a generic way. Anyway, to have an integration with small runtime overhead, there should be an integration of Akka Streams with the Camel API directly.

Thoughts?

krasserm:
Camel DSL for Akka Streams

- closes #74

### DIFF
--- a/docs/src/main/paradox/external-connectors.md
+++ b/docs/src/main/paradox/external-connectors.md
@@ -37,7 +37,11 @@ TODO: Add links to more connectors. [Contributions](https://github.com/akka/alpa
 
 ## FS2
 
-[An Akka Streams - Functional Streams for Scala (FS2) interface](https://github.com/krasserm/streamz#combinators-for-akka-stream)
+[Akka Streams - FS2 converters](https://github.com/krasserm/streamz#stream-converters)
+
+## Camel
+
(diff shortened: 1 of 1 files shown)


krasserm:
Just updated the commit to mention Camel (and FS2) directly instead of Streamz 😃

Issue #111, opened 2016-12-04
Allow Mqtt tests to run in parallel.


filosganga:
The Mqtt tests can run in parallel as long as we use a different port for each test.

2m:
Fixed in #111

PR #112, opened 2016-12-04
Implement #111 parallel Mqtt tests

filosganga:
- Add trait to discover an available TCP port
 - Change the Mqtt tests to make use of the autodiscovery TCP port trait
 - Update sbt settings to allow parallel Mqtt tests execution

### DIFF
--- a/mqtt/src/test/java/akka/stream/alpakka/mqtt/javadsl/MqttSourceTest.java
+++ b/mqtt/src/test/java/akka/stream/alpakka/mqtt/javadsl/MqttSourceTest.java
@@ -4,6 +4,7 @@
 package akka.stream.alpakka.mqtt.javadsl;
 
 import akka.stream.alpakka.mqtt.*;
+import io.moquette.BrokerConstants;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
@@ -24,6 +25,7 @@ import io.moquette.server.Server;
 
 import org.eclipse.paho.client.mqttv3.persist.MemoryPersistence;
 
+import java.nio.file.Files;
 import java.util.*;
 import java.util.concurrent.*;
 import java.util.stream.*;
@@ -32,8 +34,9 @@ public class MqttSourceTest {
 
   static ActorSystem system;
   static Materializer materializer;
-
   static Server server;
+  static int serverPort;
+  static int serverWebSocketPort;
 
   public static Pair<ActorSystem, Materializer> setupMaterializer() {
     //#init-mat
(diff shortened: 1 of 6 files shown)


mqtt/src/test/java/akka/stream/alpakka/mqtt/javadsl/MqttSourceTest.java
```diff
@@ -45,12 +47,19 @@
 
   @BeforeClass
   public static void setup() throws Exception {
+    serverPort = AvailablePort$.MODULE$.nextPort();
+    serverWebSocketPort = AvailablePort$.MODULE$.nextPort();
```
2m: This is probably a left over. `AvailablePort.nextPort();` should work fine here.

filosganga: I am not familiar with Java code calling Scala code I will take a look if `AvailablePort.nextPort();` will work, thanks

filosganga:
Tests are failing as the server are trying to write in the same folder, I need to use a different folder for each instance.

mqtt/src/test/java/akka/stream/alpakka/mqtt/javadsl/MqttSourceTest.java
```diff
@@ -45,12 +47,19 @@
...
+
     final Pair<ActorSystem, Materializer> sysmat = setupMaterializer();
     system = sysmat.first();
     materializer = sysmat.second();
 
+    Properties serverProperties = new Properties();
+    serverProperties.setProperty(BrokerConstants.PORT_PROPERTY_NAME, String.valueOf(serverPort));
+    serverProperties.setProperty(BrokerConstants.WEB_SOCKET_PORT_PROPERTY_NAME, String.valueOf(serverWebSocketPort));
+
```
2m: According the Travis CI test failures we will have to make `BrokerConstants.PERSISTENT_STORE_PROPERTY_NAME` unique as well.

2m:
An alternative would be to run mosquitto (or some other mqtt broker) in a docker container (for example https://github.com/toke/docker-mosquitto), as we do for all other integrations.

filosganga:
@2m I think, running mosquitto in docker, will be consistent with the other tests. I am happy going with it.

mqtt/src/test/java/akka/stream/alpakka/mqtt/javadsl/MqttSourceTest.java
filosganga: Does not seem I can call `AvailablePort.nextPort()` from Java

filosganga:
@2m I ended up not using the docker container for this test, as it calls a server method to publish messages and a test need the server to stop.

filosganga:
@2m Can you review it again please?

2m:
I have implemented MqttSink which allows to publish messages to mqtt broker and unblocks migration to the broker running in docker container. @filosganga, if you could review #131 I would really appreciate it.

filosganga:
The #131 supersede this one.

PR #114, opened 2016-12-09
expose cleanSession and will mqtt connection options

prettynatty:


2m:
Hi @prettynatty. Note that you have only modified the connectors data structure but have not wired it up to the [`org.eclipse.paho.client.mqttv3.MqttConnectOptions`](https://www.eclipse.org/paho/files/javadoc/org/eclipse/paho/client/mqttv3/MqttConnectOptions.html) which is being setup [here](https://github.com/akka/alpakka/blob/master/mqtt/src/main/scala/akka/stream/alpakka/mqtt/Mqtt.scala#L144).

After wiring it up, it would be great to write a test, that checks if the last will is successfully sent out.

prettynatty:
Hi @2m ,
I've wired options to MqttConnectOptions and written the test.
But something is going wrong here. Will message is not being set for some reason. I suppose that stopping source using KillSwitch does not close connection, that's why WIll message is not being set at the moment of second subscription. Am I right? If yes, there might be a bug in stream completion handling.

2m:
Accorgin to [this](http://www.hivemq.com/blog/mqtt-essentials-part-9-last-will-and-testament) LWT message is sent from the broker when one of the following happens:
* An I/O error or network failure is detected by the server.
* The client fails to communicate within the Keep Alive time.
* The client closes the network connection without sending a DISCONNECT packet first.
* The server closes the network connection because of a protocol error.

Therefore we will not be able to make broker issue LWT message by closing down the stream gracefully. I think will will need to do a TCP proxy and make MqttSource go through it. Then kill the proxy which should terminate the connection between the client and broker and issue the LWT message.

Proxy could be a simple stream like so:
```
val killSwitch = Tcp()
  .bind("127.0.0.1", 1337)
  .runWith(Sink.head)
  .map(
    _.handleWith(Tcp().outgoingConnection(brokerUrl.getHost, brokerUrl.getPort).viaMat(KillSwitches.single)(Keep.right))
  )
```

prettynatty:
Well, I've written will and persistend session specs but still no luck.
It seems that killing proxy stream does not actually close TCP connection. I've added log4j.xml to resources and see the following output:

```
NFO  NettyMQTTHandler - Received a message of type SUBSCRIBE
DEBUG ProtocolProcessor - SUBSCRIBE client <test-client> packetID 1
INFO  ClientSession - <test-client> subscribed to topicFilter <topic1> with QoS 1 - LEAST_ONE
DEBUG MapDBSessionsStore - subscriptions_test-client: {}
DEBUG MapDBSessionsStore - addNewSubscription invoked with subscription [filter:topic1, cliID: test-client, qos: LEAST_ONE, active: true]
DEBUG ProtocolProcessor - SUBACK for packetID 1
DEBUG SubscriptionsStore - root ref io.moquette.spi.impl.subscriptions.TreeNode@10253535, original root was io.moquette.spi.impl.subscriptions.TreeNode@88d148f
DEBUG MapDBMessagesStore - searchMatching scanning all retained messages, presents are 0

======
connection closed here, sleeping 1 min
======

DEBUG ProtocolProcessor - route2Subscribers republishing to existing subscribers that matches the topic topic1
DEBUG MapDBMessagesStore - storePublishForFuture store evt PublishEvent{m_msgID=2, clientID='test-client', m_retain=true, m_qos=LEAST_ONE, m_topic='topic1'}
DEBUG MapDBSessionsStore - subscriptions_test-client: {topic1=[filter:topic1, cliID: test-client, qos: LEAST_ONE, active: true]}
DEBUG MapDBSessionsStore - subscriptions_mosqsub/22611-Kuznetsov: {topic1=[filter:topic1, cliID: mosqsub/22611-Kuznetsov, qos: MOST_ONE, active: true]}
DEBUG MapDBSessionsStore - subscriptions_proxied-test-client: {topic1=[filter:topic1, cliID: proxied-test-client, qos: MOST_ONE, active: true]}
DEBUG MapDBSessionsStore - subscriptions_mosqsub/22634-Kuznetsov: {topic1=[filter:topic1, cliID: mosqsub/22634-Kuznetsov, qos: LEAST_ONE, active: true]}
DEBUG MapDBSessionsStore - subscriptions_mosqsub/22487-Kuznetsov: {topic1=[filter:topic1, cliID: mosqsub/22487-Kuznetsov, qos: MOST_ONE, active: true]}
DEBUG MapDBSessionsStore - subscriptions_mosqsub/22657-Kuznetsov: {topic1=[filter:topic1, cliID: mosqsub/22657-Kuznetsov, qos: EXACTLY_ONCE, active: true]}
DEBUG MapDBSessionsStore - subscriptions_mosqsub/22674-Kuznetsov: {topic1=[filter:topic1, cliID: mosqsub/22674-Kuznetsov, qos: EXACTLY_ONCE, active: true]}
DEBUG ProtocolProcessor - Broker republishing to client <mosqsub/22674-Kuznetsov> topic <topic1> qos <LEAST_ONE>, active false
DEBUG ProtocolProcessor - Broker republishing to client <mosqsub/22634-Kuznetsov> topic <topic1> qos <LEAST_ONE>, active false
DEBUG ProtocolProcessor - Broker republishing to client <mosqsub/22657-Kuznetsov> topic <topic1> qos <LEAST_ONE>, active false
DEBUG ProtocolProcessor - Broker republishing to client <test-client> topic <topic1> qos <LEAST_ONE>, active false
DEBUG ProtocolProcessor - Broker republishing to client <mosqsub/22611-Kuznetsov> topic <topic1> qos <MOST_ONE>, active false
DEBUG ProtocolProcessor - Broker republishing to client <proxied-test-client> topic <topic1> qos <MOST_ONE>, active false
DEBUG ProtocolProcessor - Broker republishing to client <mosqsub/22487-Kuznetsov> topic <topic1> qos <MOST_ONE>, active false
INFO  NettyMQTTHandler - Received a message of type CONNECT
DEBUG ProtocolProcessor - CONNECT for client <test-client>
DEBUG ProtocolProcessor - Connect with keepAlive 60 s
DEBUG ProtocolProcessor - Connect create session <[id: 0x73a3a6cc, /127.0.0.1:64039 => /127.0.0.1:1883]>
INFO  ClientSession - cleaning old saved subscriptions for client <test-client>
DEBUG MapDBSessionsStore - wipeSubscriptions
INFO  ProtocolProcessor - Connected client ID <test-client> with clean session true
```

prettynatty:
Ok, looks like lwm is not seen if you reconnect with the same clientId, even if lwm is retained.
Now I use different clientIds and test is passing.

prettynatty:
Ok, it passes when it's the only case running for that case (via `test-only *MqttSourceSpec -- -z will`) but fails when running the whole spec. Investigating.

2m:
Interesting. I wonder if the same client id logic is mentioned in the mqtt
spec. Have you seen it anywhere?

2m:
Take a look at my PR #131 where I migrated tests to use broker in docker which allows for all tests to be ran concurrently. It might be that your test will be able to run now as well together with all of the others.

prettynatty:
I've rebased my branch on yours but now all tests are failing. I have docker installed on my laptop.

prettynatty:
`docker-compose up` fixed the problem. My test is still failing. Investigating.

prettynatty:
Interesting thing. When I run whole MqttSubscriptionSpec, will case times out on waiting for proxied subscription, when I run only will case, proxied subscription succeeds (but will spec still fails)

mqtt/src/test/scala/akka/stream/alpakka/mqtt/scaladsl/MqttSourceSpec.scala
```diff
@@ -169,5 +170,40 @@ class MqttSourceSpec
...
+        .runWith(Sink.head)
+        .map(
+          _.handleWith(Tcp().outgoingConnection("localhost", 1883).viaMat(KillSwitches.single)(Keep.right))
+        )
+
+      {
+        val settings = MqttSourceSettings(
+            sourceSettings
+              .withBroker("tcp://localhost:1337")
+              .withWill(Will(MqttMessage(willTopic, ByteString("ohi")), MqttQoS.AtLeastOnce, retained = true)),
```
2m: Use different client ids for both sources. For example `testator` and `executor`.

mqtt/src/test/scala/akka/stream/alpakka/mqtt/scaladsl/MqttSourceSpec.scala
```diff
@@ -169,5 +170,40 @@ class MqttSourceSpec
...
+        }
+      }
+
+      {
+        val settings = MqttSourceSettings(sourceSettings, Map(willTopic -> MqttQoS.AtLeastOnce))
+        val source = MqttSource(settings, 8)
+
+        val (sub, elem) = source.toMat(Sink.head)(Keep.both).run()
+        whenReady(sub) { _ =>
+          elem.futureValue shouldBe MqttMessage(willTopic, ByteString("ohi"))
```
2m: You should start the *testator* source in this block so the *executor* does not miss the last will message.

mqtt/src/test/scala/akka/stream/alpakka/mqtt/scaladsl/MqttSourceSpec.scala
```diff
@@ -41,6 +41,7 @@ class MqttSourceSpec
   val topic1 = "source-spec/topic1"
   val topic2 = "source-spec/topic2"
   val secureTopic = "source-spec/secure-topic1"
+  val willTopic = "source-spec/will"
```
2m: Add this to `src/test/travis/acl` under the anonymous section. Otherwise no one can publish to this topic.

mqtt/src/test/scala/akka/stream/alpakka/mqtt/scaladsl/MqttSourceSpec.scala
prettynatty: lwt is retained, so *testator* should not miss the lwt even if subscribes after *executor* disconnecting

prettynatty:
Regarding using different clientIds, I found out that retained LWM is not being seen by reconnected client with the same clientId only in the case if it was subscribed to an LWM's topic.

mqtt/src/test/scala/akka/stream/alpakka/mqtt/scaladsl/MqttSourceSpec.scala
```diff
@@ -169,5 +170,42 @@ class MqttSourceSpec
...
+        .bind("localhost", 1337)
+        .runWith(Sink.head)
+        .map(
+          _.handleWith(Tcp().outgoingConnection("localhost", 1883).viaMat(KillSwitches.single)(Keep.right))
+        )
+
+      {
+        val settings = MqttSourceSettings(
+            sourceSettings
+              .withClientId("executor")
```
2m: I think these need a swap. :) testator is someone who makes a testament. And executor is the one who executes, follows through, the testament.

mqtt/src/test/scala/akka/stream/alpakka/mqtt/scaladsl/MqttSourceSpec.scala
2m: I missed the retaining part. So then we need to figure out why retaining does not work as expected. Because if we kill testator after executor has subscribed the test pass when ran together with all other tests.

mqtt/src/test/scala/akka/stream/alpakka/mqtt/scaladsl/MqttSourceSpec.scala
```diff
@@ -169,5 +170,42 @@ class MqttSourceSpec
...
+
+    "support will message" in {
+      import system.dispatcher
+
+      val ks = Tcp()
+        .bind("localhost", 1337)
+        .runWith(Sink.head)
+        .map(
+          _.handleWith(Tcp().outgoingConnection("localhost", 1883).viaMat(KillSwitches.single)(Keep.right))
+        )
```
2m: I looked at the test failures and there is a race between starting a proxy and starting a testator connection. You should keep the server binding materialized value from the proxy and run the testator source only when proxy is up and running. Something in the lines of this:

```scala
val (binding, connection) = Tcp().bind("localhost", 1337).toMat(Sink.head)(Keep.both).run()

val ks = connection.map(
  _.handleWith(Tcp().outgoingConnection("localhost", 1883).viaMat(KillSwitches.single)(Keep.right))
)

// wait for the proxy to start
whenReady(binding) { ... }
```

So it was not problem with message retaining after all. :)

prettynatty:
Finally! 🔥

2m:
Great job! This should be merged after #131

PR #117, opened 2016-12-11
FTP: Improve type safety wrt settings (closes #116)

### LINKED ISSUE #116: FTP connector: improve type safety regarding configuration settings
Currently, it's possible to build `Source` instances from wrong setting types. For instance, you could build a `Source` for `FTP` from `SftpSettings`. The error would be thrown in runtime instead of in compilation time.

juanjovazquez:
See #116

### DIFF
--- a/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpSourceFactory.scala
+++ b/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpSourceFactory.scala
@@ -13,6 +13,8 @@ import java.nio.file.Path
 
 private[ftp] trait FtpSourceFactory[FtpClient] { self =>
 
+  type S <: RemoteFileSettings
+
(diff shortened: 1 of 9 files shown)


ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpSourceFactory.scala
```diff
@@ -13,66 +13,66 @@ import java.nio.file.Path
...
 private[ftp] trait FtpSourceFactory[FtpClient] { self =>
 
+  type S <: RemoteFileSettings
+
   protected[this] final val DefaultChunkSize = 8192
 
   protected[this] def ftpClient: FtpClient
 
-  protected[this] def ftpBrowserSourceName: String
+  protected[this] val ftpBrowserSourceName: String
```
juanjovazquez: No particular reason. You're right. Changed.

Issue #98, opened 2016-11-30
Implement Amazon's Request-signing Test Suite


joearasin:
I was digging through some of my notes from originally building the S3 functionality and dug this up. Amazon provides a test suite for AWS signature libraries. It might make sense to use it.

http://docs.aws.amazon.com/general/latest/gr/signature-v4-test-suite.html

ktoso:
That would be great indeed, anyone up for jumping on this task?

devsprint:
Hi, I've started to implement a new test suite using Amazon provided test suite. First problem that I have found with the current implementation is the missing hash of the payload (hash of empty string) in building Canonical Request. In Canonical Request object:
`val hashedBody = req.headers.find(_.name == "x-amz-content-sha256")
.map(_.value).getOrElse("")`

See: [Task 1: Create a Canonical Request for Signature Version 4](http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html)

I'll come with PR tomorrow.

devsprint:
@joearasin @ktoso implementation is completed. There is only one test that is skipped (you'll see in the code). This is because I did not found a way to build a HttpRequest with body, without having it adding the charset part to content type. If you know a way to do this, then I will update the code in a second.

jrudolph:
As commented in #120 it is not possible to use the AWS signing test suite verbatim because akka-http cannot be made to render exactly the same requests. This [comment](https://github.com/akka/alpakka/pull/120#issuecomment-289786050) gives a few hints how better test coverage could be achieved anyways.

PR #120, opened 2016-12-16
#98 - end to end testing of AWS Signature 4 based on AWS test suite

devsprint:
…t validation.


typesafehub-validator:
Hi @devsprint,

ktoso:
Refs https://github.com/akka/alpakka/issues/98
Thanks, not sure if I'll be able to review today but hope someone might be able to :)

devsprint:
CLA signed

patriknw:
@agolubev is this something you would like to review?

agolubev:
@patriknw thanks for the notification. I'll review this today

s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
```diff
@@ -29,12 +36,27 @@ private[alpakka] object CanonicalRequest {
...
-  }
 
   def canonicalQueryString(query: Query): String =
-    query.sortBy(_._1).map { case (a, b) => s"${uriEncode(a)}=${uriEncode(b)}" }.mkString("&")
+    query.sortBy(_._1).map { case (a, b) => encodeQueryParams(a, b) }.mkString("&")
 
-  private def uriEncode(str: String) = URLEncoder.encode(str, "utf-8")
+  private def encodeQueryParams(name: String, value: String): String =
+    if (name.contains(' ')) {
+      s"${uriEncode(name.takeWhile(_ != ' ')).replace("%7E", "~")}="
```
agolubev: should this be changed to just string append?

s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
```diff
@@ -29,12 +36,27 @@ private[alpakka] object CanonicalRequest {
...
+    } else {
+      s"${uriEncode(name).replace("%7E", "~")}=${uriEncode(value).replace("%7E", "~")}"
+    }
+
+  private def uriEncode(str: String) = if (isAlreadyURLEncoded(str)) str else URLEncoder.encode(str, "utf-8")
+
+  // TODO: this could be removed when the Uri class in akka-http will accept utf-8
```
agolubev: I was not able to find appropriate ticket in akka-http. Should one be created?

s3/src/test/scala/akka/stream/alpakka/s3/auth/SignerSpec4.scala
```diff
@@ -0,0 +1,350 @@
...
+  }
+
+  def testStringToSign(req: HttpRequest, sts: (String, String))(implicit mat: Materializer): Unit = {
+    val date = LocalDateTime.of(2015, 8, 30, 12, 36, 0).atZone(ZoneOffset.UTC)
+    val canonicalRequest = WrappedCanonicalRequest.canonicalRequest(req)
+    val stsResult = canonicalRequest.map(cr => Signer.stringToSign("AWS4-HMAC-SHA256", signingKey, date, cr))
+    whenReady(stsResult) { builtSts =>
+      withClue(s"According to the expected string to sign stored in ${sts._1}") {
+        if (sts._1 == "post-x-www-form-urlencoded.sts") {
+          // TODO: skip this test because we can't avoid adding charset into http request as the test suite require.
```
agolubev: do we need to consider this as http ticket?

s3/src/test/scala/akka/stream/alpakka/s3/auth/SignerSpec4.scala
```diff
@@ -0,0 +1,350 @@
...
+      req <- requests
+      crq <- creqs
+    } yield req.zip(crq)
+
+    probes.toOption shouldBe defined
+
+    probes.get.foreach {
+      case (fileName, expectedResult) =>
+        testCanonicalRequest(fileName, expectedResult)
+    }
```
agolubev: All 4 test cases are different only with testMethod and expected result - can this be a single function with results collection and test function as arguments?

s3/src/test/scala/akka/stream/alpakka/s3/auth/SignerSpec4.scala
```diff
@@ -0,0 +1,350 @@
...
+  }
+
+  /**
+   * Encode the uri string, if it is requires.
+   *
+   * @param rawUri - the string URI.
+   * @return the Uri
+   */
+  private def encodeIfRequired(rawUri: String): Uri = {
+    val pattern = """^([!#$&-;=?-\[\]_a-z~]|%[0-9a-fA-F]{2})+$$""".r
```
agolubev: need some explanation here, maybe comments

s3/src/test/scala/akka/stream/alpakka/s3/auth/SignerSpec4.scala
```diff
@@ -0,0 +1,350 @@
...
+    // TODO: this look like a bug in Uri parser in akka-http ("//" is generating empty path)
```
agolubev: Another akka-http ticket?

s3/src/test/scala/akka/stream/alpakka/s3/auth/SignerSpec4.scala
```diff
@@ -0,0 +1,350 @@
...
+   * @param headers - a list of headers as string.
+   * @return a list of HttpHeaders.
+   */
+  private def loadHeaders(headers: Seq[String]): List[HttpHeader] = {
+
+    val lines: List[String] = headers.foldLeft(List.empty[String]) { (acc, current) =>
+      if (current.contains(':')) {
+        current :: acc
+      } else {
+        // TODO: trim and ',' has been added to pass the test, but it looks like a bug in HttpParser code from akka-http.
```
agolubev: same as above

agolubev:
In general - looks good

s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
devsprint: https://github.com/akka/akka-http/issues/86 - this is  a similar issue, but for Cyrillic URIs

agolubev: could you add ticket number to comment into code?

s3/src/test/scala/akka/stream/alpakka/s3/auth/SignerSpec4.scala
devsprint: https://github.com/akka/akka-http/issues/689

s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
devsprint: Sure. it will be in next PR

s3/src/test/scala/akka/stream/alpakka/s3/auth/SignerSpec4.scala
devsprint: https://github.com/akka/akka-http/issues/690

devsprint:
@agolubev please take a look to last PR.

s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
```diff
@@ -43,14 +65,21 @@ private[alpakka] object CanonicalRequest {
...
   def canonicalHeaderString(headers: Seq[HttpHeader]): String = {
     val grouped = headers.groupBy(_.lowercaseName())
-    val combined = grouped.mapValues(_.map(_.value.replaceAll("\\s+", " ").trim).mkString(","))
-    combined.toList.sortBy(_._1).map { case (k, v) => s"$k:$v" }.mkString("\n")
+    val combined = grouped.mapValues(_.map(_.value.replaceAll("\\s+", " ")).mkString(","))
+    fixContentTypeHeaderParameter(combined.toList.sortBy(_._1).map { case (k, v) => s"$k:$v" }.mkString("\n"))
   }
 
+  private def fixContentTypeHeaderParameter(headers: String): String =
+    headers.replace("charset=UTF-8", "charset=utf8")
```
jrudolph: This will break signing, as you sign a different request than what akka-http will send.

jrudolph:
I think this approach cannot work because not all of the test cases (the `.req` files) can be represented verbatim by the akka-http model. In the end, the point of the akka-http model is to remove many of the redundant representations by a single typed model which makes it harder to get many of the representation details wrong.

If you want to use the test cases, you need to check in the first step what the actually requests are that akka-http will send out and check these against the original test case. This will probably need to use `HttpRequestRenderer` or actual end-to-end testing by running the request through the complete `Http.clientLayer` to figure out what the actual request is. In many cases, it will not be the same as what the test cases need. The problem is then, that you cannot use the provided test data because you cannot make akka-http to send them unchanged, so you need to sign something else.

Worse, the changes to the signer might actually break signing because the canonical representation doesn't seem to be the same as what will be sent through akka-http, so that AWS will have to reject requests which were manipulated by the signer.

s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
```diff
@@ -43,14 +65,21 @@ private[alpakka] object CanonicalRequest {
...
   private def preprocessPath(path: Path): String =
-    uriEncode(path.toString).replace(":", "%3A").replace("%2F", "/")
+    uriEncode(removeRedundantAndRelativePath(path.toString))
+      .replace(":", "%3A")
+      .replace("%2F", "/")
+      .replace("%7E", "~")
+      .replace("+", "%20")
 
   def canonicalHeaderString(headers: Seq[HttpHeader]): String = {
     val grouped = headers.groupBy(_.lowercaseName())
```
jrudolph: This seems to be missing the value of `req.entity.contentType` which is encoded specially in the akka-http model.

devsprint:
@jrudolph hmm, I understand your reasons and thanks for clarifications.  Some parts of the test data from Amazon will not work with current constrains, but this is a small part of it, the ones that are having a body, where most of them are not having any. For this I will work out how to adapt it such that we can use it with smaller changes (.req, .creq.. changes). Are you ok with this approach ?

patriknw:
Not sure how important the comment is, do you know @ktoso ?
Can we merge and do a follow up?

devsprint:
@patriknw I will update the pull request based on @jrudolph comments but I will be only able to do it over the week-end.

patriknw:
Ok, thanks for the update, then I'll leave this open

s3/src/test/scala/akka/stream/alpakka/s3/auth/SignerSpec4.scala
```diff
@@ -0,0 +1,308 @@
...
+}
+
+case object StringToSignProbe extends ProbeType {
+  override val extension: String = ".sts"
+}
+
+/**
+ * Load the probes files from resource folder.
+ */
+object Probes {
```
ktoso: Probes is a bit of a weird name, in Akka probes are TestProbes hm.

ktoso:
Agree with @jrudolph's both contents. We must use the actual APIs to verify the test actually is testing what will be put onto the wire. Now it's a weird mix of hand constructing things which may be not the way we'd render these - please make the .req files into actual Akka HTTP model types directly. Where it's not possible we need to fix something or provide workarounds.

patriknw:
@devsprint any news here?

jrudolph:
Sorry for keeping this so long.

I think the bottom line here is:

 * it is impossible to implement the AWS test suite including the hashes given by the test suite because some of the test cases cannot be rendered by akka-http (that's by design, semantically equivalent headers can be rendered)
 * it is likely that this PR will break signing because the signer is changed in a way to make the tests pass (but the tests do not assert what akka-http renders)

Here's a sketch how to implement signing tests:
   * provide or generate a set of interesting test requests as akka-http `HttpRequest` (internal `HttpRequestParser` infrastructure could be used to parse the AWS test files, signatures of the test files cannot be used)
   * run each of the requests through akka-http's internal `HttpRequestRenderer` or use `Http.clientLayer()` flow (connect all ports of the BidiFlow to probes and then push a request and read the network side, to get access to the actual wire rendering of the request
   * use any third party library to sign this request data (which will involve parsing the request with whatever the third-party tool provides)
   * check that the alpakka AWS signer generates the same signature

Not sure if it makes sense to keep this PR open as the real solution might not have much in common with what we have here.

WDYT?

devsprint:
@jrudolph I have tried to find a way to still use the test cases from Amazon but as you also obeserved is it imposible to use them. I think that the right option is to close this PR.

jrudolph:
Ok. Let's close it for now. I'll add a comment in the ticket to see the comments here before tackling it again. Anyway, thanks for your work, @devsprint!

Issue #124, opened 2016-12-19
AMQP BindingDeclaration swapped queue and exchange name


irishshagua:
When the BindingDeclaration is being converted to a queueBind request, the queue name and exchange name are in the wrong order. Consider the below example for a newly created queue which you are trying to bind to an existing exchange:

    package akka.stream.alpakka.amqp

    import akka.actor.ActorSystem
    import akka.stream.ActorMaterializer
    import akka.stream.alpakka.amqp.scaladsl.AmqpSource
    import akka.stream.scaladsl.Sink

    object Playground extends App {

      override def main(args: Array[String]): Unit = {
        implicit val actorSystem = ActorSystem("playground")
        implicit val materializer = ActorMaterializer()

        val QueueName = "test-queue"
        val ExchangeName = "pre-existing-exchange"

        val queueDeclaration = QueueDeclaration(name = QueueName)
        val bindingDeclaration = BindingDeclaration(
          queue = QueueName,
          exchange = ExchangeName
        )
        val source = AmqpSource(
          NamedQueueSourceSettings(DefaultAmqpConnection, QueueName)
            .withDeclarations(queueDeclaration, bindingDeclaration),
          bufferSize = 10
        )

        source.map(im => im.bytes.decodeString("UTF-8")).to(Sink.foreach(println)).run()
      }
    }

This generates the below exception when run:

> [ERROR] [12/19/2016 21:35:05.535] [playground-akka.actor.default-dispatcher-2] [akka://playground/user/StreamSupervisor-0/flow-0-0-unknown-operation] Error during preStart in [akka.stream.alpakka.amqp.AmqpSourceStage@7ffb3be5]
java.io.IOException
  at com.rabbitmq.client.impl.AMQChannel.wrap(AMQChannel.java:106)
  at com.rabbitmq.client.impl.AMQChannel.wrap(AMQChannel.java:102)
  at com.rabbitmq.client.impl.AMQChannel.exnWrappingRpc(AMQChannel.java:124)
  at com.rabbitmq.client.impl.ChannelN.queueBind(ChannelN.java:945)
  at com.rabbitmq.client.impl.ChannelN.queueBind(ChannelN.java:61)
  at akka.stream.alpakka.amqp.AmqpConnectorLogic$$anonfun$preStart$1.apply(AmqpConnector.scala:73)
  at akka.stream.alpakka.amqp.AmqpConnectorLogic$$anonfun$preStart$1.apply(AmqpConnector.scala:62)
[...]

PR #125, opened 2016-12-19
Fix for issue #124

irishshagua:
This is to resolve issue https://github.com/akka/alpakka/issues/124


amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
```diff
@@ -69,7 +69,7 @@ private[amqp] trait AmqpConnectorLogic { this: GraphStageLogic =>
           arguments.asJava
         )
 
-      case BindingDeclaration(exchange, queue, routingKey, arguments) =>
+      case BindingDeclaration(queue, exchange, routingKey, arguments) =>
```
patriknw: wow

PR #131, opened 2016-12-28
Add Mqtt Sink

2m:
Also run mqtt broker in docker

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,27 +9,22 @@ scala:
 jdk:
   - oraclejdk8
 
+services:
+  - docker
+
+before_install:
+  - docker-compose up -d
+
+sbt_args: -J-XX:ReservedCodeCacheSize=128m
+
 before_cache:
   - find $HOME/.ivy2 -name "ivydata-*.properties" -print -delete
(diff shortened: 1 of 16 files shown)


.travis.yml
```diff
@@ -9,28 +9,23 @@ scala:
 jdk:
   - oraclejdk8
 
+services:
+  - docker
+
+before_install:
+  - docker-compose up -d
```
filosganga: makes lot of sense!

mqtt/src/main/scala/akka/stream/alpakka/mqtt/MqttProducerStage.scala
```diff
@@ -0,0 +1,73 @@
...
+        failStage(ex)
+    }
+
+    setHandler(in,
+      new InHandler {
+      override def onPush() = {
+        val msg = grab(in)
+        val pahoMsg = new PahoMqttMessage(msg.payload.toArray)
+        pahoMsg.setQos(qos.byteValue)
+        mqttClient match {
```
filosganga: Is not better to do it in the `preStart` hook instead?

drewhk:
LGTM. Can this be merged?

PR #144, opened 2017-01-11
Add latest version badge

ktoso:


2m:
Great idea. I think it should go next to the gitter badge instead of on a separate paragraph.

ktoso:
Moved it like Martynas suggested 👍

Issue #115, opened 2016-12-11
MongoDb Connector based on official driver


iozozturk:
Hi all,

I know that MongoDb connector is already mentioned and closed at #43 . I think not everybody is using that driver for MongoDb.

I have been using [official driver](http://mongodb.github.io/mongo-scala-driver/) for a while and it is good and reactive. It also has a compatibility with [reactive streams](http://mongodb.github.io/mongo-scala-driver/1.2/integrations/) whereas it requires a bit of manual copy paste work, which is not elegant.

I can take part if you think it is reasonable. What do you think?

johanandren:
Is there any significant pros with using the official driver over the reactivemongo driver?

iozozturk:
I don't think there is whereas people might feel safer to go with the official driver. And there are already applications using official driver. Since reactive mongo is an external connector, should they mix one more driver(reactivemongo) as a dependency to be able to use the Alpakka mongo connector?

patriknw:
@iozozturk I don't mind listing both alternatives. Will you create a PR?

iozozturk:
@patriknw Ok I will create a PR for this

stephennancekivell:
I'd also like to see this. Reactive mongo is currently based on akka (2.3.x) without scala 2.12. Where mongo-scala-driver is netty based with current 2.12.
So there maybe benefits with dependency versions.

It also looks like there are some performance benefits in the mongo-scala-driver.
http://evojam.github.io/mongo-drivers-benchmarks/

gosubpl:
Also, official driver has Kerberos support, which ReactiveMongo is missing. Quite important to those in corporate environment.
@iozozturk would be great if you could create a PR, otherwise I will probably pick this up, as I have already once done a move reactivemongo -> official driver for the reaseons pointed above :)

iozozturk:
I will propose a PR in a few days, I am struggling with the Java tests now. Thanks for reminding.

stephennancekivell:
Hey @iozozturk. I've added a PR for a simple Sink. Theres not as much to it as with your Source, but at least it points beginners in the right direction.

It should merge well with yours, I've just C&P'd your config code.

michalbogacz:
Hey, thanks for great work!
I have one additional idea: 
There is a new version of official mongo-scala-driver with support for case class conversion (http://mongodb.github.io/mongo-scala-driver/2.0/getting-started/quick-tour-case-classes/).
[...]

PR #148, opened 2017-01-16
MongoDb Source #115

iozozturk:
Source implementation for MongoDb official Scala driver. Couldn't find a practical case using it with JavaDsl and couldn't find a way to build queries in Java using Scala driver library. Hence JavaDsl is not included. Comments and reviews are welcome.


harksin:
hi iozozturk,

in my humble opinion, [reactive mongo project](http://reactivemongo.org/) provide a good [dsl](http://reactivemongo.org/releases/0.12/documentation/tutorial/streaming.html#streaming) for akka-stream. maybe you don't known it ?

stephennancekivell:
Hi @harksin, As discussed in #115 we think there exists scope for a integration based of the official scala driver.

@iozozturk Why have you chosen to return `play-json` `JsObjects` rather than just mongo `Documents`? Thats whats happening isnt it?
It would might be more reusable to just be Documents.

iozozturk:
Hi @stephennancekivell , thanks for the review. In most of my own use cases, I ended up using results as JsObjects. That's why I opinionated on the result type to be JsObjects. I can easily turn back to Documents if you feel it is more convenient/reusable and JsObjects are too opinionated to go.

johanandren:
One reason to choose to not publishing `BsonDocument`s from the source (which I guess is what you mean with `Document`) is that it is actually mutable which could lead to bad surprises if mutating in streams with async boundaries, while the play Json-ast is immutable and guaranteed to be safe.

stephennancekivell:
I just think there are lots of ways to parse these documents and alpakka doesnt really need to exclude others.

It would be easy to support a source with both Json and Document. Then Json one could be implemented as a second step using the document one.

Maybe like..
```
object MongoSource {
	def asJson(query: Observable[Document]): Source[Document, NotUsed]
	def asDocument(query: Observable[Document]): Source[Document, NotUsed]
}
```

Um, isnt it these immutable docs http://mongodb.github.io/mongo-scala-driver/1.2/bson/documents/#immutable-documents

Also, if alpakka is to include play-json its a maintenance burden to upgrade the library. Different users will be on different versions or not even using it. And thats for adding 1/2 a line of code, out of this 200 line commit.

Sorry I dont mean to be pedantic, I just want to see this get lots of use.

johanandren:
Yep, you are right about the Scala driver document being immutable, I looked at the wrong mongo sources. Given that I agree, parsing the bson documents using play-json should be optional.

mongodb/src/main/scala/akka/stream/alpakka/mongodb/Implicits.scala
```diff
@@ -0,0 +1,64 @@
...
+import java.util.concurrent.atomic.AtomicBoolean
+
+import org.mongodb.{scala => mongoDB}
+import org.{reactivestreams => rxStreams}
+
+import scala.language.implicitConversions
+
+object Implicits {
+
+  implicit def observableToPublisher[T](observable: mongoDB.Observable[T]): rxStreams.Publisher[T] = ObservableToPublisher(observable)
```
johanandren: This is used in one place, I don't think that warrants using implicits, just use the adapter explicitly instead. Also, it should be private to the mongodb connector and not public API

mongodb/src/main/scala/akka/stream/alpakka/mongodb/Mongo.scala
```diff
@@ -0,0 +1,16 @@
+package akka.stream.alpakka.mongodb
+
+import com.typesafe.config.{Config, ConfigFactory}
+import org.mongodb.scala.{MongoClient, MongoDatabase}
+
+object Mongo {
+  private val config: Config = ConfigFactory.load(getClass.getClassLoader, "application.conf")
```
johanandren: Would be better if there was a DSL for the connection details that is passed to the factory methods (possibly with an optional parse-from-config method but I'm not sure that is needed). The current solution will severly limit the usefulness of the connector since you could only ever connect to a single mongodb defined in `application.conf`. Look at how it is done in for example the amqp connector. Could be as simple as `case class MongoConnectionDetails(db, host, port)`.

mongodb/src/main/scala/akka/stream/alpakka/mongodb/scaladsl/MongoSource.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import akka.stream.scaladsl.Source
+import org.mongodb.scala.{Document, Observable}
+import play.api.libs.json.{JsObject, Json}
+import akka.stream.alpakka.mongodb.Implicits._
+
+object MongoSource {
+
+  def apply(query: Observable[Document]) ={
+    Source.fromPublisher(query).map(doc=>Json.parse(doc.toJson()).as[JsObject])
```
johanandren: Make parsing to JsObject optional (if it should be here at all), also, always explicitly define return type of public api methods.

What about removing the json parsing from the connector but providing a sample of how to use `play-json` in the docs?

mongodb/src/main/scala/akka/stream/alpakka/mongodb/Mongo.scala
```diff
@@ -0,0 +1,16 @@
...
+
+  private val dbName: String = config.getString("mongo.name")
+  private val dbAddress: String = config.getString("mongo.host")
+  private val dbPort: String = config.getString("mongo.port")
+
+  val client: MongoClient = MongoClient(s"mongodb://$dbAddress:$dbPort")
```
johanandren: If the client is to be cached it should be bound to the stage lifecycle, not put in a global singleton. This will leak the client between tests, and keep it in memory even if there is no more use of it.

project/Dependencies.scala
```diff
@@ -56,13 +56,22 @@ object Dependencies {
     )
   }
 
+
+  val MongoDb = Seq(
+    libraryDependencies ++= Seq(
+      "org.mongodb.scala" %% "mongo-scala-driver" % "1.2.1", // ApacheV2
+      "com.typesafe.play" %% "play-json" % "2.6.0-M1"
```
johanandren: Would be better if it didn't depend on a milestone, if it is to depend on a json library at all.

mongodb/src/main/scala/akka/stream/alpakka/mongodb/Mongo.scala
stephennancekivell: Should `MongoClient` just be passed as a parameter to `MongoSource.apply`, rather than doing anything with a config. Looks like `MongoClient` already provides a few different apply methods itself.

stephennancekivell:
Hi @iozozturk, what do you think about the proposed changes.

iozozturk:
Hi @stephennancekivell , I have read the proposed changes and reviews. Makes sense, thanks for those. Just didn't have the time to refactor. 

Perhaps best way is to remove `play-json` and show it as an example in the docs as @johanandren suggested. Will consider the other reviews also from @johanandren and try to look into this in the weekend, thanks for reminding.

stephennancekivell:
Thanks @iozozturk, I'll help you out by sending you a pull request adding some documentation. I think it'll help, even though it'll need updating. Its the boring part anyway.

https://github.com/iozozturk/alpakka/pull/1

patriknw:
@iozozturk I'd like to review this, but the diff looks wrong. It's changing way to many unrelated things. Can you please rebase?

stephennancekivell:
Hey @iozozturk do you want a hand finishing this off ?

iozozturk:
Hey @stephennancekivell, I think I messed up my own branch history somehow :/ Not sure how I did, I was rebasing time to time. I went over some of the reviews and was going over others (a bit busy at work). My current plan is, on my local, first get a fresh branch X from master and cherry pick my own commits from here (mongodb-ismet) to the new branch. Then delete this one and recreate with the same name and cherry pick from branch X to this recreated branch and push. Does it sound ok? Or do you know another way without recreating a new PR?

stephennancekivell:
Hi @iozozturk, yea that sounds like a good plan, you just need to force push the correct state back to your github branch `iozozturk/mongodb-ismet`.

The other way I can think of is to do a interactive rebase and drop the dodgy commits from your branch. `0fcc0fb f8d9804 cb9cb0d ac55394 85dad9f`

Alternatively just make another PR.

mongodb/src/main/scala/akka/stream/alpakka/mongodb/ObservableToPublisher.scala
```diff
@@ -0,0 +1,58 @@
...
+package akka.stream.alpakka.mongodb
+
+import java.util.concurrent.atomic.AtomicBoolean
+
+import org.mongodb.{scala => mongoDB}
+import org.{reactivestreams => rxStreams}
+
+import scala.language.implicitConversions
+
+case class ObservableToPublisher[T](observable: mongoDB.Observable[T]) extends rxStreams.Publisher[T] {
```
johanandren: `final` and `private[mongodb]`

mongodb/src/main/scala/akka/stream/alpakka/mongodb/MongoConnectionDetails.scala
```diff
@@ -0,0 +1,9 @@
+package akka.stream.alpakka.mongodb
```
johanandren: All files need the copyright header, see `akka.stream.alpakka.file.scaladsl.DirectoryChangesSource` for example.

mongodb/src/main/scala/akka/stream/alpakka/mongodb/MongoConnectionDetails.scala
```diff
@@ -0,0 +1,9 @@
...
+
+import org.mongodb.scala.{MongoClient, MongoDatabase}
+
+private[mongodb] case class MongoConnectionDetails(dbName: String,host: String, port:Int) {
+  lazy val client: MongoClient = MongoClient(s"mongodb://$host:$port")
+
+  lazy val db: MongoDatabase = client.getDatabase(dbName)
+}
```
johanandren: This is only used in the tests so can be moved there, maybe even inlined in the spec?

mongodb/src/main/scala/akka/stream/alpakka/mongodb/ObservableToPublisher.scala
```diff
@@ -0,0 +1,58 @@
...
+  def subscribe(subscriber: rxStreams.Subscriber[_ >: T]): Unit = {
+    observable.subscribe(
+      new mongoDB.Observer[T]() {
+        override def onSubscribe(subscription: mongoDB.Subscription): Unit = {
+          subscriber.onSubscribe(new rxStreams.Subscription() {
+            private final val cancelled: AtomicBoolean = new AtomicBoolean
+
+            def request(n: Long) {
```
patriknw: `override` ?

patriknw: return type, `: Unit =`

mongodb/src/main/scala/akka/stream/alpakka/mongodb/ObservableToPublisher.scala
```diff
@@ -0,0 +1,58 @@
...
+              if (!subscription.isUnsubscribed && n < 1) {
+                subscriber.onError(new IllegalArgumentException(
+                  """3.9 While the Subscription is not cancelled,
+                    |Subscription.request(long n) MUST throw a java.lang.IllegalArgumentException if the
+                    |argument is <= 0.""".stripMargin))
```
patriknw: that's a strange exception message, and no newlines in exception or log messages please

mongodb/src/main/scala/akka/stream/alpakka/mongodb/ObservableToPublisher.scala
```diff
@@ -0,0 +1,58 @@
...
+case class ObservableToPublisher[T](observable: mongoDB.Observable[T]) extends rxStreams.Publisher[T] {
+  def subscribe(subscriber: rxStreams.Subscriber[_ >: T]): Unit = {
+    observable.subscribe(
+      new mongoDB.Observer[T]() {
+        override def onSubscribe(subscription: mongoDB.Subscription): Unit = {
+          subscriber.onSubscribe(new rxStreams.Subscription() {
+            private final val cancelled: AtomicBoolean = new AtomicBoolean
+
+            def request(n: Long) {
+              if (!subscription.isUnsubscribed && n < 1) {
```
patriknw: no need to check `cancelled` here?

mongodb/src/main/scala/akka/stream/alpakka/mongodb/ObservableToPublisher.scala
patriknw: This looks like a simple `Publisher` but I'm still not 100% sure that it's completely correct. Normally we recommend such things to be implemented as `GraphStage`. Perhaps it's worth implementing it as `Publisher` because it's almost a one-to-one mapping to `mongoDB.Observable`?

mongodb/src/main/scala/akka/stream/alpakka/mongodb/scaladsl/MongoSource.scala
```diff
@@ -0,0 +1,13 @@
+package akka.stream.alpakka.mongodb.scaladsl
+
+import akka.stream.alpakka.mongodb.ObservableToPublisher
+import akka.stream.scaladsl.Source
+import org.mongodb.scala.{Document, Observable}
+
+object MongoSource {
+
+  def apply(query: Observable[Document]) = {
```
patriknw: explicit return type for all public API

mongodb/src/main/scala/akka/stream/alpakka/mongodb/MongoConnectionDetails.scala
iozozturk: Moved to the relevant test and removed this file since user will be responsible and provide for his/her connection to db.

mongodb/src/main/scala/akka/stream/alpakka/mongodb/ObservableToPublisher.scala
iozozturk: Not sure needed but checked cancelled anyway.

iozozturk:
Planning to create/update documentation after agreeing on the final state of the implementation.

project/Dependencies.scala
```diff
@@ -62,6 +62,13 @@ object Dependencies {
     )
   )
 
+  val MongoDb = Seq(
+    libraryDependencies ++= Seq(
+      "org.mongodb.scala" %% "mongo-scala-driver" % "1.2.1", // ApacheV2
+      "com.typesafe.play" %% "play-json" % "2.6.0-M1"
```
stephennancekivell: Is `play-json` still needed ?

mongodb/src/main/scala/akka/stream/alpakka/mongodb/ObservableToPublisher.scala
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package akka.stream.alpakka.mongodb
+
+import java.util.concurrent.atomic.AtomicBoolean
+
+import org.mongodb.{scala => mongoDB}
+import org.{reactivestreams => rxStreams}
+
+import scala.language.implicitConversions
```
stephennancekivell: Is this still needed ?

stephennancekivell:
Hi, what else is needed to complete this PR, just documentation right?.

The overall solution looks pretty good to me.

project/Dependencies.scala
iozozturk: Not needed. Removed.

mongodb/src/main/scala/akka/stream/alpakka/mongodb/ObservableToPublisher.scala
iozozturk: Implicit conversions not needed. Removed.

iozozturk:
Hi @patriknw , I just added documentation with looking at @stephennancekivell's pull request.Can you please review, I want to finish this implementation. Thank you

iozozturk:
Also updated the MongoDB Scala driver version, to mention.

iozozturk:
I am not planning to, this connector is based on Scala Driver.

iozozturk:
Hi @patriknw , any updates on this PR?

patriknw:
Thanks for your contribution @iozozturk 
It will be included in next release.

Issue #150, opened 2017-01-18
question: next release or daily snapshots


joost-de-vries:
I'm interested in using Dynamodb streams. But it isn't available as a part of Alpakka 0.5 it seems. 
There are no snapshot versions published either, right?

I'm curious what my options are here apart from making my own snapshot build. Any ideas?

Btw I noticed that the dynamodb subproject is not part of the overall build since it's not aggregated in the root sbt project. See [here](https://github.com/akka/alpakka/blob/master/build.sbt#L4)
Is that intentional?

patriknw:
It should have been in 0.5, but it's not https://repo1.maven.org/maven2/com/lightbend/akka/

Probably something missing in the build. We'll publish a new release.

We are also having trouble with the documentation publish and therefore was 0.5 not announced yet.

@2m can you take a look at this also?

joost-de-vries:
Btw the unittests of the Dynamodb subproject didn't run on my machine. I kept running into [this problem](http://stackoverflow.com/questions/34137043/amazon-dynamodb-local-unknown-error-exception-or-failure) with starting a local dynamodb that runs embedded in the JVM.

2m:
Indeed DynamoDB connector artifacts were not published because it was not part of the aggregate. I have published the artifacts for the 0.5 version and they are now in [Bintray repo](https://bintray.com/akka/maven/alpakka/0.5). They will appear in the Maven Central once the synchronization is completed.

I will take a look at DynamoDB unitests.

joost-de-vries:
Tx. Now I can remove my own local built version.
Btw the readme isn't quite correct: it's not
```
    libraryDependencies += "com.typesafe.akka" %% "akka-stream-alpakka-dynamodb" % "0.5"
```
But "com.lightbend.akka" etc.

patriknw:
Then we are all done here

PR #157, opened 2017-01-19
Run DynamoDB local in docker

2m:
This allows concurrent execution of all tests.

Addresses some of the issues found in #150


2m:
@areyouspiffy, as me and @joost-de-vries had some troubles running DynamoDB local embedded I have moved testing to docker container, as we do with a number of other connectors. This allows concurrent running of all test cases. Let me know what you think.

areyouspiffy:
@2m You'll see that error if you try running the tests without the native sqlite binaries that DynamoDBLocal depends on. All the binaries are [here](https://mvnrepository.com/artifact/com.almworks.sqlite4java) and you should only need the one that matches your container.  Alternatively, if you  download it directly from [here](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.html) they include everything.

2m:
The validation error of this PR was because of the missing AWS credentials. `DefaultAWSCredentialsProviderChain` throws an exception if it can not find at least one pair of credentials somewhere.

I guess in most of our development environments we have some credentials configured in `~/.aws/credentials` so we do not see this error while running tests on dev machines.

The fix for travis is to set a system property in every test. This makes `DefaultAWSCredentialsProviderChain` happy. DynamoDB local does not need any credentials anyway.

Issue #160, opened 2017-01-23
Add RPC flow for rabbit connector


Falmarri:
A `Flow[OutgoingMessage, IncomingMessage, _]` would be very useful for integrating RPC workflows into akka streams

2m:
Fixed in https://github.com/akka/alpakka/pull/161

PR #161, opened 2017-01-23
Add Flow to support RabbitMQ RPC workflow #160

Falmarri:


Falmarri:
Is this failure related to my PR? Or something else? It seems unrelated

johanandren:
the failure is from this PR, there is a missing `//#run-sink` close tag in `AmqpConnectorsSpec.scala`

Falmarri:
Any comments on this?

amqp/src/main/scala/akka/stream/alpakka/amqp/javadsl/AmqpSink.scala
```diff
@@ -17,6 +17,12 @@ object AmqpSink {
     Sink.fromGraph(new AmqpSinkStage(settings))
 
   /**
+    * Java API: Creates an [[AmqpSink]] that accepts [[OutgoingMessage]] elements.
+    */
+  def createReplyTo(settings: AmqpSinkSettings): akka.stream.javadsl.Sink[OutgoingMessage, NotUsed] =
```
johanandren: Duplicate the docs from `AmqpReplyToStage` here to make it clear how it differs from `create`.

Also, maybe we should fail the "regular" sink if `replyTo` is set on an `OutgoingMessage`, so that if someone mixes them up it will at least fail as fast as possible? (Ofc optimal would be to have two different types so doing it wrong wouldn't even compile, but to do that we'd have to break the existing method and I don't think we should do that)

amqp/src/main/scala/akka/stream/alpakka/amqp/javadsl/AmqpRpcFlow.scala
```diff
@@ -0,0 +1,30 @@
...
+import akka.util.ByteString
+
+import scala.concurrent.Future
+
+object AmqpRpcFlow {
+
+  /**
+    * Java API: Creates an [[AmqpRpcFlow]] with given settings and buffer size.
+    */
+  def create(settings: AmqpSinkSettings, bufferSize: Int) =
```
johanandren: Always use explicit return types on the public methods, all across the PR. So these should be `javadsl.Flow[...]`

Also, the scaladoc/javadoc here isn't very useful, try to explain what the flow it creates does rather than "it takes these parameters and returns a flow of this type" - that is already pretty much clear from the signature.

Something like "Create a RPC style flow (... link to rpc sample in rabbitmq docs...) allowing for bla bla communication over the message bus".

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
```diff
@@ -0,0 +1,179 @@
...
+              routingKey,
+              elem.mandatory,
+              elem.immediate,
+              props,
+              elem.bytes.toArray
+            )
+
+            // TODO: This is pretty expensive just to see if a message specifies how many responses it should be expecting
+            Option(props.getHeaders)
+            .map(_.asScala)
```
johanandren: weird indentation here

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
```diff
@@ -0,0 +1,179 @@
...
+            }
+        }
+      )
+
+      setHandler(
+        in,
+        new InHandler {
+          // We don't want to finish since we're still waiting
+          // on incoming messages from rabbit
+          override def onUpstreamFinish(): Unit = {}
```
johanandren: What if upstream cancels before emitting any messages, will we then be stuck?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala
```diff
@@ -77,3 +77,69 @@ final class AmqpSinkStage(settings: AmqpSinkSettings)
...
+        pull(in)
+      }
+
+      setHandler(in,
+        new InHandler {
+          override def onPush(): Unit = {
+            val elem = grab(in)
+            channel.basicPublish(
+              "",
+              elem.props.map(_.getReplyTo).get,
```
johanandren: What happens if no `replyTo` is set? Should it discard such messages and log a warning or maybe fail the stage even, or maybe fallback to the destination defined in the settings?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
```diff
@@ -0,0 +1,179 @@
...
+import com.rabbitmq.client.AMQP.BasicProperties
+import com.rabbitmq.client.{DefaultConsumer, Envelope, ShutdownSignalException}
+
+import scala.collection.mutable
+import scala.concurrent.{Future, Promise}
+
+object AmqpRpcFlowStage {
+
+  /**
+    * Internal API
```
johanandren: No need to mark it Internal when it is `private` as it will not be accessible anyways. We use it when something is `private[alpakka]` for example, since that does not have a Java access control counterpart and will be accessible from Java.

docs/src/main/paradox/amqp.md
```diff
@@ -115,6 +115,17 @@ We merge all sources into one and add the index of the source to all incoming me
 
 Such sink and source can be started the same way as in the previous example.
 
+### Using rabbitmq as an RPC mechanism
+
+If you have remote workers that you want to incorporate into a stream, you can do it using rabbit RPC workflow [RabbitMQ RPC](https://www.rabbitmq.com/tutorials/tutorial-six-python.html)
```
johanandren: Link to the Java one instead? ;) https://www.rabbitmq.com/tutorials/tutorial-six-java.html

docs/src/main/paradox/amqp.md
```diff
@@ -115,6 +115,17 @@ We merge all sources into one and add the index of the source to all incoming me
...
+
+Scala
+: @@snip (../../../../amqp/src/test/scala/akka/stream/alpakka/amqp/scaladsl/AmqpConnectorsSpec.scala) { #create-rpc-flow }
+
+
+Scala
+: @@snip (../../../../amqp/src/test/scala/akka/stream/alpakka/amqp/scaladsl/AmqpConnectorsSpec.scala) { #run-rpc-flow }
```
johanandren: Include Java samples as well?

amqp/src/main/scala/akka/stream/alpakka/amqp/javadsl/AmqpRpcFlow.scala
Falmarri: Sounds good. I've always been awful at writing docs. I'll give it a shot though.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
Falmarri: Good point. Any thoughts on how this should be handled? Make sure at least one element has been emitted?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala
Falmarri: I'm open to opinions. I think the best would be to remove the ability to configure a destination in the settings. I suppose it could be an option for failing the stage or discarding the element?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
Falmarri: Alright. I was just copying existing code

johanandren:
Seems like something wrong with the doc changes, you can run them locally using `docs/paradox` from sbt to verify that your changes are correct before pushing.

Falmarri:
I don't think that failure was from me

johanandren:
I think it was a timing issue on the ci-server, restarted the failed job.

Falmarri:
Any comments on this?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
```diff
@@ -0,0 +1,183 @@
...
+            channel.basicPublish(
+              exchange,
+              routingKey,
+              elem.mandatory,
+              elem.immediate,
+              props,
+              elem.bytes.toArray
+            )
+
+            // TODO: This is pretty expensive just to see if a message specifies how many responses it should be expecting
```
patriknw: I agree, try to rewrite with something more efficient (less scala idiomatic)

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala
```diff
@@ -48,32 +49,136 @@ final class AmqpSinkStage(settings: AmqpSinkSettings)
...
+              elem.mandatory,
+              elem.immediate,
+              elem.props.orNull,
+              elem.bytes.toArray
+            )
+            pull(in)
+          }
+        }
+      )
+    }, promise.future)
```
patriknw: are we sure it's completed somehow for all cases?

amqp/src/main/scala/akka/stream/alpakka/amqp/javadsl/AmqpRpcFlow.scala
```diff
@@ -0,0 +1,54 @@
...
+   * over a rabbitmq message bus. This will create a private queue, and add the reply-to header to messages sent out.
+   *
+   * This stage materializes to a Future[String], which is the name of the private exclusive queue used for RPC communication.
+   *
+   * @param repliesPerMessage The number of responses that should be expected for each message placed on the queue. This
+   *                            can be overridden per message by including `expectedReplies` in the the header of the [[OutgoingMessage]]
+   */
+  def create(settings: AmqpSinkSettings,
+             bufferSize: Int,
+             repliesPerMessage: Int): javadsl.Flow[OutgoingMessage, IncomingMessage, Future[String]] =
```
patriknw: javadsl should use `CompletionStage` instead of `Future`

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala
Falmarri: Is there a better way to know for sure? https://gitter.im/akka/dev?at=5897b6246018ccd652745a04

Falmarri:
I don't think this failure was because of me

johanandren:
Yeah, seems like a timing issue in other tests

Falmarri:
Can this be retested and accepted/commented on?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala
johanandren: `postStop` but then you won't know if the stage failed or completed.

johanandren: Could be a good idea to put a guard in there (`postStop`) try-failing the promise with an exception and "stage stopped unexpectedly" so that it does not happen silently.

ktoso:
PLS BUILD

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
```diff
@@ -0,0 +1,188 @@
...
+      override def connectionFactoryFrom(settings: AmqpConnectionSettings) = stage.connectionFactoryFrom(settings)
+
+      override def whenConnected(): Unit = {
+        import scala.collection.JavaConverters._
+        val shutdownCallback = getAsyncCallback[Option[ShutdownSignalException]] {
+          case Some(ex) =>
+            promise.tryFailure(ex)
+            failStage(ex)
+          case None =>
+            promise.trySuccess("")
```
johanandren: If it wasn't completed with a name already, shouldn't this be a failure as well?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
```diff
@@ -0,0 +1,188 @@
...
+
+      setHandler(
+        in,
+        new InHandler {
+          // We don't want to finish since we're still waiting
+          // on incoming messages from rabbit. However, if we
+          // haven't processed a message yet, we do want to complete
+          // so that we don't hang.
+          override def onUpstreamFinish(): Unit =
+            if (outstandingMessages == 0) super.onUpstreamFinish()
```
johanandren: Shouldn't we check if queue is empty as well? In case we got responses but downstream was backpressuring.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
Falmarri: Good point...

Falmarri: Oh, that's right. I guess it depends. the case None is when a queue gets deleted or something like that. I suppose that is technically a failure, huh.

Falmarri:
Ping on this please? @johanandren

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
```diff
@@ -0,0 +1,190 @@
...
+        )
+        promise.success(queueName)
+      }
+
+      def handleDelivery(message: IncomingMessage): Unit =
+        if (isAvailable(out)) {
+          pushAndAckMessage(message)
+        } else {
+          if (queue.size + 1 > bufferSize) {
+            failStage(new RuntimeException(s"Reached maximum buffer size $bufferSize"))
```
johanandren: Hmm, shouldn't we rather back pressure if there is no room in the buffer for the response?

Falmarri: I took this from the AmqpSourceStage that currently exists. I believe the idea is that the else should never happen. This would be a situation where for some reason we have more responses coming in from rabbit than we asked for.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala
```diff
@@ -48,32 +49,143 @@ final class AmqpSinkStage(settings: AmqpSinkSettings)
...
+  }
 
   override def toString: String = "AmqpSink"
 }
+
+object AmqpReplyToStage {
+
+  private val defaultAttributes =
+    Attributes.name("AmqpReplyToSink").and(ActorAttributes.dispatcher("akka.stream.default-blocking-io-dispatcher"))
```
Falmarri: What is the right name?

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpSink.scala
```diff
@@ -3,23 +3,33 @@
...
-   * Scala API: Creates an [[AmqpSink]] that accepts [[OutgoingMessage]] elements.
+   * Scala API:
+   *
+   * Connects to an AMQP server upon materialization and sends incoming messages to the server.
+   * Each materialized sink will create one connection to the broker. This stage sends messages to
+   * the queue named in the replyTo options of the message instead of from settings declared at construction.
+   *
    */
-  def apply(settings: AmqpSinkSettings): Sink[OutgoingMessage, NotUsed] =
+  def apply(settings: AmqpSinkSettings): Sink[OutgoingMessage, Future[Done]] =
```
Falmarri: Just following what's already in the file. Should i explicitly type everything in this file?

johanandren: Just document in what circumstances that `Future[Done]` is completed/failed in the scaladoc.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala
johanandren: Thie name of the stage is `AmqpReplyToStage`, no? `AmqpReplyToSink` is an old name or copy-pasta, I think.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
johanandren: Alright, I think you are right. We shouldn't have asked for an element if the buffer didn't have space for it it or out was available.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala
Falmarri: Oh. I meant for the object to be AmqpReplyToSinkStage. Basically the existing AmqpSinkStage, but it's a ReplyTo sink.

johanandren:
Needs a rebase now, not sure what got merged that changed `AqmpSinkStage`...

johanandren:
And a compile after the rebase to update formatting, which is causing the build to fail now. Almost there!

Falmarri:
Any other comments on this?

2m:
All green. Great job!

PR #163, opened 2017-01-23
Limit parallelism for the SqsSource

### LINKED ISSUE #135: SQS Source has unbounded parallelism
Every time downstream pulls when the buffer is not full, a new async request is spawned using `AmazonSQSAsyncClient`, thus ignoring the `longPollingDuration` or the `maxBufferSize` configuration. This makes it unusable and dangerous.

I understand that throughput of this Source is achieved with parallelism of requests of `AmazonSQSAsyncClient`. This client can be instantiated with a custom thread pool but if the downstream is too fast, this will effectively starve other usages of the same client instance. For example:

```
implicit val sqsClient : AmazonSQSAsyncClient = ???

alpakkaSqsSource(queueName1) ~> } ~> fastFlow ~> deleteSqsMessage
alpakkaSqsSource(queueName2) ~> }
```

The sqsSources will exhaust all the resources of the sqsClient. This can be easily fixed using a different sqsClient, **but IMHO this should be properly advertised on the documentation** because it's clearly not how you'd use this api:

```
alpakkaSqsSource(queueName1)(sqsClient1) ~> } ~> fastFlow ~> deleteSqsMessage(sqsClient3)
alpakkaSqsSource(queueName2)(sqsClient2) ~> }
```

Now, we've avoided starvation BUT if we check our SQS metrics on AWS console we'll find astronomic polling rates (which are billed to your AWS account). This happens because the Source polls whenever its buffer it's not full on every downstream pull, as I previously said.

One would expect that the Source honors the `longPollingDuration` and the `maxBufferSize` configuration in order to control its internal logic and throughput, but it fails under the most trivial scenarios:

## Scenario 1:
Max buffer: 100
Current buffer: 85
Batch size: 10
Downstream pulls: 2

Outcome: 2 requests are performed, when only 1 of the responses would fit in the local buffer

## Scenario 2:
Max buffer: 100
Current buffer: 0
Batch size: 10
Downstream pulls: 2

Outcome: 2 requests are performed, even when the application has no clue about the current messages on the server, making  `longPollingDuration` parameter useless

---

Solution? 
Source may keep local state to track if there are pending async requests and the latest server response.
I think that would be enough, and it'd make the Source behave as expected.

dpfeiffer:
Fixes #135 

The solution to the issue, as proposed by @aserrallerios was to create a custom `AmazonSQSAsyncClient` with a thread pool which is limited to `maxBufferSize / maxBatchSize` threads. To make sure the maxBufferSize is not exceeded, I keep track of the number of current requests and only allow new requests if `maxBatchSize` elements would fit into the buffer for another request.

To further optimize AWS resource usage, as soon as there is a request which returns 0 messages, I assume that the queue is empty and limit the number of concurrent requests to 1. This makes sure that there will be only one request long polling for new messages.

### DIFF
--- a/docs/src/main/paradox/sqs.md
+++ b/docs/src/main/paradox/sqs.md
@@ -35,7 +35,7 @@ Gradle
 
 ## Usage
 
-Sources provided by this connector need a prepared `AmazonSQSAsyncClient` to load messages from a queue.
+Sources and Sinks provided by this connector need a prepared `AmazonSQSAsyncClient` to load messages from a queue.
 
 Scala
 : @@snip (../../../../sqs/src/test/scala/akka/stream/alpakka/sqs/scaladsl/DefaultTestContext.scala) { #init-client }
@@ -78,6 +78,22 @@ Options:
  - `maxBufferSize` - internal buffer size used by the `Source`. Default: 100 messages
  - `waitTimeSeconds` - the duration for which the call waits for a message to arrive in the queue before
     returning (see `WaitTimeSeconds` in AWS docs). Default: 20 seconds
+
+Be aware that the `SqsSource` runs multiple requests to Amazon SQS in parallel. The maximum number of concurrent
+requests is limited by `parallelism = maxBufferSize / maxBatchSize`. E.g.: By default `maxBatchSize` is set to 10 and
+`maxBufferSize` is set to 100 so at the maximum, `SqsSource` will run 10 concurrent requests to Amazon SQS. `AmazonSQSAsyncClient`
+uses a fixed thread pool with 50 threads by default. To tune the thread pool used by
+`AmazonSQSAsyncClient` you can supply a custom `ExecutorService` on client creation.
+
+Scala
+: @@snip (../../../../sqs/src/test/scala/akka/stream/alpakka/sqs/scaladsl/SqsSourceSpec.scala) { #init-custom-client }
+
+Java
+: @@snip (../../../../sqs/src/test/java/akka/stream/alpakka/sqs/javadsl/SqsSourceTest.java) { #init-custom-client }
+
+Please make sure to configure a big enough thread pool to avoid resource starvation. This is especially important
+if you share the client between multiple Sources, Sinks and Flows. For the SQS Sinks and Sources the sum of all
+`parallelism` (Source) and `maxInFlight` (Sink) must be less than or equal to the thread pool size.
 
 ### Stream messages to a SQS queue
 
--- a/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
+++ b/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
@@ -4,28 +4,20 @@
 package akka.stream.alpakka.sqs
 
 import java.util
-import java.util.concurrent.Executors
 
 import akka.stream.stage.{GraphStage, GraphStageLogic, OutHandler}
 import akka.stream.{Attributes, Outlet, SourceShape}
-import com.amazonaws.auth.{AWSCredentials, DefaultAWSCredentialsProviderChain}
 import com.amazonaws.handlers.AsyncHandler
-import com.amazonaws.services.sqs.AmazonSQSAsyncClient
+import com.amazonaws.services.sqs.AmazonSQSAsync
 import com.amazonaws.services.sqs.model.{Message, ReceiveMessageRequest, ReceiveMessageResult}
 
 import scala.collection.JavaConverters._
 
 object SqsSourceSettings {
-  val Defaults = SqsSourceSettings(20, 100, 10, None)
+  val Defaults = SqsSourceSettings(20, 100, 10)
 
-  def create(waitTimeSeconds: Int,
-             maxBufferSize: Int,
-             maxBatchSize: Int,
-             credentials: AWSCredentials): SqsSourceSettings =
-    SqsSourceSettings(waitTimeSeconds, maxBufferSize, maxBatchSize, Some(credentials))
-
-  private def create(waitTimeSeconds: Int, maxBufferSize: Int, maxBatchSize: Int): SqsSourceSettings =
-    SqsSourceSettings(waitTimeSeconds, maxBufferSize, maxBatchSize, None)
+  def create(waitTimeSeconds: Int, maxBufferSize: Int, maxBatchSize: Int): SqsSourceSettings =
+    SqsSourceSettings(waitTimeSeconds, maxBufferSize, maxBatchSize)
 
 }
 
@@ -33,8 +25,7 @@ object SqsSourceSettings {
 final case class SqsSourceSettings(
     waitTimeSeconds: Int,
     maxBufferSize: Int,
-    maxBatchSize: Int,
-    credentials: Option[AWSCredentials]
+    maxBatchSize: Int
 ) {
   require(maxBatchSize <= maxBufferSize, "maxBatchSize must be lower or equal than maxBufferSize")
   // SQS requirements
@@ -45,7 +36,8 @@ final case class SqsSourceSettings(
 }
 //#SqsSourceSettings
 
-final class SqsSourceStage(queueUrl: String, settings: SqsSourceSettings) extends GraphStage[SourceShape[Message]] {
+final class SqsSourceStage(queueUrl: String, settings: SqsSourceSettings)(implicit sqsClient: AmazonSQSAsync)
+    extends GraphStage[SourceShape[Message]] {
 
   val out: Outlet[Message] = Outlet("SqsSource.out")
   override val shape: SourceShape[Message] = SourceShape(out)
@@ -53,12 +45,7 @@ final class SqsSourceStage(queueUrl: String, settings: SqsSourceSettings) extend
   override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
     new GraphStageLogic(shape) {
 
-      private val credentials =
-        settings.credentials.getOrElse(DefaultAWSCredentialsProviderChain.getInstance().getCredentials)
-
       private val maxConcurrency = settings.maxBufferSize / settings.maxBatchSize
-      private val threadPool = Executors.newFixedThreadPool(maxConcurrency)
-      private val sqsClient = new AmazonSQSAsyncClient(credentials, threadPool)
       private val buffer = new util.ArrayDeque[Message]()
 
       private val successCallback = getAsyncCallback[ReceiveMessageResult](handleSuccess)
@@ -122,8 +109,5 @@ final class SqsSourceStage(queueUrl: String, settings: SqsSourceSettings) extend
             receiveMessages()
           }
       })
-
-      override def postStop(): Unit =
-        sqsClient.shutdown() //shuts down the underlying thread pool as well
     }
 }
--- a/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
+++ b/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
@@ -17,6 +17,16 @@ import scala.collection.JavaConverters._
 
 object SqsSourceSettings {
   val Defaults = SqsSourceSettings(20, 100, 10, None)
+
+  def create(waitTimeSeconds: Int,
+             maxBufferSize: Int,
+             maxBatchSize: Int,
+             credentials: AWSCredentials): SqsSourceSettings =
+    SqsSourceSettings(waitTimeSeconds, maxBufferSize, maxBatchSize, Some(credentials))
+
+  private def create(waitTimeSeconds: Int, maxBufferSize: Int, maxBatchSize: Int): SqsSourceSettings =
+    SqsSourceSettings(waitTimeSeconds, maxBufferSize, maxBatchSize, None)
+
 }
 
 //#SqsSourceSettings
--- a/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
+++ b/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
@@ -36,10 +36,12 @@ final case class SqsSourceSettings(
     maxBatchSize: Int,
     credentials: Option[AWSCredentials]
 ) {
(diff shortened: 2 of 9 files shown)


aserrallerios:
I really like the new constructor :+1:

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -4,44 +4,77 @@
...
 
 //#SqsSourceSettings
-final case class SqsSourceSettings(waitTimeSeconds: Int, maxBufferSize: Int, maxBatchSize: Int) {
+final case class SqsSourceSettings(
+    waitTimeSeconds: Int,
+    maxBufferSize: Int,
+    maxBatchSize: Int,
+    credentials: Option[AWSCredentials]
+) {
   require(maxBatchSize <= maxBufferSize)
```
patriknw: include error message string in those checks

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -4,44 +4,77 @@
...
   override val shape: SourceShape[Message] = SourceShape(out)
 
   override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
     new GraphStageLogic(shape) {
 
+      private val credentials =
+        settings.credentials.getOrElse(DefaultAWSCredentialsProviderChain.getInstance().getCredentials)
+
+      private val maxConcurrency = settings.maxBufferSize / settings.maxBatchSize
+      private val threadPool = Executors.newFixedThreadPool(maxConcurrency)
```
patriknw: shutdown the thread pool in `postStop`

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -4,44 +4,77 @@
...
   require(maxBatchSize <= maxBufferSize)
   // SQS requirements
   require(waitTimeSeconds >= 0 && waitTimeSeconds <= 20)
   require(maxBatchSize >= 1 && maxBatchSize <= 10)
 }
 //#SqsSourceSettings
 
-final class SqsSourceStage(queueUrl: String, settings: SqsSourceSettings, sqsClient: AmazonSQSAsyncClient)
-    extends GraphStage[SourceShape[Message]] {
+final class SqsSourceStage(queueUrl: String, settings: SqsSourceSettings) extends GraphStage[SourceShape[Message]] {
```
patriknw: no need for sharing the same `sqsClient` between different stages?

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -4,44 +4,77 @@
...
 
   override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
     new GraphStageLogic(shape) {
 
+      private val credentials =
+        settings.credentials.getOrElse(DefaultAWSCredentialsProviderChain.getInstance().getCredentials)
+
+      private val maxConcurrency = settings.maxBufferSize / settings.maxBatchSize
+      private val threadPool = Executors.newFixedThreadPool(maxConcurrency)
+      private val sqsClient = new AmazonSQSAsyncClient(credentials, threadPool)
```
patriknw: do we have to stop the sqsClient?

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -61,13 +94,16 @@ final class SqsSourceStage(queueUrl: String, settings: SqsSourceSettings, sqsCli
 
       def handleSuccess(result: ReceiveMessageResult): Unit = {
 
+        currentRequests = currentRequests - 1
+        maxCurrentConcurrency = if (result.getMessages.size() == 0) 1 else maxConcurrency
```
patriknw: `result.isEmpty` or `result.getMessages.isEmpty` ?

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -61,13 +94,16 @@ final class SqsSourceStage(queueUrl: String, settings: SqsSourceSettings, sqsCli
...
+
         result.getMessages.asScala.reverse.foreach(buffer.addFirst)
```
patriknw: why isn't the `ArrayDeque` used as a FIFO `Queue`, i.e. using `offer` and `poll`?

sqs/src/test/java/akka/stream/alpakka/sqs/javadsl/SqsSourceTest.java
```diff
@@ -13,6 +14,7 @@
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
+import scala.Some;
```
patriknw: not used?

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
dpfeiffer: I'll try to introduce another way to initialize the `SqsSource` with a `AmazonSQSAsync` instead of only creating one explicitly. @aserrallerios came up with the idea to initialize the client within the `SqsSourceStage` so the threadpool can be configured properly. Parallelism is now tied to `maxBufferSize` and `maxBatchSize` so the maximum size for the thread pool can be defined with `maxBufferSize / maxBatchSize`.

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
drewhk: Can we have an Executor from a dispatcher? That would be more flexible, plus it allows sharing the executor when it makes sense.

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
drewhk: hence the postStop at L126?

dpfeiffer: as @drewhk already mentioned the client is being stopped in `postStop`

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
dpfeiffer: The javadoc for `AmazonSQSAsyncClient` says following: 

> Shuts down the client, releasing all managed resources.This includes forcibly terminating all pending asynchronous service calls.

Should I rely on that or shutdown everything manually?

dpfeiffer:
@patriknw @aserrallerios I finally decided to go back to the external `AmazonSQSAsnyc` because it would overcomplicate handling the internal creation of a client. When you take a look at other places where `AmazonX` clients are passed, they chose the same way, even if they (as in my case) had to handle parallelism internally. I think that especially the AWS modules should stick to the same way of handling that (s3 and dynamodb don't use the aws-java-sdk as far as I can see that)

I also don't think that it will hurt too much if the client uses a too big or too small thread pool (in comparison to what the maxConcurrency for the SqsSource is). Maybe we should add information about the internal caclulation (`maxConcurrency = maxBufferSize / maxBatchSize`) to the documentation so users know about it.

sqs/src/test/scala/akka/stream/alpakka/sqs/scaladsl/SqsSourceSpec.scala
```diff
@@ -35,7 +37,7 @@ class SqsSourceSpec extends AsyncWordSpec with ScalaFutures with Matchers with D
       }
 
       //#run
-      SqsSource(queue).take(100).runWith(Sink.seq).map(_ should have size 100)
+      SqsSource(queue, sqsSourceSettings).take(100).runWith(Sink.seq).map(_ should have size 100)
```
viktorklang: The assertion is done within a Future.

aserrallerios:
I'm OK with it, as long as the documentation explicitly states that the provided `AmazonSQSAsnyc` should have higher parallelism than `maxBufferSize / maxBatchSize`. In addition, if the same `AmazonSQSAsnyc` instance is used for several alpakka `Sources` or for handling direct interactions with SQS (for example, delete messages), the parallelism should be higher than the sum of _all_ of them, otherwise some of them may starve.

I think that the easiest would be to add some opinionated usage examples.

docs/src/main/paradox/sqs.md
```diff
@@ -79,6 +79,22 @@ Options:
...
+Scala
+: @@snip (../../../../sqs/src/test/scala/akka/stream/alpakka/sqs/scaladsl/SqsSourceSpec.scala) { #init-custom-client }
+
+Java
+: @@snip (../../../../sqs/src/test/java/akka/stream/alpakka/sqs/javadsl/SqsSourceTest.java) { #init-custom-client }
+
+Please make sure to configure a big enough thread pool to avoid resource starvation. This is especially important
+if you share the client between multiple Sources, Sinks and Flows. For the SQS Sinks and Sources the sum of all
+`parallelism` (Source) and `maxInFlight` (Sink) must be less than or equal to the thread pool size.
+
```
ktoso: the added docs seem to address the discussion in the main issue well 👍

sqs/src/test/java/akka/stream/alpakka/sqs/javadsl/SqsSourceTest.java
```diff
@@ -77,4 +79,29 @@ public void streamFromQueue() throws Exception {
...
+
+        //#init-custom-client
+        final ExecutorService executor = Executors.newFixedThreadPool(10);
+        final AmazonSQSAsyncClient sqsClient = new AmazonSQSAsyncClient(credentials, executor)
+            .withEndpoint("http://localhost:9324");
+        //#init-custom-client
+
+        final String queueUrl = randomQueueUrl();
+
+        final List<String> input = IntStream.range(0, 100).boxed().map(i -> String.format("alpakka-%s", i)).collect(Collectors.toList());
```
ktoso: why the `boxed()`?

ktoso: The entire thing could just be:

`Source.range(0, 100).runForeach(i -> String.format("alpakka-%s", i), mat)` 

😉 

Could you replace that? Much simpler IMO

sqs/src/test/scala/akka/stream/alpakka/sqs/scaladsl/SqsSourceSpec.scala
```diff
@@ -24,6 +28,20 @@ class SqsSourceSpec extends AsyncWordSpec with ScalaFutures with Matchers with D
...
+      //#init-custom-client
+      val executor: ExecutorService = Executors.newFixedThreadPool(10)
+      implicit val sqsClient: AmazonSQSAsyncClient =
+        new AmazonSQSAsyncClient(credentials, executor).withEndpoint[AmazonSQSAsyncClient]("http://localhost:9324")
+      //#init-custom-client
+
+      val queue = randomQueueUrl()
+      sqsClient.sendMessage(queue, "alpakka")
+
+      SqsSource(queue, sqsSourceSettings).take(1).runWith(Sink.seq).map(_.map(_.getBody) should contain("alpakka"))
```
ktoso: FYI: `runWith(Sink.head, mat)` :)

dpfeiffer:
@ktoso Thanks for reviewing I hope I got everything you mentioned.

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -8,25 +8,35 @@ import java.util
...
+    maxBatchSize: Int
+) {
+  require(maxBatchSize <= maxBufferSize, "maxBatchSize must be lower or equal than maxBufferSize")
   // SQS requirements
-  require(waitTimeSeconds >= 0 && waitTimeSeconds <= 20)
-  require(maxBatchSize >= 1 && maxBatchSize <= 10)
+  require(waitTimeSeconds >= 0 && waitTimeSeconds <= 20,
+    "Invalid value for waitTimeSeconds. Requirement: 0 <= waitTimeSeconds <= 20 ")
+  require(maxBatchSize >= 1 && maxBatchSize <= 10,
+    "Invalid value for maxBatchSize. Requirement: 1 <= waitTimeSeconds <= 10 ")
```
ktoso: Typo in the message, should be:

```
    "Invalid value for maxBatchSize. Requirement: 1 <= maxBatchSize <= 10 ")
```

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -35,13 +45,25 @@ final class SqsSourceStage(queueUrl: String, settings: SqsSourceSettings, sqsCli
   override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
     new GraphStageLogic(shape) {
 
+      private val maxConcurrency = settings.maxBufferSize / settings.maxBatchSize
```
ktoso: good

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -35,13 +45,25 @@ final class SqsSourceStage(queueUrl: String, settings: SqsSourceSettings, sqsCli
...
+      private var currentRequests = 0
+
+      private def canReceiveNewMessages = {
+        val currentFreeRequests = (settings.maxBufferSize - buffer.size) / settings.maxBatchSize
+        currentFreeRequests > currentRequests && maxCurrentConcurrency > currentRequests
+      }
+
       def receiveMessages(): Unit = {
 
+        currentRequests = currentRequests + 1
```
ktoso: Would `inflightRequests` easier to grasp what it means for future maintanance?

PR #164, opened 2017-01-24
Add mongodb sink

stephennancekivell:
Based off #115 #148  from @iozozturk, which it should merge well with.

No Java DSL as this is based off of the mongo-scala-driver. A java connecter should possibly be from the mongo-java-driver.

Pending Documentation

### DIFF
--- a/docs/src/main/paradox/cassandra.md
+++ b/docs/src/main/paradox/cassandra.md
@@ -1,6 +1,6 @@
 # Cassandra Connector
 
-The Cassandra connector provides a way to provide the result of a Cassandra query as a stream of rows.
+The Cassandra connector allows you to read and write to Cassandra. You can query a stream of rows from @scaladoc[CassandraSource](akka.stream.alpakka.cassandra.scaladsl.CassandraSource$) or use prepared statements to insert or update with @scaladoc[CassandraSink](akka.stream.alpakka.cassandra.scaladsl.CassandraSink$).
 
 ## Artifacts
 
--- a/docs/src/main/paradox/mongodb.md
+++ b/docs/src/main/paradox/mongodb.md
@@ -2,6 +2,8 @@
 
 The MongoDB connector allows you to read and save documents. You can query as a stream of documents from @scaladoc[MongoSource](akka.stream.alpakka.mongodb.scaladsl.MongoSource$) or update documents in a collection with @scaladoc[MongoSink](akka.stream.alpakka.mongodb.scaladsl.MongoSink$).
 
+This connector is based off the [mongo-scala-driver](https://github.com/mongodb/mongo-scala-driver) and does not have a java interface.
+
 ## Artifacts
 
 @@dependency [sbt,Maven,Gradle] {
@@ -57,7 +59,7 @@ Scala
 
 #### Update
 
-We can update documents with a Source of @scaladoc[DocumentUpdate](akka.stream.alpakka.mongodb.scaladsl.MongoSink$#DocumentUpdate) which is a filter and a update definition.
+We can update documents with a Source of @scaladoc[DocumentUpdate](akka.stream.alpakka.mongodb.scaladsl.DocumentUpdate) which is a filter and a update definition.
 Use either @scaladoc[updateOne](akka.stream.alpakka.mongodb.scaladsl.MongoSink$#updateOne) or @scaladoc[updateMany](akka.stream.alpakka.mongodb.scaladsl.MongoSink$#updateMany) if the filter should target one or many documents.
 
 Scala
--- a/docs/src/main/paradox/mongodb.md
+++ b/docs/src/main/paradox/mongodb.md
@@ -1,6 +1,6 @@
 # MongoDB Connector
 
-The MongoDB connector provides a way to provide the result of a MongoDB query as a stream of documents.
+The MongoDB connector allows you to read and save documents. You can query as a stream of documents from @scaladoc[MongoSource](akka.stream.alpakka.mongodb.scaladsl.MongoSource$) or save documents to a collection with @scaladoc[MongoSink](akka.stream.alpakka.mongodb.scaladsl.MongoSink$).
 
 ## Artifacts
 
@@ -36,7 +36,14 @@ And finally we can run it.
 Scala
 : @@snip ($alpakka$/mongodb/src/test/scala/akka/stream/alpakka/mongodb/MongoSourceSpec.scala) { #run-source }
 
-Here we used a basic sink to complete the stream by collecting all of the stream elements to a collection. The power of streams comes from building larger data pipelines which leverage backpressure to ensure efficient flow control. Feel free to edit the example code and build @extref[more advanced stream topologies](akka-docs:scala/stream/stream-introduction).
+Here we used a basic sink to complete the stream by collecting all of the stream elements to a collection.
+
+### Sink Usage
+
+We can take a Source of documents and save them to a mongo collection using MongoSink.
+
+Scala
+: @@snip (../../../../mongodb/src/test/scala/akka/stream/alpakka/mongodb/MongoSinkSpec.scala) { #create-sink }
 
 ### Running the example code
 
--- a/docs/src/main/paradox/mongodb.md
+++ b/docs/src/main/paradox/mongodb.md
@@ -26,7 +26,9 @@ Scala
 
 This is all preparation that we are going to need.
 
-### Source Usage
+### Source and Flow Usage
+
+Each of these sink factory methods have a corresponding factory in @scaladoc[insertOne](akka.stream.alpakka.mongodb.scaladsl.MongoFlow) which will emit the written document or result of the operation downstream.
 
 Let's create a source from a MongoDB collection observable, which can optionally take a filter.
 
--- a/docs/src/main/paradox/mongodb.md
+++ b/docs/src/main/paradox/mongodb.md
@@ -1,6 +1,6 @@
 # MongoDB Connector
 
-The MongoDB connector allows you to read and save documents. You can query as a stream of documents from @scaladoc[MongoSource](akka.stream.alpakka.mongodb.scaladsl.MongoSource$) or save documents to a collection with @scaladoc[MongoSink](akka.stream.alpakka.mongodb.scaladsl.MongoSink$).
+The MongoDB connector allows you to read and save documents. You can query as a stream of documents from @scaladoc[MongoSource](akka.stream.alpakka.mongodb.scaladsl.MongoSource$) or update documents in a collection with @scaladoc[MongoSink](akka.stream.alpakka.mongodb.scaladsl.MongoSink$).
 
 ## Artifacts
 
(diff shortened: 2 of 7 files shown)


jrudolph:
Hi @stephennancekivell, thanks for the contribution and sorry for keeping this around for so long.

This will probably need to be rebased and adapted once #148 was merged so let's revisit this once #148 has been merged.

peterschrott:
Hi @stephennancekivell, I just cherrypicked your Mongo Sink. I must say: Very good work! 

There are three features I am missing:
* I can not provide the database name in the connection URL (I need to provide that, otherwise I get some authentication issues)
* I can not provide authentication for database
I am thinking about something like: `MongoClient(s"mongodb://$user:$pw@$dbAddress:$dbPort/$dbName" )`

* In version 2.0.0 of the Mongo Scala Driver they added a very nice feature, the Macros: http://mongodb.github.io/mongo-scala-driver/2.0/bson/macros/
Therefore, It is possible to not only write Documents to the Mongo, you also can write Case Classes after registering them. For that I adjusted the `MongoSink` and simply typed it:
```
def apply[T](parallelism: Int, collection: MongoCollection[T])
            (implicit ex: ExecutionContext): Sink[T, Future[Done]] = {
  Flow[T].mapAsyncUnordered(parallelism)(element => collection.insertOne(element).toFuture())
    .toMat(Sink.ignore)(Keep.right)
}
```
This works like charm!

stephennancekivell:
Thanks @peterschrott,

Regarding points 1&2, I think it'd be better for this alpakka library to not be responsible for creating the MongoClient or MongoDatabase and just take them in as method parameters. The mongo library already provides many ways to create each. This would address your issues. I think we talked about this in #148  at one point.

Regarding the Generic Case Class sink, I agree that would be nice to add but shouldnt replace saving raw Documents. Im surprised to see they didnt use shapeless for that, but thats up to them.

Anyway I wouldnt want to extend this to support their macro until something gets merged in.

stephennancekivell:
I've updated this to fit with whats currently merged in.

I wonder if we should support more of the mongo collection methods. Currently its only insertOne. But could have insertMany, updateOne, updateMany.

stephennancekivell:
I've completed adding the update and delete methods along with documentation.

I think this is ready to be reviewed or merged in. :)

raboof:
I'm not completely up to speed, but noticed that the mongo driver already has some `org.reactivestreams` support (e.g. https://github.com/mongodb/mongo-scala-driver/blob/master/examples/src/test/scala/rxStreams/Implicits.scala). Perhaps it'd be good to describe what this alpakka module offers on top of that?

stephennancekivell:
Hi @raboof, Thanks for assigning me the ticket. Can we merge it in now. I dont have seem to have permission to do that.

Regarding your earlier question that `examples/src/test/scala/rxStreams` is what the mongo source is based off, but that alone is not a akka streams integration. This connector takes it that little bit further. Also those examples are not a published library. This PR is for adding a sink which cannot be done with their rxStreams example.

Sorry for the late reply.
Thanks.

gosubpl:
I am a bit worried that the Mongo Sinks don't materialize the value of respective operations (see: http://mongodb.github.io/mongo-scala-driver/2.1/scaladoc/org/mongodb/scala/MongoCollection.html ). Would be great if one could get the update result, not only done, as it is especially important in cases of `upsert` operations.

stephennancekivell:
Do you have any suggestion as to how we could do that @gosubpl ?

Perhaps that feature could come in a later pr. I think there are a lot of use cases where it isn't s concern.

stephennancekivell:
Thanks for the review @johanandren. Good points. Ill do that in the next day or two.

gosubpl:
I think we can add the materialized values in a separate PR. I would be very happy to submit one :)

ennru:

docs/src/main/paradox/cassandra.md
```diff
@@ -1,6 +1,6 @@
 # Cassandra Connector
 
-The Cassandra connector provides a way to provide the result of a Cassandra query as a stream of rows.
+The Cassandra connector allows you to read and write to Cassandra. You can query a stream of rows from @scaladoc[CassandraSource](akka.stream.alpakka.cassandra.scaladsl.CassandraSource$) or use prepared statements to insert or update with @scaladoc[CassandraSink](akka.stream.alpakka.cassandra.scaladsl.CassandraSink$).
```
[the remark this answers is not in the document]
stephennancekivell: I dont see any of the other connectors doing that. It renders as a inline link. Im not sure about how to style it. It would be weird to interrupt every sentence with java-link & scala-link.

Do you have any suggestion? I'd rather leave it.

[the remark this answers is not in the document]
raboof: It looks weird only in the markdown source, not in the produced page, right?

I would be cool to have a `@unidoc` directive that figures out how to link to both Scala and Java docs automatically. Preferably even checking that those classes indeed exist? Way outside the scope of this PR though :).

mongodb/src/main/scala/akka/stream/alpakka/mongodb/scaladsl/MongoSink.scala
```diff
@@ -0,0 +1,61 @@
...
+    run[Bson](parallelism)(collection.deleteOne)
+
+  def deleteMany(parallelism: Int, collection: MongoCollection[Document]): Sink[Bson, Future[Done]] =
+    run[Bson](parallelism)(collection.deleteMany)
+
+  private def run[A](parallelism: Int)(fn: A => SingleObservable[_]): Sink[A, Future[Done]] =
+    Flow[A]
+      .mapAsyncUnordered(parallelism)(fn(_).toFuture())
+      .toMat(Sink.ignore)(Keep.right)
+}
```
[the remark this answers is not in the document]
stephennancekivell: I've done this. Im not sure its that useful though.

Arent these results only useful if the elements are ordered, which they aren't because of mapAsyncUnordered ?

docs/src/main/paradox/mongodb.md
```diff
@@ -1,6 +1,6 @@
 # MongoDB Connector
 
-The MongoDB connector provides a way to provide the result of a MongoDB query as a stream of documents.
+The MongoDB connector allows you to read and save documents. You can query as a stream of documents from @scaladoc[MongoSource](akka.stream.alpakka.mongodb.scaladsl.MongoSource$) or update documents in a collection with @scaladoc[MongoSink](akka.stream.alpakka.mongodb.scaladsl.MongoSink$).
```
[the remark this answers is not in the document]
stephennancekivell: Done.

mongodb/src/main/scala/akka/stream/alpakka/mongodb/scaladsl/MongoSink.scala
```diff
@@ -0,0 +1,153 @@
...
+   */
+  def insertOne(parallelism: Int, collection: MongoCollection[Document]): Sink[Document, Future[Done]] =
+    insertOneFlow(parallelism, collection).toMat(Sink.ignore)(Keep.right)
+
+  /**
+   * A [[Flow]] that will insert documents into a collection.
+   * @param parallelism number of documents to insert in parallel.
+   * @param collection mongo db collection to insert to.
+   */
+  def insertOneFlow(parallelism: Int, collection: MongoCollection[Document]): Flow[Document, Completed, NotUsed] =
```
johanandren: Make a `MongoFlow` and put these in that without prefix!

mongodb/src/main/scala/akka/stream/alpakka/mongodb/scaladsl/MongoSink.scala
```diff
@@ -0,0 +1,153 @@
...
+   */
+  def insertMany(parallelism: Int, collection: MongoCollection[Document]): Sink[Seq[Document], Future[Done]] =
+    insertManyFlow(parallelism, collection).toMat(Sink.ignore)(Keep.right)
+
+  /**
+   * A [[Flow]] that will insert batches documents into a collection.
+   * @param parallelism number of batches of documents to insert in parallel.
+   * @param collection mongo db collection to insert to.
+   */
+  def insertManyFlow(parallelism: Int, collection: MongoCollection[Document]) =
```
johanandren: Make sure all public API methods have an explicit return type.

docs/src/main/paradox/mongodb.md
```diff
@@ -36,7 +38,38 @@ And finally we can run it.
 Scala
 : @@snip ($alpakka$/mongodb/src/test/scala/akka/stream/alpakka/mongodb/MongoSourceSpec.scala) { #run-source }
 
-Here we used a basic sink to complete the stream by collecting all of the stream elements to a collection. The power of streams comes from building larger data pipelines which leverage backpressure to ensure efficient flow control. Feel free to edit the example code and build @extref[more advanced stream topologies](akka-docs:scala/stream/stream-introduction).
+Here we used a basic sink to complete the stream by collecting all of the stream elements to a collection.
+
+### Sink Usage
```
johanandren: If you do not want to duplicate examples for flow vs sink I think We could call this "Flow and Sink Usage" and then add a paragraph directly below saying something like "Each of these sink factory methods have a corresponding factory in `MongoFlow` which will emit the written document or result of the operation downstream"

mongodb/src/main/scala/akka/stream/alpakka/mongodb/scaladsl/MongoSink.scala
stephennancekivell: 👍  thanks

mongodb/src/main/scala/akka/stream/alpakka/mongodb/scaladsl/MongoSink.scala
stephennancekivell: Great idea.

docs/src/main/paradox/mongodb.md
stephennancekivell: Sounds good.

stephennancekivell:
Thanks @johanandren, I've moved the flows as mentioned and made them return their inputs, so they can be used.

Its a bit annoying how now the Sink methods need to take a execution context. They could almost be a completely seperate implementation. What do you think?

mongodb/src/main/scala/akka/stream/alpakka/mongodb/scaladsl/MongoSink.scala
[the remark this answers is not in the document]
johanandren: If you cannot correlate them they would not be very useful, but doesn't the result coming "out" of the Mongo API, `DeleteResult` for example, contain information about what deletes/updates/inserts happened?

gosubpl:
LGTM, I have just realised that there are `Flow` s that allow one to obtain relevant operation results, and in that case it is natural that `Sink` s don't materialize anything.

mongodb/src/main/scala/akka/stream/alpakka/mongodb/scaladsl/MongoFlow.scala
```diff
@@ -0,0 +1,116 @@
...
+  /**
+   * A [[Flow]] that will insert documents into a collection.
+   * @param parallelism number of documents to insert in parallel.
+   * @param collection mongo db collection to insert to.
+   */
+  def insertOne(parallelism: Int, collection: MongoCollection[Document])(
+      implicit executionContext: ExecutionContext
+  ): Flow[Document, Document, NotUsed] =
+    Flow[Document]
+      .mapAsyncUnordered(parallelism)(doc => collection.insertOne(doc).toFuture().map(_ => doc))
```
patriknw: Isn't the unordered part pretty important to document?
Looks like they are inserted in order but the results (out Document) can be unordered.
Wonder if it wouldn't be performant enough to use ordinary `mapAsync` and avoid the confusion that might be caused by the unordered?

stephennancekivell: Fair point. Ill take it out. Maybe in a later PR someone can speed test it and add back unordered variants.

stephennancekivell:
Thanks guys. As discussed I've just changed the mapAsyncUnordered to mapAsync. If someone really wants it they can add it back in another PR.

docs/src/main/paradox/mongodb.md
johanandren: Still missing this one, I think it is done for merging except for this.

stephennancekivell: Ah sorry, thanks. All fixed up.

ennru:
A long way coming...
Thank you for this contribution.

Issue #165, opened 2017-01-24
FTP - `list` fails to reach nested dirs


eharik:
Current recursive listing will not reach deeply nested folders.  This happens in the case where a nested directory contains no files but only a single directory, i.e.

```
dir1/
  dir2/
    dir3/
      foo.txt
```

juanjovazquez:
@eharik I'll look into this ASAP. Thanks for catching this!.

PR #166, opened 2017-01-24
Fix FTP#list to recurse through sparse dirs, #165

eharik:
* Change `fillBuffer` to ensure that `buffer`
  head has a file (as opposed to file or dir).
* Clean up pattern matching on Vector.

### DIFF
--- a/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpBrowserGraphStage.scala
+++ b/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpBrowserGraphStage.scala
@@ -62,6 +62,7 @@ private[ftp] trait FtpBrowserGraphStage[FtpClient, S <: RemoteFileSettings] exte
       private[this] def initBuffer(basePath: String) =
         getFilesFromPath(basePath)
 
+      @scala.annotation.tailrec
       private[this] def fillBuffer(): Unit = buffer match {
         case head +: tail if head.isDirectory => {
           buffer = getFilesFromPath(head.path) ++ tail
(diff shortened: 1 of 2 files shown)


eharik:
@juanjovazquez looks like you are the original contributor on this component, can you take a look at this PR?

juanjovazquez:
LGTM. Thanks for catching this corner case!.

ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpBrowserGraphStage.scala
```diff
@@ -75,14 +62,13 @@ private[ftp] trait FtpBrowserGraphStage[FtpClient, S <: RemoteFileSettings] exte
...
-        buffer match {
-          case Seq() => // Nothing to do
-          case head +: tail =>
-            if (head.isDirectory) {
-              buffer = getFilesFromPath(head.path) ++ tail
-            }
+      private[this] def fillBuffer(): Unit = buffer match {
+        case head +: tail if head.isDirectory => {
+          buffer = getFilesFromPath(head.path) ++ tail
+          fillBuffer()
```
patriknw: `@tailrec` ?

eharik: Good call, just pushed an update with the annotation.

eharik:
Specs failing, looks to be the same comment as [here](https://github.com/akka/alpakka/pull/167.  I'll rebase and update the timeout.

johanandren:
Refs #165

PR #179, opened 2017-02-10
Add Local config to build docs with local api docs

2m:
Uses the new paradox which has support for custom configurations. This PR adds a new `local` configuration, which is used to remap API doc links.

This allows developers to use `docs/local:paradox` to generate reference documentation with working links to the API docs of the working tree code.


patriknw:
PR validation failure was the large ftp file test, fixed elsewhere I think

PR #183, opened 2017-02-12
S3: provide access to the returned response

2beaucoup:


s3/src/main/scala/akka/stream/alpakka/s3/scaladsl/S3Client.scala
```diff
@@ -39,6 +38,8 @@ final class S3Client(credentials: AWSCredentials, region: String)(implicit syste
 
   private[this] val impl = S3Stream(credentials, region)
 
+  def request(bucket: String, key: String): Future[HttpResponse] = impl.request(S3Location(bucket, key))
```
patriknw: is there a corresponding thing in `javadsl`?

is this the purpose of the PR? the other changes are "only" refactoring?

2beaucoup: > is there a corresponding thing in javadsl?

just added with the latest commit. 

> is this the purpose of the PR? the other changes are "only" refactoring?

yep, that's the core change of this PR. the refactorings were needed to make it work.

Issue #182, opened 2017-02-12
FTP connector - add toPath Sink


svezfaz:
I would find it useful to have a `Ftp.toPath` sink - mirroring `Ftp.fromPath` source.

Has this been considered already? I would be happy to provide a PR if not.

johanandren:
Sounds like a good idea to me.

eharik:
Agreed on the proposal to add "modes".  In addition to "overwrite", "append", "create", and "safe overwrite" I would imagine we'd want something along the lines of "don't overwrite."  Not sure if appropriate behavior is to return a failed Future in the case where the file already exists?

A couple other options that might be useful would be "atomic" and "chmod".  I don't know the underlying libraries well enough to know if atomic writes are the default option, or potentially the *only* option.  But if not I could see a user wanting to ensure that the file is created atomically.  And a user may also want to set the file permissions of the file they are creating.

johanandren:
Modes sounds like a good idea if the underlying FTP clients support it. I think "Don't overwrite" would be "create" and should fail if the file already exists. 

I think there is a level where you should just drop down to an imperative FTP client, we can and should not cover every thinkable use case for putting files on a FTP server.

NicolaeNMV:
Will be really interested to start using the FTP sink. Does anybody know if there are any plans for the a new alpakka release ?

johanandren:
Thanks for this one @svezfaz !

PR #184, opened 2017-02-13
FTP - toPath sink #182

svezfaz:
Handles #182

### DIFF
--- a/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpIOGraphStage.scala
+++ b/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpIOGraphStage.scala
@@ -4,24 +4,20 @@
 package akka.stream.alpakka.ftp
 package impl
 
-import akka.stream.stage.{GraphStageWithMaterializedValue, OutHandler}
-import akka.stream.{Attributes, IOResult, Outlet, SourceShape}
 import akka.stream.impl.Stages.DefaultAttributes.IODispatcher
+import akka.stream.stage.{GraphStageWithMaterializedValue, InHandler, OutHandler}
+import akka.stream.{Attributes, IOResult, Inlet, Outlet, Shape, SinkShape, SourceShape}
 import akka.util.ByteString
 import akka.util.ByteString.ByteString1C
 import scala.concurrent.{Future, Promise}
-import scala.util.control.NonFatal
-import java.io.InputStream
-import java.nio.file.Path
+import java.io.{InputStream, OutputStream}
 
-private[ftp] trait FtpIOGraphStage[FtpClient, S <: RemoteFileSettings]
-    extends GraphStageWithMaterializedValue[SourceShape[ByteString], Future[IOResult]] {
+private[ftp] trait FtpIOGraphStage[FtpClient, S <: RemoteFileSettings, Sh <: Shape]
+    extends GraphStageWithMaterializedValue[Sh, Future[IOResult]] {
 
   def name: String
 
-  def path: Path
-
-  def chunkSize: Int
+  def path: String
 
   def connectionSettings: S
 
@@ -32,7 +28,16 @@ private[ftp] trait FtpIOGraphStage[FtpClient, S <: RemoteFileSettings]
   override def initialAttributes: Attributes =
     super.initialAttributes and Attributes.name(name) and IODispatcher
 
-  val shape = SourceShape(Outlet[ByteString](s"$name.out"))
+  override def shape: Sh
+}
+
+private[ftp] trait FtpIOSourceStage[FtpClient, S <: RemoteFileSettings]
+    extends FtpIOGraphStage[FtpClient, S, SourceShape[ByteString]] {
+
+  def chunkSize: Int
+
+  val shape: SourceShape[ByteString] = SourceShape(Outlet[ByteString](s"$name.out"))
+  val out: Outlet[ByteString] = shape.outlets.head.asInstanceOf[Outlet[ByteString]]
 
   def createLogicAndMaterializedValue(inheritedAttributes: Attributes) = {
 
@@ -76,13 +81,8 @@ private[ftp] trait FtpIOGraphStage[FtpClient, S <: RemoteFileSettings]
           super.postStop()
         }
 
-      protected[this] def doPreStart(): Unit = {
-        val tryIs = ftpLike.retrieveFileInputStream(path.toAbsolutePath.toString, handler.get)
-        if (tryIs.isSuccess)
-          isOpt = tryIs.toOption
-        else
-          tryIs.failed.foreach { case NonFatal(t) => throw t }
-      }
+      protected[this] def doPreStart(): Unit =
+        isOpt = Some(ftpLike.retrieveFileInputStream(path, handler.get).get)
 
       protected[this] def matSuccess(): Boolean =
         matValuePromise.trySuccess(IOResult.createSuccessful(readBytesTotal))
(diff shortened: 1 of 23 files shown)


ftp/src/test/scala/akka/stream/alpakka/ftp/CommonFtpStageSpec.scala
```diff
@@ -97,4 +99,47 @@ trait CommonFtpSourceSpec extends BaseSpec {
     }
   }
 
+  "FtpIOSink" should {
```
eharik: is it worth having a spec showing the "sad path?"  maybe an IOException when trying to write to a location without sufficient permissions?

svezfaz: thanks for reviewing, that's a fair point

maybe I could leave the anon user with writepermissions=false for the sad path + setup a different, write-enabled user for the happy path?

juanjovazquez: @svezfaz That would be great!.

juanjovazquez:
BTW, It seems that #185 is still happening. I'm not able to reproduce it on my side. Maybe we should relax the `PatienceConfig` a bit more on `CommonFtpSourceSpec` trait. Maybe something like this?:

```scala
    PatienceConfig(timeout = Span(20, Seconds), interval = Span(900, Millis))
```

ftp/src/test/scala/akka/stream/alpakka/ftp/CommonFtpStageSpec.scala
svezfaz: I can easily manage FtpServer permissions with the users.properties file, but still haven't worked out how to do it for SFTP.

svezfaz:
@juanjovazquez I have mixed-in `IntegrationPatience` to the relevant test to address the flaky tests.

ftp/src/test/scala/akka/stream/alpakka/ftp/CommonFtpStageSpec.scala
juanjovazquez: I haven't done yet either. It seems that you can control the permissions set with [chmod](http://stackoverflow.com/questions/13150122/set-directory-permissions-chmod-with-jsch).

svezfaz: @juanjovazquez @eharik I tried several approaches, chmod being one of them. But ultimately, the underlying filesystem is Jimfs - which [does not support posix-like permissions](https://github.com/google/jimfs).

> For example, while setting and reading POSIX file permissions is possible with the "posix" view, those permissions will not actually affect the behavior of the file system.

I don't see an easy way of testing this without moving to - say - [Memoryfilesystem](https://github.com/marschall/memoryfilesystem).
The good news is that - somehow obviously - when trying it with my local filesystem, I get IOExceptions for permission errors.

WDYT?

juanjovazquez: I don't know what Akka members team will think, but seeing that it is so difficult to test this particular case, I would let it be. I'd stick to `jimfs` as it's been very rewarding up until now.

ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpIOGraphStage.scala
```diff
@@ -114,3 +121,70 @@ private[ftp] trait FtpIOGraphStage[FtpClient, S <: RemoteFileSettings]
...
+          tryOs.failed.foreach { case NonFatal(t) => throw t }
+      }
+
+      protected[this] def matSuccess(): Boolean =
+        matValuePromise.trySuccess(IOResult.createSuccessful(writtenBytesTotal))
+
+      protected[this] def matFailure(t: Throwable): Boolean =
+        matValuePromise.trySuccess(IOResult.createFailed(writtenBytesTotal, t))
+
+      /** BLOCKING I/O WRITE */
```
johanandren: So, it already runs on a separate dispatcher from the stage superclass right?

ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpIOGraphStage.scala
```diff
@@ -114,3 +121,70 @@ private[ftp] trait FtpIOGraphStage[FtpClient, S <: RemoteFileSettings]
...
+        }
+
+      protected[this] def doPreStart(): Unit = {
+        val tryOs = ftpLike.storeFileStream(path.toAbsolutePath.toString, handler.get)
+        if (tryOs.isSuccess) {
+          osOpt = tryOs.toOption
+          pull(in)
+        } else
+          tryOs.failed.foreach { case NonFatal(t) => throw t }
+      }
```
johanandren: This just mimics regular exception handling in a more complicated way. What about just calling `get` on the returned `Try` of `storeFileStream`, it will throw the exception if there is one and return the `OutputStream` if there is one.

ftp/src/main/scala/akka/stream/alpakka/ftp/javadsl/FtpApi.scala
```diff
@@ -149,7 +151,22 @@ sealed trait FtpApi[FtpClient] { _: FtpSourceFactory[FtpClient] =>
...
+
+  /**
+   * Java API: creates a [[Sink]] of [[ByteString]] to some file [[Path]].
+   *
+   * @param path the file path
+   * @param connectionSettings connection settings
+   * @return A [[Sink]] of [[ByteString]] that materializes to a [[CompletionStage]] of [[IOResult]]
+   */
+  def toPath(
+      path: Path,
```
johanandren: Is `java.nio.Path` really a good type to use here? The parameter refers to the path on the FTP server, while `Path` instances always has a `FileSystem` (which does not belong to the FTP server). Won't that mean that `path.toAbsolutePath.toString` likely will contain "C:\my\path" on windows for example?

I think it should just be a string, or a domain specific path-like type.

ftp/src/test/java/org/apache/ftpserver/filesystem/jimfs/impl/JimfsFtpFile.java
```diff
@@ -236,7 +236,11 @@ public boolean mkdir() {
         boolean retVal = false;
         try {
             if (isWritable()) {
-                Files.createDirectory(path);
+                Set<PosixFilePermission> perms =
+                        PosixFilePermissions.fromString("rwxrwxrwx");
+                FileAttribute<Set<PosixFilePermission>> attr =
+                        PosixFilePermissions.asFileAttribute(perms);
+                Files.createDirectory(path, attr);
```
johanandren: Are you sure this will only ever be called on a Posix/unixy file system?

ftp/src/main/scala/akka/stream/alpakka/ftp/javadsl/FtpApi.scala
svezfaz: I think that makes sense. String path is also used in the `ls` Sinks already. 

@juanjovazquez any particular reason why this was applied to `fromPath` ?

I can go ahead and  use String in both `fromPath` and `toPath`.

ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpIOGraphStage.scala
svezfaz: Correct.

ftp/src/test/java/org/apache/ftpserver/filesystem/jimfs/impl/JimfsFtpFile.java
svezfaz: This bit is actually only used in tests within a Unix-configured Jimfs instance. So my take is that it should be safe to do..

ftp/src/main/scala/akka/stream/alpakka/ftp/javadsl/FtpApi.scala
juanjovazquez: @svezfaz IIRC that idea also came from Camel. But, I think @johanandren 's considerations make sense. I had that doubt during the first implementation and now I realize that I took the wrong path after all. So, @svezfaz go ahead with these changes if you are willing to. Thanks!.

ftp/src/test/scala/akka/stream/alpakka/ftp/CommonFtpStageSpec.scala
patriknw: sounds good to keep it simple, and stick to `jimfs`

patriknw:
@svezfaz @juanjovazquez anything more remaining here apart from a rebase?

svezfaz:
@patriknw I think this one is ready to go. Further expansions will be handled in future tickets. Thanks for reviewing.

juanjovazquez:
@svezfaz @patriknw LGTM

patriknw:
thanks, please rebase @svezfaz

svezfaz:
@patriknw rebased

Issue #192, opened 2017-02-16
S3: Implement Get List Bucket call


jotsif:
As an Alpakka S3 user I would like to use Alpakka for listing the contents of S3 buckets, so that I can see what is there before I decide what to download.

Scalahansolo:
Myself and a couple of co-workers have a forked version that does exactly this. https://github.com/andrewrapp/alpakka/blob/master/s3/src/main/scala/akka/stream/alpakka/s3/scaladsl/S3Client.scala

Reason for not merging this back in - Akka Http does not have proxy support currently which is a blocker for us, so we have a our version until that exists. 

There is an open ticket to get proxy support, but it just doesn't seem like its a priority to get done currently. https://github.com/akka/akka-http/pull/583

johanandren:
Free for picking up as PR #193 was closed.

francisdb:
Listing bucket contents is currently available including object metadata (#347). Please close.

ennru:
As @francisdb stated in July, this is supported nowadays.

PR #193, opened 2017-02-16
Implement issue #192: List S3 buckets

jotsif:
Code for basic implementation of listing contents of a S3 bucket with prefix and paginated results.
I have confirmed that it works by using my own credentials in S3NoMock-test, both with and without truncation (using "&max-keys=10" this is fairly easy to test even with small buckets)

I will be happy to add more tests though, given some pointers from maintainers on what to put effort testing.

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
@@ -12,7 +12,6 @@ import akka.stream.alpakka.s3.S3Settings
 import akka.stream.alpakka.s3.acl.CannedAcl
 import akka.stream.scaladsl.Source
 import akka.util.ByteString
-
 import scala.collection.immutable
 import scala.collection.immutable.Seq
 import scala.concurrent.{ExecutionContext, Future}
@@ -28,16 +27,24 @@ private[alpakka] object HttpRequests {
       continuation_token: Option[String] = None
   )(implicit conf: S3Settings): HttpRequest = {
     val uri: Uri = {
-      val prefixstring = if (prefix.isDefined) "&prefix=" + prefix.get else ""
+      val prefixString = if (prefix.isDefined) "&prefix=" + prefix.get else ""
       val continuation_token_string =
         if (continuation_token.isDefined)
           "&continuation-token=" + continuation_token.get.replaceAll("=", "%3D").replaceAll("[+]", "%2B")
         else ""
-      Uri("/?list-type=2" + prefixstring + continuation_token_string)
-        .withHost(Uri.Host(s"${bucket}.s3-$region.amazonaws.com"))
-        .withScheme("https")
+
+      val uriBeforeProxy =
+        Uri(s"/$bucket/?list-type=2" + prefixString + continuation_token_string)
+
+      conf.proxy match {
+        case None => uriBeforeProxy.withHost(requestHost(region)).withScheme("https")
+        case Some(proxy) => uriBeforeProxy.withHost(proxy.host).withPort(proxy.port).withScheme(proxy.scheme)
+      }
     }
-    HttpRequest(HttpMethods.GET).withHeaders(Host(requestHost(bucket, region))).withUri(uri)
+
+    HttpRequest(HttpMethods.GET)
+      .withHeaders(Host(requestHost(region)))
+      .withUri(uri)
   }
 
   def getDownloadRequest(s3Location: S3Location, region: String)(implicit conf: S3Settings): HttpRequest =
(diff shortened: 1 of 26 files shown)


typesafehub-validator:
Hi @jotsif,

johanandren:
Refs #192

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -58,6 +60,32 @@ private[alpakka] final class S3Stream(credentials: AWSCredentials, region: Strin
...
   val signingKey = SigningKey(credentials, CredentialScope(LocalDate.now(), region, "s3"))
 
+  /** listBucket lists files in a bucket, filtered by an optional prefix 
+    * 
+    * @param bucket name of the bucket
+    * @param prefix optional prefix
+    * @return A akka Source with keys
+    * 
+    */
+  def listBucket(bucket: String, prefix: Option[String] = None): Source[String, NotUsed] = {
```
Scalahansolo: First off, before seeing this I posted https://github.com/akka/alpakka/pull/214 as the team Im on also needed this and we have forked implementation right now. A couple concerns/questions about this implementation. 

1. This is assuming that the caller wants ALL keys under a given bucket. Maybe this is okay because you could cut off the stream later if you didn't want all keys?
2. With this, if I have a very large bucket, do I start getting keys downstream while this is paging? Or does this page until the end before I see anything downstream?

jotsif: Nice! Always fun to compare implementations. We are also using a fork of alpakka internally to get listBucket functionality.

1) I use prefix to limit the number of keys returned. But use .take(k) if you want to cancel stream after k keys.
2) Yes, first page emits all its elements before getting next page because of the use of concat (http://doc.akka.io/api/akka/2.4/akka/stream/scaladsl/Source.html#concat)

Scalahansolo: Good deal. With that in mind, I think I am more of a fan of your implementation. There's some stuff in mine about the URLs that I still need to land. 

One last thing. S3 is known to fail for for random reasons and most of the time the prescribed way to fix it is to retry. With that, would it make sense to put a recoverWithRetries somewhere in this call chain to deal with failed calls? Or should that be something that happens on using the source?

Scalahansolo:
Just noticed one other thing. This is missing the javaDSL method. Easy to add and that would block this from landing.

s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
```diff
@@ -21,6 +20,33 @@ case class MetaHeaders(headers: Map[String, String])
...
+      bucket: String,
+      region: String,
+      prefix: Option[String] = None,
+      continuation_token: Option[String] = None
+  )(implicit conf: S3Settings): HttpRequest = {
+    val uri: Uri = {
+      val prefixString = if (prefix.isDefined) "&prefix=" + prefix.get else ""
+      val continuation_token_string =
+        if (continuation_token.isDefined)
+          "&continuation-token=" + continuation_token.get.replaceAll("=", "%3D").replaceAll("[+]", "%2B")
```
jotsif: This should use Uri-methods from http://doc.akka.io/docs/akka-http/current/scala/http/common/uri-model.html so we don't have to use .replaceAll. Will fix asap

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
jotsif: Good question! Sounds like a design decision for the maintainers, I am not really sure what the best practice is ....

Scalahansolo:
With the changes that I merged in, this PR is possibly a little overloaded based on the PR name. 

There were some issues with how the proxy was used that were fixed. I.e. The host header on the request should always be the s3 host, and never the proxy host. The host in the URI should be the proxy. 

I also changed how the URI's and Host Headers were constructed to use path style instead of virtual hosting style. Path style allows for non-DNS compliant bucket names such as "sean.mysweetbucket".

johanandren:
Any chance you could take the unrelated improvements out and PR them separately, too make it easier to keep track of things?

jotsif:
@johanandren Thanks, but we have decided to move our platform to GCP so will close this PR, as I won't spend more time on it. Perhaps there will a future PR about alpakka functionality for cloud storage, let's see ;)

@Scalahansolo Please feel free to take over my list bucket implementation and revive your own PR at your convenience.

Issue #149, opened 2017-01-17
JMS Selectors


evan108108:
How can I add simple JMS Selectors? Like "myfilter='abc'"? 

```scala
session.createConsumer(destination, "myfilter='abc'")
```

patriknw:
@larousso Do you know the answer here?

larousso:
This not implemented, but it could be easily added. 

A workaround is to use the filter from the akka stream api.

evan108108:
@larousso This is what I am doing now (simply filtering each message in the stream). This works fine but it is perhaps not as efficient for some workflows. I hope this can be added in the future.

johanandren:
@evan108108 Maybe the perfect task for a first alpakka PR? ;)

larousso:
@evan108108 @johanandren  I can do PR for this issue but if you want to participate tell me I let you work on this.

rafacm:
@larousso @evan108108 Have a look the wip-jms-selectors-issue-149 branch [here](afacm/alpakka/tree/wip-jms-selectors-issue-149). 

Implemented:
1. Introduced a case class JmsTextMessage to model JMS messages with properties. I followed the same pattern as in the MQTT connector.
2. Added a JmsTextMessageSink and refactored JmsSinkStage to add support for "sending" messages with properties
3. Refactored Java and Scala unit tests and added Scala unit test for the selector feature

Open questions/issues/tasks:
- [ ] I have left the current `JmsSink` that accept String(s). Should we leave it or should we have only one Sink that accepts `JmsTextMessage`s?
- [ ] Add Java unit tests for the selector feature
- [ ] Update documentation with the selector feature

I can implement the above points once we have clarified the options.

evan108108:
@rafacm +1 | This looks great! 

About the first point, I would probably only ever use the JmsTextMessage Sink. I had to implement something very similar as it is now.

larousso:
@evan108108  @rafacm This seems nice. 

For the first point, may be you should only keep the JmsSink that take a TextMessage and add a method like stringSink that take string and do the transformation even if that break the api.

May be TextMessage should extend a JmsMessage trait and the Sink use this trait. For the future we could implement the other type of messages supported by jms extending this trait. 

What do you think ?

evan108108:
@rafacm Where are you with the "Open questions"? Do you need help?

johanandren:
[...]

Issue #199, opened 2017-02-23
Support other datatypes than String for JmsSink


eirirlar:
As of now, the only API support for Jms Sinks is this:

```
val jmsTopicSink: Sink[String, NotUsed] = JmsSink(
  JmsSinkSettings(connectionFactory).withTopic("topic")
)
```

Other datatypes should be supported, at least the various message types in the Jms spec like TextMessage.

On that note, an API that exposes the JmsSession to the client should be implemented, so one could call i.e session.createTextMessage.

eirirlar:
As a workaround for jms Messages, I created this for personal use.
´´´
package akka.stream.alpakka.jms

import javax.jms.{Message, MessageProducer}

import akka.NotUsed
import akka.stream.scaladsl.Sink
import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler}
import akka.stream.{ActorAttributes, Attributes, Inlet, SinkShape}

object Jms {
  def sink(jmsSettings: JmsSinkSettings): Sink[Message, NotUsed] =
    Sink.fromGraph(new JmsMessageSinkStage(jmsSettings))
}

final class JmsMessageSinkStage(settings: JmsSettings) extends GraphStage[SinkShape[Message]] {

  private val in = Inlet[Message]("JmsSink.in")

  override def shape: SinkShape[Message] = SinkShape.of(in)

  override protected def initialAttributes: Attributes =
    ActorAttributes.dispatcher("akka.stream.default-blocking-io-dispatcher")

  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
    new GraphStageLogic(shape) with JmsConnector {

      private var jmsProducer: MessageProducer = _

      override private[jms] def jmsSettings = settings

      override def preStart(): Unit = {
        jmsSession = openSession()
        jmsProducer = jmsSession.session.createProducer(jmsSession.destination)
        pull(in)
      }

      setHandler(in,
        new InHandler {
          override def onPush(): Unit = {
            val elem = grab(in)
            jmsProducer.send(elem)
            pull(in)
          }
        })

      override def postStop(): Unit =
        Option(jmsSession).foreach(_.closeSession())
    }
}
´´´

johanandren:
This has been discussed in #149, I think it will be in the PR for that.

d2e8f9:
Please take a look at the commit above and let me know if something is missing. The JmsSink and JmsSource supports now Byte-, Map- and ObjectMessages according to the Jms spec. 

[...]

PR #200, opened 2017-02-23
Add support for JMS Selectors (#149)

rafacm:
Introduced JmsTextMessage case class with body and properties.
Refactored JmsSink to support both string and JmsTextMessages.
Added support for JMS selectors to the JmsSource.
Added and refactored Java and Scala JMS connector tests.
Updated JMS documentation.


rafacm:
Refs #149

evan108108:
🤘@rafacm This looks great. Any word on getting this merged? @patriknw @johanandren

johanandren:
Sorry about not being good with the response times on reviews we are knee deep in work to prepare for the 2.5 Akka release, so please have some patience with us here, thanks! :)

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
```diff
@@ -15,6 +17,19 @@ sealed trait Destination
 final case class Topic(name: String) extends Destination
 final case class Queue(name: String) extends Destination
 
+final case class JmsTextMessage(body: String, properties: Map[String, Any] = Map.empty) {
+  def withProperty(name: String, value: Any) = copy(properties = properties + (name -> value))
```
johanandren: I think we in general use `add` as prefix when it does not replace the entire thing it updates but adds to it (in this case it would be `withProperties(map)` which may also make sense for Java users as they can't really use `copy`)

Also, mark it with "Java API"

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
```diff
@@ -15,6 +17,19 @@ sealed trait Destination
...
+}
+
+object JmsTextMessage {
+
+  /**
+   * Java API: create  [[JmsTextMessage]]
+   */
+  def create(body: String) = JmsTextMessage(body, Map.empty)
+  def create(body: String, properties: util.Map[String, Any]) = JmsTextMessage(body, properties.toMap)
```
jrudolph: That seems to be Java API as well, please add JavaDoc for this method as well.

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
```diff
@@ -15,6 +17,19 @@ sealed trait Destination
 final case class Topic(name: String) extends Destination
 final case class Queue(name: String) extends Destination
 
+final case class JmsTextMessage(body: String, properties: Map[String, Any] = Map.empty) {
```
jrudolph: Could you quickly explain what's the reason to introduce this new type instead of using the JMS message type directly?

docs/src/main/paradox/jms.md
```diff
@@ -40,7 +40,11 @@ Gradle
 
 ## Usage
 
-### Sending messages to JMS server
```
jrudolph: :+1: lot's of good documentation changes.

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
rafacm: Will do! I used the `width` since I saw if being used already in other connectors like [on the MQTT connector](https://github.com/akka/alpakka/blob/master/mqtt/src/main/scala/akka/stream/alpakka/mqtt/Mqtt.scala#L83):
```scala
final case class MqttConnectionSettings(
    broker: String,
    clientId: String,
    persistence: MqttClientPersistence,
    auth: Option[(String, String)] = None,
    socketFactory: Option[SSLSocketFactory] = None,
    cleanSession: Boolean = true,
    will: Option[Will] = None
) {
  def withBroker(broker: String) =
    copy(broker = broker)

  def withAuth(username: String, password: String) =
    copy(auth = Some((username, password)))

  def withCleanSession(cleanSession: Boolean) =
    copy(cleanSession = cleanSession)

  def withWill(will: Will) =
    copy(will = Some(will))

  def withClientId(clientId: String) =
    copy(clientId = clientId)
}
```

jrudolph:
Would be nice to push that over the finish line. Let us know, @rafacm if we can help with something.

evan108108:
@jrudolph @johanandren @rafacm : Any status update?

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
rafacm: The reason is mainly due to the fact that in order to create a JMS message you need to do it via the JmsSession 
```
val textMessage: TextMessage = jmsSession.session.createTextMessage(elem.body)
```
and I wanted to be able to pass JMS message properties. 

I am not happy with the solution in part because the API is currently not symmetric: `Sink[akka.stream.alpakka.jms.JmsTextMessage, NotUsed]` vs. `Source[javax.jms.Message, NotUsed]`. Furthermore the new type currently just adds a convenience method to add JMS message properties. 

We should either get rid of `akka.stream.alpakka.jms.JmsTextMessage` or refactor `JmsSource` to use the new type `akka.stream.alpakka.jms.JmsTextMessage`. I am inclined to go for the first option (getting rid of `akka.stream.alpakka.jms.JmsTextMessage`), but for this we would need to refactor the `JmsConnector` (see related #199).

Any suggestions?

rafacm:
@jrudolph @johanandren @evan108108: DONE!

Issue #197, opened 2017-02-20
It should be possible to specify custom private key when creating SFTP source.


argast:
At the moment only way of authenticating using SFTP is username/password or existing host keys. It should be possible to specify custom key when configuring SFTP source.

argast:
I will try to contribute a PR soon.

PR #202, opened 2017-02-23
Added possibility configure SFTP connection using private key #197.

argast:
* Added SftpIdentity case class to allow configuring private/public key,
* Added Java API methods to configure SftpIdentity from Java
* Added Scala spec and Java test  that should fail password based authentication and revert to private key one,
* Added docs paragraph to describe this option.


ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
```diff
@@ -97,7 +97,8 @@ object RemoteFileSettings {
       host: InetAddress,
       port: Int = DefaultSftpPort,
       credentials: FtpCredentials = AnonFtpCredentials,
-      strictHostKeyChecking: Boolean = true
+      strictHostKeyChecking: Boolean = true,
```
juanjovazquez: Should we deal with the case of strict host key checking in the context of this issue?. On that case, we probably should let specify a `known_hosts` custom file.

ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
```diff
@@ -129,3 +130,37 @@ object FtpCredentials {
...
+  /** Java API */
+  def createSftpIdentity(name: String, privateKey: Array[Byte]): Option[SftpIdentity] =
+    Some(SftpIdentity(name, privateKey))
+
+  /** Java API */
+  def createSftpIdentity(name: String, privateKey: Array[Byte], password: Array[Byte]): Option[SftpIdentity] =
+    Some(SftpIdentity(name, privateKey, Some(password)))
+
+  /** Java API */
+  def createSftpIdentity(name: String,
```
juanjovazquez: I think the format on signatures is not following the previous ones along the project. It's just a nitpick but important in the long term.

ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
```diff
@@ -129,3 +130,37 @@ object FtpCredentials {
...
+ *
+ * @param name name of identity
+ * @param privateKey private key to use when connecting
+ * @param password password to use to decrypt private key
+ * @param publicKey public key to use when connecting
+ */
+case class SftpIdentity(name: String,
+                        privateKey: Array[Byte],
+                        password: Option[Array[Byte]] = None,
+                        publicKey: Option[Array[Byte]] = None)
```
juanjovazquez: I wonder if it would be interesting to offer variants of the identity specification based on files on the classpath in addition to these `Array[Byte]`.

ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
argast: Do you mean to use withFoo methods similar to amqp model?

ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
argast: It's a good idea,  I will add additional case class with possibility to specify location of key files and passphrase.

ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
argast: Yes, that would be useful as well. I will add it to SftpSettings then.

ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
juanjovazquez: Nop, more silly than that. Just enter a carriage return after the parentheses so that `scalafmt` line up left on the next line.

ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
juanjovazquez: Some additional testing covering this would be welcome :)

argast: Of course, will do :)

ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
argast: Actually I like the way withFoo works, it's a nicer way to handle Option in Java so I think I will migrate java API methods to withFoo way.

argast:
Thanks for the review. I rebased the code.

2m:
Great. Thanks!

Issue #204, opened 2017-02-25
AWS SNS connector


fg-devs:
Provide a Sink to publish a message to an AWS SNS topic.

PR #205, opened 2017-02-25
add SNS connector with publish sink #204

fg-devs:
- Includes a new connector called sns
- Includes a sink to publish messages to an AWS SNS topic


johanandren:
REFS #204

sns/src/main/scala/akka/stream/alpakka/sns/SnsPublishSinkStage.scala
```diff
@@ -0,0 +1,77 @@
...
+import akka.Done
+import akka.stream.stage._
+import akka.stream.{Attributes, Inlet, SinkShape}
+import com.amazonaws.handlers.AsyncHandler
+import com.amazonaws.services.sns.AmazonSNSAsync
+import com.amazonaws.services.sns.model.{PublishRequest, PublishResult}
+
+import scala.concurrent.{Future, Promise}
+
+final class SnsPublishSinkStage(topicArn: String, snsClient: AmazonSNSAsync)
```
johanandren: Is the `AmazonSNSAsync` thread safe? A stage can be materialized any number of times and all of them will share the same `snsClient`.  If it cannot be shared a factory lambda is better (`() => AmazonSNSAsync`), that way every materialization can have its own client (which it will be responsible to manage/close etc in sync with the stage lifecycle).

Another option might be to take a `AmazonSNSAsyncClientBuilder` and `.build()` that each materialization (if the builder is immutable that is).

sns/src/main/scala/akka/stream/alpakka/sns/SnsPublishSinkStage.scala
```diff
@@ -0,0 +1,77 @@
...
+        }
+      })
+
+      override def preStart(): Unit = {
+        setKeepGoing(true)
+        pull(in)
+      }
+
+      def handleFailure(ex: Throwable): Unit = {
+        log.error(ex, "AmazonSNSAsync failure: {}", ex.getMessage)
```
johanandren: No need to log the failure since we fail the stage and the materialized value with it. You could wrap it with a en exception to decorate with extra information from the stage though.

sns/src/test/scala/akka/stream/alpakka/sns/SnsPublishSinkSpec.scala
```diff
@@ -0,0 +1,125 @@
...
+      new Answer[AnyRef] {
+        override def answer(invocation: InvocationOnMock): Future[PublishResult] = {
+          val publishRequest = invocation.getArgument[PublishRequest](0)
+          invocation
+            .getArgument[AsyncHandler[PublishRequest, PublishResult]](1)
+            .onSuccess(
+              publishRequest,
+              new PublishResult().withMessageId("message-id")
+            )
+          new CompletableFuture()
```
johanandren: This future is never completed though, shouldn't that cause the `matVal` to never complete, is there a bug around waiting for ack on publishing?

sns/src/main/scala/akka/stream/alpakka/sns/SnsPublishSinkStage.scala
johanandren: An additional thought: It could be better to make this a flow, emitting the values downstream as they have been ack:ed from AWS, the factory for the simple use case as sink could then be achieved through composition (`SNSPublishFlow.to(Sink.ignore)`).

sns/src/main/scala/akka/stream/alpakka/sns/scaladsl/SnsPublishSink.scala
```diff
@@ -0,0 +1,18 @@
...
+import com.amazonaws.services.sns.AmazonSNSAsync
+
+import scala.concurrent.Future
+
+object SnsPublishSink {
+
+  /**
+   * Scala API: creates a [[SnsPublishSinkStage]] for a SNS topic using an [[AmazonSNSAsync]]
+   */
+  def apply(topicArn: String)(implicit snsClient: AmazonSNSAsync): Sink[String, Future[Done]] =
```
johanandren: I don't think the client should be implicit, especially considering the note about thread safety.

build.sbt
```diff
@@ -77,6 +77,14 @@ lazy val simpleCodecs = project
     name := "akka-stream-alpakka-simple-codecs"
   )
 
+lazy val sns = project
+  .in(file("sns"))
```
fg-devs: Alright, I will try it out.

sns/src/main/scala/akka/stream/alpakka/sns/SnsPublishSinkStage.scala
fg-devs: Yes the AmazonSNSAsync is ThreadSafe regarding to the AWS [docs.](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/sns/AmazonSNSAsyncClient.html) That's why I think it is fine to share a single AmazonSNSAsync among multiple materialization stages.

Also regarding your last comment about not making the AmazonSNSAsync implicit, I try to be consistent with other connectors, like the SQS one. There we also pass the AmazonSQSClient implicitly. I think it will be easier for people who use the connectors if the initialization of the GraphStages work the same way.

Good point to change the Sink to a Flow. I thought about keeping the Sink and provide an additional Flow GraphStage which will return the messageId of the published message, so clients can do further processing with the returned messageId.

sns/src/main/scala/akka/stream/alpakka/sns/SnsPublishSinkStage.scala
fg-devs: Alright I will remove it then.

sns/src/test/scala/akka/stream/alpakka/sns/SnsPublishSinkSpec.scala
fg-devs: I will further investigate it.

docs/src/main/paradox/sns.md
```diff
@@ -0,0 +1,77 @@
+# AWS SNS Connector
+
+The AWS SNS connector provides an Akka Stream sink for AWS SNS.
```
fg-devs: Good point, I will add some more information.

fg-devs:
@johanandren I updated the PR regarding your comments. Furthemrore, I also added a SnsPublishFlow what I mentioned in one of my replies of your comments.

sns/src/main/scala/akka/stream/alpakka/sns/SnsPublishSinkStage.scala
johanandren: Ah, Hadn't noticed that how many connectors that do that. Thats a valid argument to keep it implicit.

johanandren: Thanks for verifying that the client is threadsafe.

johanandren: Maybe even just let it emit PublishResult downstream and then both String => String, and String => MessageId can be achieved through combinations with other stages.

fg-devs: Updated the SnsPublishFlow to emit a PublishResult.

johanandren: This means we could dlete the sink stage and express it by using the flow in the factory method instead:
```
val flow: Flow[String, PublishResult, NotUsed] 
val sink: Sink[String, Future[Done]] = flow.toMat(Sink.ignore)(Keep.right)
```

sns/src/main/scala/akka/stream/alpakka/sns/SnsPublishFlowStage.scala
```diff
@@ -0,0 +1,59 @@
...
+          if (!hasBeenPulled(in)) tryPull(in)
+          push(out, result)
+        }
+      }
+
+      override def onPush(): Unit = {
+        val request = new PublishRequest().withTopicArn(topicArn).withMessage(grab(in))
+
+        snsClient.publishAsync(request,
+          new AsyncHandler[PublishRequest, PublishResult] {
```
johanandren: Not necessarily anything to act on but, as this gets the values passed in on success/failure, it could be a single instance of `AsyncHandler` that you reuse rather than a new instance for every message.

sns/src/main/scala/akka/stream/alpakka/sns/SnsPublishFlowStage.scala
```diff
@@ -0,0 +1,59 @@
...
+          override def onError(exception: Exception): Unit =
+            failureCallback.invoke(exception)
+
+          override def onSuccess(request: PublishRequest, result: PublishResult): Unit =
+            successCallback.invoke(result)
+        })
+      }
+
+      override def onPull(): Unit = {
+        if (isClosed(in)) completeStage()
```
johanandren: There could be an ongoing publish that did not yet complete here, even if in was closed, you need to keep track of that and emit that last publish result when it comes, and then complete the stage.

sns/src/main/scala/akka/stream/alpakka/sns/javadsl/SnsPublishSink.scala
```diff
@@ -0,0 +1,26 @@
...
+
+import akka.Done
+import akka.stream.alpakka.sns.SnsPublishSinkStage
+import akka.stream.javadsl.{Sink => JSink}
+import akka.stream.scaladsl.Sink
+import com.amazonaws.services.sns.AmazonSNSAsync
+
+import scala.compat.java8.FutureConverters._
+
+object SnsPublishSink {
```
johanandren: Maybe this should then instead be `SnsPublisher` and have both `.sink` and `.flow` in the same place?

fg-devs:
@johanandren Updated regarding your comments. Removed the SnsPublishSink and provide a SnsPublisher with .flow and .sink factory methods. Also implemented a way to keep track of in flight messages for proper stage closing.

sns/src/main/scala/akka/stream/alpakka/sns/SnsPublishFlowStage.scala
```diff
@@ -0,0 +1,62 @@
...
+      private val successCallback = getAsyncCallback[PublishResult](handleSuccess)
+
+      private def handleFailure(ex: Throwable): Unit =
+        failStage(ex)
+
+      private def handleSuccess(result: PublishResult): Unit = {
+        log.debug("Published SNS message: {}", result.getMessageId)
+        inFlight -= 1
+        if (isAvailable(out)) {
+          if (!hasBeenPulled(in)) tryPull(in)
```
patriknw: why `pull` here? shouldn't that only be done from `onPull`?

sns/src/main/scala/akka/stream/alpakka/sns/SnsPublishFlowStage.scala
```diff
@@ -0,0 +1,62 @@
...
+      private val failureCallback = getAsyncCallback[Throwable](handleFailure)
+      private val successCallback = getAsyncCallback[PublishResult](handleSuccess)
+
+      private def handleFailure(ex: Throwable): Unit =
+        failStage(ex)
+
+      private def handleSuccess(result: PublishResult): Unit = {
+        log.debug("Published SNS message: {}", result.getMessageId)
+        inFlight -= 1
+        if (isAvailable(out)) {
```
patriknw: when will it not be `available`? should this be `!isClosed` 
we are just dropping the message if it's not `available` so I want to be sure that we don't do that accidentally

sns/src/main/scala/akka/stream/alpakka/sns/SnsPublishFlowStage.scala
```diff
@@ -0,0 +1,62 @@
...
+
+      private val asyncHandler = new AsyncHandler[PublishRequest, PublishResult] {
+        override def onError(exception: Exception): Unit =
+          failureCallback.invoke(exception)
+        override def onSuccess(request: PublishRequest, result: PublishResult): Unit =
+          successCallback.invoke(result)
+      }
+
+      override def onPush(): Unit = {
+        inFlight += 1
```
patriknw: can `inFlight` be > 1? perhaps use `boolean` if it can only be 0 or 1

sns/src/main/scala/akka/stream/alpakka/sns/SnsPublishFlowStage.scala
```diff
@@ -0,0 +1,62 @@
...
+        val request = new PublishRequest().withTopicArn(topicArn).withMessage(grab(in))
+        snsClient.publishAsync(request, asyncHandler)
+      }
+
+      override def onPull(): Unit = {
+        if (isClosed(in) && inFlight == 0) completeStage()
+        if (!hasBeenPulled(in)) tryPull(in)
```
patriknw: `else if`

sns/src/main/scala/akka/stream/alpakka/sns/SnsPublishFlowStage.scala
```diff
@@ -0,0 +1,62 @@
...
+ */
+package akka.stream.alpakka.sns
+
+import akka.stream._
+import akka.stream.stage._
+import com.amazonaws.handlers.AsyncHandler
+import com.amazonaws.services.sns.AmazonSNSAsync
+import com.amazonaws.services.sns.model.{PublishRequest, PublishResult}
+
+final class SnsPublishFlowStage(topicArn: String, snsClient: AmazonSNSAsync)
```
patriknw: Internal API?
`private[akka]`

sns/src/main/scala/akka/stream/alpakka/sns/javadsl/SnsPublisher.scala
```diff
@@ -0,0 +1,28 @@
...
+import akka.stream.alpakka.sns.SnsPublishFlowStage
+import akka.stream.javadsl.{Flow, Keep, Sink}
+import akka.{Done, NotUsed}
+import com.amazonaws.services.sns.AmazonSNSAsync
+import com.amazonaws.services.sns.model.PublishResult
+
+object SnsPublisher {
+
+  /**
+   * Java API: creates a [[Flow]] from a [[SnsPublishFlowStage]] for a SNS topic using an [[AmazonSNSAsync]]
```
patriknw: users should not care about the implementation `SnsPublishFlowStage`

sns/src/main/scala/akka/stream/alpakka/sns/SnsPublishFlowStage.scala
fg-devs: Wouldn't this limit the Flow to process elements only one by one? Otherwise we would always have to wait for the output port to pull?

Thats why I also made inFlight as an Int and not a Boolean.

ktoso:
Hi there @fg-devs, are you going to have some time to address the small comments made?
I'd hope so, would be awesome to include this in the next release :)

fg-devs:
@ktoso Hey, yes I will do the changes today.

fg-devs:
@ktoso @patriknw Update the PR regarding your comments. The build for Scala 2.12 failed (there is already an open [ticket)](https://github.com/akka/alpakka/issues/235)

sns/src/main/scala/akka/stream/alpakka/sns/SnsPublishFlowStage.scala
```diff
@@ -0,0 +1,59 @@
...
+ */
+package akka.stream.alpakka.sns
+
+import akka.stream._
+import akka.stream.stage._
+import com.amazonaws.handlers.AsyncHandler
+import com.amazonaws.services.sns.AmazonSNSAsync
+import com.amazonaws.services.sns.model.{PublishRequest, PublishResult}
+
+private[akka] final class SnsPublishFlowStage(topicArn: String, snsClient: AmazonSNSAsync)
```
johanandren: Does the client have a lifecycle, is is thread safe, should this be a client factory instead (`() => AmazonSNSAsync`) and have its lifecycle managed by the stage (`client.close() when the stage stops)?

johanandren: NVM, I already asked that previously but github hid all comments and I had forgot.

PR #210, opened 2017-03-05
logrotation file sink

tg44:
It's a basic logrotation type file sink. I'm still a newbie so feel free to comment/discuss my code.

### DIFF
--- a/docs/src/main/paradox/file.md
+++ b/docs/src/main/paradox/file.md
@@ -68,6 +68,36 @@ Scala
 Java
 : @@snip (../../../../file/src/test/java/akka/stream/alpakka/file/javadsl/DirectoryChangesSourceTest.java) { #minimal-sample }
 
+### LogRotationSink
+
+The `LogRotationSink` will create and write to multiple files.  
+This sink will takes a function as parameter which returns a
+ `Bytestring => Option[Path]` function. If the generated function returns a path
+ the sink will rotate the file output to this new path and the actual `ByteString` will be
+  written to this new file too.
+ With this approach the user can define a custom stateful file generation implementation.
+
+The java implementation is a bit different. The inner function must return null or Path instead of the Option.
+
+A small snippet for the usage:
+
+Scala
+: @@snip (../../../../file/src/test/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSinkSpec.scala) { #LogRotationSink-sample }
+
+In this sample we create a size based rotation function:
+
+Scala
+: @@snip (../../../../file/src/test/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSinkSpec.scala) { #LogRotationSink-filesize-sample }
+Java
+: @@snip (../../../../file/src/test/java/akka/stream/alpakka/file/javadsl/LogRotatorSinkTest.java) { #LogRotationSink-filesize-sample }
+
+In this sample we create a time based rotation function:
+
+Scala
+: @@snip (../../../../file/src/test/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSinkSpec.scala) { #LogRotationSink-timebased-sample }
+Java
+: @@snip (../../../../file/src/test/java/akka/stream/alpakka/file/javadsl/LogRotatorSinkTest.java) { #LogRotationSink-timebased-sample }
+
 
 ### Running the example code
 
--- a/file/src/main/scala/akka/stream/alpakka/file/javadsl/LogRotatorSink.scala
+++ b/file/src/main/scala/akka/stream/alpakka/file/javadsl/LogRotatorSink.scala
@@ -1,0 +1,46 @@
+/*
+ * Copyright (C) 2016-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.file.javadsl
+
+import java.nio.file.{Path, StandardOpenOption}
+import java.util
+import java.util.concurrent.CompletionStage
+
+import akka.Done
+import akka.stream.javadsl
+import akka.stream.javadsl.Sink
+import akka.util.ByteString
+import akka.japi.{function}
+
+import scala.collection.JavaConverters._
+
+object LogRotatorSink {
+  def createFromFunction(
+      f: function.Creator[function.Function[ByteString, Path]]
+  ): javadsl.Sink[ByteString, CompletionStage[Done]] =
+    new Sink(
+      akka.stream.alpakka.file.scaladsl
+        .LogRotatorSink({ () =>
+          val fun = f.create()
+          elem =>
+            Option(fun(elem))
+        })
+        .toCompletionStage()
+    )
+
+  def createFromFunctionAndOptions[T](
+      f: function.Creator[function.Function[ByteString, Path]],
+      options: util.Set[StandardOpenOption]
+  ): javadsl.Sink[ByteString, CompletionStage[Done]] =
+    new Sink(
+      akka.stream.alpakka.file.scaladsl
+        .LogRotatorSink({ () =>
+          val fun = f.create()
+          elem =>
+            Option(fun(elem))
+        }, options.asScala.toSet)
+        .toCompletionStage()
+    )
+}
--- a/file/src/main/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSink.scala
+++ b/file/src/main/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSink.scala
@@ -1,0 +1,161 @@
+/*
+ * Copyright (C) 2016-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.file.scaladsl
+
+import java.nio.file.{OpenOption, Path, StandardOpenOption}
+
+import akka.Done
+import akka.stream.ActorAttributes.SupervisionStrategy
+import akka.stream._
+import akka.stream.scaladsl.{FileIO, Sink, Source}
+import akka.stream.stage._
+import akka.util.ByteString
+
+import scala.concurrent.{Future, Promise}
+import scala.util.{Failure, Success, Try}
+
+object LogRotatorSink {
+  def apply(
+      functionGeneratorFunction: () => ByteString => Option[Path],
+      fileOpenOptions: Set[OpenOption] = Set(StandardOpenOption.APPEND, StandardOpenOption.CREATE)
+  ): Sink[ByteString, Future[Done]] =
+    Sink.fromGraph(new LogRotatorSink(functionGeneratorFunction, fileOpenOptions))
+}
+
+final private[scaladsl] class LogRotatorSink(functionGeneratorFunction: () => ByteString => Option[Path],
+                                             fileOpenOptions: Set[OpenOption])
+    extends GraphStageWithMaterializedValue[SinkShape[ByteString], Future[Done]] {
+
+  val in = Inlet[ByteString]("FRotator.in")
+  override val shape = SinkShape.of(in)
+
+  override def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, Future[Done]) = {
+    val promise = Promise[Done]()
+    val logic = new GraphStageLogic(shape) {
+      val pathGeneratorFunction: ByteString => Option[Path] = functionGeneratorFunction()
+      var sourceOut: SubSourceOutlet[ByteString] = _
+      var fileSinkCompleted: Option[Future[IOResult]] = Option.empty
+      val decider =
+        inheritedAttributes.get[SupervisionStrategy].map(_.decider).getOrElse(Supervision.stoppingDecider)
+
+      def failThisStage(ex: Throwable): Unit =
+        if (!promise.isCompleted) {
+          if (sourceOut != null) {
+            sourceOut.fail(ex)
+          }
+          cancel(in)
+          promise.failure(ex)
+        }
+
+      def generatePathOrFailPeacefully(data: ByteString): Option[Path] = {
+        var ret = Option.empty[Path]
+        try {
+          ret = pathGeneratorFunction(data)
+        } catch {
+          case ex: Throwable =>
+            failThisStage(ex)
+        }
+        ret
+      }
+
+      def fileSinkFutureCallbackHandler(result: Try[IOResult], future: Future[IOResult]): Unit =
+        result match {
+          case Success(IOResult(_, Failure(ex))) if decider(ex) == Supervision.Stop =>
+            promise.failure(ex)
+          case Success(x) if fileSinkCompleted.isDefined && fileSinkCompleted.get == future =>
+            promise.trySuccess(Done)
+            completeStage()
+          case Failure(ex) =>
+            failThisStage(ex)
+          case _ =>
+        }
+
+      //init stage where we are waiting for the first path
+      setHandler(
+        in,
+        new InHandler {
+          override def onPush(): Unit = {
+            val data = grab(in)
+            val pathO = generatePathOrFailPeacefully(data)
+            pathO.fold(
+              if (!isClosed(in)) pull(in)
+            )(
+              switchPath(_, data)
+            )
+          }
+
+          override def onUpstreamFinish(): Unit =
+            completeStage()
+
+          override def onUpstreamFailure(ex: Throwable): Unit =
+            failThisStage(ex)
+        }
+      )
+
+      //we must pull the first element cos we are a sink
+      override def preStart(): Unit = {
+        super.preStart()
+        pull(in)
+      }
+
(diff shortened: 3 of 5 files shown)


tg44:
I cleaned up the code, the last fail is not my fault :)
I documented it, but no java api.
Still waiting for any review.

docs/src/main/paradox/file.md
```diff
@@ -68,6 +68,28 @@ Scala
...
+The `LogRotationSink` will create and write to multiple files. There is two sink right now.
+
+The `functionBasedRotation` will call the given function every time when its 
+receive a new `ByteString` and if this function give new location the outputfile
+will change according to it. This is good for time based rotations 
+(like start a new file in every hour).
+
+The `sizeBasedRotation` will call the given function every time when the given size limit
+would be reached (so this generate lower then the limit filesizes). This is obiuosly good
+for size based rotations (like 'I want lower then 10Mb files').
```
johanandren: "The `sizeBasedRotation` works like the `functionBasedRotation` except that it only calls the provided path selection function when the file size plus the current bytes would go over a specific number of bytes. This is useful when you want to put an upper limit to the size of the written files.

file/src/main/scala/akka/stream/alpakka/file/internal/ChangeableFileSink.scala
```diff
@@ -0,0 +1,42 @@
...
+import akka.stream.impl.SinkModule
+import akka.stream.impl.Stages.DefaultAttributes.IODispatcher
+import akka.stream.impl.StreamLayout.Module
+
+import scala.concurrent.{Future, Promise}
+
+private[akka] final class ChangeableFileSink(options: Set[StandardOpenOption],
+                                             val attributes: Attributes,
+                                             shape: SinkShape[ChangeableFileSubscriberMessages])
+    extends SinkModule[ChangeableFileSubscriberMessages, Future[IOResult]](shape) {
```
johanandren: `SinkModule` is internal in Akka and not good to extend, use the public API `GraphStage`s instead

file/src/main/scala/akka/stream/alpakka/file/internal/ChangeableFileSubscriber.scala
```diff
@@ -0,0 +1,110 @@
...
+
+  case class ChangePathMessage(path: Path) extends ChangeableFileSubscriberMessages
+
+}
+
+protected[akka] class ChangeableFileSubscriber(
+    completionPromise: Promise[IOResult],
+    bufSize: Int,
+    openOptions: Set[StandardOpenOption]
+) extends akka.stream.actor.ActorSubscriber
```
johanandren: `ActorSubscriber` will be deprecated in 2.5.0, all that it can do can be done with `GraphStage` instead, so use that. There is plenty of samples in Alpakka to look at for inspiration already.

file/src/main/scala/akka/stream/alpakka/file/internal/FileSizeRotatorFlow.scala
```diff
@@ -0,0 +1,54 @@
...
+import akka.stream.alpakka.file.internal.ChangeableFileSubscriber.{
+  ByteStringWrapper,
+  ChangePathMessage,
+  ChangeableFileSubscriberMessages
+}
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
+import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
+import akka.util.ByteString
+
+class FileSizeRotatorFlow(max: Long, f: () => Path)
```
johanandren: "If you liked it, make it final, also if it is internal, make it `private[file]`.
A better name for `f` would be good, "selectPath" for example.

If you abstract the decision when to ask for a new path should be asked for into a class, you could have just one stage and then just provide different factories for the case of always asking and or based on counting bytes.

I think you should be able to encapsulate all the logic needed in this stage rather than combining it with other stages. Alternatively use substreams and the built in akka streams file writer if you want to avoid touching the IO/NIO apis directly.

tg44:
Thanks for the reply!

I basicly copied the akka streams file writer and extend it with the "dinamicly change path" functionality, thats why I used `ActorSubscriber`, and thats the reason behind the `SinkModule` too :) I tried to use the build in file writer first, but the actor/sink implementation was not created to support this kind of functionality (as I saw).

Extracting all the logic from the functionBased and fileSize rotator to a predefined abstract class is an intresting idea, I will rethink it.

johanandren:
If you use `SubSinkInlet` (which is also internal API but likely more safe to use) you should be able to re-use the existing filesink, and just complete the substream and materialize a new one when the file path changes. Look at the sources of for example `RecoverWith` in Akka streams to see how it can be used (it is tricky though).

tg44:
I rewrote it as you mentioned, but the CI get a strange testfail... Any idea what I missed? (I think this could be a memorybuffered file, but I'm not sure. I think I use proper future completition. I can't reproduce it in local.)

Any more suggestions? Maybe more test ideas?

If you think it's worth a merge, I close this PR and start a new one with a cleaned commit.

tg44:
Any progression?

johanandren:
The test failure seems related, did you look into it?

```
[info] - must write lines to multiple files due to filesize *** FAILED *** (81 milliseconds)
[info]   List(0, 2002, 2002) did not contain the same elements as List(2002, 2002, 2002) (LogRotatorSinkSpec.scala:148)
[info]   at org.scalatest.MatchersHelper$.indicateFailure(MatchersHelper.scala:340)
[info]   at org.scalatest.words.ResultOfContainWord.theSameElementsAs(ResultOfContainWord.scala:184)
[info]   at akka.stream.alpakka.file.scaladsl.LogRotatorSinkSpec.$anonfun$new$7(LogRotatorSinkSpec.scala:148)
```

tg44:
Yapp, and I can't reproduce it. Locally every time its just run fine. Maybe paralell execution messed up things so I rewrite the test to a more safer form.  
These are my favorite type of tests, the ones those fail only on ci :D

mqtt test failed this time.

docs/src/main/paradox/file.md
```diff
@@ -68,6 +68,29 @@ Scala
 Java
 : @@snip (../../../../file/src/test/java/akka/stream/alpakka/file/javadsl/DirectoryChangesSourceTest.java) { #minimal-sample }
 
+### LogRotationSink
+
+The `LogRotationSink` will create and write to multiple files.  
+This sink will wait for a function as parameter which returns a
```
johanandren: `takes` or `accepts`, the sink will not exist before creating an instance so it doesn't really wait for anything.

docs/src/main/paradox/file.md
```diff
@@ -68,6 +68,29 @@ Scala
...
+
+In this sample we create a size based rotation function:
+
+Scala
+: @@snip (../../../../file/src/test/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSinkSpec.scala) { #LogRotationSink-filesize-sample }
+
+In this sample we create a time based rotation function:
+
+Scala
+: @@snip (../../../../file/src/test/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSinkSpec.scala) { #LogRotationSink-timebased-sample }
```
johanandren: We need the corresponding Java factory and Java samples as well, should be piece of cake with everything else in place! 🍰

file/src/main/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSink.scala
```diff
@@ -0,0 +1,119 @@
...
+import scala.util.{Failure, Success}
+
+object LogRotatorSink {
+  def apply(functionGeneratorFunction: () => ByteString => Option[Path],
+            fileOpenOptions: Set[StandardOpenOption] = Set(StandardOpenOption.APPEND, StandardOpenOption.CREATE))
+    : LogRotatorSink =
+    new LogRotatorSink(functionGeneratorFunction, fileOpenOptions)
+}
+
+class LogRotatorSink(functionGeneratorFunction: () => ByteString => Option[Path],
```
johanandren: `final` and possibly `private[scaladsl]` since we have the factory

file/src/main/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSink.scala
```diff
@@ -0,0 +1,119 @@
...
+import akka.util.ByteString
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.{Future, Promise}
+import scala.util.{Failure, Success}
+
+object LogRotatorSink {
+  def apply(functionGeneratorFunction: () => ByteString => Option[Path],
+            fileOpenOptions: Set[StandardOpenOption] = Set(StandardOpenOption.APPEND, StandardOpenOption.CREATE))
+    : LogRotatorSink =
```
johanandren: return type can be more secretive, `Sink[ByteString, Future[Done]]`, that is good as it does not lock down future changes, there is anyways no specific public API on `LogRotatorSink`.

file/src/main/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSink.scala
```diff
@@ -0,0 +1,119 @@
...
+            } else {
+              switchPath(pathO.head, data)
+            }
+          }
+
+          override def onUpstreamFinish(): Unit = {
+            sourceOut.complete()
+            fileSinkCompleted onComplete {
+              case Success(x) => promise.trySuccess(Done)
+              case Failure(ex) => promise.trySuccess(Done)
```
johanandren: Shouldn't it rather be failed here?

file/src/main/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSink.scala
```diff
@@ -0,0 +1,119 @@
...
+            }
+            completeStage()
+          }
+
+          override def onUpstreamFailure(ex: Throwable): Unit = {
+            sourceOut.fail(ex)
+            failStage(ex)
```
johanandren: The promise needs to be failed here as well.

file/src/test/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSinkSpec.scala
```diff
@@ -0,0 +1,164 @@
...
+              None
+            }
+          }
+      }
+      val completion =
+        Source(testByteStrings).runWith(LogRotatorSink(testFunction))
+      val result = Await.result(completion, 3.seconds)
+
+      val sizes = readUpFileSizesAndClean(files)
+      sizes should contain theSameElementsAs Seq(2002L, 2002L, 2002L)
```
johanandren: Would be nice to assert that the file contents is what we expect it to be as well, just to be sure the boundary didn't duplicate data or something like that.

file/src/test/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSinkSpec.scala
```diff
@@ -0,0 +1,164 @@
...
+
+  def readUpFileSizesAndClean(files: Seq[Path]): Seq[Long] = {
+    var ret = Seq.empty[Long]
+    files.foreach { path =>
+      ret = ret :+ Files.size(path)
+      Files.delete(path)
+    }
+    ret
+  }
+}
```
johanandren: Missing test coverage the unhappy case when:
 * upstream fails
 * downstream fails (write to a path that doesn't exist for example?)
 * (maybe) the path function throws exception (is downstream and upstream torn down correctly)

file/src/main/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSink.scala
tg44: Thats a good question. In fact its doesn't matter if it fails or success(es?), cos its just wait for the (proper) release of the file. So it's a fifty fifty if its need to fail if the filesinks onComplete fails or not.

But if you think I can fail the whole stage :)

file/src/main/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSink.scala
tg44: Yapp good catch.

file/src/main/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSink.scala
johanandren: Ah, good point, hmm, yeah, I think that it should fail the stage and the promise even if this is just closing downstream. I think it would be wrong to let that go unnoticed. Note that it will require an async-handler, it's not safe touching internal stage stuff from future callbacks.

tg44:
Ok, I solved most of the mentions, but I have two questions:
1. If I want to fail the stage when the last filesink future is failing and I don't want to do it in a callback I need to use Await.result? (Most of the cases it's an antipattern so I want be sure that there are no better way :) ) 

2. If I want to fail the stage when any filesink fails I need to either know; if its promise is completed with fail, or if the downstream closed (with an error). Is there any inner thing that I didn't find for the second option? I tried the onDownstreamFinished in the OutHandler, but that function is not called after the FileSubscriber cancel, and I don't really want to go deep in the Future hacking if there is an easy way.

Now just need a small cleanup in the newly written tests, a proper downstream fail handling, and a java api :)

johanandren:
1. You could attach an identifier with "current", say a counter/id, to the file, then close over that when closing the filesink, and pass back through async callback and then only fail the stage if the id is the same as current.
2. pass the result of the `Future[IOResult]`s through the async callback and you can deal with it in the context of this stage safely.

file/src/main/scala/akka/stream/alpakka/file/javadsl/LogRotatorSink.scala
```diff
@@ -0,0 +1,26 @@
...
+
+import akka.Done
+import akka.stream.javadsl
+import akka.stream.javadsl.Sink
+import akka.util.ByteString
+
+import scala.collection.JavaConverters._
+
+object LogRotatorSink {
+  def createFromFunction[T](
```
tg44: Any suggestion how to implement properly a javadsl for this?

file/src/test/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSinkSpec.scala
```diff
@@ -0,0 +1,227 @@
...
+      val path = Files.createTempFile(fs.getPath("/"), "test", ".log")
+      val testFunction = () => { (x: ByteString) =>
+        {
+          Option(path)
+        }
+      }
+      val (probe, completion) =
+        TestSource.probe[ByteString].toMat(LogRotatorSink(testFunction, Set(StandardOpenOption.READ)))(Keep.both).run()
+      probe.sendNext(ByteString("test"))
+      probe.expectCancellation()
```
tg44: This cancellation never happens and I have no idea why. I tried a lot of callback failstage implementation, but its stuck in a waiting stage somewhere and I'm not as expert yet to find out what is the problem.

tg44:
I commented in the source, the javadsl is a totaly black area for me, and the stage completition is not working when the downstream fails :(

file/src/main/scala/akka/stream/alpakka/file/javadsl/LogRotatorSink.scala
johanandren: For name `create` is good for both of them and is how the Java-DSL factories most often looks.

For the parameters, there is `java.util.function.Supplier` for the factory function and `java.util.function.Function` for the actual bytes=>path function, instead of `Option` use `Optional`. This will force you to do some toJava/asJava conversions and explicit calls to `apply` of the Java types but should not be too invasive.

file/src/test/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSinkSpec.scala
johanandren: So the test just times out here?

tg44: Yapp, its looks like that the cancle(in) (or the failstage) in the callback function not closes the upstream properly.

johanandren:
I wont be available to look at this for one or two weeks I'm afraid. I'll get back to you as soon as I can.

tg44:
No problem, my first implementation still running on my server, so its not urgent to me. 
I'm trying to figure out whats wrong on the next weekend again but I'm busy too.

tg44:
Long time ago...

So I had a small freetime, and find this unresolved task. I think I fixed all the previous problems.

johanandren:
Fixes #555

Issue #60, opened 2016-11-15
CSV Component


skhatri:
I can write CSV component, if you like.

I was thinking of creating a CSVFlow as well as CSVSourceStage.
Where CSVFlow will take a ByteString and chunk it, the CSVSourceStage could be a wrapper around FileIO. Let me know if this will be an acceptable component.

hseeberger:
AFAIK Camel has it, too, so this makes sense to me.

stephennancekivell:
Great idea!
Might be nice to wrap open csv or something to take care of the escaping nuance's.

Or scala-csv / pure-csv / mighty csv.

joearasin:
Not sure if it's optimal, but we've been using something along the lines of this to write CSVs, built around commons-csv:

```
  def forFormat(csvFormat: CSVFormat, charset: Charset = StandardCharsets.UTF_8): Flow[Seq[Any], ByteString, NotUsed] = {
    Flow[Seq[Any]].map {
      cells => {
        val buf: ByteStringBuilder = ByteString.newBuilder
        val writer = new OutputStreamWriter(buf.asOutputStream, charset)
        csvFormat.printRecord(writer, cells.map(_.asInstanceOf[AnyRef]): _*)
        writer.close()
        buf.result()
      }
    }
  }
```

On top of this, one thing worth keeping ind mind with an API sort of like this is possibly padding the inputs, to deal with varying-length sequences that may be truncated.

ennru:
@skhatri Did you put any more effort into this?

What I find in the existing CSV libraries (commons-csv, open csv, scala-csv) is that they tend to take control over file IO, which is not what we want in an Akka Stream. I've done some parsing Stages with scala-csv, but converting to CSV is a bit like what @joearasin shows for commons-csv.

johanandren:
On the parser side we'd want it to be push/record based, in that we can push bytestrings on it and potentially get a record back, just like the JSON framing in Akka stream core: https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/scaladsl/JsonFraming.scala

ennru:
So the expected signature could be something like 
`val csvFlow: Flow[ByteString, Source[ByteString, NotUsed], NotUsed] = CsvFraming()` 
where the created stream contains one element per column in the incoming CSV ByteString and a new element for every "line" (CSV columns may contain line breaks).

Or should it make things simpler and create elements of `immutable.Seq[ByteString]`.

[...]

PR #213, opened 2017-03-07
CSV data transformation module

ennru:
This is a proposal to add Comma-separated Values handling support to Alpakka (see #60).

For now there is a framing stage to frame `ByteString` data into `List[ByteString]`. The CSV lines must be separated with either LF or CR + LF, the stage will try parsing the same line again until it finds a line ending.

Other things I'd like to add:
* stage to create a `Map[String, ByteString]` by either providing column names or treating the first line in the flow as column names
* stage to select columns from the `Map` into a List again
* stage to convert a `List` into valid CSV `ByteString`s for output

There is an initial documentation page.

Please have a look and point things out that could be done in a better way.


typesafehub-validator:
Hi @ennru,

johanandren:
Isn't the order of columns relevant (so that it would have to be `Seq[(String, ByteString)]` rather than `Map`?

csv/src/main/java/akka/stream/alpakka/csv/javadsl/CsvFraming.java
```diff
@@ -0,0 +1,27 @@
...
+import akka.util.ByteString;
+import scala.collection.JavaConversions;
+
+import java.util.Collection;
+
+public class CsvFraming {
+
+    public static final byte backslash = '\\';
+    public static final byte comma = ',';
+    public static final byte doubleQuote = '"';
```
johanandren: Java style for constants is UPPER_SNAKE_CASES

docs/src/main/paradox/csv.md
```diff
@@ -0,0 +1,65 @@
+# Comma-Separated Values - CSV
+
+Comma-Separated Values are used as interchange format for tabular data 
+of text. This format is supported by most
```
johanandren: by most who?

csv/src/test/scala/akka/stream/alpakka/csv/CsvParserSpec.scala
```diff
@@ -0,0 +1,295 @@
...
+
+  "CSV parser" should {
+    "read comma separated values into a list" in {
+      val in = ByteString("one,two,three\n")
+      val parser = new CsvParser()
+      parser.offer(in)
+      val res = parser.poll()
+      res.value.map(_.utf8String) should be(List("one", "two", "three"))
+      parser.poll() should be('empty)
+    }
```
johanandren: Maybe a nice little abstraction would be in place here, as it is pretty much a long repetition of the same logic with different input and output? (Not necessarily something to do anything about though)

csv/src/main/java/akka/stream/alpakka/csv/javadsl/CsvFraming.java
```diff
@@ -0,0 +1,27 @@
+package akka.stream.alpakka.csv.javadsl;
```
johanandren: Copyright headers in all files, thanks!

csv/src/main/scala/akka/stream/alpakka/csv/scaladsl/CsvFraming.scala
```diff
@@ -0,0 +1,64 @@
...
+import akka.util.ByteString
+
+import scala.util.control.NonFatal
+
+/** Provides CSV framing stages that can separate CSV lines from incoming [[ByteString]] objects. */
+object CsvFraming {
+
+  val backslash: Byte = '\\'
+  val comma: Byte = ','
+  val doubleQuote: Byte = '"'
```
johanandren: Uppercase first letter for Scala constants

csv/src/main/scala/akka/stream/alpakka/csv/scaladsl/CsvFraming.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  def lineScanner(delimiter: Byte = comma,
+                  quoteChar: Byte = doubleQuote,
+                  escapeChar: Byte = backslash): Flow[ByteString, List[ByteString], NotUsed] =
+    Flow[ByteString].via(new GraphStage[FlowShape[ByteString, List[ByteString]]] {
```
johanandren: I'd rather see an actual class than anonymous in the factory like this (and that is the pattern in the other connectors in Alpakka).

ennru:
Tack @johanandren,
So I'm not too far off. I will try to give it a bit more love the next days.
Coding Java was hard...

ennru:
@johanandren 
What is a good way to tackle Scala<->Java collection conversions in the Flow?
The parsing flow is `Flow<ByteString, Collection<ByteString>, NotUsed>`, but the Scala implementation of my to map conversion expects `scala.List`. Best choice to implement a clean stage with Java collection APIs?

johanandren:
In general we tend to be on the Scala side and use the JavaConverters to make the Java APIs happen, not quite sure about this case. Would it be easier to combine the two if you accepted `Seq` perhaps?

ennru:
I ended up implementing a stage with Java-collections maybe not the DRYest solution, but better than converting back and forth between Scala and Java collections.

csv/src/main/scala/akka/stream/alpakka/csv/CsvToMapJavaStage.scala
```diff
@@ -0,0 +1,70 @@
...
+import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
+import akka.util.ByteString
+
+/**
+ * Java API: Converts incoming {@link Collection}<{@link ByteString}> to {@link java.util.Map}<String, ByteString>.
+ *
+ * @param columnNames If given, these names are used as map keys; if not first stream element is used
+ * @param charset Character set used to convert header line ByteString to String
+ */
+class CsvToMapJavaStage(columnNames: Option[ju.Collection[String]], charset: Charset)
```
johanandren: As it is the Java api it should be `Optional` here rather than `Option`

Also, "if you liked it, make it final" ;)

csv/src/main/scala/akka/stream/alpakka/csv/CsvToMapJavaStage.scala
```diff
@@ -0,0 +1,70 @@
...
+              emit(out, map)
+            }
+          }
+        }
+      )
+
+      setHandler(out,
+        new OutHandler {
+        override def onPull(): Unit = pull(in)
+      })
```
johanandren: We often let the logic itself implement the InHandler and OutHandler to avoid the two extra allocations per materialization. Not sure it matters much here.

csv/src/main/scala/akka/stream/alpakka/csv/CsvToMapJavaStage.scala
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  private def zipWithHeaders(elem: ju.Collection[ByteString]): ju.Map[String, ByteString] = {
+    val map = new ju.HashMap[String, ByteString]()
+    val hIter = headers.get.iterator()
+    val colIter = elem.iterator()
+    while (hIter.hasNext && colIter.hasNext) {
+      map.put(hIter.next(), colIter.next())
+    }
+    map
```
johanandren: If the number of elements differ, is that an error perhaps?

csv/src/main/scala/akka/stream/alpakka/csv/CsvToMapStage.scala
```diff
@@ -12,15 +12,16 @@ import akka.util.ByteString
...
- * Converts incoming List[ByteString] to Map[String, ByteString].
+ * Scala API: Converts incoming [[List[ByteString]]] to [[Map[String, ByteString]]].
+ * @see akka.stream.alpakka.csv.CsvToMapJavaStage
  *
  * @param columnNames If given, these names are used as map keys; if not first stream element is used
  * @param charset Character set used to convert header line ByteString to String
  */
-class CsvToMapStage(columnNames: Option[immutable.Seq[String]], charset: Charset)
-    extends GraphStage[FlowShape[immutable.List[ByteString], Map[String, ByteString]]] {
+class CsvToMapStage(columnNames: Option[List[String]], charset: Charset)
```
johanandren: isn't immutable `Seq` better here, to give more flexibility to what can be accepted ("be liberal in what you accept and conservative in what you send")?

I hope I didn't say "Just use List" when we last talked...

csv/src/main/scala/akka/stream/alpakka/csv/CsvParser.scala
```diff
@@ -0,0 +1,299 @@
...
+  private final val LineEnd = 3
+  private final val QuoteStarted = 4
+  private final val QuoteEnd = 5
+  private final val WithinQuotedField = 6
+
+  private final val LF: Byte = '\n'
+  private final val CR: Byte = '\r'
+}
+
+class CsvParser(delimiter: Byte, quoteChar: Byte, escapeChar: Byte) {
```
johanandren: final, maybe private[csv] also rather than making this a public api?

csv/src/main/scala/akka/stream/alpakka/csv/CsvToMapJavaStage.scala
```diff
@@ -0,0 +1,70 @@
...
+        in,
+        new InHandler {
+          override def onPush(): Unit = {
+            val elem = grab(in)
+            if (headers.isEmpty) {
+              headers = Some(decode(elem))
+              pull(in)
+            } else {
+              val map = zipWithHeaders(elem)
+              emit(out, map)
```
johanandren: No need to emit here, just push as its a single element we push downstream, and we know we had a downstream pull or else we wouldn't get here.

csv/src/main/scala/akka/stream/alpakka/csv/CsvToMapJavaStage.scala
ennru: This part is not API, I'll change it to `private[csv]`, but change to `j.u.Optional` anyway.

csv/src/main/scala/akka/stream/alpakka/csv/CsvToMapJavaStage.scala
ennru: No, the rest i just ignored. Just as in `Seq.zip`.

csv/src/main/java/akka/stream/alpakka/csv/javadsl/CsvToMap.java
```diff
@@ -0,0 +1,47 @@
...
+        return Flow.fromGraph(stage).named("csvToMap");
+    }
+
+    /**
+     * A flow translating incoming {@link Collection<ByteString>} to a {@link Map<String, ByteString>} using the given headers
+     * as keys.
+     * @param headers column names to be used as map keys
+     */
+    public static Flow<Collection<ByteString>, Map<String, ByteString>, ?> withHeaders(String... headers) {
+        CsvToMapJavaStage stage = new CsvToMapJavaStage(Option.apply(Arrays.asList(headers)), StandardCharsets.UTF_8);
```
johanandren: Do'h. Ands it's just the varargs, nevermind that comment.

ennru: "Just the varargs" could be constructed as mutable array, so to really be sure we'd need a protective copy. I do not consider it necessary, though. Agreed?

johanandren: Yeah

ennru:
I'd say this CSV module is feature complete to become part of an Alpakka release now.
Please have an eye on the Java interfaces and bridging to Scala.

project/Dependencies.scala
```diff
@@ -39,6 +39,11 @@ object Dependencies {
     )
   )
 
+  val Csv = Seq(
+    libraryDependencies ++= Seq(
```
jrudolph: This one is not needed ;)

csv/src/main/scala/akka/stream/alpakka/csv/CsvFormattingStage.scala
```diff
@@ -0,0 +1,43 @@
...
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
+import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
+import akka.util.ByteString
+
+import scala.collection.immutable
+
+/**
+ * Internal API: Use [[akka.stream.alpakka.csv.scaladsl.CsvFormatting]] instead.
+ */
+private[csv] class CsvFormattingStage(delimiter: Char,
```
jrudolph: Couldn't this be represented using the built-in  `map` combinator?

ennru:
Hallo Johannes,
Github hat mir nicht verraten, dass du etwas kommentiert hattest.
Ich habe gestern @patriknw und @johanandren getroffen.

A few things I'd like to discuss are
- The framing/parsing works on byte level as of now, this could blow up if the ByteStrings have a less ASCII-like encoding. The safes solution is of course to decode to String and operate on String instead.
- As an earlier comment by @johanandren: The toMap might be more handy to result in Seq[(String, ByteString)]. At least for Scala users that might be more flexible, but for Java-folks?
- A friend told me about some list of CSV-strangenesses, I'll look into it to add broarder support.

Any comments?

jrudolph:
> Github hat mir nicht verraten, dass du etwas kommentiert hattest.
> Ich habe gestern @patriknw und @johanandren getroffen.

:)

> The framing/parsing works on byte level as of now, this could blow up if the ByteStrings have a less ASCII-like encoding. The safes solution is of course to decode to String and operate on String instead.

Yes, that would probably make sense. On the other hand, the framing should probably work with most encodings unchanged on the byte-level. It would only be the interpretation of the cell values that need to take a charset into account.

> As an earlier comment by @johanandren: The toMap might be more handy to result in Seq[(String, ByteString)]. At least for Scala users that might be more flexible, but for Java-folks?

Hmm not sure about that. Might make sense as a faster low-level variant. On the high-level, having each row as a map seems handy (though maybe slow).

ennru:
Hej @jrudolph,
I did some work on charsets and Byte Order Mark support. Let's see if it is useful and improve from there.
Please merge when you have a slot.

jrudolph:
Changes LGTM. Let's merge after PR validation. Do you care for the individual commits? Otherwise, we could stash them into a single one.

A few things that could be improved later:
 * automatically select the right BOM based on charset in formatter
 * parsing from characters / support for multi-byte charsets
 * builder pattern for multi-parameter constructors like `format` for the Java API

jrudolph:
I fixed the broken merge I did, autoformatted some files and simplified the dependency declaration.

ennru:
I'm fine with a single commit for the whole CSV module. 


Am 12. April 2017 12:11:02 MESZ schrieb Johannes Rudolph <notifications@github.com>:

jrudolph:
Great, merged. Thanks, @ennru for working through this :)

PR #214, opened 2017-03-07
Added listBuckets method. Change URI style to path-style URL from virtual-hosted–style URL

Scalahansolo:
The listBuckets method is something my team has been using for awhile to get the list of all keys under and given bucket with the ability to page. 

Our team also has had to use a fork of alpakka for a while due to alpakka using virtual-hosted–style URLs which causes issues for non DNS compliant bucket names. http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html

Path-style URLs solves this issue and should not cause issues for others.


Scalahansolo:
Closing in favor of https://github.com/akka/alpakka/pull/193. Will open another pull request with just the changes to how the URLs are constructed.

PR #217, opened 2017-03-13
[FTP] Critical fix for infinite loop of traversing "." and ".." directories

nktpro:
Current (".") and parent ("..") directories are returned as part of the list, which causes FtpBrowserGraphStage (https://github.com/akka/alpakka/blob/master/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpBrowserGraphStage.scala#L67) to infinitely trying to call listFiles again on these directories.

SftpOperations.scala has this safeguard (https://github.com/nktpro/alpakka/blob/master/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/SftpOperations.scala#L47) but it's missing in FtpOperations.scala


typesafehub-validator:
Hi @nktpro,

PR #218, opened 2017-03-13
[FTP] Critical fix for infinite loop of traversing "." and ".." directories

nktpro:
Current (".") and parent ("..") directories are returned as part of the list, which causes `FtpBrowserGraphStage` (https://github.com/akka/alpakka/blob/master/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpBrowserGraphStage.scala#L67) to infinitely try to call `listFiles()` again on these directories.

`SftpOperations.scala` has this filter (https://github.com/akka/alpakka/blob/master/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/SftpOperations.scala#L47) but it's missing in `FtpOperations.scala`

### DIFF
--- a/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpOperations.scala
+++ b/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpOperations.scala
@@ -4,7 +4,7 @@
 package akka.stream.alpakka.ftp
 package impl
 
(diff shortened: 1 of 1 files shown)


juanjovazquez:
LGTM

johanandren:
Fixes  #223

Issue #219, opened 2017-03-13
Mqtt - ssl socket factory


tdroxler:
Currently we cannot use ssl to connect to a mqtt broker (i.e. we cannot use `setSocketFactory` function on `MqttConnectOptions`).

I have a PR ready for this. #220

PR #220, opened 2017-03-13
Allow to pass a `SSLSocketFactory` to `MqttConnectionSettings`

tdroxler:
with this we can connect with ssl to mqtt broker, see #219


ktoso:
LGTM, thanks

PR #221, opened 2017-03-17
Elasticsearch connector

### LINKED ISSUE #99: Elasticsearch connectors
It could be interesting to have source and flow for the corresponding scroll and bulk in elasticsearch.

takezoe:
fixes #99


elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSinkStage.scala
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.http.entity.StringEntity
+import org.elasticsearch.client.RestClient
+import spray.json._
+
+import scala.collection.JavaConverters._
+import scala.concurrent.{Future, Promise}
+
+final case class ElasticsearchSinkSettings(bufferSize: Int = 10)
+
+final case class IncomingMessage[T](id: Option[String], source: T)
```
jrudolph: This class might not be necessary. Can't we just provide it as a special case of the `Typed` variant giving it the `DefaultJsonProtocol.RootJsObjectFormat` (or, if that doesn't fit type-wise, a simple identity `JsonWriter`)?

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSourceStage.scala
```diff
@@ -0,0 +1,144 @@
...
+      case Some(error) => {
+        failStage(new IllegalStateException(error.toString))
+      }
+    }
+  }
+
+  setHandler(out,
+    new OutHandler {
+    override def onPull(): Unit = {
+      if (buffer.isEmpty) {
```
jrudolph: This can be simplified using `emit` to push the complete list of elements received by `receiveMessages`.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSourceStage.scala
```diff
@@ -0,0 +1,144 @@
...
+import spray.json._
+import DefaultJsonProtocol._
+
+import scala.collection.JavaConverters._
+
+final case class ElasticsearchSourceSettings(bufferSize: Int = 10)
+
+final case class OutgoingMessage[T](id: String, source: T)
+
+final class ElasticsearchSourceStage(indexName: String,
```
jrudolph: Same as above, can probably be provided by creating a `JsonReader[JsObject]` and then using the typed variant.

elasticsearch/src/test/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSpec.scala
```diff
@@ -0,0 +1,156 @@
...
+import org.scalatest.{BeforeAndAfterAll, Matchers, WordSpec}
+import spray.json.JsString
+
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+import scala.collection.JavaConverters._
+import spray.json._
+import DefaultJsonProtocol._
+
+class ElasticsearchSpec extends WordSpec with Matchers with BeforeAndAfterAll {
```
jrudolph: Nice tests.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSinkStage.scala
takezoe: Hmm... In my opinion, holding id and object (which is converted to document json) separately is natural when using Elasticsearch. It's possible to replace with tuple like `(Option[String], T)`.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSinkStage.scala
```diff
@@ -0,0 +1,124 @@
...
+    private def sendBulkRequest(messages: Seq[IncomingMessage[T]]): Unit =
+      try {
+        val json = messages.map { message =>
+          s"""{"index": {"_index": "${indexName}", "_type": "${typeName}"${message.id.map { id =>
+               s""", "_id": "${id}""""
+             }.getOrElse("")}}
+               |${convert(message.source).toString}""".stripMargin
+        }.mkString("", "\n", "\n")
+
+        client.performRequest(
```
takezoe: Exactly. It is almost Apache HTTP Components, but it offers handling multiple nodes when accessing to Elasticsearch cluster via HTTP.
https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/_initialization.html

However it seems to be possible to use [CloseableHttpAsyncClient](http://hc.apache.org/httpcomponents-asyncclient-dev/httpasyncclient/apidocs/org/apache/http/impl/nio/client/CloseableHttpAsyncClient.html). Should I use it instead?

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSourceStage.scala
takezoe: Thanks! I will try that.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSinkStage.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  override def preStart(): Unit =
+    pull(in)
+
+  setHandler(in,
+    new InHandler {
+    override def onPush(): Unit = {
+      val message = grab(in)
+      buffer.addLast(message)
+      if (buffer.size >= settings.bufferSize) {
```
takezoe: That's right. I will try to make that pattern.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSinkStage.scala
jrudolph: Yes, using an asynchronous client would be best. Make sure to deliver callbacks through `getAsyncCallback` as explained here: http://doc.akka.io/docs/akka/2.4.14/scala/stream/stream-customize.html#Using_asynchronous_side-channels

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSinkStage.scala
```diff
@@ -0,0 +1,175 @@
...
+                  case Finished => handleSuccess()
+                  case _ => state.set(Idle)
+                }
+              } else {
+                sendBulkUpdateRequest(messages)
+              }
+
+              sentHandler.invoke(())
+            }
+          }
```
takezoe: Although I tried to move entire this part to an async callback, callback wasn't invoked in second time. It works to call only `tryPull()` in an async callback like this. I expect that something called from that async callback might be blocking the graph, but I can't find a correct way. Hmm...

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSinkStage.scala
jrudolph: Oh, I actually meant the class below. Do we need both `ElasticsearchSinkStage` and `ElasticsearchSinkStageTyped`? Isn't `ElasticsearchSinkStage` just `ElasticsearchSinkStageTyped[JsObject]`?

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSinkStage.scala
jrudolph: Hmm that's still quite dangerous to do because you can run in all kinds of race conditions. The one that I can see is the `state.set(Idle)` case. When the stage fills the buffer after the `val messages` in line 150 but before line 155 is executed, then the state will be set to `Idle` but `tryPull` won't do anything and the stream is locked because it will never be woken up again. Not very likely but still possible.

Once you do all handling in the async callback you can remove all other synchronization like the `AtomicReferences` and the `ConcurrentLinkedHashmap`.

jrudolph: So, can you try to execute everything in the async callback? If it still blocks the graph as you say, can you post the code and maybe a stack trace gathered with `jstack <pid>` on the shell (or any other means of gathering stack traces), so we can figure out what's going on?

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSourceStage.scala
```diff
@@ -0,0 +1,148 @@
...
+            override def onSuccess(response: Response) = responseHandler.invoke(response)
+          }
+        )
+      } else {
+        client.performRequestAsync(
+          "POST",
+          s"/_search/scroll",
+          Map[String, String]().asJava,
+          new StringEntity(Map("scroll" -> "5m", "scroll_id" -> scrollId).toJson.toString),
+          new ResponseListener {
```
jrudolph: The GraphStageLogic could implement the `ResponseListener` directly. That would also remove the code duplication with the above branch.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSourceStage.scala
```diff
@@ -0,0 +1,148 @@
...
+      case Some(error) => {
+        failStage(new IllegalStateException(error.toString))
+      }
+    }
+  }
+
+  setHandler(out,
+    new OutHandler {
+    override def onPull(): Unit =
+      if (started == false) {
```
jrudolph: `if (!started)`

jrudolph:
Looks good! Only small things to fix.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSinkStage.scala
takezoe: Got it! That's right.

takezoe:
@jrudolph Thanks for your review. I fixed almost points you mentioned.

About async callback problem, this is my code to test handling response in async callback.
https://github.com/takezoe/alpakka/commit/193eb0621fe385e83f2ce5612ec493d112c5c02e

And this is a result of jstack. I'm checking now:
https://gist.github.com/takezoe/e3a4b3d4ecb424f69cbc14c599544dab

takezoe:
@jrudolph I'm debugging my sink implementation. An actor to run callback looks to be stopped when upstream is complete even if buffered messages remain, so `AsyncInput` message which has callback is sent to the dead letter. How can I implement in this situation?

takezoe:
@jrudolph Finished to fix. In addition, I added `ElasticsearchFlow` and make sink based on it.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSourceStage.scala
```diff
@@ -0,0 +1,124 @@
...
+      } finally {
+        out.close()
+      }
+    }
+
+    val jsObj = json.parseJson.asJsObject
+
+    jsObj.fields.get("error") match {
+      case None => {
+        val hits = jsObj.fields("hits").asJsObject.fields("hits").asInstanceOf[JsArray]
```
jrudolph: Might make sense here to model the result wrapper also using spray-json?

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSourceStage.scala
```diff
@@ -0,0 +1,124 @@
...
+        } else {
+          scrollId = jsObj.fields("_scroll_id").asInstanceOf[JsString].value
+          val messages = hits.elements.reverse.map { element =>
+            val doc = element.asJsObject
+            val id = doc.fields("_id").asInstanceOf[JsString].value
+            val source = doc.fields("_source").asJsObject
+            OutgoingMessage(id, source.convertTo[T])
+          }
+          emitMultiple(out, messages)
+          sendScrollScanRequest()
```
jrudolph: AFAICS doing it like this would disable backpressure because there's basically a loop `sendScrollScanRequest` -> `handleResponse` -> `sendScrollScanRequest` that immediately slurps the whole data source instead of waiting until the first results are drained by the stream and a new pull comes in.

Fortunately, it should be very simple to fix. Couldn't you remove the `sendScrollScanRequest` here and also remove the condition in `onPull` below. What then should happen is that when you use `emitMultiple`, your below `OutHandler` gets exchanged with an internal `EmitMultiple` handler that will send out the current results. Once all of those results are drained your original `OutHandler` is swapped in again which will then use `sendScrollScanRequest` on the next `onPull`.

takezoe:
> I wonder if we could move the Flow/Sink part (which I haven't reviewed in depth yet) to another PR to unblock this one. WDYT?

It's reasonable. I will do so after fix source part.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSourceStage.scala
takezoe: Do you mean creating case classes to map Elasticsarch's results?

jrudolph: Yes, exactly. Would that be feasible?

takezoe: It would be possible. But the REST client of Elasticsearch is improving rapidly. I expect it will become to return Java models instead of json string as same as the transport client (it's an another type of Elasticsearch client) in the future. So now creating our models might be overinvestment.

takezoe:
@jrudolph Separating this pull request to source and flow/sink is reasonable, but separating testcases might be waste... Since I'm never in a hurry, I can wait your review!

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/scaladsl/ElasticsearchFlow.scala
```diff
@@ -0,0 +1,37 @@
...
+   * Scala API: creates a [[ElasticsearchFlowStage]] that accepts as JsObject
+   */
+  def apply(indexName: String, typeName: String, settings: ElasticsearchSinkSettings)(
+      implicit client: RestClient
+  ): Flow[IncomingMessage[JsObject], Response, NotUsed] =
+    Flow
+      .fromGraph(
+        new ElasticsearchFlowStage(indexName, typeName, client, settings)(DefaultJsonProtocol.RootJsObjectFormat)
+      )
+      .mapAsync(settings.parallelism)(identity)
```
jrudolph: The parallelism setting has no effect as it doesn't matter how many times you call `identity` concurrently. Just use `1` instead.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
```diff
@@ -0,0 +1,146 @@
...
+            "_bulk",
+            Map[String, String]().asJava,
+            new StringEntity(json),
+            this
+          )
+        } catch {
+          case ex: Exception => failStage(ex)
+        }
+
+      setHandler(out, new OutHandler {
```
jrudolph: Just extend the GraphStageLogic with `InHandler with OutHandler` and put the handlers directly on the logic. Then use

```scala
setHandlers(in, out, this)
```

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
```diff
@@ -0,0 +1,146 @@
...
+
+          client.performRequestAsync(
+            "POST",
+            "_bulk",
+            Map[String, String]().asJava,
+            new StringEntity(json),
+            this
+          )
+        } catch {
+          case ex: Exception => failStage(ex)
```
jrudolph: No need for explicit exception handling if it just defers to `failStage`. This will be done automatically.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
```diff
@@ -0,0 +1,146 @@
...
+    settings: ElasticsearchSinkSettings
+)(implicit writer: JsonWriter[T])
+    extends GraphStage[FlowShape[IncomingMessage[T], Future[Response]]] {
+
+  private val in = Inlet[IncomingMessage[T]]("messages")
+  private val out = Outlet[Future[Response]]("result")
+  override val shape = FlowShape(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) with ResponseListener {
```
jrudolph: Could you add a comment that explains the basic workflow of this stage?

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/scaladsl/ElasticsearchFlow.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import akka.NotUsed
+import akka.stream.alpakka.elasticsearch.{ElasticsearchFlowStage, ElasticsearchSinkSettings, IncomingMessage}
+import akka.stream.scaladsl.Flow
+import org.elasticsearch.client.{Response, RestClient}
+import spray.json.{DefaultJsonProtocol, JsObject, JsonWriter}
+
+object ElasticsearchFlow {
+
+  /**
```
jrudolph: It's unclear what the exact semantics of the flow are. How many responses do you get, one per source element or one per batch?

jrudolph: Reading the source code of the GraphStage, I see that it would be one `Response` per batch. Would there be anything you could do when a batch fails? Could it make sense to simplify and just make the GraphStage a sink that fails the flow when the request fails? Or maybe configure it with an exception handler (similar to a supervisor) that knows what to do when a single batch request fails?

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSourceStage.scala
jrudolph: It's not really a big investment and would document the current view about how the API is structured, which fields are expected, etc.

Would it really be more than that:

```scala
case class Hit[T](_id: String, _source: T)
case class Response[T](hits: Seq[Hit[T]], _scroll_id: String)

object Protocol {
  import DefaultJsonProtocol._
  implicit def hitFormat[T: JsonFormat] = jsonFormat2(Hit.apply _)
  implicit def responseFormat[T: JsonFormat] = jsonFormat2(Response.apply _)
}
```

An then it's just

```scala
val response = jsObj.convertTo[Response[T]]
scrollId = response._scroll_id
val messages = response.hits.map(h => OutgoingMessage(h._id, h._source)
```

But maybe I'm missing something ;)

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
```diff
@@ -0,0 +1,146 @@
...
+
+      override def onFailure(exception: Exception): Unit = failureHandler.invoke(exception)
+
+      override def onSuccess(response: Response): Unit = responseHandler.invoke(response)
+
+      private def sendBulkUpdateRequest(messages: Seq[IncomingMessage[T]]): Unit =
+        try {
+          val json = messages
+            .map { message =>
+              s"""{"index": {"_index": "${indexName}", "_type": "${typeName}"${message.id
```
jrudolph: Might make sense to model these using spray-json as well, otherwise, you will have to add escaping for all of those fields.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
```diff
@@ -0,0 +1,146 @@
...
+import scala.collection.JavaConverters._
+import spray.json._
+
+import scala.concurrent.Future
+
+final case class ElasticsearchSinkSettings(bufferSize: Int = 10, parallelism: Int = 1)
+
+final case class IncomingMessage[T](id: Option[String], source: T)
+
+class ElasticsearchFlowStage[T](
```
jrudolph: I think the basic approach here is sound.

jrudolph:
Another thing that will be needed in the end is documentation, see https://github.com/akka/alpakka/blob/master/CONTRIBUTING.md#documentation.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSourceStage.scala
takezoe: This demands `JsonFormat` for Source despite only `JsonReader` is necessary essentially. And hand written readers for `Hit` and `Response` are not much distance from current code. I'm not sure whether I should do it.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/scaladsl/ElasticsearchFlow.scala
takezoe: Response of bulk request contains results per command. So it's possible to handle errors if some commands in bulk request failed. I fixed to fail `ElasticsearchFlowStage` when at least one command failed, but I'm not sure this is the best solution.

takezoe:
Fixed codes which are commented by @jrudolph. In addition, I updated following stuffs as well:

- Added abstraction layer for JSON conversion
  - since spray-json is hard to use in Java, I fixed to use Jackson in Java API instead.
- Added tests for ElasticsearchFlow and Java API

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
```diff
@@ -0,0 +1,187 @@
...
+  private val in = Inlet[IncomingMessage[T]]("messages")
+  private val out = Outlet[Future[R]]("failed")
+  override val shape = FlowShape(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new TimerGraphStageLogic(shape) with InHandler with OutHandler {
+
+      private var state: State = Idle
+      private val queue = new mutable.Queue[IncomingMessage[T]]()
+      private val failureHandler = getAsyncCallback[(Seq[IncomingMessage[T]], Throwable)](handleFailure)
```
jrudolph: I guess you could have used a single handler with type `getAsyncCallback[(Seq[IncomingMessage[T]], Try[Response])]` instead, but that's fine for now as well.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSourceStage.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  val out: Outlet[OutgoingMessage[T]] = Outlet("ElasticsearchSource.out")
+  override val shape: SourceShape[OutgoingMessage[T]] = SourceShape(out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new ElasticsearchSourceLogic[T](indexName, typeName, query, client, settings, out, shape, reader)
+
+}
+
+sealed class ElasticsearchSourceLogic[T](indexName: String,
```
jrudolph: `final` instead of `sealed`?

Issue #191, opened 2017-02-16
AMQP: Provide full configuration of the ConnectionFactory


avandendaele:
Currently, the `AmqpConnector` only supports `setUri`, `setHost`, `setPort`, `setUsername`, `setPassword`, `setVirtualHost` and `useSslProtocol` to configure the  `ConnectionFactory`.
I would like to be able to configure all settings of the `ConnectionFactory` such as `setRequestedHeartbeat `, `setConnectionTimeout `, `setHandshakeTimeout`, `setShutdownTimeout `, `setNetworkRecoveryInterval `, `setAutomaticRecoveryEnabled `, `setTopologyRecoveryEnabled ` and `setExceptionHandler`.

johanandren:
Sounds like a relatively simple change, up for making a PR adding those to the connection DSL?

avandendaele:
@johanandren Yes it is. I'll make a PR as soon as I have time to.

johanandren:
Awesome!

matsluni:
There is also a method called newConnection, which takes an Address or a list of addresses. I think it would also be a good addition to support those, e.g. to list all cluster nodes.

avandendaele:
@johanandren Any naming convention or preference about the variable pattern [here](https://github.com/akka/alpakka/blob/master/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala#L18) ? I mean with or without `maybe` prefix ?

johanandren:
I'd skip the `maybe`, the type makes it obvious, but it doesn't matter much with a small local scope IMO.

matsluni:
I could also help out on this. Let me know if this helps.

PR #222, opened 2017-03-19
AMQP: add more options to configuration of the ConnectionFactory, #191

matsluni:
Hi,

I added additional configuration options as described in #191. 

Please let me know what you think.

Matthias

### DIFF
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
@@ -15,18 +15,46 @@ private[amqp] trait AmqpConnector {
     val factory = new ConnectionFactory
     settings match {
       case AmqpConnectionUri(uri) => factory.setUri(uri)
-      case AmqpConnectionDetails(host, port, maybeCredentials, maybeVirtualHost, sslProtocol) =>
-        factory.setHost(host)
-        factory.setPort(port)
+      case AmqpConnectionDetails(_,
+                                 maybeCredentials,
+                                 maybeVirtualHost,
+                                 sslProtocol,
+                                 requestedHeartbeat,
+                                 connectionTimeout,
+                                 handshakeTimeout,
+                                 shutdownTimeout,
+                                 networkRecoveryInterval,
+                                 automaticRecoveryEnabled,
+                                 topologyRecoveryEnabled,
+                                 exceptionHandler) =>
         maybeCredentials.foreach { credentials =>
           factory.setUsername(credentials.username)
           factory.setPassword(credentials.password)
         }
         maybeVirtualHost.foreach(factory.setVirtualHost)
         sslProtocol.foreach(factory.useSslProtocol)
+        requestedHeartbeat.foreach(factory.setRequestedHeartbeat)
+        connectionTimeout.foreach(factory.setConnectionTimeout)
+        handshakeTimeout.foreach(factory.setHandshakeTimeout)
+        shutdownTimeout.foreach(factory.setShutdownTimeout)
+        networkRecoveryInterval.foreach(factory.setNetworkRecoveryInterval)
+        automaticRecoveryEnabled.foreach(factory.setAutomaticRecoveryEnabled)
+        topologyRecoveryEnabled.foreach(factory.setTopologyRecoveryEnabled)
+        exceptionHandler.foreach(factory.setExceptionHandler)
       case DefaultAmqpConnection => // leave it be as is
     }
     factory
+  }
+
+  def newConnection(factory: ConnectionFactory, settings: AmqpConnectionSettings): Connection = settings match {
+    case a: AmqpConnectionDetails => {
+      import scala.collection.JavaConverters._
+      if (a.hostAndPortList.nonEmpty)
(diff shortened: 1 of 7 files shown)


amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
```diff
@@ -15,19 +15,42 @@ private[amqp] trait AmqpConnector {
...
-      case AmqpConnectionDetails(host, port, maybeCredentials, maybeVirtualHost, sslProtocol) =>
-        factory.setHost(host)
-        factory.setPort(port)
+      case AmqpConnectionDetails(hostAndport, maybeCredentials, maybeVirtualHost, sslProtocol, requestedHeartbeat,
+          connectionTimeout, handshakeTimeout, shutdownTimeout, networkRecoveryInterval, automaticRecoveryEnabled,
+          topologyRecoveryEnabled, exceptionHandler) =>
+        if (hostAndport.size == 1) {
+          factory.setHost(hostAndport.head._1)
+          factory.setPort(hostAndport.head._2)
+        }
```
2m: If more than one host and port provided, then no host nor port will be set on factory?

amqp/src/main/scala/akka/stream/alpakka/amqp/model.scala
```diff
@@ -121,15 +121,25 @@ object AmqpConnectionUri {
...
-    sslProtocol: Option[String] = None
+    sslProtocol: Option[String] = None,
+    requestedHeartbeat: Option[Int] = None,
+    connectionTimeout: Option[Int] = None,
+    handshakeTimeout: Option[Int] = None,
+    shutdownTimeout: Option[Int] = None,
+    networkRecoveryInterval: Option[Int] = None,
+    automaticRecoveryEnabled: Option[Boolean] = None,
+    topologyRecoveryEnabled: Option[Boolean] = None,
+    exceptionHandler: Option[ExceptionHandler] = None
```
2m: We have to provide a way to set these from Java, which means adding `with***` methods for every one of these properties.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
matsluni: Yes, if there are more than 1 entries, the newConnection() method in the AmqpConnector will be used, where all the entries from the list can be given to create a newConnection.

Or do you think it would be better to set at least one host and port here, just in case....?

amqp/src/main/scala/akka/stream/alpakka/amqp/model.scala
matsluni: Ok, I added the `with***` methods.

matsluni: I also added a `addHostAndPort(hostAndPorts: (String, Int))` and the Java one `addHostAndPort(hostAndPort: akka.japi.Pair[String, Int])` to add host/port combinations to the already given list. I am a bit unsure about the name `addHostAndPort`. What do you think?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
2m: Alright, now I understand the idea. However I still see some discrepancy between the way `hostAndPortList` is handled here and in `newConnection`. 

Here it checks if the size is one. And in newConnection it checks if the list is non empty. So if the hostAndPortList has only one element the host and port will be set here as well as passed to the `factory.newConnection` method. But if the hostAndPortList has more than one element then these will be passed only when calling `factory.newConnection`. 

Can we remove the `factory.setHost` and `factory.setPort` altogether?

amqp/src/main/scala/akka/stream/alpakka/amqp/model.scala
2m: Looking at other connectors (amqp and mqtt) the convention is still to use `with***` method name, but also make it a vararg method, like so:

```scala
@annotation.varargs
def withHostAndPorts(hostAndPorts: akka.japi.Pair[String, Int]*) = ...
```

2m:
@matsluni I have pushed a commit with a formatting fix so CI could run. After a couple of nitpick fixes this can go in.

matsluni:
I was on vacation, but now I am back. I will have a look at it tomorrow.

matsluni:
I have updated the PR with the comments from above, which mainly are:

- stick to the `with***` pattern 
- also removed `setHost` and `setPort` on the `factory` as these are now arguments to `factory.newConnection`

I hope this is good now. Thank you very much for your patience.

Best regards,
Matthias

matsluni:
hmmm.....travis build failed due to formatting error. I did run `test:compile` again but this didn't changed the formatting to the one travis log showed. Any ideas?

2m:
I see. Formatting rules have changed a tiny bit in master and Travis builds PRs after a merging them onto current master giving unformatted code error.

Rebase this PR onto master and then `test:compile` will reformat code according to the latest rules. After committing these formatting changes it should be all good.

matsluni:
I rebased the branch onto current master. Looks better now. :-)

2m:
Good stuff. Thanks a lot. :)

PR #225, opened 2017-03-21
Make Travis fail build on code format differences

bantonsson:


bantonsson:
So in its infinite wisdom Travis continues to run the `sbt test` even after the `diff` has *failed* the build. Argh. I'll have another look at why.

bantonsson:
Excellent. A proper failure that stopped the rest of the build. Let's see if it works without the broken indentation.

2m:
Awesomeness. I think we should have a more contributor friendly message when the diff is not clean. Something like: "Unformatted code found. Please run 'test:compile' and commit the reformatted code."?

bantonsson:
Valid point. I'll change the text.

hseeberger:
LGTM

PR #229, opened 2017-03-23
Make external libs more visible (reactive kafka)

ktoso:
Now people don't notice Kafka is in here since it's "external", expanding the TOC one more level makes it more visible.

WDYT?

### DIFF
--- a/docs/src/main/paradox/index.md
+++ b/docs/src/main/paradox/index.md
@@ -4,11 +4,12 @@ Welcome to the home of the Alpakka initiative, which harbours various Akka Strea
 and data transformations for integration use cases. Here you can find documentation of the components that are
 part of this project as well as links to components that are maintained by other projects.
 
-@@ toc { .main depth=3 }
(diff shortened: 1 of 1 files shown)


2m:
I am not sure if this is the best way to make Reactive Kafka (and other external connectors) more visible. Here is the screenshot of the result:

alpakka

Reactive Kafka and other external connectors became visible in the TOC, but not in the left menu. Also every connector now has *Artifacts* and *Usage* sections visible, which clutters TOC quite a bit.

My proposal would be to move *External Connectors* section one level up. Then external connectors would show up on the left menu and in the TOC with *deph=2*.

ktoso:
Good idea, let's do that.

ktoso:
Split in 2 PRs since done in browser, so we can merge this and this https://github.com/akka/alpakka/pull/237 to achieve the effect @2m explained. 👍

ktoso:
Done, thanks!

PR #230, opened 2017-03-23
=pro update akka http to 10.0.5

ktoso:


ktoso:
Nothing to discuss here I think

PR #242, opened 2017-03-29
S3 path style access

### LINKED ISSUE #64: S3: add support of bucket names with period (.)
based on comment @andrewrapp https://github.com/akka/alpakka/pull/24#pullrequestreview-8701281

"Hi all, looking forward to the s3 support! I played around with it a bit and noticed a cert related error. The issues seems to be related to having periods "." in bucket names, which breaks the s3 wild card certificate. For example, if I have a bucket "has.dots" then I'll get the following error: Caused by: java.security.cert.CertificateException: No subject alternative DNS name matching has.dots.s3.amazonaws.com found.

I was able to work around the issue by moving the bucket out of the host and placing it in the path (ie not using virtual hosted–style buckets):

  ```
def requestHost(s3Location: S3Location): Uri.Host = Uri.Host("s3-us-west-1.amazonaws.com")

  def requestUri(s3Location: S3Location): Uri = Uri(s"/${s3Location.bucket}/${s3Location.key}").withHost(requestHost(s3Location)).withScheme("https")

```
This issue is described in http://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html
Specifically, "When using virtual hosted–style buckets with SSL, the SSL wild card certificate only matches buckets that do not contain periods. To work around this, use HTTP or write your own certificate verification logic."

background: http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html
relates to Ref #24

### PR DESCRIPTION
**Pull request by kobmic**
Added configuration to allow path style access to s3 buckets, i.e. "http://s3-eu-west-1.amazonaws.com/mybucket/myobject".  Default is virtual hosted style. Fixes #64 
(signed Ligthbend CLA yesterday)

### CODE REVIEW

File: s3/src/main/resources/reference.conf [hunk 1]
```diff
@@ -18,4 +18,8 @@ akka.stream.alpakka.s3 {
     secret-access-key = ""
     default-region = "us-west-2"
   }
+
+  # enable path style access to s3, i.e. "http://s3-eu-west-1.amazonaws.com/mybucket/myobject"
+  # default is virtual-hosted style
```
johanandren: add a note about when it is required for things to work!

thereisnospoon:
Hi. I see that the changes are approved. It would be great if you could merge this PR, because changes I want to implement depend on these changes.

johanandren:
Absolutely, sorry for the delay.

Issue #247, opened 2017-04-03
Support of HTTP schema for custom S3 endpoint


thereisnospoon:
In order to use Alpakka S3 with various S3 mocks it would be convenient to have an option to specify `http` schema for HTTP requests to S3 endpoints. I think, it can be achieved by adding `schema` subproperty to `proxy` property which will contain schema name (`https` by default).

PR #248, opened 2017-04-03
Added configuration support for requests over HTTP #247

thereisnospoon:


johanandren:
Refs #247

johanandren:
Since it is only ever `http` or `https`, wouldn't a boolean setting make more sense?

svezfaz:
+1 on the boolean param. Maybe we could have the setting as
```
   proxy {
      ...
      secure = true
    }
```

thereisnospoon:
Thanks for review. I've made changes according to your comments.
I've accidentally committed some housekeeping comments (I've worked from two stations). If there is such a need I can repair that and squash them.

Issue #245, opened 2017-03-31
Failed: SqsSourceTest


2m:
https://travis-ci.org/akka/alpakka/jobs/217164066

```
[info] Test akka.stream.alpakka.sqs.javadsl.SqsSourceTest.streamFromQueueWithCustomClient started
[error] Test akka.stream.alpakka.sqs.javadsl.SqsSourceTest.streamFromQueueWithCustomClient failed: java.util.concurrent.TimeoutException: null, took 5.311 sec
[error]     at java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1771)
[error]     at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1915)
[error]     at scala.concurrent.java8.FuturesConvertersImpl$CF.super$get(FutureConvertersImpl.scala:84)
[error]     at scala.concurrent.java8.FuturesConvertersImpl$CF.$anonfun$get$2(FutureConvertersImpl.scala:84)
[error]     at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:53)
[error]     at scala.concurrent.package$.blocking(package.scala:123)
[error]     at scala.concurrent.java8.FuturesConvertersImpl$CF.get(FutureConvertersImpl.scala:84)
[error]     at akka.stream.alpakka.sqs.javadsl.SqsSourceTest.streamFromQueueWithCustomClient(SqsSourceTest.java:101)
[error]     ...
```

dpfeiffer:
@2m what should we do about that one. As I can see it worked. Do you know if that test is flaky in general? I tried to reproduce it but wasn't able to do so locally?

2m:
When running that one test suite repeatedly I could not reproduce it either. So I think we can just increase the timeout for this particular test and see if it happens more.

dpfeiffer:
Should I create a PR? We could also reduce the amount of messages to the queue in that test (currently we send and receive 100 messages) because the real tests are already done in the `SqsSourceSpec` and the `SqsSourceTest` is only there for testing the Java API.

2m:
PR would be great! Decreasing the number of messages also sounds good.

PR #249, opened 2017-04-04
Fix SqsSourceTest

dpfeiffer:
- Reduce amount of sent messages to 1 (multiple batch streaming is tested in the SqsSourceSpec)
- Increase timeout
- Solves #245


2m:
Thank you!

PR #252, opened 2017-04-06
Initial commit for apache geode connector

cheleb:
First attempt.

### DIFF
--- a/geode/src/main/scala/akka/stream/alpakka/geode/internal/stage/GeodeCQueryGraphLogic.scala
+++ b/geode/src/main/scala/akka/stream/alpakka/geode/internal/stage/GeodeCQueryGraphLogic.scala
@@ -1,0 +1,119 @@
+/*
+ * Copyright (C) 2016-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.alpakka.geode.internal.stage
+
+import java.util
+import java.util.concurrent.Semaphore
+
+import akka.stream.stage.{AsyncCallback, StageLogging}
+import akka.stream.{Outlet, SourceShape}
+import org.apache.geode.cache.client.ClientCache
+import org.apache.geode.cache.query.{CqAttributesFactory, CqEvent, CqQuery, Struct}
+import org.apache.geode.cache.util.CqListenerAdapter
+
+import scala.collection.mutable
+import scala.util.Try
+
+abstract class GeodeCQueryGraphLogic[V](val shape: SourceShape[V],
+                                        val clientCache: ClientCache,
+                                        val queryName: Symbol,
+                                        val sql: String)
+    extends GeodeSourceStageLogic[V](shape, clientCache)
+    with StageLogging {
+
+  /**
+   * Queue containing, only
+   */
+  private val incomingQueue = mutable.Queue[V]()
+
+  private val semaphore = new Semaphore(10)
+
+  val onElement: AsyncCallback[V]
+
+  private var query: CqQuery = _
+
+  override def executeQuery() = Try {
+
+    val cqf = new CqAttributesFactory()
+
+    val eventListener = new CqListenerAdapter() {
+      override def onEvent(ev: CqEvent): Unit =
+        onGeodeElement((ev.getNewValue().asInstanceOf[V]))
+
+      override def onError(ev: CqEvent): Unit =
+        log.error(ev.getThrowable, s"$ev")
+
+      override def close(): Unit = {
+        log.debug("closes")
+        inFinish.invoke(())
+      }
+    }
+    cqf.addCqListener(eventListener)
+
+    val cqa = cqf.create()
+
+    query = qs.newCq(queryName.name, sql, cqa)
+
+    buildInitialResulsIterator(query)
+
+  }
+
+  private def buildInitialResulsIterator(q: CqQuery) = {
+    val res = q.executeWithInitialResults[Struct]
+    val it = res.iterator()
+    new util.Iterator[V] {
+      override def next(): V =
+        it.next().getFieldValues()(1).asInstanceOf[V]
+
+      override def hasNext: Boolean = it.hasNext
+    }
+  }
+
+  /**
+   * May lock on semaphore.acquires().
+   */
+  protected def onGeodeElement(v: V): Unit = {
+    semaphore.acquire()
+    onElement.invoke(v)
+  }
+
+  protected def enqueue(v: V): Unit =
+    incomingQueue.enqueue(v)
+
+  protected def dequeue(): Option[V] =
+    if (incomingQueue.isEmpty)
+      None
+    else
+      Some(incomingQueue.dequeue())
+
+  /**
+   * Pushes an element downstream and releases a semaphore acquired in onGeodeElement.
+   */
+  protected def pushElement(out: Outlet[V], element: V) = {
+    push(out, element)
+    semaphore.release()
+  }
+
+  override def postStop(): Unit = {
+    if (clientCache.isClosed)
+      return
+    qs.closeCqs()
+  }
+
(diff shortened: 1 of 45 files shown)


geode/src/test/scala/akka/stream/alpakka/geode/scaladsl/GeodeContinuousSourceSpec.scala
```diff
@@ -0,0 +1,56 @@
...
+        //#continuousQuery
+        val source =
+          reactiveGeode
+            .continuousQuery[Person]('test, s"select * from /persons")
+            .runWith(Sink.fold(0) { (c, p) =>
+              log.debug(s"$p $c")
+              if (c == 19) {
+                reactiveGeode.closeContinuousQuery('test).foreach { _ =>
+                  log.info("test cQuery is closed")
+                }
```
2m: Is there a reason such style has been chosen? A more stream-like approach to take a finite number of stream elements from a never ending continuous stream would be:

    reactiveGeode.continuosQuery(...).take(19).runWith(Sink.fold(...))

geode/src/main/scala/akka/stream/alpakka/geode/internal/stage/GeodeFlowStage.scala
```diff
@@ -0,0 +1,54 @@
...
+      setHandler(out, new OutHandler {
+        override def onPull() =
+          pull(in)
+      })
+
+      setHandler(in, new InHandler {
+        override def onPush() = {
+          val msg = grab(in)
+
+          put(msg)
```
2m: Do you know how thread friendly is `put` here? Is it going to block and not return until the object has been put to geode?

I was thinking that we could use `put(K key, V value, Object aCallbackArgument)` here and only `pull` when the stage gets called back. WDYT?

geode/src/main/scala/akka/stream/alpakka/geode/internal/stage/GeodeCQueryGraphLogic.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  private def handleQuery(q: CqQuery) = {
+    val res = q.executeWithInitialResults[Struct]()
+
+    val it = res.iterator()
+    while (it.hasNext) {
+      val str = it.next()
+      val v = str.getFieldValues()(1).asInstanceOf[V]
+      onElement.invoke(v)
+    }
```
2m: I see now why the test case for the continuous query was that style. There is actually no way to backpressure this `GeodeCQueryGraphLogic`. It starts and continues to emit elements to the stream from its preStart, which is not the best behaviour for a GraphStage.

Also why is `handleQuery` needed if above there is an `cqf.addCqListener(eventListener)` which also gets elements of type `V`?

geode/src/main/scala/akka/stream/alpakka/geode/internal/stage/GeodeCQueryGraphLogic.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  var query: Option[CqQuery] = None
+
+  override def preStart(): Unit = {
+
+    val cqf = new CqAttributesFactory()
+
+    val eventListener = new CqListenerAdapter() {
+      override def onEvent(ev: CqEvent): Unit =
+        onElement.invoke(ev.getNewValue().asInstanceOf[V])
```
2m: Here `onElement` is:

```
if (isAvailable(out)) pushElement(out, element)
else enqueue(element)
```
which is dangerous as it will grow the queue without any bounds.

Here we have an opportunity to block on the Geode library thread and in such way signal backpressure to the Geode library. Take a look at this [blog post](http://blog.akka.io/integrations/2016/09/05/flow-control-at-the-akka-stream-boundary#blocking-the-library-thread) for a more thorough description how this should be done.

Also you can take a look at Mqtt connector here in Alpakka project which uses [exactly the same strategy](https://github.com/akka/alpakka/blob/master/mqtt/src/main/scala/akka/stream/alpakka/mqtt/MqttSourceStage.scala#L63-L66) to signal backpressure to the Mqtt library.

geode/src/main/scala/akka/stream/alpakka/geode/internal/stage/GeodeQueryGraphLogic.scala
```diff
@@ -0,0 +1,34 @@
...
+import akka.stream.stage.StageLogging
+import org.apache.geode.cache.client.ClientCache
+import org.apache.geode.cache.query.SelectResults
+
+abstract class GeodeQueryGraphLogic[V](val shape: SourceShape[V], val clientCache: ClientCache, val query: String)
+    extends GeodeSourceStageLogic[V](shape, clientCache)
+    with StageLogging {
+
+  override def preStart(): Unit =
+    new Thread() {
```
2m: GraphStages that spawn threads should be avoided. Can you use `CqListenerAdapter` here as you did in `GeodeCQueryGraphLogic` to get notified on a new element?

cheleb:
Thanks a lot, I'll try to fix those issues this week end.

geode/src/test/scala/akka/stream/alpakka/geode/scaladsl/GeodeContinuousSourceSpec.scala
cheleb: Ignorance is the main raison :)
Also I wanted to stop the explictly the continuous query, but that is not the good (stream centric) way

geode/src/main/scala/akka/stream/alpakka/geode/internal/stage/GeodeFlowStage.scala
cheleb: It is blocking (hence the IO dispatcher). The aCallbackArgument, is a serializable argument that will be given to callback *inside* the grid (if they are configured).
AFAIK there is no async / no-blocking client yet.

geode/src/main/scala/akka/stream/alpakka/geode/internal/stage/GeodeQueryGraphLogic.scala
cheleb: It is *not a Continuous query*, I've spawn a thread to apply backpressure on it, without blocking the stage (AFAIU)

viktorklang: I'd suggest using the configured dispatcher and wrapping the blocking logic in a `blocking {}`

cheleb:
I've pushed a new commit without thread issues, sorry for that.
I'm not used at all with docker [0] but I think I will be able to provide
better integration soon (with help of friends :) )

[0]:
https://github.com/cheleb/alpakka/blob/geode-docker/docker-compose.yml#L26


Le ven. 28 avr. 2017 à 11:42, Viktor Klang (√) <notifications@github.com> a
écrit :

2m:
Looking better! And docker setup seems a bit involved but good as well.

cheleb:
Docker geode is up, and less involved...

cheleb:
Geode does support transaction for writing, but I'm not sure how to implement transaction in a streaming context (begin ? / end ?). Is it relevant here ?

2m:
Great updates! Thanks a lot @cheleb 

For the inspiration for tx management I think it would be useful to take a look at the API that is provided by Reactive Kafka (akka-streams Kafka connector): http://doc.akka.io/docs/akka-stream-kafka/current/atleastonce.html

The API provides a committable Source which then allows to acknowledge emmited elements later in the stream.

cheleb:
Thanks a lot too for the review @2m , I will take a look to kafka connector asap.

PR #253, opened 2017-04-06
Added list bucket method to S3 library.

Scalahansolo:
Adds the ability to get a source of all keys under a given prefix.


Scalahansolo:
When running sbt test:compile locally, I get successful runs. Im not sure what exactly is going wrong in the Travis build.

cheleb:
IMHO there it is an "external" issue, I have *sometime* the same error !

Le ven. 7 avr. 2017 à 18:42, Sean Callahan <notifications@github.com> a
écrit :

When running sbt test:compile locally, I get successful runs. Im not sure
what exactly is going wrong in the Travis build.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/akka/alpakka/pull/253#issuecomment-292588161>, or mute
the thread

jrudolph:
I tried to resolve the conflicts (hopefully correctly...).

s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
```diff
@@ -69,42 +92,41 @@ private[alpakka] object HttpRequests {
...
-      val uri = if (conf.pathStyleAccess) {
-        Uri(s"/${s3Location.bucket}/${s3Location.key}").withHost(requestHost(s3Location, region))
-      } else {
-        Uri(s"/${s3Location.key}").withHost(requestHost(s3Location, region))
-      }
-      conf.proxy match {
-        case None => uri.withScheme("https")
-        case Some(proxy) => uri.withPort(proxy.port).withScheme(proxy.scheme)
-      }
+  private[this] def requestHost(bucket: String, region: String)(implicit conf: S3Settings): Uri.Host =
```
jrudolph: Good idea, to put extract that method.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -43,6 +41,8 @@ final case class FailedUpload(reasons: Seq[Throwable]) extends Exception
 
 final case class CompleteMultipartUploadResult(location: Uri, bucket: String, key: String, etag: String)
 
+final case class ListBucketResult(is_truncated: Boolean, continuation_token: Option[String], keys: Seq[String])
```
jrudolph: Maybe we should keep the camel case style for parameters?

s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
```diff
@@ -11,12 +11,35 @@ import akka.http.scaladsl.model.{ContentTypes, RequestEntity, _}
...
 private[alpakka] object HttpRequests {
 
+  def listBucket(
+      bucket: String,
+      region: String,
+      prefix: Option[String] = None,
+      continuation_token: Option[String] = None
+  )(implicit conf: S3Settings): HttpRequest = {
+
+    val listTypeQuery: (String, String) = "list-type" -> "2"
```
jrudolph: I guess this implementation is fine as is. 

It could probably be simplified by collecting all parameters as a `Seq[(String, Option[String])` and then converting that to a Map, filtering out the `None` values.

Something like

```scala
Seq("list-type" -> Some("2"),
  "prefix" -> prefix,
  "continuation-token" -> continuation_token.map(_.replace...)
).collect { case (k, Some(v)) => k -> v }.toMap
```

Scalahansolo:
All still looks good at first glance.

s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
```diff
@@ -11,12 +11,35 @@ import akka.http.scaladsl.model.{ContentTypes, RequestEntity, _}
...
+    val prefixQuery: Option[(String, String)] = prefix.map("prefix" -> _)
+    val continuationTokenQuery: Option[(String, String)] =
+      continuation_token.map("continuation-token" -> _.replaceAll("=", "%3D"))
```
jrudolph: Why does only this character need to be replaced? Are you use that's correct at all, because `Uri.query` will also encode characters, so that the percent character may then be double encoded?

s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
```diff
@@ -11,12 +11,35 @@ import akka.http.scaladsl.model.{ContentTypes, RequestEntity, _}
...
 import scala.collection.immutable.Seq
 import scala.concurrent.{ExecutionContext, Future}
 
 private[alpakka] object HttpRequests {
 
+  def listBucket(
+      bucket: String,
+      region: String,
+      prefix: Option[String] = None,
+      continuation_token: Option[String] = None
```
jrudolph: Let's use camel case there as well.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -65,6 +65,24 @@ private[alpakka] final class S3Stream(credentials: AWSCredentials,
...
   }
 
+  def listBucket(bucket: String, prefix: Option[String] = None): Source[String, NotUsed] = {
+    def listBucketCall(continuation_token: Option[String] = None): Future[ListBucketResult] =
+      signAndGetAs[ListBucketResult](HttpRequests.listBucket(bucket, region, prefix, continuation_token))
+
+    def fileSourceFromFuture(f: Future[ListBucketResult]): Source[String, NotUsed] =
+      Source
+        .fromFuture(f)
+        .flatMapConcat((res: ListBucketResult) => {
```
jrudolph: The recursive calls here look like a nice pattern but it doesn't run in constant memory as expected for streams and may lead to OOM eventually. See the comment https://github.com/akka/akka/issues/22352#issuecomment-281624764 here for a potential workaround.

s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
```diff
@@ -11,12 +11,35 @@ import akka.http.scaladsl.model.{ContentTypes, RequestEntity, _}
 import akka.stream.alpakka.s3.S3Settings
 import akka.stream.scaladsl.Source
 import akka.util.ByteString
-
 import scala.collection.immutable.Seq
 import scala.concurrent.{ExecutionContext, Future}
 
 private[alpakka] object HttpRequests {
 
+  def listBucket(
```
jrudolph: I wonder if it wouldn't make sense to use separate methods for the initial call and the continuation. It seems that you either would want to add a prefix or a token both never both?

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -65,6 +65,24 @@ private[alpakka] final class S3Stream(credentials: AWSCredentials,
     Source.fromFuture(request(s3Location, range).flatMap(entityForSuccess).map(_.dataBytes)).flatMapConcat(identity)
   }
 
+  def listBucket(bucket: String, prefix: Option[String] = None): Source[String, NotUsed] = {
+    def listBucketCall(continuation_token: Option[String] = None): Future[ListBucketResult] =
+      signAndGetAs[ListBucketResult](HttpRequests.listBucket(bucket, region, prefix, continuation_token))
+
+    def fileSourceFromFuture(f: Future[ListBucketResult]): Source[String, NotUsed] =
```
jrudolph: Why is it called `fileSourceFromFuture`? ;)

s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
Scalahansolo: From how I understand this, the continuation token needs to be passed with the same exact same request as the one before it in order to properly page, so it is valid that on the first request you would pass a prefix and no token, but on the next call, you would be passing both.

s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
Scalahansolo: Will fix

s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
Scalahansolo: Ya, this is a little nicer. Still leveling up on my use of the collections in scala and using them properly.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
Scalahansolo: Not the best name now that you mention it. Hah. Got any better ideas?

Scalahansolo:
Also, for what its worth, I have been using a forked version of the repo for a while now using this method with success.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
Scalahansolo: Interesting. Ill dig in and try to get something working.

Scalahansolo: After digging a little here, Im not quite sure how you use this technique to page through the results properly. I dont how statefulMapConcat gives you access to the previous value than continues to page until completed.

jrudolph: In the meantime I figured that using `Source.unfoldAsync` would probably even simpler. It allows you to thread state through between each call. So, it should basically look similar to this:

```scala
def getNextBucket(continuationToken: Option[String]): Future[Option[(String, Seq[...])]] // needs to return a future of a tuple `Some(next_continuation_token, sequence of results)` (or `None` if complete).

Source.unfoldAsync(None)(getNextBucket)
   .mapConcat(identity) // flatten sequence of results
```

Can you try that?

jrudolph:
The increased memory usage will probably only show for particular use cases. @Scalahansolo, could you try `unfoldAsync`? It might even simplify the code. Otherwise, I think this only needs the smallish fixups I commented last time.

Scalahansolo:
The unfoldAsync ended up working out. Thats a really cool API method that I was unaware of. The type on the listBucket call ended up being pretty brutal, but the logic seems sound.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -65,6 +65,31 @@ private[alpakka] final class S3Stream(credentials: AWSCredentials,
...
+    import system.dispatcher
+
+    /**
+     *
+     * @param args The Option[String] is our continuation token and the Boolean is our holder
+     *             for wether this is the first time the method was called. When the token is None
+     *             and the first call Boolean is false, we know that we are done getting keys.
+     * @return
+     */
+    def listBucketCall(args: (Option[String], Boolean)): Future[Option[((Option[String], Boolean), Seq[String])]] =
```
jrudolph: Ah, I see, you need to distinguish "Start", "Running", and "End" states. Maybe it would make sense to create a three state ADT instead? That would read a bit cleaner:

```scala
sealed trait State
case object Starting extends State
case class Running(continuationTokenL String) extends State
case object Finished extends State
```

WDYT?

Scalahansolo: Ya, that would be a little cleaner for sure. Ill see what I can come up. Will update this soon.

Scalahansolo:
I think that you'll like this  a bit better. I definitely feel this is cleaner and much easier to read.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -65,6 +70,27 @@ private[alpakka] final class S3Stream(credentials: AWSCredentials,
...
+      signAndGetAs[ListBucketResult](HttpRequests.listBucket(bucket, region, prefix, token))
+        .map { (res: ListBucketResult) =>
+          Some(
+            res.continuationToken
+              .fold[(ListBucketState, Seq[String])]((Finished, res.keys))(t => (Running(t), res.keys))
+          )
+        }
+
+    Source
+      .unfoldAsync[ListBucketState, Seq[String]](Starting) {
```
jrudolph: Yes, this is great now.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -43,6 +41,13 @@ final case class FailedUpload(reasons: Seq[Throwable]) extends Exception
 
 final case class CompleteMultipartUploadResult(location: Uri, bucket: String, key: String, etag: String)
 
+final case class ListBucketResult(isTruncated: Boolean, continuationToken: Option[String], keys: Seq[String])
+
+sealed trait ListBucketState
```
jrudolph: Can you move those into a private place so that it doesn't clutter the public namespace? Would be ok from my side to but the whole hierarchy into the `listBucket` method if it's only used there.

Scalahansolo: Done :)

jrudolph:
Great one, thanks a lot @Scalahansolo.

Issue #235, opened 2017-03-27
Failed: AwsLambdaFlowSpec


2m:
Failed while validating #233

https://travis-ci.org/akka/alpakka/jobs/215442139

```
[info] AwsLambdaFlow
[info] - should call a single invoke request *** FAILED *** (226 milliseconds)
[info]   org.mockito.exceptions.verification.TooLittleActualInvocations: aWSLambdaAsyncClient.invokeAsync(
[info]     {FunctionName: test_function,Payload: java.nio.HeapByteBuffer[pos=0 lim=12 cap=12],},
[info]     <any>
[info] );
[info] Wanted 1 time:
[info] -> at akka.stream.alpakka.awslambda.scaladsl.AwsLambdaFlowSpec$$anonfun$1$$anonfun$apply$mcV$sp$1.apply(AwsLambdaFlowSpec.scala:69)
[info] But was 0 times.
[info]   at akka.stream.alpakka.awslambda.scaladsl.AwsLambdaFlowSpec$$anonfun$1$$anonfun$apply$mcV$sp$1.apply(AwsLambdaFlowSpec.scala:69)
[info]   at akka.stream.alpakka.awslambda.scaladsl.AwsLambdaFlowSpec$$anonfun$1$$anonfun$apply$mcV$sp$1.apply(AwsLambdaFlowSpec.scala:54)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.WordSpecLike$$anon$1.apply(WordSpecLike.scala:1078)
[info]   at org.scalatest.TestSuite$class.withFixture(TestSuite.scala:196)
[info]   at akka.stream.alpakka.awslambda.scaladsl.AwsLambdaFlowSpec.withFixture(AwsLambdaFlowSpec.scala:24)
[info]   at org.scalatest.WordSpecLike$class.invokeWithFixture$1(WordSpecLike.scala:1075)
[info]   at org.scalatest.WordSpecLike$$anonfun$runTest$1.apply(WordSpecLike.scala:1088)
[info]   at org.scalatest.WordSpecLike$$anonfun$runTest$1.apply(WordSpecLike.scala:1088)
[...]

PR #254, opened 2017-04-07
s3: support for encryption, storage class, custom headers

### LINKED ISSUE #109: S3 Encryption support
Add Encryption support for S3, PR started against s3-stream:
https://github.com/bluelabsio/s3-stream/pull/15

kobmic:
Implemented server side encryption, added s3 storage class and custom headers. Fixes #109

Tried not to break the current S3Client multipartUpload API, even if I'm not to happy with always passing CannedAcl and MetaHeaders. Would be nicer to have Options for these arguments, i.e. always passing the default 'CannedAcl.Private' makes it impossible to add a custom 'x-amz-grant-read' header (conflicting).


kobmic:
Saw that Travis build of my PR failed with an error in AwsLambdaFlowSpec and that other PRs had the same issue - anything I can do to make Travis happy?

johanandren:
We're tracking that in #235, I guess someone re-triggered the build because this PR is green now.

svezfaz:
@kobmic I agree, the S3Client API looks a bit brittle. I don't think we should go great length not to break it.. this should be all experimental ATM? Might be a good idea to change `S3Client` API now to use optional params.

kobmic:
If it's ok to change the current API I would refactor my PR, move the params to the new S3Header case class and make them optional. What do you think?

svezfaz:
It makes sense to me. @johanandren WDYT?

johanandren:
I agree that this is better, but could we provide a `@deprecated` overload of the method that keeps the old signature? I'ts always unpleasant to have end user code break.

kobmic:
The compiler won't let me have multiple overloaded alternatives of method 'multipartUpload' with default arguments (which I still want in both cases). So I ended up with a 'multipartUploadWithHeaders' alternative without deprecating the old API.

Issue #189, opened 2017-02-13
failed: MqttSourceSpec


patriknw:
https://travis-ci.org/akka/alpakka/jobs/201082367

```
[info] - should support will message *** FAILED *** (5 seconds, 234 milliseconds)
[info]   A timeout occurred waiting for a future to complete. Queried 51 times, sleeping 100000000 nanoseconds between each query. (MqttSourceSpec.scala:193)
[info]   org.scalatest.concurrent.Futures$FutureConcept$$anon$1:
[info]   at org.scalatest.concurrent.Futures$FutureConcept$class.tryTryAgain$1(Futures.scala:538)
[info]   at org.scalatest.concurrent.Futures$FutureConcept$class.futureValueImpl(Futures.scala:550)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:275)
[info]   at org.scalatest.concurrent.Futures$class.whenReady(Futures.scala:676)
[info]   at akka.stream.alpakka.mqtt.scaladsl.MqttSourceSpec.whenReady(MqttSourceSpec.scala:21)
[info]   at akka.stream.alpakka.mqtt.scaladsl.MqttSourceSpec$$anonfun$1$$anonfun$apply$mcV$sp$10$$anonfun$apply$11.apply(MqttSourceSpec.scala:193)
[info]   at akka.stream.alpakka.mqtt.scaladsl.MqttSourceSpec$$anonfun$1$$anonfun$apply$mcV$sp$10$$anonfun$apply$11.apply(MqttSourceSpec.scala:183)
```

2m:
Closing it as #263 got merged. WIll open a new one if this spec fails again.

PR #263, opened 2017-04-13
Fail MqttSourceStage mat. value on connection loss

raboof:
And increase the timeout. Might help with #189, or otherwise help generate a
better error message when it does happen again.


2m:
Interestingly the spec under question did not fail on my machine after 200 test runs after applying this change. To increasing the timeout appears to really have helped.

LGTM

Issue #264, opened 2017-04-13
Support of partial file download from S3


thereisnospoon:
It would be great to have an ability to create stream for download only of a given bytes range of a file.

johanandren:
As far as I understand, the range is just headers, would be good if this was solved the same way as #254, allowing for more custom headers to be added without changing the API every time. WDYT @thereisnospoon ?

jrudolph:
Fixed by #265.

jrudolph:
@johanandren not sure if the range header is as general as the ones tackled in #254. AFAICS the range header is mostly relevant for downloads so it seems to fit somewhat well here. The header added in #254 seemed to be more meta-level headers so it makes sense to configure them in a single place.

johanandren:
Alright, thanks for clarifying @jrudolph

PR #265, opened 2017-04-13
Added support for partial file download from S3 #264

thereisnospoon:
Please review

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
@@ -10,6 +10,7 @@ import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.model._
+import akka.http.scaladsl.model.headers.ByteRange
 import akka.http.scaladsl.unmarshalling.{Unmarshal, Unmarshaller}
 import akka.stream.Materializer
 import akka.stream.alpakka.s3.acl.CannedAcl
@@ -59,13 +60,18 @@ private[alpakka] final class S3Stream(credentials: AWSCredentials,
   val MinChunkSize = 5242880 //in bytes
   val signingKey = SigningKey(credentials, CredentialScope(LocalDate.now(), region, "s3"))
 
-  def download(s3Location: S3Location): Source[ByteString, NotUsed] = {
+  def download(s3Location: S3Location, range: Option[ByteRange] = None): Source[ByteString, NotUsed] = {
     import mat.executionContext
-    Source.fromFuture(request(s3Location).flatMap(entityForSuccess).map(_.dataBytes)).flatMapConcat(identity)
+    Source.fromFuture(request(s3Location, range).flatMap(entityForSuccess).map(_.dataBytes)).flatMapConcat(identity)
   }
 
-  def request(s3Location: S3Location): Future[HttpResponse] =
-    signAndGet(getDownloadRequest(s3Location, region))
+  def request(s3Location: S3Location, rangeOption: Option[ByteRange] = None): Future[HttpResponse] = {
+    val downloadRequest = getDownloadRequest(s3Location, region)
+    signAndGet(rangeOption match {
+      case Some(range) => downloadRequest.withHeaders(headers.Range(range))
+      case _ => downloadRequest
+    })
+  }
 
   /**
    * Uploads a stream of ByteStrings to a specified location as a multipart upload.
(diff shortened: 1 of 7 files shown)


s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -59,13 +60,16 @@ private[alpakka] final class S3Stream(credentials: AWSCredentials,
...
     import mat.executionContext
-    Source.fromFuture(request(s3Location).flatMap(entityForSuccess).map(_.dataBytes)).flatMapConcat(identity)
+    Source.fromFuture(request(s3Location, range).flatMap(entityForSuccess).map(_.dataBytes)).flatMapConcat(identity)
   }
 
-  def request(s3Location: S3Location): Future[HttpResponse] =
-    signAndGet(getDownloadRequest(s3Location, region))
+  def request(s3Location: S3Location, range: Option[ByteRange] = None): Future[HttpResponse] = {
+    val downloadRequest = getDownloadRequest(s3Location, region)
+    val rangeHeader = range.map(headers.Range(_))
```
svezfaz: pattern match might be a bit clearer here?

```
signAndGet (range match {
  case Some(r) => downloadRequest.withHeaders(Range(r))
  case _       => downloadRequest
})
```

s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3WireMockBase.scala
```diff
@@ -35,19 +36,35 @@ abstract class S3WireMockBase(_system: ActorSystem, _wireMockServer: WireMockSer
 
   val body = "<response>Some content</response>"
   val bucketKey = "testKey"
+  val bucketKeyForRanged = "testKeyForRanged"
```
svezfaz: can we reuse `val bucketKey = "testKey"` instead?

s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3SourceSpec.scala
```diff
@@ -17,14 +18,28 @@ class S3SourceSpec extends S3WireMockBase with S3ClientIntegrationSpec {
...
+  it should "download a range of file's bytes from S3 if bytes range given" in {
+
+    mockRangedDownload()
+
+    //#rangedDownload
+    val s3Source: Source[ByteString, NotUsed] =
+      s3Client.download(bucket, bucketKeyForRanged, ByteRange(bytesRangeStart, bytesRangeEnd))
+    //#rangedDownload
+
+    val result: Future[Array[Byte]] = s3Source.map(_.toArray).runWith(Sink.head)
```
svezfaz: minor: this could be consistent with the previous tests and deal with strings (doing `map(_.utf8string)`).
this will need
`val rangeOfBody = body.substring(bytesRangeStart, bytesRangeEnd + 1)`

lightbend-cla-validator:
Hi @thereisnospoon,

s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3WireMockBase.scala
thereisnospoon: taking into account that we use Wiremock and that `bucketKey` is used as a part of request URI, it would be hard to tweak Wiremock to return different mocked response for the same endpoint based just on presence of `Range` header

s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3SourceSpec.scala
thereisnospoon: Taking into account that `Range` header operates on bytes level, I believe, It's better not to couple to `String`s and concrete strings encodings (UTF-8 in this case, but it could be easily be another encoding).
In short, I think, the test should verify that given byte range of data is returned regardless of what this bytes represent.

thereisnospoon:
signed the CLA

s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3WireMockBase.scala
svezfaz: I see what you mean. Then could it be a good idea to reset the wiremock server mappings after each test instead? This way we'll avoid unwanted mappings to mess with future tests

thereisnospoon: That could work. Thanks! I will try

jrudolph:
Thanks a lot @thereisnospoon.

jrudolph:
Refs #264.

Issue #272, opened 2017-04-21
directory listing source


ennru:
In parallell to `Ftp.ls(...)` Alpakka should support local directory listing into streams.

juanjovazquez:
It might be considered to define a common API for both components FTP and File/Directory. I don't know if this has been discussed before but it could be worth it since the semantics in both cases is very similar.

PR #274, opened 2017-04-21
Directory sources

johanandren:
Refs  #272

Adds `Directory.ls` and `Directory.walk` to stream file paths from the filesystem.

### DIFF
--- a/file/src/main/java/akka/stream/alpakka/file/javadsl/Directory.java
+++ b/file/src/main/java/akka/stream/alpakka/file/javadsl/Directory.java
@@ -5,7 +5,13 @@ package akka.stream.alpakka.file.javadsl;
 
 import akka.NotUsed;
 import akka.stream.javadsl.Source;
+import akka.stream.javadsl.StreamConverters;
+import scala.None;
+import scala.None$;
+import scala.collection.immutable.Nil$;
 
+import java.nio.file.FileVisitOption;
+import java.nio.file.Files;
 import java.nio.file.Path;
 
 public final class Directory {
@@ -21,6 +27,15 @@ public final class Directory {
    * Recursively list files and directories in the given directory, depth first.
    */
   public static Source<Path, NotUsed> walk(Path directory) {
-    return akka.stream.alpakka.file.scaladsl.Directory.walk(directory).asJava();
+    return StreamConverters.fromJavaStream(() -> Files.walk(directory));
+  }
+
+  /**
+   * Recursively list files and directories in the given directory, depth first,
+   * with a maximum directory depth limit and a possibly set of options (See {@link java.nio.file.Files#walk} for
+   * details.
+   */
+  public static Source<Path, NotUsed> walk(Path directory, int maxDepth, FileVisitOption... options) {
+    return StreamConverters.fromJavaStream(() -> Files.walk(directory, maxDepth, options));
   }
 }
--- a/file/src/main/java/akka/stream/alpakka/file/javadsl/Directory.java
+++ b/file/src/main/java/akka/stream/alpakka/file/javadsl/Directory.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright (C) 2016-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.alpakka.file.javadsl;
+
+import akka.NotUsed;
+import akka.stream.javadsl.Source;
+
+import java.nio.file.Path;
+
+public final class Directory {
+
+  /**
+   * List all files in the given directory
+   */
+  public static Source<Path, NotUsed> ls(Path directory) {
+    return akka.stream.alpakka.file.scaladsl.Directory.ls(directory).asJava();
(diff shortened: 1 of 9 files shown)


file/src/main/java/akka/stream/alpakka/file/javadsl/Directory.java
```diff
@@ -0,0 +1,26 @@
...
+
+import akka.NotUsed;
+import akka.stream.javadsl.Source;
+
+import java.nio.file.Path;
+
+public final class Directory {
+
+  /**
+   * List all files in the given directory
```
ktoso: files and directories

file/src/main/scala/akka/stream/alpakka/file/scaladsl/Directory.scala
```diff
@@ -0,0 +1,36 @@
...
+  /**
+   * Recursively list files in the given directory and its subdirectories. Listing is done
+   * depth first.
+   */
+  def walk(directory: Path): Source[Path, NotUsed] = {
+    require(Files.isDirectory(directory), s"Path must be a directory, $directory isn't")
+    StreamConverters.fromJavaStream(() => Files.walk(directory))
+  }
+
+  private def allFilesFilter = new BiPredicate[Path, BasicFileAttributes] {
```
ktoso: this is not used

file/src/main/scala/akka/stream/alpakka/file/scaladsl/Directory.scala
```diff
@@ -0,0 +1,36 @@
...
+  def ls(directory: Path): Source[Path, NotUsed] = {
+    require(Files.isDirectory(directory), s"Path must be a directory, $directory isn't")
+    StreamConverters.fromJavaStream(() => Files.list(directory))
+  }
+
+  /**
+   * Recursively list files in the given directory and its subdirectories. Listing is done
+   * depth first.
+   */
+  def walk(directory: Path): Source[Path, NotUsed] = {
```
ktoso: I think it would be good to expose the overloads too: 
https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#walk-java.nio.file.Path-int-java.nio.file.FileVisitOption...-
the visit options and depth seem very useful when someone would be using this API

file/src/test/scala/akka/stream/alpakka/file/scaladsl/FileTailSourceSpec.scala
```diff
@@ -13,7 +13,7 @@ import akka.stream.{ActorMaterializer, Materializer}
 
 import scala.concurrent.duration._
 
-object FileTailSourceSpec extends {
```
ktoso: hah funny

file/src/main/scala/akka/stream/alpakka/file/scaladsl/Directory.scala
johanandren: I started out with that, but then thought that it is so simple to create it yourself, so maybe more specific needs can be met by just looking at the sources of this and copying that single line of code...

ktoso: I think with alpakka we're targeting an "works out of the box" and "I would not look at the sources" use cases/people, so I'd really add those parameters IMHO

johanandren: This was rather from the not covering every conceivable use case under the sun, just the "normal" ones point of view, especially the filter which could be achieved by composition (with a bit of alloc overhead, but hey this is disk io we are talking about here).

Could you come up with a common use case for depths other than 0, which is covered by ls and all the depth (`find ./`) which is covered by this?

ktoso: Depth is pretty important to protect from entering infinite loops with symlinks to an upper directory if I remember correctly how these APIs work?

johanandren: Alright, I'll add it. Skipping filter though, as that can be achieved with composition.

file/src/main/scala/akka/stream/alpakka/file/scaladsl/Directory.scala
```diff
@@ -23,14 +23,24 @@ object Directory {
...
    * Recursively list files in the given directory and its subdirectories. Listing is done
    * depth first.
+   *
+   * @param maxDepth If defined limits the depth of the directory structure to walk through
+   * @param fileVisitOptions See `java.nio.files.Files.walk()` for details
    */
-  def walk(directory: Path): Source[Path, NotUsed] = {
+  def walk(directory: Path,
+           maxDepth: Option[Int] = None,
+           fileVisitOptions: immutable.Seq[FileVisitOption] = Nil): Source[Path, NotUsed] = {
```
ktoso: great thanks :)

Issue #280, opened 2017-05-01
azure storage queue


akreuzer:
Hi, I can add a connector for the Azure storage queue.
The api would be similar to the existing connector for Amazon SQS.

johanandren:
Sounds good!

raboof:
Has been merged in 0.9

PR #287, opened 2017-05-06
Initial commit for azure-storage-queue connector

akreuzer:
This is the initial first attempt for the azure storage queue connector.

#280


azure-storage-queue/src/main/scala/akka/stream/alpakka/azure/storagequeue/AzureQueueSourceStage.scala
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    def receiveMessages(): Unit =
+      if (canReceiveNewMessages) {
+        implicit val executionContext = materializer.executionContext
+        currentRequests += 1
+
+        Future {
+          import scala.collection.JavaConverters._
+          val res = cloudQueue.retrieveMessages(settings.maxBatchSize, settings.waitTimeSeconds, null, null)
```
johanandren: As this stage doesn't actually run multiple of these requests concurrently (even if the counter hints that there is an idea to do that), and they seem to be blocking it would be better to not fork off futures and run the stage on a dispatcher suitable for blocking (you can use `ActorAttributes.dispatcher("akka.stream.default-blocking-io-dispatcher")` as attribute for this stage)

If the idea is indeed to change something else to run concurrent requests (might be problematic due to ordering of the elements perhaps?) then this future should instead be run on a blocking dispatcher.

azure-storage-queue/src/test/scala/akka/stream/alpakka/azure/storagequeue/scaladsl/AzureQueueSpec.scala
```diff
@@ -0,0 +1,179 @@
...
+import akka.stream._
+import akka.testkit._
+import akka.actor.ActorSystem
+import scala.util.Properties
+import akka.stream.alpakka.azure.storagequeue._
+import scala.collection.JavaConverters._
+
+
+// These tests are all live since the Azure Storage Emulator
+// does not run on Linux/Docker yet
```
johanandren: Does this mean the tests will not be able to run on travis or that they need an actual azure account for this to work?

azure-storage-queue/src/main/scala/akka/stream/alpakka/azure/storagequeue/FlowMapECStage.scala
```diff
@@ -0,0 +1,34 @@
...
+ */
+package akka.stream.alpakka.azure.storagequeue
+
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
+import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
+import scala.concurrent.{ExecutionContext, Future}
+
+/** This flow does just a map
+  * with the exception that is also hands the ExecutionContext
+  * of the materializer to the function f
```
akreuzer: Yes, one could replace this with `mapAsync`.
Basically, what one would have then is that `AzureQueueSink.fromFunction` would be implemented by something like `Flow.create().mapAsync(…)` and one would have to provide an `ExecutionContext` as an additional parameter to `AzureQueueSink.fromFunction`.
This would mean that all the `apply`-functions creating Sinks would also require an `ExecutionContext` since they are implemented using `AzureQueueSink.fromFunction`.

I tried to avoid that since I have not seen that in any other api. But if you think that it is better to just use `mapAsync`, I will change that.

azure-storage-queue/src/test/scala/akka/stream/alpakka/azure/storagequeue/scaladsl/AzureQueueSpec.scala
akreuzer: One does need an azure storage account. With this, it will run on travis.

azure-storage-queue/src/main/scala/akka/stream/alpakka/azure/storagequeue/AzureQueueSourceStage.scala
akreuzer: It can run multiple requests. Every time pull is called it will spawn of a new request (if there are not already `maxConcurrency` many or the buffer is full).
Since a request generates more than one output eventually the stage will run `maxConcurrency` many requests.
It is actually the same logic as in the SQS package.

Ordering is not an issue since Azure Stroge Queue itself does not guarantee it.

akreuzer:
Thank you for your feedback. 
I submitted a revision which fixes the points I have not commented on.

azure-storage-queue/src/main/scala/akka/stream/alpakka/azure/storagequeue/AzureQueueSourceStage.scala
johanandren: There is only ever one pull until a graph stage pushes, so that means that as this stage only fetches on pull it does not do more concurrent requests. To run concurrent requests you'd have to introduce a buffer and fetch batches from the queue, there's a few different ways to do this but you should make sure it really makes sense before introducing that complexity.

akreuzer: Maybe I miss a detail, but I think the process is like this:

(Assume for the sake of argument that the azure queue contains enough messages.)

- The Source is created.
- The Source is pulled from for the first time, i.e., `onPull`-handler is called.
- `onPull` calls `receiveMessage`, and this functions forks off a `Future` waiting for messages from Azure.
- If the future is successful it will receive `maxBatchSize` (default: 10) messages and schedules a call to `handleSuccess`.
- `handleSuccess` puts the 10 messages in `buffer`, pushes 1 message out, and forks off a new `Future` (through a call to `receiveMessage`).
  Now there are 9 messages in `buffer`.
- Another Message is being pulled, i.e., `onPull`-handler is called:
  Since `buffer` is not empty a message is pushed, ***and*** `receiveMessage` is called. 
This call forks off a second `Future`.
  Now there are 8 message in `buffer`.
- This continues until the buffer is empty.
  At this time 10 many futures are running.
  (Assuming that they do not have completed yet.)

johanandren: Ah, you are correct, not sure how I could miss out on that when I reviewed it originally, sorry about that!

azure-storage-queue/src/main/scala/akka/stream/alpakka/azure/storagequeue/scaladsl/AzureQueueSink.scala
```diff
@@ -0,0 +1,65 @@
...
+import akka.stream.scaladsl.{Flow, Keep, Sink}
+import akka.Done
+import scala.concurrent.{ExecutionContext, Future}
+
+object AzureQueueSink {
+
+  /**
+   * ScalaAPI: creates a [[akka.stream.scaladsl.Sink]] which queues message to an Azure Storage Queue.
+   */
+  def apply(cloudQueue: CloudQueue, maxInFlight: Int = 4): Sink[CloudQueueMessage, Future[Done]] =
```
johanandren: Magical number here, why 4 is a good default? Should it be a constant somewhere - (with an explanation attached)?

azure-storage-queue/src/main/scala/akka/stream/alpakka/azure/storagequeue/AzureQueueSinkFunctions.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import com.microsoft.azure.storage.queue.{CloudQueue, CloudQueueMessage}
+import scala.concurrent.{ExecutionContext, Future}
+import akka.Done
+
+sealed trait DeleteOrUpdateMessage
+case object Delete extends DeleteOrUpdateMessage
+case class UpdateVisibility(timeout: Int) extends DeleteOrUpdateMessage
+
+private[storagequeue] object AzureQueueSinkFunctions {
```
johanandren: All these operations are blocking on network and must not be run on the default dispatcher but on a dedicated dispatcher for blocking operations, just like the future inside the `AzureQueueSourceStage`. 

I think the easiest way would be to actually get this right would be to get rid of that `FlowMapECStage`, take an implicit `ActorSystem` to the individual methods, look up the `akka.stream.default-blocking-io-dispatcher` from it and pass that as EC to the `mapAsync`s.

Another option would be to rather wrap these in a custom minimal stage and configure that to run on the blocking dispatcher using attributes (see other existing connectors for samples of this)

azure-storage-queue/src/main/scala/akka/stream/alpakka/azure/storagequeue/AzureQueueSourceStage.scala
```diff
@@ -0,0 +1,89 @@
...
+    AzureQueueSourceSettings(waitTimeSeconds, maxBufferSize, maxBatchSize)
+}
+
+private[storagequeue] final class AzureQueueSourceStage(cloudQueue: CloudQueue, settings: AzureQueueSourceSettings)
+    extends GraphStage[SourceShape[CloudQueueMessage]] {
+  val out: Outlet[CloudQueueMessage] = Outlet("AzureCloudQueue.out")
+  override val shape: SourceShape[CloudQueueMessage] = SourceShape(out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+    val buffer = new Queue[CloudQueueMessage]
```
johanandren: private here as well

build.sbt
```diff
@@ -3,6 +3,7 @@ lazy val alpakka = project
   .enablePlugins(PublishUnidoc)
   .aggregate(amqp,
              awslambda,
+	           azureStorageQueue,
```
johanandren: something weird with the indentation here?

docs/src/main/paradox/azure-storage-queue.md
```diff
@@ -0,0 +1,39 @@
+# Azure Storage Queue Connector
+
+The Azure Storage Queue connector provides as Akka Stream Source and Sinks for Azure Storage Queue integration.
```
johanandren: Would be great with a one sentence what-is-Azure-Storage-Queue here in addition to the link where one can read more.

azure-storage-queue/src/main/scala/akka/stream/alpakka/azure/storagequeue/AzureQueueSourceStage.scala
```diff
@@ -0,0 +1,89 @@
...
+package akka.stream.alpakka.azure.storagequeue
+
+import com.microsoft.azure.storage.queue.{CloudQueue, CloudQueueMessage}
+import akka.stream.stage.{GraphStage, GraphStageLogic, OutHandler}
+import akka.stream.{Attributes, Outlet, SourceShape}
+import scala.collection.mutable.Queue
+import scala.concurrent.Future
+import scala.util.{Failure, Success}
+
+final case class AzureQueueSourceSettings(
```
akreuzer: Added a scaladoc comment.

azure-storage-queue/src/main/scala/akka/stream/alpakka/azure/storagequeue/AzureQueueSourceStage.scala
```diff
@@ -0,0 +1,89 @@
...
+  val default = AzureQueueSourceSettings(20, 100, 10)
+
+  /**
+   *  Java API
+   */
+  def create(waitTimeSeconds: Int, maxBufferSize: Int, maxBatchSize: Int): AzureQueueSourceSettings =
+    AzureQueueSourceSettings(waitTimeSeconds, maxBufferSize, maxBatchSize)
+}
+
+private[storagequeue] final class AzureQueueSourceStage(cloudQueue: CloudQueue, settings: AzureQueueSourceSettings)
```
akreuzer: I switched to a factory. (Actually it should be thread-safe but are not guaranteed for future releases.)

azure-storage-queue/src/main/scala/akka/stream/alpakka/azure/storagequeue/AzureQueueSourceStage.scala
akreuzer: Done

azure-storage-queue/src/main/scala/akka/stream/alpakka/azure/storagequeue/AzureQueueSourceStage.scala
```diff
@@ -0,0 +1,89 @@
...
+      }
+      receiveMessages
+    }
+
+    setHandler(out, new OutHandler {
+      override def onPull: Unit = {
+        if (!buffer.isEmpty) {
+          push(out, buffer.dequeue)
+        }
+        receiveMessages
```
akreuzer: Done

build.sbt
akreuzer: Done

docs/src/main/paradox/azure-storage-queue.md
akreuzer: I added another sentence.

akreuzer:
I made the following changes

- Made all requests to `CloudQueue` single-threaded
  - As you hinted this is much simpler. I though about it and I believe that this is sufficient.
  - Removed `FlowMapECStage`.
  - Removed all uses of Futures and Dispatchers in Stages.
- Added the option `retrieveRetryTimeout` to `AzureQueueSource`. This option allows to specify what happens when the queue runs empty.
- Using factories for `CloudQueue` now.

azure-storage-queue/src/main/scala/akka/stream/alpakka/azure/storagequeue/AzureQueueSourceStage.scala
```diff
@@ -4,86 +4,85 @@
...
-        currentRequests += 1
-
-        Future {
-          import scala.collection.JavaConverters._
-          val res = cloudQueue.retrieveMessages(settings.maxBatchSize, settings.waitTimeSeconds, null, null)
-          res.asScala.toList
-        } onComplete {
-          case Success(res) => successCallback.invoke(res)
-          case Failure(t) => failureCallback.invoke(t)
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new TimerGraphStageLogic(shape) {
```
johanandren: Alright, this simplified the stage quite a bit. Now it needs to run on a dispatcher for blocking stuff, you can look at https://github.com/akka/alpakka/blob/20ee323a22d5cf8e0ac53948d6b9f323a4b832ae/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpIOGraphStage.scala#L28 for an example how to do that through the initial attributes of the stage

azure-storage-queue/src/main/scala/akka/stream/alpakka/azure/storagequeue/javadsl/AzureQueueSink.scala
```diff
@@ -44,27 +40,21 @@ object AzureQueueWithTimeoutsSink {
...
    */
-  def create(cloudQueue: CloudQueue, maxInFlight: Int): Sink[MessageWithTimeouts, CompletionStage[Done]] =
+  def create(cloudQueue: Supplier[CloudQueue]): Sink[MessageWithTimeouts, CompletionStage[Done]] =
     AzureQueueSink.fromFunction(
-      (input: MessageWithTimeouts, ec: ExecutionContext) =>
-        AzureQueueSinkFunctions.addMessage(cloudQueue)(input.message, input.timeToLive, input.initialVisibility)(ec),
-      maxInFlight
+      { input: MessageWithTimeouts =>
+        AzureQueueSinkFunctions.addMessage(cloudQueue.get())(input.message, input.timeToLive, input.initialVisibility)
+      }
```
johanandren: `fromFunction` expects the function to be non-blocking, so use the blocking IO dispatcher through attributes here as well, something like this: `Flow.fromFunction(...).addAttributes(Attributes(DefaultAttributes.IODispatcher))`

azure-storage-queue/src/main/scala/akka/stream/alpakka/azure/storagequeue/scaladsl/AzureQueueSource.scala
```diff
@@ -14,8 +14,8 @@ object AzureQueueSource {
    * Scala API: creates a [[AzureQueueSource]] for a Azure CloudQueue.
    */
   def apply(
-      cloudQueue: CloudQueue,
+      cloudQueue: () => CloudQueue,
       settings: AzureQueueSourceSettings = AzureQueueSourceSettings.default
   ): Source[CloudQueueMessage, NotUsed] =
-    Source.fromGraph(new AzureQueueSourceStage(cloudQueue, settings))
+    Source.fromGraph(new AzureQueueSourceStage(cloudQueue(), settings))
```
johanandren: As I said before, the Stage can be created once but materialized any number of times concurrently. This means the factory must be passed to the stage, and called to create a queue client when the `GraphStageLogic` is created to be safe.

akreuzer:
I made the following changes in the revision.
- The `IODispatcher` attribute was added to the `Source` and the `Sink`s.
- The `CloudQueue` is now created inside of the `GraphStageLogic` in `AzureQueueSourceStage`.
- The call to `CloudQueue` inside the `Flow.map(...)` also uses the factory now, since the `Sink`s might also be reused.

azure-storage-queue/src/main/scala/akka/stream/alpakka/azure/storagequeue/AzureQueueSourceStage.scala
```diff
@@ -0,0 +1,95 @@
...
+final case class AzureQueueSourceSettings(
+    initialVisibilityTimeout: Int,
+    batchSize: Int,
+    retrieveRetryTimeout: Option[FiniteDuration]
+)
+
+object AzureQueueSourceSettings {
+  // default initialVisibilityTimeout (30) is from
+  // com.microsoft.azure.storage.queue.QueueConstants.DEFAULT_VISIBILITY_MESSAGE_TIMEOUT_IN_SECONDS
+  val default = AzureQueueSourceSettings(30, 10, None)
```
johanandren: Good comment, make it a Scaladoc instead! Also, in Akka constants in Scala use uppercase as first letter, so `Default`

azure-storage-queue/src/main/scala/akka/stream/alpakka/azure/storagequeue/javadsl/AzureQueueSink.scala
```diff
@@ -0,0 +1,77 @@
...
+
+object AzureQueueSink {
+
+  /**
+   * JavaAPI: creates a [[akka.stream.javadsl.Sink]] which queues message to an Azure Storage Queue.
+   */
+  def create(cloudQueue: Supplier[CloudQueue]): Sink[CloudQueueMessage, CompletionStage[Done]] =
+    fromFunction(AzureQueueSinkFunctions.addMessage(() => cloudQueue.get)(_))
+
+  private[javadsl] def fromFunction[T](f: T => Unit): Sink[T, CompletionStage[Done]] = {
```
johanandren: Please also add: 
```
/** 
 * Internal API 
 */ 
```
To the internal api things.

akreuzer:
Great! 
Thank you for the help Johan.

PR #288, opened 2017-05-08
sse: Upgrade to Akka SSE 3 and make test more robust

hseeberger:
This is an important fix for the SSE connector, because in Akka SSE 2 heartbeats, which are used by this connector as framing delimiter, aren't ignored during unmarshalling.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -2,9 +2,9 @@ import sbt._, Keys._
 
 object Dependencies {
 
-  val ScalaVersions = Seq("2.11.8", "2.12.1")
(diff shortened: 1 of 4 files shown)


hseeberger:
Thanks for merging. I'd like to release a bugfix release 0.8.1. Is there anything written how to do that?

jrudolph:
I think it's not much more difficult than adding the tag and travis will do its job.

However, we just made a release four days ago. Would be nice to get a few more changes in this week and release on Thursday or Friday, wdyt?

hseeberger:
Thursday/Friday sounds great! Thanks.

PR #289, opened 2017-05-08
Add Alpakka banner

2m:
that mentions a couple of frequently used sbt tasks.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -3,7 +3,7 @@ import sbt._, Keys._
 object Dependencies {
 
   val ScalaVersions = Seq("2.11.11", "2.12.2")
-  val AkkaVersion = "2.4.17"
+  val AkkaVersion = "2.4.18"
   val AkkaHttpVersion = "10.0.6"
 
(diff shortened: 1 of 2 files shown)


project/Dependencies.scala
```diff
@@ -3,7 +3,7 @@ import sbt._, Keys._
 object Dependencies {
 
   val ScalaVersions = Seq("2.11.11", "2.12.2")
-  val AkkaVersion = "2.4.17"
+  val AkkaVersion = "2.4.18"
```
hseeberger: Why not 2.5.1?

2m: We need to make sure that Alpakka works with 2.5.x and 2.4.x, therefore leaving the older version here. We could use 2.5.x here if we could run tests with jars from an older series.

johanandren: In fact we do not know they do work with 2.5.1 as there are a few cases we are stricter with in graph stages now.

2m:
Validation failure was: https://github.com/akka/alpakka/issues/290

PR #294, opened 2017-05-09
CSV: Fixes ignored second double quote

### LINKED ISSUE #293: CSV: Second double quote is ignored
The CSV parser in 0.8 ignores a second quote as in `"Venture ""Extended Edition"""` which is expected to be parsed to `Venture "Extended Edition"`.

ennru:
Fixes #293

### DIFF
--- a/csv/src/test/resources/correctness.csv
+++ b/csv/src/test/resources/correctness.csv
@@ -1,0 +1,7 @@
+Year,Make,Model,Description,Price
+1997,Ford,E350,"ac, abs, moon",3000.00
+1999,Chevy,"Venture ""Extended Edition""","",4900.00
+1996,Jeep,Grand Cherokee,"MUST SELL!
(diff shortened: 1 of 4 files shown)


csv/src/test/resources/correctness.csv
```diff
@@ -0,0 +1,7 @@
+Year,Make,Model,Description,Price
+1997,Ford,E350,"ac, abs, moon",3000.00
+1999,Chevy,"Venture ""Extended Edition""","",4900.00
+1996,Jeep,Grand Cherokee,"MUST SELL!
+air, moon roof, loaded",4799.00
+1999,Chevy,"Venture ""Extended Edition, Very Large""",,5000.00
+,,"Venture ""Extended Edition""","",4900.00
```
jrudolph: Didn't now about double quote escaping in CSV, but good to support it.

PR #296, opened 2017-05-10
(PENDING/DON'T MERGE) sse: Avoid unnecessary retries

hseeberger:


sse/src/main/scala/akka/stream/alpakka/sse/scaladsl/EventSource.scala
```diff
@@ -94,11 +94,9 @@ object EventSource {
...
         send(request).flatMap(Unmarshal(_).to[EventSource]).fallbackTo(noEvents)
       }
-      def recover(eventSource: EventSource) = eventSource.recoverWithRetries(1, { case _ => Source.empty })
-      def delimit(eventSource: EventSource) = eventSource.concat(singleDelimiter)
-      Flow[Option[String]]
-        .mapAsync(1)(getEventSource)
-        .flatMapConcat((recover _).andThen(delimit))
+      def recoverAndDelimit(eventSource: EventSource) =
+        eventSource.recoverWithRetries(0, { case _ => Source.empty }).concat(singleDelimiter)
+      Flow[Option[String]].mapAsync(1)(getEventSource).flatMapConcat(recoverAndDelimit)
```
johanandren: The old one recovered with empty, the new one recovers with empty, empty will never fail, not sure how the old and new would behave differently with regards to retries?

hseeberger: The old one: `recoverWithRetries(1, ...`
The new one: `recoverWithRetries(0`

Makes sense?

johanandren: Ah, my bad, didn't know that behaviour with retries `0`, which is somewhat useless, as it never retries but just fails the stage with the exception upstream failed with it makes more sense to just remove the recovery. `Source.failed(ex)` will give you the same behaviour as `Source.failed(ex).recoverWithRetries(0, { ... whatever ...})`.

hseeberger: I'm not sure I understand. Or maybe you don't ;-)

`recoverWithRetries` doesn't **fail** the stream, it does the opposite: it **recovers** from failure.

In my case retries don't make sense, hence 0 as argument.

johanandren: If you use `recoverWithRetries(0, {... whatever ...})` that will always fail the stream and never retry, because of the retry `0`. Try this one out for example:

```
Source.failed(new RuntimeException("bork")).recoverWithRetries(0, {
    case _: Throwable =>
      println("recovering")
      Source.empty
  }).runForeach(println)
    .onComplete(t => println("done: " + t))
```

You should see it always fails the stream with the `RuntimeException`, so that is the same as just not using `recoverWithRetries` at all, if you want it to recover with the given source the value of `attempts` must be `1` (or higher, that doesn't matter as `Source.empty` will never fail)

johanandren: (that the stage doesn't enforce attempts > 0 or see 0 as "try one time" was what surprised me)

hseeberger: Wow, thanks for bringing that up. Looking at the source code I'm pretty sure this is a bug in `RecoverWith`: I'll create a ticket and a PR.

So for the time being – until the above is fixed – keep this PR pending.

johanandren: To complete this PR you could just set `attempts` to whatever number > 0 and it will work, since `Source.empty` will never fail the number of retries is irrelevant as long as it will happen at least once.

hseeberger: OK, then I misunderstood how `recoverWithRetries` is working and this PR is obsolete. Thanks!

hseeberger:
As a value of `0` doesn't recover at all (see the docs), the old value `1` is perfect.

Issue #297, opened 2017-05-10
S3: Remove duplicated region argument in client methods


zaharidichev:
Currently we have: 

```scala
final class S3Settings(val bufferType: BufferType,
                       val diskBufferPath: String,
                       val proxy: Option[Proxy],
                       val awsCredentials: AWSCredentials,
                       val s3Region: String,
                       val pathStyleAccess: Boolean)
```

But then we have methods such as: 

```scala
  private[this] def s3Request(s3Location: S3Location,
                              region: String,
                              method: HttpMethod = HttpMethods.GET,
                              uriFn: (Uri => Uri) = identity)(implicit conf: S3Settings): HttpRequest
```

Whats the point of passing a region arg if we already have it configured into the implicit configuration. This is confusing. IMO, we need to omit the region argument and keep the one in S3Settings and use it.

zaharidichev:
As a matter of fact I think it will be better to proceed with this before addressing #271

johanandren:
Would be good to have one of the previous s3 committers comment on this as it will break the code of pretty much everyone who used the current s3 client API. 

Ping @thereisnospoon, @kobmic, @Scalahansolo, @svezfaz, @2beaucoup

svezfaz:
I think cleaning up the API now might make sense. It seems like there is more than one wrong thing with it:

```
object S3Stream {

  def apply(credentials: AWSCredentials, region: String)(implicit system: ActorSystem, mat: Materializer): S3Stream =
    new S3Stream(credentials, region, S3Settings(system))
}

private[alpakka] final class S3Stream(credentials: AWSCredentials,
                                      region: String,
                                      val settings: S3Settings)(implicit system: ActorSystem, mat: Materializer) {
...
}
```

1. neither system nor materializer are actually needed to create a `S3Stream`. Only thing needed is an executionContext, and that could as well be passed to every method separately, instead of passing it at creation
2. as per @zaharidichev comment, we only really need the `S3Settings` object to create an `S3Stream`, having this hybrid approach makes for a confusing API
3. `S3Settings.apply` taking an `ActorSystem` does not seem good practice to me. If we want a default way to create `S3Settings` we could just have an empty `apply()` that performs `ConfigFactory.load()` for us

Moreover, issues 1 and 2 gets propagated to `S3Client` as well.

+1 to sort this up now :)

zaharidichev:
[...]

PR #298, opened 2017-05-11
Remove duplicated region argument in client methods

zaharidichev:
In my opinion an S3 client should be configured for a region via the config file.

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/S3Settings.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/S3Settings.scala
@@ -5,16 +5,16 @@ package akka.stream.alpakka.s3
 
 import akka.actor.ActorSystem
 import akka.stream.alpakka.s3.auth.AWSCredentials
-import com.typesafe.config.Config
+import com.typesafe.config.{Config, ConfigFactory}
 
 final case class Proxy(host: String, port: Int, scheme: String)
 
-final class S3Settings(val bufferType: BufferType,
-                       val diskBufferPath: String,
-                       val proxy: Option[Proxy],
-                       val awsCredentials: AWSCredentials,
-                       val s3Region: String,
-                       val pathStyleAccess: Boolean) {
+case class S3Settings(bufferType: BufferType,
+                      diskBufferPath: String,
+                      proxy: Option[Proxy],
+                      awsCredentials: AWSCredentials,
+                      s3Region: String,
+                      pathStyleAccess: Boolean) {
 
   override def toString: String =
     s"S3Settings($bufferType,$diskBufferPath,$proxy,$awsCredentials,$s3Region,$pathStyleAccess)"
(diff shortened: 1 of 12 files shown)


johanandren:
Refs #297

s3/src/main/scala/akka/stream/alpakka/s3/S3Settings.scala
```diff
@@ -5,16 +5,16 @@ package akka.stream.alpakka.s3
...
 
 final case class Proxy(host: String, port: Int, scheme: String)
 
-final class S3Settings(val bufferType: BufferType,
-                       val diskBufferPath: String,
-                       val proxy: Option[Proxy],
-                       val awsCredentials: AWSCredentials,
-                       val s3Region: String,
-                       val pathStyleAccess: Boolean) {
+case class S3Settings(bufferType: BufferType,
```
johanandren: Should still be final

s3/src/main/scala/akka/stream/alpakka/s3/S3Settings.scala
```diff
@@ -31,27 +31,33 @@ case object DiskBufferType extends BufferType {
 }
 
 object S3Settings {
-  def apply(system: ActorSystem): S3Settings =
-    apply(system.settings.config.getConfig("akka.stream.alpakka.s3"))
+  //def apply(system: ActorSystem): S3Settings = apply(system.settings.config.getConfig("akka.stream.alpakka.s3"))
```
johanandren: I think this signature should be kept, and be available as `apply()(implicit system: ActorSystem)` it's nice to be able to use the same config as your actor system rather than loading a separate config every time you create settings. And then I think the one below should be removed. If someone wants that it can be achieved with the third signature - `S3Settings(ConfigFactory.load())`, but I don't think that is common.

s3/src/main/scala/akka/stream/alpakka/s3/S3Settings.scala
```diff
@@ -31,27 +31,33 @@ case object DiskBufferType extends BufferType {
...
+        )
       } else None
     },
-    awsCredentials = AWSCredentials(config.getString("aws.access-key-id"), config.getString("aws.secret-access-key")),
-    s3Region = config.getString("aws.default-region"),
-    pathStyleAccess = config.getBoolean("path-style-access")
+    awsCredentials = AWSCredentials(config.getString("akka.stream.alpakka.s3.aws.access-key-id"),
+                                    config.getString("akka.stream.alpakka.s3.aws.secret-access-key")),
+    s3Region = config.getString("akka.stream.alpakka.s3.aws.default-region"),
+    pathStyleAccess = config.getBoolean("akka.stream.alpakka.s3.path-style-access")
```
johanandren: There's no Java API ones here, in addition to the previous comment in this object, would be great if you could add `.create(ActorSystem)` and `create(Config)` and mark those with `Java API`, thanks!

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -45,20 +45,18 @@ final case class ListBucketResult(isTruncated: Boolean, continuationToken: Optio
...
-  def apply(credentials: AWSCredentials, region: String)(implicit system: ActorSystem, mat: Materializer): S3Stream =
-    new S3Stream(credentials, region, S3Settings(system))
+  def apply(settings: S3Settings)(implicit system: ActorSystem, mat: Materializer): S3Stream =
+    new S3Stream(settings)
 }
 
-private[alpakka] final class S3Stream(credentials: AWSCredentials,
-                                      region: String,
-                                      val settings: S3Settings)(implicit system: ActorSystem, mat: Materializer) {
+private[alpakka] final class S3Stream(val settings: S3Settings)(implicit system: ActorSystem, mat: Materializer) {
```
svezfaz: can `val settings: S3Settings` just be `settings: S3Settings` here?

zaharidichev:
Currently travelling, will take a look tomorrow morning and address comments

PR #304, opened 2017-05-16
Add attribute parameters to sqs source settings

### LINKED ISSUE #302: SQS: Allow requesting attributes for messages
I've been using alpakka's sqs connector (great work btw!), and need to be able to request message attributes for the messages that come in.

Would y'all accept a PR with this functionality?

lnr0626:
This fixes #302


sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -19,20 +19,69 @@ object SqsSourceSettings {
...
+    messageAttributeNames: Seq[String] = Seq()
 ) {
   require(maxBatchSize <= maxBufferSize, "maxBatchSize must be lower or equal than maxBufferSize")
   // SQS requirements
   require(0 <= waitTimeSeconds && waitTimeSeconds <= 20,
           s"Invalid value ($waitTimeSeconds) for waitTimeSeconds. Requirement: 0 <= waitTimeSeconds <= 20 ")
   require(1 <= maxBatchSize && maxBatchSize <= 10,
           s"Invalid value ($maxBatchSize) for maxBatchSize. Requirement: 1 <= maxBatchSize <= 10 ")
+
+  private val validNames = Seq(
```
johanandren: How about enumerating them as an ADT instead of stringly typed: 
```
sealed abstract class AttributeName(val name: String)
case object All extends AttributeName("All")
```

and then `attributeNames: Set[AttributeName]` etc. ?

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
```diff
@@ -19,20 +19,69 @@ object SqsSourceSettings {
...
+  )
+
+  require(attributeNames.forall(name => validNames.contains(name)), s"AttributeName must be one of $validNames")
+
+  require(
+    messageAttributeNames.forall(name => name.matches("[0-9a-zA-Z_\\-.]+")),
+    "MessageAttributeNames may only contain alphanumeric characters and the underscore (_), hyphen (-), and period (.)"
+  )
+
+  require(
```
johanandren: This could also be done with a type instead to fail "faster", do the require directly in the constructor and know that it is safe here. Not sure if it is the other names plus custom names? Could either be a sealed trait which `AttributeName` extends plus an extra `CustomAttributeName` subtype or just single type which always takes that string.

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceStage.scala
lnr0626: that's better, making the change now.

out of curiosity (don't have a ton of scala experience), what's the benefit of sealed class hierarchy vs an enumeration?

lnr0626:
I made the changes and pulled SqsSourceSettings into it's own file with the supporting classes - this is similar to the other Settings classes. Happy to revert pulling that into it's own file if you want to keep it the way it was.

Issue #290, opened 2017-05-08
Failed: MqttSourceSpec should support will message


2m:
While validating https://github.com/akka/alpakka/pull/289

https://travis-ci.org/akka/alpakka/jobs/229954352

```
[info] - should support will message *** FAILED *** (251 milliseconds)
[info]   The future returned an exception of type: org.eclipse.paho.client.mqttv3.MqttException, with message: Unable to connect to server. (MqttSourceSpec.scala:198)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.concurrent.Futures$FutureConcept$class.tryTryAgain$1(Futures.scala:523)
[info]   at org.scalatest.concurrent.Futures$FutureConcept$class.futureValueImpl(Futures.scala:550)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:275)
[info]   at org.scalatest.concurrent.Futures$class.whenReady(Futures.scala:676)
[info]   at akka.stream.alpakka.mqtt.scaladsl.MqttSourceSpec.whenReady(MqttSourceSpec.scala:21)
[info]   at akka.stream.alpakka.mqtt.scaladsl.MqttSourceSpec$$anonfun$1$$anonfun$apply$mcV$sp$10$$anonfun$apply$11.apply(MqttSourceSpec.scala:198)
[info]   at akka.stream.alpakka.mqtt.scaladsl.MqttSourceSpec$$anonfun$1$$anonfun$apply$mcV$sp$10$$anonfun$apply$11.apply(MqttSourceSpec.scala:187)
[info]   at org.scalatest.concurrent.Futures$class.whenReady(Futures.scala:677)
[info]   at akka.stream.alpakka.mqtt.scaladsl.MqttSourceSpec.whenReady(MqttSourceSpec.scala:21)
[info]   at akka.stream.alpakka.mqtt.scaladsl.MqttSourceSpec$$anonfun$1$$anonfun$apply$mcV$sp$10.apply(MqttSourceSpec.scala:187)
[info]   at akka.stream.alpakka.mqtt.scaladsl.MqttSourceSpec$$anonfun$1$$anonfun$apply$mcV$sp$10.apply(MqttSourceSpec.scala:178)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
...
```

2m:
This one started to fail almost constantly now:
[...]

PR #308, opened 2017-05-18
Fix releasing

jrudolph:


2m:
Validation failure was https://github.com/akka/alpakka/issues/290 and #310

PR #311, opened 2017-05-19
#310 exclude newer Akka dependency

### LINKED ISSUE #310: NoClassDefFoundError: akka/stream/impl/StreamLayout$Module
While validating https://github.com/akka/alpakka/pull/308

https://travis-ci.org/akka/alpakka/jobs/233906736
https://travis-ci.org/akka/alpakka/jobs/233906734

2m:
So older version is not evicted.

Fixes #310

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -115,7 +115,12 @@ object Dependencies {
   val Sqs = Seq(
     libraryDependencies ++= Seq(
       "com.amazonaws"     %  "aws-java-sdk-sqs"    % "1.11.109",             // ApacheV2
-      "org.elasticmq"     %% "elasticmq-rest-sqs"  % "0.13.4"        % Test, // ApacheV2
+      "org.elasticmq"     %% "elasticmq-rest-sqs"  % "0.13.4"        % Test excludeAll(
+        // elasticmq-rest-sqs depends on Akka 2.5, exclude it, so we can choose Akka version
+        ExclusionRule(organization = "com.typesafe.akka") //
+      ), // ApacheV2
+      // elasticmq-rest-sqs depends on akka-slf4j which was excluded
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -115,7 +115,12 @@ object Dependencies {
   val Sqs = Seq(
     libraryDependencies ++= Seq(
       "com.amazonaws"     %  "aws-java-sdk-sqs"    % "1.11.109",             // ApacheV2
-      "org.elasticmq"     %% "elasticmq-rest-sqs"  % "0.13.4"        % Test, // ApacheV2
+      "org.elasticmq"     %% "elasticmq-rest-sqs"  % "0.13.4"        % Test excludeAll(
+        // elasticmq-rest-sqs depends on Akka 2.5, exclude it, so we can choose Akka version
+        ExclusionRule(organization = "com.typesafe.akka") //
```
patriknw: That's an interesting one. strictly speaking I think that it might work when running with 2.4, because the library is compiled with an later akka version and we are only binary **backwards** compatible.

2m: The problem was mixing Akka 2.4 and 2.5 versions. Without this exclusion rule and when alpakka is compiled with 2.4, `elasticmq-rest-sqs` would evict akka-stream and akka-actor to 2.5.1 version, but not akka-stream-testkit. Therefore we would end up with newer akka-actor and akka-stream and older akka-stream-testkit which would cause `NoClassDefFoundError` mentioned in the original ticket.

By adding this ExclusingRule, we deny any control over Akka version from this library and then we test with Akka 2.4 and Akka 2.5.

patriknw: I understand. Yes, that's the best we can do, but I still think it's not completely safe to run that with 2.4.

Issue #170, opened 2017-01-27
Add Amazon Kinesis support


alexmnyc:

sarveshseri:
Is there anyone working on it ?

If not then I would like to take this up.

areyouspiffy:
I pretty much finished it [here](https://github.com/areyouspiffy/reactive-aws/tree/master/kinesis/src/main/scala/reactive/kinesis). You're welcome to take whatever you like from it. It uses the same AwsClient used by the existing DynamoDB connector.

areyouspiffy:
@sarveshseri Are you working on this? If not, I will submit a PR for this in the next couple days.

sarveshseri:
@areyouspiffy I had the notion that you were already working on it so I did not pick it up.

You should go ahead with your PR.

easel:
This looks awesome! Any thoughts of doing a Sink as well?

areyouspiffy:
@easel I don't personally has a use for a sink but I'm more than happy to add that if you think find it useful 👍

easel:
@areyouspiffy I'd definitely use it and be willing to collaborate on it and test it -- I'm going to have to write one otherwise =p

aserrallerios:
We've implemented one Flow that publishes messages to a Kinesis stream and we're currently testing it on production. If you're patient I can port it to alpakka.

2m:
I will close this as initial Kinesis support (Source) has been merged. Please open tickets for sink and flow discussions correspondingly.

PR #318, opened 2017-06-05
Add AWS Kinesis Support

areyouspiffy:
Adds an Amazon Kinesis source per #170 that consumes a single shard, with support for consuming multiple shards via merging.

### DIFF
--- a/docs/src/main/paradox/kinesis.md
+++ b/docs/src/main/paradox/kinesis.md
@@ -1,0 +1,93 @@
+# AWS Kinesis Connector
+
+The AWS Kinesis connector provides an Akka Stream Source for consuming Kinesis Stream records.
+
+For more information about Kinesis please visit the [official documentation](https://aws.amazon.com/documentation/kinesis/).
+
+## Artifacts
+
+sbt
+:   @@@vars
+    ```scala
+    libraryDependencies += "com.lightbend.akka" %% "akka-stream-alpakka-kinesis" % "$version$"
+    ```
+    @@@
+
+Maven
+:   @@@vars
+    ```xml
+    <dependency>
+      <groupId>com.lightbend.akka</groupId>
+      <artifactId>akka-stream-alpakka-kinesis$scalaBinaryVersion$</artifactId>
+      <version>$version$</version>
+    </dependency>
+    ```
+    @@@
+
+Gradle
+:   @@@vars
+    ```gradle
+    dependencies {
+      compile group: "com.lightbend.akka", name: "akka-stream-alpakka-kinesis$scalaBinaryVersion$", version: "$version$"
+    }
+    ```
+    @@@
+
+## Usage
+
+Sources provided by this connector need a `AmazonKinesisAsync` instance to consume messages from a shard.
+
+@@@ note
+The `AmazonKinesisAsync` instance you supply is thread-safe and can be shared amongst multiple `GraphStages`. As a result, individual `GraphStages` will not automatically shutdown the supplied client when they complete.
+@@@
+
+Scala
+: @@snip (../../../../kinesis/src/test/scala/akka/stream/alpakka/kinesis/scaladsl/Examples.scala) { #init-client }
+
+Java
+: @@snip (../../../../kinesis/src/test/java/akka/stream/alpakka/kinesis/javadsl/Examples.java) { #init-client }
+
+We will also need an @scaladoc[ActorSystem](akka.actor.ActorSystem) and an @scaladoc[ActorMaterializer](akka.stream.ActorMaterializer).
+
+Scala
+: @@snip (../../../../kinesis/src/test/scala/akka/stream/alpakka/kinesis/scaladsl/Examples.scala) { #init-system }
+
+Java
+: @@snip (../../../../kinesis/src/test/java/akka/stream/alpakka/kinesis/javadsl/Examples.java) { #init-system }
+
+### Using the Source
+
+The `KinesisSource` creates one `GraphStage` per shard. Reading from a shard requires an instance of `ShardSettings`. 
+
+Scala
+: @@snip (../../../../kinesis/src/test/scala/akka/stream/alpakka/kinesis/scaladsl/Examples.scala) { #settings }
+
+Java
+: @@snip (../../../../kinesis/src/test/java/akka/stream/alpakka/kinesis/javadsl/Examples.java) { #settings }
+
+You have the choice of reading from a single shard, or reading from multiple shards. In the case of multiple shards the results of running a separate `GraphStage` for each shard will be merged together. 
+
+@@@ warning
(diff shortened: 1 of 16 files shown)


aserrallerios:
I'm no expert in Kinesis but I think that it's possible to save different sequence numbers for different shards, so each shard can be consumed independently.

The current implementation of `basicMerge`, if used in combination with `startingSequenceNumber` settings, can be missleading or dangerous.

A possible solution would be to use a Map of shardId -> settings instead of a single instance of settings for the `basicMerge` methods.

Otherwise, It'll force users to do something like that:

```scala
    getShardsByStreamName(kinesisSettings.streamName, amazonKinesisAsync)
      .map(_.map(_.getShardId))
      .map(shardId => basicMerge(shardId, kinesisSettings.copy(startingSequenceNumber = getSequenceNumber(shardId)), amazonKinesisAsync)) 
```

This is not so bad, but I think this use case will be so common that it'd be nice to be support it out-of-the-box in alpakka.

WDYT?

areyouspiffy:
@aserrallerios Hey, thanks for your feedback! And nice catch. I agree that's it's misleading and I like your suggestion. I'll update the PR accordingly.

areyouspiffy:
@aserrallerios Updated. Now it takes an instance of settings, including shard id, for each shard.

aserrallerios:
Thank you! looks very good, nice addition to the lib!

docs/src/main/paradox/kinesis.md
```diff
@@ -0,0 +1,93 @@
...
+    }
+    ```
+    @@@
+
+## Usage
+
+Sources provided by this connector need a `AmazonKinesisAsync` instance to consume messages from a shard.
+
+@@@ note
+The `AmazonKinesisAsync` instance you supply is thread-safe and can be shared amongst multiple `GraphStages`. As a result, individual `GraphStages` will not be automatically shutdown the supplied client when they complete.
```
2m: .. will not ~be~ automatically shutdown ...

kinesis/src/test/java/akka/stream/alpakka/kinesis/javadsl/Examples.java
```diff
@@ -0,0 +1,43 @@
...
+import com.amazonaws.services.kinesis.model.ShardIteratorType;
+import scala.concurrent.duration.FiniteDuration;
+
+import java.util.Arrays;
+import java.util.concurrent.TimeUnit;
+
+public class Examples {
+
+    //#init-client
+    AmazonKinesisAsync amazonKinesisAsync = AmazonKinesisAsyncClientBuilder.defaultClient();
```
2m: Add `final` to all of the variables that are immutable.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisSourceStage.scala
```diff
@@ -0,0 +1,164 @@
...
+      private[this] def getShardIterator(): Future[GetShardIteratorResult] = {
+        val request = Function.chain(
+          Seq(
+            (r: GetShardIteratorRequest) => r.withStreamName(streamName),
+            (r: GetShardIteratorRequest) => r.withShardId(shardId),
+            (r: GetShardIteratorRequest) => r.withShardIteratorType(shardIteratorType),
+            (r: GetShardIteratorRequest) => startingSequenceNumber.fold(r)(r.withStartingSequenceNumber),
+            (r: GetShardIteratorRequest) => atTimestamp.fold(r)(r.withTimestamp)
+          )
+        )(new GetShardIteratorRequest())
```
2m: Nice usage of `Function.chain`. You can reduce verbosity by giving a type parameter to `chain` like so:

```scala
val request = Function.chain[GetShardIteratorRequest](
  Seq(
    _.withStreamName(streamName),
    ...
    r => atTimestamp.fold(r)(r.withTimestamp)
  )
)(new GetShardIteratorRequest())
```

kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisSourceStage.scala
```diff
@@ -0,0 +1,164 @@
...
+        (result: GetRecordsResult) => self.ref ! GetRecordsSuccess(result)
+      )
+
+      private[this] val getRecordsAsyncHandler = getAmazonKinesisHandler[GetRecordsRequest, GetRecordsResult](
+        getRecordsSuccessCallback,
+        getRecordsFailureCallback
+      )
+
+      private[this] def requestRecords(): Unit =
+        amazonKinesisAsync.getRecordsAsync(new GetRecordsRequest().withShardIterator(iterator), getRecordsAsyncHandler)
```
2m: You can simplify and remove the usage of async callbacks above by calling `getRecordsAsyncHandler` with `self.ref` and then sending messages to that ActorRef straight from the handler.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisSourceStage.scala
```diff
@@ -0,0 +1,164 @@
...
+
+      private def awaitingRecords(in: (ActorRef, Any)): Unit = in match {
+        case ((_, GetRecordsSuccess(result))) => {
+          val records = result.getRecords.asScala
+          Option(result.getNextShardIterator).fold {
+            log.info("Shard {} returned a null iterator and will now complete.", shardId)
+            completeStage()
+          } { iterator = _ }
+          if (records.size > 0) {
+            records.foreach(buffer.enqueue(_))
```
2m: I am a bit wary about this unbounded buffer. I see that `GetRecordsRequest` supports specifying a limit of how many records to get. With that we could batch up the results. One complication with this approach is that a sequence shard iterator type then needs to be used.

Do you think batching results could be doable?

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSource.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  def getShardsByStreamName(streamName: String, amazonKinesisAsync: AmazonKinesisAsync): Try[List[Shard]] = {
+    @tailrec
+    def getShards(streamDescription: StreamDescription, shards: java.util.ArrayList[Shard]): List[Shard] =
+      if (streamDescription.getHasMoreShards) {
+        shards.addAll(streamDescription.getShards)
+        getShards(streamDescription, shards)
+      } else shards.asScala.toList
+
+    Try(amazonKinesisAsync.describeStream(streamName).getStreamDescription)
```
2m: We should call the async variant of describeStream here and return a scala future. Also this method is missing in the javadsl.KinesisSource object. Therefore it makes sense to move the implementation that returns java future to the javadsl package and change the implementation here to call to javadsl and then convert to scala future.

areyouspiffy:
Thanks for the feedback @2m! I'll get to work on making these changes.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisSourceStage.scala
areyouspiffy: Good idea. I've added this to `ShardSettings` and the associated `GetRecords` request in the stage.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisSourceStage.scala
areyouspiffy: Handy!

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSource.scala
areyouspiffy: I just removed this because it seemed a bit out of place. Maybe if anybody comes up with a helpful usage pattern that they like it can be added in another PR.

2m:
Validation failure was https://github.com/akka/alpakka/issues/290

Otherwise this can go in. Good job @areyouspiffy!

PR #320, opened 2017-06-07
WIP Reactive marshalling

jypma:
Moved over from [ts-reaktive](https://github.com/Tradeshift/ts-reaktive/tree/master/ts-reaktive-marshal) as discussed in akka-http call https://github.com/akka/akka-meta/issues/43 .

Features:
- Reactive parsers and generators, converting `Source[ByteString]` into `Source[*Event]` and vice-versa, with `*` being JSON, XML and CSV
- DSL for reactive (un)marshalling of such events, converting `Source[*Event]` into `Source[T]` and vice-versa (without the need to do any framing)

This will enable the following example use cases:
- Handling streams of nested JSON / XML / CSV objects that can't be easily framed, e.g. `{"items": {"a": 1, "b": 2, ...etc...}}`
- Binding a single data class to both JSON and XML, with "nice" formats for both, without using reflection
- The `alpakka-s3` module can use this to implement listing a bucket without waiting for a 300KB http response to be fully parsed

**TODO**
- [ ] Merge `AaltoReader` with the existing aalto wrapper already in alpakka, use the existing `XMLEvent` case classes, but add location information to them
- [ ] Compare the CSV parser in this PR to the one already in alpakka, and select which one to keep
- [ ] Decide on packaging / module structure. Should `alpakka-csv` and `alpakka-xml` both have the xml- and csv- specific marshalling classes, and then make those modules depend on a generic `alpakka-marshal`? Which would mean introducing `alpakka-json` as well.
- [ ] Decide whether to add a [Scala DSL](https://github.com/Tradeshift/ts-reaktive/pull/56) in this PR, or later
- [ ] Move over the unit tests
- [ ] Use sbt-boilerplate to generate the N-ary combinators for JSON `object` and XML `tag`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -126,6 +126,14 @@ lazy val hbase = project
     fork in Test := true
   )
 
+lazy val marshal = project
+  .enablePlugins(AutomateHeaderPlugin)
+  .settings(
+    name := "akka-stream-alpakka-marshal",
+    Dependencies.Marshal,
+    fork in Test := true
+  )
+
 lazy val ironmq = project
   .enablePlugins(AutomateHeaderPlugin)
   .settings(
--- a/marshal/src/main/java/akka/stream/alpakka/marshal/Protocol.java
+++ b/marshal/src/main/java/akka/stream/alpakka/marshal/Protocol.java
@@ -1,0 +1,88 @@
+package akka.stream.alpakka.marshal;
+
+import javaslang.Function1;
+import javaslang.control.Try;
+
+public interface Protocol<E,T> extends ReadProtocol<E,T>, WriteProtocol<E,T> {
+    /**
+     * Returns a protocol that considers all events emitted results, and vice-versa.
+     */
+    public static <T> Protocol<T,T> identity(Class<T> type) {
+        return new Protocol<T,T>() {
+            @Override
+            public Reader<T, T> reader() {
+                return Reader.identity();
+            }
+
+            @Override
+            public Class<? extends T> getEventType() {
+                return type;
+            }
+
+            @Override
+            public Writer<T, T> writer() {
+                return Writer.identity();
+            }
+        };
+    }
+    
+    /**
+     * Combines a read-only and a write-only protocol into a read/write protocol
+     */
+    public static <E,T> Protocol<E,T> of(ReadProtocol<E,T> read, WriteProtocol<E,T> write) {
+        return new Protocol<E,T>() {
+            @Override
+            public Writer<E,T> writer() {
+                return write.writer();
+            }
+
+            @Override
+            public Class<? extends E> getEventType() {
+                return write.getEventType();
+            }
+
+            @Override
+            public Reader<E,T> reader() {
+                return read.reader();
+            }
+            
+            @Override
+            public Try<T> empty() {
+                return read.empty();
+            }
+            
+            @Override
+            public String toString() {
+                return read.toString();
+            }
+        };
+    }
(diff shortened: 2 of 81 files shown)


2m:
Sorry for taking so long to get to this. I have played with ts-reaktive protocols before (https://github.com/2m/ts-reaktive/tree/wip-parser) for parsing massive xml files, and I have to say it was a fun experience!

Regarding packaging, I think having `alpakka-{csv, json, xml}` is great. When adding corresponding marshalling classes to these projects, would that mean that these modules will have to depend on akka-http-core?

Merging with csv parser might get a bit more tricky as the current parser does not parse CSV to events but to lines and then to a map of key (header name) value pairs. Will have to discuss with @ennru for the best way forward for csv.

Regarding this PR, I think it should be about adding alpakka-json which would contain a parser and marshalling classes and alpakka-marshal which would have generic marshal classes. Then we would have a streaming json marshaller that other Alpakka modules could use and a clear path forward for the already existing csv and xml parsers. WDYT?

ennru:
Interesting with an event-based approach. For CSV it might not be as useful as for Json and XML as it doesn't contain nested structures.

The current CSV parser works line-based - what I reckon people would expect. If there's a good way to hide the event-based parser underneath a line-based experience it might be a good thing. I don't think just having the event-based CSV parser gives a good out-of-the-box experience for Alkpakka.

The current CSV parser does depend on Akka libraries only. Dependening on Jackson and Javaslang/Vavr can easily lead to conflicts for users.

2m:
Taking a look at this again. It would be good to split this into a couple of subprojects before merging it. What do you think about the following:
* alpakka-marshal - would contain contents of the following packages:
  * akka.stream.alpakka.marshal._
  * akka.stream.alpakka.marshal.generic._
* alpakka-marshal-csv:
  * akka.stream.alpakka.marshal.csv._
  * akka.stream.alpakka.marshal.csv.javadsl._
* alpakka-marshal-json:
  * akka.stream.alpakka.marshal.json._
  * akka.stream.alpakka.marshal.json.javadsl._
* alpakka-marshal-xml:
  * akka.stream.alpakka.marshal.xml._
  * akka.stream.alpakka.marshal.xml.javadsl._

This would leave the current alpakka's xml and csv connectors untouched for future consideration. Also this layout would leave a spot for scaladsl.

I am not sure how we should deal with Http classes. `HttpFlow` and `HttpStreamingMarshallers` could go to the `alpakka-marshal` sub-project. The `JSONMarshallers` and `XMLMarshallers` could go to the `json` and `xml` sub-projects. This would mean that to all of the projects akka-http dependency should be added. I think it is fine, considering that the alternative would be to either add three more sub-projects or a sub-project for akka-http that depends on all marshal sub-projects.

Let me know what you think.

jypma:
Thanks a lot for your suggestions @2m , I think I'm mostly onboard 👍 .
As for the akka-http dependency, I think I'd prefer that to be separate. Http is only one of the kinds of I/O where streaming marshalling makes sense; reading large files, or any of the other alpakka streaming connectors, are just as valid. 

Would having a separate `alpakka-marshal-http` be that bad? It'd transiently pull in `alpakka-marshall`, so in you app's build you'd just say `http` and `xml` either way.

jypma:
Hmm disregard my previous comment. Indeed the marshalling part would explode the dependencies.
Let's go with @2m 's suggestion of just the 4 new sub-projects.

ennru:

Issue #309, opened 2017-05-19
Clean up build


jrudolph:
* sort projects alphabetically in build.sbt and Dependencies.scala
 * factor out dependencies to object

PR #323, opened 2017-06-07
Alphabetically order dependencies

Corbel:
https://github.com/akka/alpakka/issues/309

Looks like the projects in ´build.sbt´ are already ordered, and the dependencies are obviously in an ´Object´, so all that's left of this issue is to order the projects ´Dependencies.scala´


2m:
Validation failure was https://github.com/akka/alpakka/issues/290

Issue #326, opened 2017-06-08
Support multiple Puts in the HBase Connector


yiftachdy:
Add support for batching inside the HBase Connector, meaning if one element from the stream is translated into multiple puts (for example - when Grouping in the stream, you will receive a collection of elements).

2m:
Fixed in #1050 which was released in `v0.20`.

PR #328, opened 2017-06-08
Add support for multi put method for HBase connector

yiftizur:
I want to add support for batching inside the HBase Connector, meaning if one element from the stream is translated into multiple puts (for example - when Grouping in the stream, you will receive a collection of elements).


johanandren:
Hey @yiftizur please do not open new pull requests for the same thing over and over again. If you do updates, just push new commits to the branch you PR:ed from and they will be included in this PR.

johanandren:
Refs #326

yiftachdy:
Hi @johanandren Sorry for the mess, tried to follow the rules and created some pull requests first from the wrong user, then from the wrong branch....

2m:
Sorry it took us so long to come to this. If I understand it correctly multi put is for sending multiple records to HBase at a time. Do you think it makes sense to have different factories with and without `multi` prefix? 

AFAICS the only difference is in the converter. Therefore we could use the same source of flow factories but changing the converter itself to return either one or more put statements. I think this would make the API more general and reduce the overall API surface. WDYT?

yiftizur:
Hi @2m ,
I've added the 'multi' API because i saw (and im no Scala expert) that due to type erasure, 2 methods with different Function signatures, end up as a duplicate method definition:
`def create[T](conf: Configuration,
                tableName: TableName,
                columnFamilies: java.util.List[String],
                converter: java.util.function.Function[T, Put]): HTableSettings[T] = {
    HTableSettings(conf, tableName, immutable.Seq(columnFamilies.asScala: _*), (t: T) => List(converter.apply(t)))
  }`
And
`def create[T](conf: Configuration,
                     tableName: TableName,
                     columnFamilies: java.util.List[String],
                     converter: java.util.function.Function[T, java.util.List[Put]]): HTableSettings[T] = {
    HTableSettings(conf,
                   tableName,
                   immutable.Seq(columnFamilies.asScala: _*),
                   (t: T) => converter.apply(t).asScala.toList)
  }`

end up as the same method signature - which does not compile....

2m:
You are correct. My proposal is to only have the second method. And if a user wants to have only one Put, then the converter function could return `Collections.singletonList(put)`

cheleb:
hi @yiftachdy,
  Sorry for this late review.
  IMHO it would be simpler to provide:

```
sealed trait HbaseCommand[T]

case class SimpleCommand[T](put: Put) extends HbaseCommand[T]

case class MultiCommand[T](puts: java.util.List[Put]) extends HbaseCommand[T]

final case class HTableSettings[T](conf: Configuration,
                                   tableName: TableName,
                                   columnFamilies: immutable.Seq[String],
                                   converter: T => HbaseCommand[T])
```


And in the stage just match the convertion result ?
```
settings.converter(msg) match {
            case SimpleCommand(put) =>
              table.put(put)
            case MultiCommand(puts) =>
              table.put(puts)
          }
```

WDYT ?

yiftizur:
@cheleb Sounds great to me, i just did not want to break existing API...

cheleb:
There is a typo in the master branch at 
docs/src/main/paradox/hbase.md

raboof:
@yiftizur thanks for your work on this feature! It seems this PR has developed some conflicts with master though.. would you be willing to resolve those?

2m:
This has been superseded by #328 which implemented multi put support. However this PR also contains changes to run the HBase tests in a docker container. I'll cherry-pick that commit to a separate PR.

PR #332, opened 2017-06-11
Make amqp recover from broker errors and set as default

JonathanKnight:
The stream fails on an amqp broker failure. Amqp is now able to recover from broker failures, so the stream should keep running while the broker recovers and continue processing when amqp comes back.

### DIFF
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
@@ -65,6 +65,7 @@ private[amqp] trait AmqpConnectorLogic { this: GraphStageLogic =>
 
   private var connection: Connection = _
   protected var channel: Channel = _
+  protected var autoRecoverEnabled: Boolean = _
 
   def settings: AmqpConnectorSettings
   def connectionFactoryFrom(settings: AmqpConnectionSettings): ConnectionFactory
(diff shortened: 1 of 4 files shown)


lightbend-cla-validator:
Hi @JonathanKnight,

2m:
Interesting. I did not know about `addRecoveryListener` before.

How could we test this? Currently tests are running against a RabbitMQ instance that is started by docker. I am not sure if It would be possible control that instance from the test code. Do you know if that could be done (via some control channel)?

Another approach would be to add an in memory amqp broker. We could use it only for the recovery test therefore it should not interfere or require modifications to other tests.

JonathanKnight:
There is a post [here](https://tamasgyorfi.net/2016/04/21/writing-integration-tests-for-rabbitmq-based-components/) which describes doing exactly that. Unfortunately it isn't easy. I don't know if I will have time in the next little while to do this.

2m:
Another resource worth checking out is [Management Plugin](https://www.rabbitmq.com/management.html) for RabbitMQ which has HTTP API and supports forcibly closing connections.

Could be possible to add that to the RabbitMQ instance we run in docker container and then forcibly closing a connection during the recovery test.

ennru:
@JonathanKnight Quite a lot has happened in the AMQP connecotr since you prepared this PR. Do you have any plans do revive it?

ennru:
I just filed #1034 to see if someone else wants to take this further.
Closing this now.

PR #333, opened 2017-06-12
Add kairosdb Sink

psycho-ir:
KairosDB Sink is here 🍨 
Will do the documentation after getting feedback.


2m:
This looks good! Thanks for bringing one more connector to the Alpakka family. :) Currently I do not have time to review this in more detail. Ping me if no one from the community has not reviewed this by the next week.

kairosdb/src/main/scala/akka/stream/alpakka/kairosdb/KairosDBSinkStage.scala
```diff
@@ -0,0 +1,118 @@
...
+private[kairosdb] class KairosSinkStageLogic(
+    in: Inlet[MetricBuilder],
+    shape: SinkShape[MetricBuilder],
+    kairosClient: Client,
+    promise: Promise[Done],
+    settings: KairosSinkSettings
+)(executionContext: ExecutionContext)
+    extends GraphStageLogic(shape)
+    with StageLogging {
+
```
francisdb: Would it not be better to force an explicit EC as this uses blocking code?

kairosdb/src/main/scala/akka/stream/alpakka/kairosdb/javadsl/KairoSink.scala
```diff
@@ -0,0 +1,44 @@
...
+import akka.stream.alpakka.kairosdb.{KairosDBSinkStage, KairosSinkSettings, NullExecutionContext}
+import akka.stream.javadsl.Sink
+import org.kairosdb.client.Client
+import org.kairosdb.client.builder.MetricBuilder
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/**
+ * Created by SOROOSH on 3/22/17.
+ */
```
francisdb: This doc should probably be removed / replaced with something more useful?

francisdb:
Might take some time later to test this on our staging kairos

kairosdb/src/main/scala/akka/stream/alpakka/kairosdb/KairosDBSinkStage.scala
psycho-ir: Actually in the first implementation I did it in this way. 
But later on decided to keep it similar to what already is there in alpakka. e.g. IranMQ implementation
https://github.com/akka/alpakka/blob/master/ironmq/src/main/scala/akka/stream/alpakka/ironmq/IronMqPullStage.scala#L48
But still is still more flexible as client can pass different EC as well.

kairosdb/src/main/scala/akka/stream/alpakka/kairosdb/javadsl/KairoSink.scala
psycho-ir: Yes of course :) 
I will document the API after we finalize the document.

psycho-ir:
@francisdb No I haven't tried this version yet on the production. but in staging works well.

kairosdb/src/main/scala/akka/stream/alpakka/kairosdb/KairosDBSinkStage.scala
francisdb: _Let me first fix my wording. I mean removing the **default** value for the implicit ec but keeping it implicit._

The IronMQ client is async using callbacks, there is no blocking{...} in there so this should not be an issue. I wonder if the internal ec from akka streams is set up to grow when there are blocking calls (like the global one).
Anyway, when implicit without default the user can still decide to provide mat.executionContext.

```scala
class KairosDBSinkStage(settings: KairosSinkSettings, kairosClient: Client)(implicit executionContext: ExecutionContext)
```

francisdb:
Could even be better to just write a client using akka http instead of using the blocking java client. Copying some stuff from https://github.com/waylayio/kairosdb-scala which would avoid the play dependencies.

psycho-ir:
@francisdb 
executionContext has no default valye anymore.

2m:
I have taken a closer took and now wonder if this could be implemented without a graph stage. Since you are wrapping calls to the kairos db in a future, then it could be implemented as

```scala
Source[MetricBuilder](...).mapAsync(parallelism, Future { kairosClient.pushMetrics(builder) })
```

Since the client is blocking this would still require a more careful management of execution contexts.

I also liked @francisdb suggestion to reimplement kairosdb client using akka-http. For starters it could be only the `pushMetrics` operation and that one could also be implemented with `Http.singleRequest` (which takes a request and returns a future of response) and mapAsync.

Let me know what you thing.

yaroot:
Hi, though I don't use kairosdb anymore but I'm use akka-stream for a lot of similar stuff. It could be implemented by just using flow.

```
    def submitDatapoints(...) = // call kairosdb http endpoint, just need a json library and use akka-http here

    case class Datapoint(name: String, datapoints: Seq[Seq[String]], tags: Seq[String], ttl: Long)

    Flow[Datapoint]
      .batch(100, Vector.newBuilder[Datapoint].+=)(_ += _) // batch http calls for efficiency
      .mapAsync(5) { datapoints =>
        submitDatapoints(datapoints.result())
      }
      .to(Sink.ignore)
```

And if you have to block you could use the built-in blocking pool via `Flow.addAttributes(Attributes(IODispatcher))`

just my 2 cents

ennru:

PR #344, opened 2017-06-15
Fix #343 by handling preStart failures explicitly

### LINKED ISSUE #343: AMQP: Sink promise never completes if connection fails
If opening a connection to rabbit during stage preStart fails, the AmpqSinkStage's promise will never be completed.

Falmarri:
If an error happened in preStart, the future wouldn't be completed and anyone mapping on the result would just hang on that future.

### DIFF
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala
@@ -91,6 +91,15 @@ final class AmqpSinkStage(settings: AmqpSinkSettings)
           }
         }
       )
+
+      override def postStop(): Unit = {
+        promise.tryFailure(new RuntimeException("stage stopped unexpectedly"))
+        super.postStop()
+      }
+
+      override def onFailure(ex: Throwable): Unit =
+        promise.tryFailure(ex)
+
     }, promise.future)
   }
 
@@ -148,6 +157,9 @@ final class AmqpReplyToSinkStage(settings: AmqpReplyToSinkSettings)
         super.postStop()
       }
 
+      override def onFailure(ex: Throwable): Unit =
+        promise.tryFailure(ex)
+
       setHandler(
         in,
         new InHandler {
@@ -186,6 +198,7 @@ final class AmqpReplyToSinkStage(settings: AmqpReplyToSinkSettings)
           }
         }
       )
+
     }, promise.future)
   }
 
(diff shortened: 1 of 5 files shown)


amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala
```diff
@@ -91,6 +91,15 @@ final class AmqpSinkStage(settings: AmqpSinkSettings)
...
         }
       )
+
+      override def postStop(): Unit = {
+        promise.tryFailure(new RuntimeException("stage stopped unexpectedly"))
+        super.postStop()
+      }
+
+      override def onFailure(ex: Throwable): Unit =
+        promise.tryFailure(ex)
```
ktoso: `super.onFailure` call too?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala
```diff
@@ -148,6 +157,9 @@ final class AmqpReplyToSinkStage(settings: AmqpReplyToSinkSettings)
         super.postStop()
       }
 
+      override def onFailure(ex: Throwable): Unit =
+        promise.tryFailure(ex)
```
ktoso: super call?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSourceStage.scala
```diff
@@ -129,7 +129,7 @@ final class AmqpSourceStage(settings: AmqpSourceSettings, bufferSize: Int)
           false // just this single message
         )
       }
-
+      override def onFailure(ex: Throwable): Unit = {}
```
ktoso: are you sure?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala
ktoso: Meh, nevermind I see that's your custom method and not the GraphStageLogic one

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala
ktoso: Meh, nevermind I see that's your custom method and not the GraphStageLogic one

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSourceStage.scala
ktoso: Meh, nevermind I see that's your custom method and not the GraphStageLogic one

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala
Falmarri: Should i rename the method since it apparently is somewhat confusing?

Falmarri:
@ktoso ping =]

PR #346, opened 2017-06-18
WIP - Add AWS SES Connector

naferx:


awsses/src/main/scala/akka/stream/alpakka/awsses/AwsSesFlowStage.scala
```diff
@@ -0,0 +1,71 @@
...
+              failureHandler.invoke(ex)
+          }
+
+        override def onSuccess(request: SendEmailRequest, result: SendEmailResult): Unit =
+          successHandler.invoke(result)
+      }
+
+      override def onPush(): Unit = {
+        inFlight += 1
+        awsSesClient.sendEmailAsync(grab(in), asyncHandler)
```
2m: `sendEmailAsync` returns a Java Future. Therefore a simpler implementation of this stage could be written using only stream combinators as:

    Flow[SendEmailRequest].mapAsync(parallelism)(awsSesClient.sendEmailAsync)

and using https://github.com/scala/scala-java8-compat to convert from Java Future to a Scala one.

naferx: No worries. I also have been busy these days. 
I will give it a try to this approach

Thanks Martynas!

ennru:

PR #354, opened 2017-06-21
Upgrade to Akka HTTP 10.0.8 and remove dep on akka-sse

hseeberger:


hseeberger:
Don't think so, we are in the 0.x era.

PR #358, opened 2017-06-22
fix #347 return more info when listing bucket

### LINKED ISSUE #347: AWS S3 Listing bucket should return more info than only the key
The official client returns this:

```java
    /** The name of the bucket in which this object is stored */
    protected String bucketName;

    /** The key under which this object is stored */
    protected String key;

    /** Hex encoded MD5 hash of this object's contents, as computed by Amazon S3 */
    protected String eTag;

    /** The size of this object, in bytes */
    protected long size;

    /** The date, according to Amazon S3, when this object was last modified */
    protected Date lastModified;

    /** The class of storage used by Amazon S3 to store this object */
    protected String storageClass;
```

more info here:
http://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html

francisdb:
Is it ok to use `java.time.Instant` (JDK 8)?
This is a breaking change as the return type for listing changes, is that ok? If not what is the proposed solution?
Can the java and scala api's share a case class? Currently I just copied the case class and added a map to convert

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/Marshalling.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/Marshalling.scala
@@ -3,9 +3,12 @@
  */
 package akka.stream.alpakka.s3.impl
 
+import java.time.Instant
+
 import akka.http.scaladsl.marshallers.xml.ScalaXmlSupport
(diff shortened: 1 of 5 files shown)


francisdb:
Any feedback on the questions in this PR would be welcome

francisdb:
build fails unrelated: see #363

francisdb:
ref #347

PR #366, opened 2017-06-27
Ignore sftp test

### LINKED ISSUE #365: Secure FTP connection intermittently fails
We use version 0.1.54 of jsch, which [appears to contain a bug](https://sourceforge.net/p/jsch/bugs/111/) that is triggered [since jdk 1.8.0_121](https://bugs.openjdk.java.net/browse/JDK-8174719).

We're impacted by this as shown by the test failures tracked in #361.

It looks like we'll either have to move away from jsch, wait for the bug to be fixed there, or help them fix it.

raboof:
To reduce test flakiness until we fix #365


raboof:
Likely we also need to ignore the other case in that test, but let's see evidence of that happening first.

francisdb:
👍🏻

PR #376, opened 2017-07-03
fix #374 correctly encodes non-ascii characters for signing s3 requests

### LINKED ISSUE #374: S3 signature doesn't match expected when path contains special chars
Hi guys.

Currently I'm facing an issue with alpakka s3 client 0.10 when accessing a file like `files/überall.csv`
```
The request signature we calculated does not match the signature you provided. Check your key and signing method.
```
When I pass the key with encoded special chars, e.g. `files/%C3%BCberall.csv` I get another exception: `The specified key does not exist.`

Supposedly the issue is related to https://github.com/akka/alpakka/issues/357 and https://github.com/akka/alpakka/issues/316 because in version 0.9 I had exactly the error
`akka.http.scaladsl.model.IllegalUriException: Illegal URI reference: Invalid input 'ü', expected '/', 'EOI', '#', '?' or pchar` for exactly same configuration.

francisdb:
fix #374 correctly encodes non-ascii characters for signing s3 requests

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
@@ -8,8 +8,6 @@ import java.net.URLEncoder
 import akka.http.scaladsl.model.Uri.{Path, Query}
 import akka.http.scaladsl.model.{HttpHeader, HttpRequest}
 
-import scala.annotation.tailrec
-
 // Documentation: http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html
 private[alpakka] case class CanonicalRequest(method: String,
                                              uri: String,
--- a/s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
@@ -59,7 +59,13 @@ private[alpakka] object CanonicalRequest {
       pathEncodeRec(builder ++= uriEncodePath(head), tail)
   }
 
-  private def toHexUtf8(ch: Char): String = "%" + Integer.toHexString(ch.toInt)
+  private def toHexUtf8(ch: Char): String = ch match {
+    case asciiChar if ch >= 0 && ch <= 127 =>
+      "%" + Integer.toHexString(asciiChar.toInt).toUpperCase
+    case otherChar =>
+      // I guess this adds quite some overhead but the detailed api is not public
+      Path(otherChar.toString).toString()
+  }
 
   // translated from java example at http://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html
   private def uriEncodePath(input: String, encodeSlash: Boolean = true): String =
@@ -73,4 +79,5 @@ private[alpakka] object CanonicalRequest {
       case ch =>
         toHexUtf8(ch)
     }
+
 }
--- a/s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
@@ -47,37 +47,12 @@ private[alpakka] object CanonicalRequest {
   def signedHeadersString(headers: Seq[HttpHeader]): String =
     headers.map(_.lowercaseName()).distinct.sorted.mkString(";")
 
-  private def pathEncode(path: Path): String = path match {
-    case p if p.isEmpty => "/"
-    case nonEmptyPath => pathEncodeRec(new StringBuilder, nonEmptyPath).toString()
-  }
-
-  @tailrec private def pathEncodeRec(builder: StringBuilder, path: Path): StringBuilder = path match {
-    case Path.Empty ⇒ builder
-    case Path.Slash(tail) ⇒ pathEncodeRec(builder += '/', tail)
-    case Path.Segment(head, tail) ⇒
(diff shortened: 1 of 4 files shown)


s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
```diff
@@ -59,7 +59,13 @@ private[alpakka] object CanonicalRequest {
       pathEncodeRec(builder ++= uriEncodePath(head), tail)
   }
 
-  private def toHexUtf8(ch: Char): String = "%" + Integer.toHexString(ch.toInt)
+  private def toHexUtf8(ch: Char): String = ch match {
+    case asciiChar if ch >= 0 && ch <= 127 =>
+      "%" + Integer.toHexString(asciiChar.toInt).toUpperCase
+    case otherChar =>
+      // I guess this adds quite some overhead but the detailed api is not public
+      Path(otherChar.toString).toString()
```
francisdb: Personally not too fond of this part but I want to avoid re-inventing the wheel on uri encoding

francisdb:
🙄 No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
francisdb: might take the time to use create a modified version of the akka-http uri encoder that users paraboiled

s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
```diff
@@ -59,7 +59,13 @@ private[alpakka] object CanonicalRequest {
       pathEncodeRec(builder ++= uriEncodePath(head), tail)
   }
 
-  private def toHexUtf8(ch: Char): String = "%" + Integer.toHexString(ch.toInt)
+  private def toHexUtf8(ch: Char): String = ch match {
```
gipeshka: maybe it makes sense to put this directly into list of cases at line 80

francisdb: well wanted to keep the code close to the aws example

s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
raboof: I suppose it would be neat to expose a `UriRendering.renderPath` that accepts a `keep: CharPredicate` instead of a `encodeFirstSegmentColons: Boolean`, that would make this a lot easier right?

I guess another approach might be to reuse the Path rendering and treat the `sub-delims` characters specially, like this:
```
private def pathEncode(path: Path): String =
    if (path.isEmpty) "/"
    else
      path.toString().flatMap {
        case ch if "!$&'()*+,;=".contains(ch) => "%" + Integer.toHexString(ch.toInt).toUpperCase
        case other => other.toString
      }
```

francisdb: let me do some tests but that indeed seems like a simpler solution 👍

francisdb: Feel free to submit your own pull request if you mind me copy pasting the above code :-)

francisdb:
Should I smash or do you guys handle that on merge?

s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
raboof: Haha no go ahead :)

raboof:
We'll squash & merge

francisdb:
@raboof are we still waiting for something?

s3/src/main/scala/akka/stream/alpakka/s3/auth/CanonicalRequest.scala
```diff
@@ -47,30 +47,12 @@ private[alpakka] object CanonicalRequest {
   def signedHeadersString(headers: Seq[HttpHeader]): String =
     headers.map(_.lowercaseName()).distinct.sorted.mkString(";")
 
-  private def pathEncode(path: Path): String = path match {
-    case p if p.isEmpty => "/"
-    case nonEmptyPath => pathEncodeRec(new StringBuilder, nonEmptyPath).toString()
-  }
-
-  @tailrec private def pathEncodeRec(builder: StringBuilder, path: Path): StringBuilder = path match {
```
raboof: `scala.annotation.tailrec` is now an unused import above

s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3NoMock.scala
```diff
@@ -11,17 +11,22 @@ import akka.stream.scaladsl.{Sink, Source}
...
 import com.typesafe.config.ConfigFactory
 import org.scalatest.concurrent.ScalaFutures
-import org.scalatest.{BeforeAndAfterAll, FlatSpecLike, Matchers}
+import org.scalatest.time.{Millis, Seconds, Span}
+import org.scalatest.{BeforeAndAfterAll, FlatSpecLike, Ignore, Matchers}
 
 import scala.concurrent.Await
 import scala.concurrent.duration._
 
+@Ignore
```
raboof: :+1: perhaps add a comment on how to use/enable/run these tests?

PR #385, opened 2017-07-06
S3 - Fix signature bug when the 'Range' header is used (fixes #384)

### LINKED ISSUE #384: S3 - an invalid signature is generated when the Range header is specified
When the range (https://github.com/akka/alpakka/pull/265) is being added to the download request, it replaces the "Host" header; it causes the generated signature to fail validation on S3.

In S3Stream.scala@101, the original headers should be concatenated
```
      case Some(range) => downloadRequest.withHeaders(downloadRequest.headers :+ headers.Range(range))
```

nberthet:
When a ByteRange is provided for a download request, the "Range" header was replacing the "Host" instead of being concatenated, this caused the request signer to generate an invalid signature.

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
@@ -98,7 +98,7 @@ private[alpakka] final class S3Stream(settings: S3Settings)(implicit system: Act
   def request(s3Location: S3Location, rangeOption: Option[ByteRange] = None): Future[HttpResponse] = {
     val downloadRequest = getDownloadRequest(s3Location)
     signAndGet(rangeOption match {
-      case Some(range) => downloadRequest.withHeaders(headers.Range(range))
+      case Some(range) => downloadRequest.addHeader(headers.Range(range))
       case _ => downloadRequest
     })
   }


raboof:
Refs #384

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -98,7 +98,7 @@ private[alpakka] final class S3Stream(settings: S3Settings)(implicit system: Act
   def request(s3Location: S3Location, rangeOption: Option[ByteRange] = None): Future[HttpResponse] = {
     val downloadRequest = getDownloadRequest(s3Location)
     signAndGet(rangeOption match {
-      case Some(range) => downloadRequest.withHeaders(headers.Range(range))
+      case Some(range) => downloadRequest.withHeaders(downloadRequest.headers :+ headers.Range(range))
```
raboof: Might be a bit neater to use [addHeader](http://doc.akka.io/api/akka-http/current/akka/http/javadsl/model/HttpRequest.html#addHeader(header:akka.http.javadsl.model.HttpHeader):Self) instead.

nberthet: Agreed, just pushed the update

nberthet:
@raboof I'm not very comfortable with scala, but I'll be happy to give it a shot, however, I'll be away for a over week, so I'm not sure I'll be able to contribute before that.
I guess to make this testable it will require to push the range parameter down to `getDownloadRequest` (ie method signature change with optional parameter), because once wrapped with in the Future, I don't see how the headers would be accessible

johanandren:
This same fix was done in #461 which I merged yesterday, sorry, missed out that this one existed.

PR #387, opened 2017-07-07
Sqs send message in batch mode

mehmetgunturkun:
As far as I can see, there was no support for sending messages to SQS in batch mode so I wanted to implement new stage for it.

I added one GraphStage, `SqsBatchFlowStage` and two different ways to use it such as:

- `SqsFlow.batch(queueUrl)`, which returns Flow[Seq[String], Seq[Result]] gets the sequence of messages, send as a batch and return list of Results.
- `SqsFlow.grouped(queueUrl, batchSize)`, which returns Flow[String, Result] as the usual SqsFlow but groups messages according to the batch size, sends the batch to SQS and flattens the list of Result.


sqs/src/main/scala/akka/stream/alpakka/sqs/scaladsl/SqsFlow.scala
```diff
@@ -18,6 +21,34 @@ object SqsFlow {
...
+        builder.add(Flow[Future[List[Result]]].mapAsync(settings.maxInFlight)(identity))
+
+      val flattenResults: FlowShape[List[Result], Result] = builder.add(Flow[List[Result]].mapConcat(identity))
+      groupingStage ~> sqsStage ~> flattenFutures ~> flattenResults
+
+      FlowShape(groupingStage.in, flattenResults.out)
+    }
+
+    Flow.fromGraph(graph)
+  }
```
aserrallerios: I fail to see why this his not just

```scala
Flow[String]
.grouped(batchSize)
.via(new SqsBatchFlowStage(queueUrl, sqsClient))
.mapAsync(settings.maxInFlight)(identity))
```

mehmetgunturkun: That works too. Honestly, I did it just because I like to see these steps explicitly.

aserrallerios:
I know that the documented max batch size of the SQS api is 10. I've never tested it, but probably using a batch greater than 10 will make the request fail. Should we validate the `batchSize` parameter?

IMHO alpakka really shines when the developer can save time integrating with a third party, so parameters like the max-batch-size should be transparent to them if it's only a implementation detail of the AWS api. Why should we force every alpakka developer to know that the max-batch-size is 10 if we can spare them from it?

Other than that, I'd remove the `SqsFlow.batch(queueUrl)` Flow. What are the odds that you're already working with `Seq`s that you want to publish to SQS and they have size <= 10?

mehmetgunturkun:
For the batch size,
I thought to put max batch size check in `grouped` function. However, I checked AWS SDK and realized that there is no constant value for that limit and also during running a couple of examples with ElasticMQ, I found out that this check is executed in server-side. Then, I thought that value might change in the future and setting a hardcoded value into code might be a bad idea. However, if you think otherwise, I can add that logic too.

For `SqsFlow.batch` function I thought users might want to group messages with different logic or somehow their function results sequence of messages. I did not want them to flatten their sequences before sending just for the sake of the api. Also, in that case flattening will be for nothing since we will group them again.

aserrallerios:
Yeah, is hardcoded on the server side, so it must be also hardcoded on the client side, either on alpakka or the alpakka user's code.

My point is that my usage of the lib in it's current state would be:

```scala
Flow[String]
.groupedWithin(10, 0.5 second)
.via(SqsFlow.batch("my-queue"))
```

And it could be:

```scala
Flow[String]
.via(SqsFlow("my-queue"))
```

Making life easier to alpakka users. Also note that your `grouped` method will wait until a batch is completed, forcing the users to make a decision whether they want to wait or not:

```scala
slowFlow.via(SqsFlow.grouped("slow-queue", 1))
fastFlow.via(SqsFlow.grouped("fast-queue", 10))
```

Because there's virtually no reason to use baches < 10 or > 10 whenever you can, and there's no reason to not taking advantage of the natural backpressure of the stream to make dynamic batches, I'd merge your methods into one, and maybe let the user configure it, but providing sane default settings:

```
def batch(queueName: String, settings: SqsFlowSettings = SqsFlowSettings.default): Flow[String, Result]
```

Default settings:
* maxBatchSize = 10
* maxBatchWait = 500 millis
* concurrentRequests = 1

Note that `maxBatchWait = 0` can still use batching, if the SQS server backpressures.

Also note that `concurrentRequests` can be really needed, if the queue's throughput is high (using a `Source` with concurrent requests).

mehmetgunturkun:
Yeah, grouping logic should be definitely with `groupWithin` instead. And also, putting these parameters in settings and providing sane default makes perfect sense.

mehmetgunturkun:
@s12v @kciesielski It would be awesome if you can also review this PR.

aserrallerios:
Looking good. Documentation should be updated, specially warning users about `batch` method.

mehmetgunturkun:
Awesome, I will update documentation.

BTW if you have other suggestions for names of the functions instead of `grouped` and `batch`, I am happy to change. I am not really good at naming functions :)

sqs/src/main/scala/akka/stream/alpakka/sqs/scaladsl/SqsFlow.scala
calvinlfer: How about using groupedWithin instead since this will just hold the messages until the batchSize is reached

mehmetgunturkun:
@s12v, @kciesielski have you had a chance to look at this?

Also if any other users come to your mind for review, you can tag them, I asked to you because I guess you're the top contributors of Sqs module.

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsBatchFlowStage.scala
```diff
@@ -0,0 +1,157 @@
...
+              new util.ArrayList[SendMessageBatchRequestEntry]()
+            var id = 0
+            messages.foreach { message =>
+              messageRequestEntries.add(new SendMessageBatchRequestEntry(id.toString, message))
+              id += 1
+            }
+
+            val batchRequest = new SendMessageBatchRequest(queueUrl, messageRequestEntries)
+            batchRequest
+          }
```
SergeiShabanau: You can write this function without mutating local vars. Something like this:
```scala

  private def createMessageBatch(messages: Array[String]): SendMessageBatchRequest =
    new SendMessageBatchRequest(
      queueUrl,
      messages
        .toList
        .zipWithIndex
        .map { case (message, id) => new SendMessageBatchRequestEntry(id.toString, message) }
        .asJava
    )
```

mehmetgunturkun: @SergeiShabanau, the reason why I do not use `zipWithIndex` with `map` is to avoid unnecessary memory allocations and also please correct me if I am wrong but basically this code is iterating three times, but with one local variable I guess we can do that with one iteration. Even though according to Asymptotic Notation they are both O(n), I thought fewer iterations might be better. However, if it confuses the logic here, I can change that part.

calvinlfer:
Sorry to de-rail this change but isn't it possible to achieve automatic batching behaviour through Amazon's Buffered SQS Client?

```scala
val sqsClient: AmazonSQSAsync = {
  val client = AmazonSQSAsyncClientBuilder.standard()
    .withCredentials(new DefaultAWSCredentialsProviderChain())
    .build()
  new AmazonSQSBufferedAsyncClient(client)
}

val normalSource: Source[Message, NotUsed] = SqsSource(settings.sqs.queueUrl)(sqsClient)
val normalAckSink: Sink[MessageActionPair, Future[Done]] = SqsAckSink(settings.sqs.queueUrl)(sqsClient)
```

aserrallerios:
@calvinlfer totally miss it, but looking at documentation at http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-client-side-buffering-request-batching.html looks like you're right

calvinlfer:
I'm actually using the Buffered Client and it works quite well 👍

mehmetgunturkun:
@calvinlfer  Frankly, I missed that buffered client. My only concern is that as far as I can see from documentation AmazonSQSBufferedAsyncClient hasn't supported FIFO queues yet. It might be a problem since contributors probably want this library to work with all queue types. 

Also, maybe users want to have a stage to send the batch of messages explicitly without waiting batch timeout.

calvinlfer:
@mehmetgunturkun I agree, it's still useful to have a client that can handle explicit batch messages so this is still useful. I'm using people using alpakka would love to take advantage of FIFO queues and if we can give that to them then this is another reason 👍 . 

I do want to mention in the documentation that you can take advantage of the `AmazonSQSBufferedAsyncClient` using the existing API to enable automatic batching. Should I do this as part of another PR?

mehmetgunturkun:
@calvinlfer I believe you should definitely add this information to documentation 👍

calvinlfer:
alright will do @mehmetgunturkun 
Look forward to seeing your change in, been wanting to experiment with FIFO queues

calvinlfer:
I came back after doing some more extensive tests on the buffered client. We placed over 8,000,000 messages in a non FIFO SQS queue. The Buffered Client pulls 20 messages and just fails to pull the remaining messages after so we need the explicit batch Sink.

mehmetgunturkun:
@calvinlfer I couldn't understand. It failed with an exception or just hung like there is no message on queue while there was?

calvinlfer:
@mehmetgunturkun sorry for the delay, it just hung like there is no messages on the queue when there was.

calvinlfer:
Hey guys, what's preventing this from going through 😛

johanandren:
Vacation times and none of us in the Akka team having had time to review yet I'm afraid, sorry, we'll get to it.

calvinlfer:
@johanandren thanks for the update :-)

adamsar:
Any movement on this? Anything I can do to help out?

ennru:
Some conflicts need resolving. 
We'll bake a release in two weeks time and it would be great, if we could get your contribution in.

mehmetgunturkun:
@raboof I added a note to point that users can use `AmazonSQSBufferedAsyncClient` for client-side buffering.

@ennru solved the conflicts and made it ready for merge. Sorry for the delay.

ennru:
Thank you for this contribution. We'll probably cut a release within the next two weeks.

PR #389, opened 2017-07-10
SSE: fix retry handling (closes #375)

### LINKED ISSUE #375: SSE reconnect issue
Hi,

`EventSource` is unable to reconnect to the server in the following scenarios.

1. The server is unavailable, the client starts - the client is unable to connect to the server when the server starts up. 
2. The server is available, the client connects to the server. Then the server gets restarted before sending any message to the client - the client is unable to reconnect to the server.

hseeberger:


sse/src/main/scala/akka/stream/alpakka/sse/scaladsl/EventSource.scala
```diff
@@ -103,6 +104,7 @@ object EventSource {
 
     val lastEventId =
       Flow[ServerSentEvent]
+        .prepend(Source.single(heartbeat)) // to make sliding and collect-matching work
```
hseeberger: This is the only relevant change, the rest is cosmetics.

PR #386, opened 2017-07-06
fix #380 handling IOExceptions in FTP sink onPush

frj:
This pull request fixes the problem detailed in #380. There are two issues addressed:
* IOExceptions thrown by the ftp client during onPush were not being handled, the interpreter handles the exception and calls postStop on the graph stage, which by default completes the IOResult promise with a success. This is fixed by intercepting the the exception in onPush and failing the promise with the exception. 
* postStop was throwing an additional exception, in the case of a failure in onPush, as by default it closes the OutputStream which is in a failed state. This seems unnecessary, so this change fixes that by cleaning up the IOStream in the exception handling in onPush.

PR #407, opened 2017-07-15
FTP Homogenize model

### LINKED ISSUE #395: FTP: RemoteFileSettings classes inside or outside companion object?
Most subclasses of `RemoteFileSettings` are defined inside the `RemoteFileSettings` companion object, but `SftpSettings` are defined outside.

What do we prefer?

juanjovazquez:
This is a proposal to make the model of the FTP component more homogeneous. It remains to be decided if we want to replicate in `FTP` and` FTPs` the [fluent API](https://github.com/akka/alpakka/blob/master/ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala#L116) that has been defined in `sFTP`, especially for Java use. Feedback is appreciated.

Closes #395

### DIFF
--- a/ftp/src/test/scala/akka/stream/alpakka/ftp/CommonFtpStageSpec.scala
+++ b/ftp/src/test/scala/akka/stream/alpakka/ftp/CommonFtpStageSpec.scala
@@ -71,7 +71,7 @@ trait CommonFtpStageSpec extends BaseSpec with Eventually {
   implicit val system = getSystem
   implicit val mat = getMaterializer
   implicit val defaultPatience =
-    PatienceConfig(timeout = Span(10, Seconds), interval = Span(300, Millis))
+    PatienceConfig(timeout = Span(30, Seconds), interval = Span(600, Millis))
 
   "FtpBrowserSource" should {
     "list all files from root" in {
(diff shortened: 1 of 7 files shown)


juanjovazquez:
It seems that #386 might have destabilized Travis construction

raboof:
I think moving these to the top level makes a lot of sense, nice.

That 'fluent API' seems nice, too - I'd be in favour of adding it for the other configuration classes as well. Any reason not to want it?

juanjovazquez:
@raboof No reason at all. In fact, I've already applied the fluent constructors to the other settings. Thanks. This PR would be ready for merging if you agree.

ftp/src/test/scala/akka/stream/alpakka/ftp/CommonFtpStageSpec.scala
```diff
@@ -71,7 +71,7 @@ trait CommonFtpStageSpec extends BaseSpec with Eventually {
   implicit val system = getSystem
   implicit val mat = getMaterializer
   implicit val defaultPatience =
-    PatienceConfig(timeout = Span(10, Seconds), interval = Span(300, Millis))
+    PatienceConfig(timeout = Span(30, Seconds), interval = Span(600, Millis))
```
raboof: Might be wise, I've seen timeouts as well. Are those actually to be expected? Do we know why those are so slow?

juanjovazquez: @raboof I don't know why they're so slow on Travis. Of course, it has to do with the fact that these tests need to start and stop FTP servers but we should do fine tuning on these instances in order to configure the servers properly for working on Travis builds.

raboof: Actually I've seen tests be quite slow on my local machine at times as well - but haven't taken the time to really dig into that yet.

PR #411, opened 2017-07-17
Explicitly expect some logging

raboof:
Makes explicit that we expect this logging, and suppresses it
when running the tests.

We still see some errors in the logging because sshj also logs
to slf4j itself. Perhaps we want something similar for that
output? Not sure if slf4j also has a kind of testkit/approach for
this?


raboof:
Hmm `FtpIOSinkStage` fails twice (once when failing to write to the stream and once when failing to gracefully disconnect from the server), not sure if that's problematic but makes this test tricky to write. Perhaps it's no big deal anyway :)

PR #412, opened 2017-07-17
Don't use `***` in the sbt header

raboof:
Makes it harder to search for `***` to find test failures

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -25,7 +25,7 @@ lazy val alpakka = project
   .settings(
     onLoadMessage :=
       """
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -25,7 +25,7 @@ lazy val alpakka = project
   .settings(
     onLoadMessage :=
       """
-        |*** Welcome to the sbt build definition for Alpakka! ***
+        |** Welcome to the sbt build definition for Alpakka! **
```
johanandren: ```[ERROR] Welcome to the set build definition for Alpakka [WARNING]```

Issue #414, opened 2017-07-17
RecordIOFramingSpec: parse a series of records


raboof:
https://travis-ci.org/akka/alpakka/jobs/254392980

Looks like an overly tight timeout setting, will widen.

```
[info] RecordIOFramingSpec:
[info] RecordIO framing
[info] - should parse a series of records *** FAILED *** (562 milliseconds)
[info]   A timeout occurred waiting for a future to complete. Queried 3 times, sleeping 15 milliseconds between each query. (RecordIOFramingSpec.scala:53)
[info]   org.scalatest.concurrent.Futures$FutureConcept$$anon$1:
[info]   at org.scalatest.concurrent.Futures$FutureConcept$class.tryTryAgain$1(Futures.scala:538)
[info]   at org.scalatest.concurrent.Futures$FutureConcept$class.futureValueImpl(Futures.scala:550)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:275)
[info]   at org.scalatest.concurrent.Futures$FutureConcept$class.futureValue(Futures.scala:476)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:275)
[info]   at akka.stream.alpakka.recordio.RecordIOFramingSpec$$anonfun$2.apply(RecordIOFramingSpec.scala:53)
[info]   at akka.stream.alpakka.recordio.RecordIOFramingSpec$$anonfun$2.apply(RecordIOFramingSpec.scala:43)
```

PR #415, opened 2017-07-17
Widen scalatest timeout for simple-codes

raboof:
Refs #414

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -168,7 +168,11 @@ lazy val simpleCodecs = project
   .in(file("simple-codecs"))
   .enablePlugins(AutomateHeaderPlugin)
   .settings(
-    name := "akka-stream-alpakka-simple-codecs"
+    name := "akka-stream-alpakka-simple-codecs",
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -168,7 +168,11 @@ lazy val simpleCodecs = project
   .in(file("simple-codecs"))
   .enablePlugins(AutomateHeaderPlugin)
   .settings(
-    name := "akka-stream-alpakka-simple-codecs"
+    name := "akka-stream-alpakka-simple-codecs",
+    // By default scalatest futures time out in 150 ms, dilate that to 600ms.
+    // This should not impact the total test time as we don't expect to hit this
+    // timeout, and indeed it doesn't appear to.
+    testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-F", "4")
```
johanandren: You think 600ms is enough to not get false positives (considering that we have 3 seconds in the Akka build)

raboof: Good question, maybe not but until now it has been *reasonably* stable at the pretty tight 150ms, so let's try this for now.

PR #416, opened 2017-07-17
Avoid overlapping filenames in tests

raboof:


raboof:
Hmm doesn't eliminate the `fail and report the exception in the result status if connection fails` failure though I'm afraid..

juanjovazquez:
@raboof I haven't had time to look into this but I can see [here](https://github.com/akka/alpakka/blob/57d865be865664cef34c756ab3e6d0ca43924197/ftp/src/test/scala/akka/stream/alpakka/ftp/CommonFtpStageSpec.scala#L260) and [here](https://github.com/akka/alpakka/blob/57d865be865664cef34c756ab3e6d0ca43924197/ftp/src/test/scala/akka/stream/alpakka/ftp/CommonFtpStageSpec.scala#L262) that the server is explicitly stopped and started. You need to take into account that currently every Scala test instance is initializing its own server instance. Once per test class. See [this](https://github.com/akka/alpakka/blob/57d865be865664cef34c756ab3e6d0ca43924197/ftp/src/test/scala/akka/stream/alpakka/ftp/BaseSpec.scala#L43). So, I wonder if this explicit server management is responsible for these errors.

raboof:
@juanjovazquez hmm, it certainly could be related (after all it's indeed the test that stops/starts the server that fails), but I don't see how: indeed the stream should fail because we stop the server, that's exactly what's being tested here.

PR #418, opened 2017-07-18
Disable parallel testing for projects that use mockito

raboof:
Not sure if those projects will still interfere with each other though. We'll
see. We might have to fork... or make sure this deadlock gets fixed upstream.

Refs #390


raboof:
(failure was #290, PR seems OK)

PR #430, opened 2017-08-01
#199 Support other datatypes than String for JMS-Sink

d2e8f9:
Please take a look and let me know if something is missing. The JmsSink and JmsSource supports now Byte-, Map- and ObjectMessages according to the Jms spec. 

Therefore  a JmsMessage trait is introduced and refactored out to a separate file. 
The logic to create and populate the Jms messages is still inside the JmsSinkStage. I would move it from there to another place, but I don't know which direction to go... 

- Passing the Java.jms.Session to the JmsMessages,so that this classes could create their javax message representation 

or    

-  introduce a MessageFactory (or something like that)

### DIFF
--- a/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
@@ -16,8 +16,6 @@ sealed trait Destination
 final case class Topic(name: String) extends Destination
 final case class Queue(name: String) extends Destination
 
-
-
 object JmsSourceSettings {
 
   def create(connectionFactory: ConnectionFactory) = JmsSourceSettings(connectionFactory)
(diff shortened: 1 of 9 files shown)


jugalps:
I am using a similar patch in 0.11, it would be good if this can be merged and made available in 0.14.

ennru:
Looks good as far as I can see. 
Would you have the chance to rebase and incorporate the changes in master?
Please suggest updates to the documentation, as well.

d2e8f9:
Okay, i will do it at the weekend.

ennru:
Just another ping. We'll bake a release in two weeks time and it would be great, if we could get your contribution in.

d2e8f9:
Sorry. I had trouble with my git repo. So I closed this pull request and created a new one. Please see therefore #612

Issue #282, opened 2017-05-03
[MQTT] Same client for subscribe and publish


tdroxler:
Currently it seems to be impossible to use a same client for subscribe and publish, which is not convenient since we can't connect the same client twice to a MQTT broker, the second connection drop the first one.

We can't because the client is automatically created inside the mqtt `Source` and the `Sink`. 

To fit my need, I materialized the client in the `Source`. (see https://github.com/tdroxler/alpakka/commit/e83a641b18b17bf3368772836f793dd003c5c869)

So I can directly use my client to publish, but using directly the `IMqttAsyncClient` paho functions.

But this is not a clean solution, since we still can't create a `Source` and `Sink` at the same time. I don't know if you have plan to tackle this problem or maybe have some hints so I can make it cleaner and useful for others?

Thx, cheers

2m:
Have you tried using a different `clientId` [like so](https://github.com/akka/alpakka/blob/399b968ae5fcb003364a828226df31406fb95aa4/mqtt/src/test/scala/akka/stream/alpakka/mqtt/scaladsl/MqttSinkSpec.scala#L40). When `clientId`s are different, a second connection should not drop the first one.

tdroxler:
Yes I tried, this is actually when I have a connection with an [auth](https://github.com/akka/alpakka/blob/399b968ae5fcb003364a828226df31406fb95aa4/mqtt/src/main/scala/akka/stream/alpakka/mqtt/Mqtt.scala#L78), as well as a [socketFactory](https://github.com/akka/alpakka/blob/399b968ae5fcb003364a828226df31406fb95aa4/mqtt/src/main/scala/akka/stream/alpakka/mqtt/Mqtt.scala#L79) that I get disconnected.

PR #432, opened 2017-08-04
Implement an MQTT flow for bidirectional communication with the same client ID

mwegrz:
Existing implementation provides separate `MqttSource` and `MqttSink`. This causes problems if connecting to an MQTT server that requires a single client ID per user/client (e.g. [IBM Watson IoT Platform](https://console.bluemix.net/docs/services/IoT/reference/security/connect_devices_apps_gw.html#connect_devices_apps_gw)). This issue is already tracked in the following issue: https://github.com/akka/alpakka/issues/282.

This PR implements an MQTT flow that combines the logic of `MqttSource` and `MqttSink` and opens just a single client connection/session to the remote server using a single client ID. 

Public API remains unchanged with the only addition of `MqttFlow`. Internal implementations of `MqttSource` and `MqttSink` are now based on the new flow in order to keep the code DRY. 

`MqttProducerStage` and `MqttSourceStage` are no longer used and are removed in this PR. Current tests `MqttSourceSpec` and `MqttSinkSpec` remain unchanged but now they test the new implementations based on `MqttFlow`.


raboof:
Hi @mwegrz , sorry for not looking into this sooner - this looks like a pretty interesting feature! I'm afraid this PR has developed some merge conflicts with master, could you fix those?

mwegrz:
Thanks @raboof. Will fix those soon.

mwegrz:
Hi @raboof, the PR is ready to be reviewed

mqtt/src/main/scala/akka/stream/alpakka/mqtt/MqttFlowStage.scala
```diff
@@ -0,0 +1,125 @@
...
+
+import scala.collection.mutable
+import scala.concurrent.{Future, Promise}
+import scala.util.{Failure, Success, Try}
+
+object MqttFlowStage {
+  final object NoClientException extends Exception("No MQTT client.")
+}
+
+class MqttFlowStage(sourceSettings: MqttSourceSettings, bufferSize: Int, qos: MqttQoS)
```
johanandren: If you like it, make it final ;)

mqtt/src/main/scala/akka/stream/alpakka/mqtt/MqttFlowStage.scala
```diff
@@ -0,0 +1,125 @@
...
+      override def handleConnectionLost(ex: Throwable): Unit = {
+        failStage(ex)
+        subscriptionPromise.tryFailure(ex)
+      }
+
+      override def postStop(): Unit =
+        mqttClient.foreach {
+          case c if c.isConnected => c.disconnect()
+          case c => ()
+        }
```
johanandren: You also need to `tryFail` the subscriptionPromise here in case the stage is abruptly terminated (`materializer.shutdown()`) before it managed to subscribe

mwegrz:
Thx for the quick review @johanandren

Issue #367, opened 2017-06-27
S3 download fails when credentials are provided at EC2 level


shyamsalimkumar:
```scala
val credentialsProvider = new DefaultAWSCredentialsProviderChain()
val credentials = credentialsProvider.getCredentials
val awsCredentials = AWSCredentials(
  accessKeyId = credentials.getAWSAccessKeyId,
  secretAccessKey = credentials.getAWSSecretKey
)
val settings = new S3Settings(
  MemoryBufferType,
  "",
  None,
  awsCredentials,
  appConfiguration.awsRegion,
  false
)
val s3Client = new S3Client(settings)

s3Client
  .download(bucketName, keyName)
  .toMat(FileIO.toPath(Paths.get(fileName)))(Keep.right)
  .run()
  .foreach { x =>
    if (x.wasSuccessful) {
      println(s"Downloaded")
    } else {
      throw x.getError
    }
  }
```

The above works when credentials are placed at `~/.aws/credentials` (and probably at lower levels too) but failed when they are set at a higher level (Amazon ECS container credentials/Instance profile credentials) with the message `The AWS Access Key Id you've used does not exist in our records.`. I don't think it has something to do with `DefaultAWSCredentialsProviderChain` as the Java SDK also uses the same thing internally afaik, but that was working fine. 

Code that I'd tried out using Java SDK 

```scala
val s3ClientBuilder = AmazonS3ClientBuilder.standard()
s3ClientBuilder.setRegion(region)
val s3 = s3ClientBuilder.build()

try {
  val o = s3.getObject(bucketName, keyName)
  val s3is = o.getObjectContent

  val in = scala.io.Source.fromInputStream(s3is)
  val out = new java.io.PrintWriter(new File(fileName))
  try {
    in.getLines().foreach(out.println)
  } catch {
    case NonFatal(t) => println(t)
  } finally {
    s3is.close()
    out.close()
  }
} catch {
  case e: AmazonServiceException =>
    System.err.println(e.getErrorMessage)
    System.exit(1)
  case e: FileNotFoundException =>
    System.err.println(e.getMessage)
    System.exit(1)
  case e: IOException =>
    System.err.println(e.getMessage)
    System.exit(1)
}
```

Another thing I noticed (not sure if its important) is that the credential value returned by `DefaultAWSCredentialsProviderChain` changes when it is at a higher level periodically, but that too doesn't seem to affect the Java SDK version.

Tvaroh:
Having same issue.

[...]

PR #440, opened 2017-08-11
[S3] Use AWS Core SDK credentials for auth

tkroman:
Use AWS Core SDK credentials mechanism to fully support all possible auth variants.

This will automatically fix e.g. #367 and solve all the potential problems with token refresh etc.

I took a liberty to change the config as well. Probably will need more docs as well.

Just let me know if this is even the right approach. Tests didn't break (after some adjustments), but compatibility (source/binary) is certainly broken.

I also ran `scalafmt` on the repo root, hence some seemingly unrelated changes.


tkroman:
Here's the ammonite script which I ran on an EC2 instance to check that everything is fine: https://gist.github.com/tkroman/b41a54ff0753fe8be13474e2829eb688. As you can see, no credentials are used, but this works perfectly.

shyamsalimkumar:
Just curious, but is it possible to implement the same but avoid having to include `aws-java-sdk` as a dependency or is that too much of a hassle? There might've been a good reason for removing the SDK dependency in the first place.

tkroman:
Too much. With AWS SDK “we” support all the available ways to get credentials - from .aws/credentials, properties, env variables, ec2 metadata service, anonymous access & profile credentials. Some of these support expiration, others do not etc. This would take too much effort to implement and support.

avietrov:
Would it make sense to split into this PR into two? as you've mentioned, scalafmt did lots of changes.

avietrov:
I see you've used AWSCredentials. Just for my understanding, for whatever reason only AWSSessionCredentials work in my case, using simple credentials fails with signing errors. Is this something specific to my setup?

tkroman:
I’m not sure if PRs that don’t pass formatting checks (if there is such a thing) can be accepted.

On account of your credentials question - sorry, need more context.

tkroman:
Oh wait, I got the question. Sure, there is a difference. Session credentials also include a token, which is ignored by basic credentials, but required in the process of signing. Basically, this is what I’m talking about - too much to understand, too much to implement anew, too much to take care about.

idzivinskyi:
@tkroman it looks so good. When do you plan to release it?

tkroman:
@idzivinskyi when the akka team has time to review, I think. 

+1 on a PR could make a difference. I'm going to go and shamelessly self-promote on gitter :D.

tkroman:
@johanandren @raboof (tagging you because looks like you are the active maintainers, sorry if no) any chance of getting some feedback on this?

s3/src/main/scala/akka/stream/alpakka/s3/scaladsl/S3Client.scala
```diff
@@ -48,10 +45,13 @@ object S3Client {
   def apply()(implicit system: ActorSystem, mat: Materializer): S3Client =
     new S3Client(S3Settings(system.settings.config))
 
-  def apply(credentials: AWSCredentials, region: String)(implicit system: ActorSystem, mat: Materializer): S3Client = {
-    val settings = S3Settings
-      .apply(system.settings.config)
-      .copy(awsCredentials = credentials, s3Region = region)
+  def apply(credentials: AWSCredentialsProvider, region: String)(implicit system: ActorSystem,
```
johanandren: Could we keep the old signature as well, and just call this method wrapping the AWSCredentials with a static provider, that way we avoid breaking everyones code.

johanandren:
Sorry about the long delay for review, we have been very busy with other things.

s3/src/main/scala/akka/stream/alpakka/s3/scaladsl/S3Client.scala
tkroman: Will do. However, the code relying on config will still break, do you want me to do a compatibility layer for that?

tkroman:
Merged, rebased on the latest master, ran scalafmt. Should be 👌 now.

tkroman:
Also, added compat layer for config, so old users should not suffer from this change (I hope) Check out s3/src/test/scala/akka/stream/alpakka/s3/S3SettingsSpec.scala & comments in reference.conf.

idzivinskyi:
Hi @tkroman @johanandren, 

I have a problem when I use this approach. It looks like after expiration of token or after a big timeout between requests ```s3Client.download``` starts to return new token instead of the original file. (I have one instance of client per application)

What should I do in that case? Do I need to init s3Client for each request or I need to add an additional configuration to aws auth?

Thanks.

tkroman:
1) wdym “token instead of file”, is that a typo or a really really weird behavior?
2) what is your current configuration?

idzivinskyi:
```
val s3Client: S3Client = S3Client()
```

configuration from config:
```
stream.alpakka.s3 {
    buffer = "memory"
    aws {
      credentials.provider = default
      default-region = "us-east-2"
    }
  }
```

tkroman:
I don't think tokens should expire when used properly. Can you ping me on gitter so we can discuss this in more detail? I'm not affiliated with the Akka team, just feeling sort of responsible for the behavior my changes might be related to.

idzivinskyi:
I will link a solution here: https://github.com/akka/alpakka/pull/545/files

PR #455, opened 2017-08-31
Initial support for Neo4j

cheleb:
Source and Flow from simple cypher statement.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -16,6 +16,7 @@ lazy val modules: Seq[ProjectReference] = Seq(
   kinesis,
   mongodb,
   mqtt,
+  neo4j,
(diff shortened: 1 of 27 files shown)


cheleb:
I wanted to handle "generic" solution for simple use case. BTW it would be much simpler to "leave how to serialize/deserialize elements to the user" !
Perhaps, I should implement the two options ... WDYT ?

johanandren:
I think in some other connector, um, the MongoDB one perhaps, we decided to have a generic serialization interface, and then had a sample (test dependency on the actual library) showing how to use some specific library. WDYT about that?

ennru:

Issue #276, opened 2017-04-24
DynamoDB: add  support for temporary security credentials in AwsClient


mczubak:
I'd like to add ```x-amz-security-token``` header in AwsClient implementation. It needs to be added if AWS credentials are obtained using AWS Security Token Service API. Otherwise requests to dybamodb service cannot be properly signed. 

You can read more [here](http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html#UsingTemporarySecurityCredentials)

I can create a PR with small update to AwsClient

PR #470, opened 2017-09-16
[DDB] Support for AWS Session Credentials - Using the Token  session auth

teroxik:
Current implementation uses the DefaultAWSCredentialsProviderChain which provides support for obtaining the credentials from the instance. 

When the headers are added on the http request the  "X-Amz-Security-Token" is ommited but it is required.

Small DRY of the host address and update for the documentation for the port settings.

Tested in AWS. Any suggestion how to test it locally welcome.

See #276


cornelf:
👍

PR #292, opened 2017-05-09
Added a source that doesn't automatically ACK messages to rabbit #291

Falmarri:
There's some parts of this that I'm not sure about. Specifically the implementation of Ack and Nack being subclasses of an arbitrary Command function.

Issue #104, opened 2016-11-30
define Committable


patriknw:
For sources that read from a queue and acknowledge consumption later downstream.

Something like `Committable` in akka-stream-kafka.
https://github.com/akka/reactive-kafka/blob/master/core/src/main/scala/akka/kafka/ConsumerMessage.scala#L37

johanandren:
Should it go in Alpakka and not directly in Akka Streams?

ktoso:
It can start here, kind of as an incubator?

drewhk:
It should not be just Committable, but Repliable. Commit is a specific use case of in-stream replies.

drewhk:
Also, an adapter stage is needed that takes a flow of Reply-Request and turns it into a Source[Repliable]

patriknw:
I was creating the issue here to spark discussion and explore it with concrete examples from the Alpakka connectors. When it has matured the interfaces should move to akka-stream so that same can be used in Alpakka and akka-stream-kafka (and other places)

filosganga:
I need it as well for ironmq

pspiegel:
We (SBB) also need this feature in order to acknowledge (commit) or discard (rollback) messages from the JMSConnector manually later downstreams.
Furthermore, after a few (unsuccessful) retries we wanna put the message in some kind of a backout Q.

Falmarri:
I have a PR out that is a first draft of this functionality, but I didn't see this issue before I wrote it. It's not nearly as safe as the design suggested in this version.

https://github.com/akka/alpakka/pull/292

ssturdivant:
Where does this ticket lie on the development roadmap? I have a project requiring application-level JMS ack, and need to decide between waiting for this ticket or implementing another solution.

Thanks.

patriknw:
Akka team is not working on this and we don't have it in plans for the next month.

EmilDafinov:
Since https://github.com/akka/alpakka/pull/483 went in, would that count as a good example of Commitable / Repliable as described here? Would the `CommittableIncomingMessage` introduced work as an interface for such things?

juanjoDiaz:
I've also added a committable message to MQTT following the same pattern in #515 (still WIP).

I think that all the systems are different: some can reject and other not, some require some parameters and some require others, etc.

So I'm not totally convinced of a general committable trait being a good idea.

EmilDafinov:
[...]

PR #483, opened 2017-09-21
Allow message ack downstream

### LINKED ISSUE #97: AMQP Source acknowledges messages too early
Currently the AMQP source sends an acknowledgement back to the RabbitMQ server as soon as it sends the received message to the outlet. 
https://github.com/akka/alpakka/blob/master/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSourceStage.scala#L126

If anything happens to the message during the flow which would cause the flow to explode then the message will not be re-queued (although other messages that were in the buffer will). It should be possible to reserve the acknowledgement of the message until it has reached the sink. This way there would be no possibility of losing a message due to some non related runtime exception.

I haven't fully thought out how this would deal with invalid messages which could cause repeated exceptions, but it'd be cool to see if acking after the flow had been considered.

juanjoDiaz:
Fix #97 by adding the possibility of deactivate `autoAck` and allowing to `ack` or `nack` messages downstream.

Nack allow to optionally requeue the message or not (by default it does).

I've noticed that #292 was doing the same but it seems stale and the work seems bloated to me (it introduce a lot of duplicities and unrelated changes).

This PR introduce a very minimal change.
No breaking changes at API level.
Fully tested.

Only "breaking" change is that `IncomingMessage` becomes a trait and `AckedIncomingMessage` and `UnackedIncomingMessage` are case classes extending it. (I'm any case I don't think that anyone should be using `IncomingMessage` so this shouldn't even be a real breaking change)
The advantage of having `AckedIncomingMessage` and `UnackedIncomingMessage` as case classes inheriting from `IncomingMessage` is that is minimize changes, maximize backwards compatibility and allow pattern matching to act differently depending on the type of message.

### DIFF
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
@@ -66,7 +66,7 @@ private[amqp] trait AmqpConnector {
 private[amqp] trait AmqpConnectorLogic { this: GraphStageLogic =>
 
   private var connection: Connection = _
-  protected var channel: Channel = _
+  protected implicit var channel: Channel = _
 
   def settings: AmqpConnectorSettings
   def connectionFactoryFrom(settings: AmqpConnectionSettings): ConnectionFactory
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
@@ -66,7 +66,7 @@ private[amqp] trait AmqpConnector {
 private[amqp] trait AmqpConnectorLogic { this: GraphStageLogic =>
 
   private var connection: Connection = _
-  protected implicit var channel: Channel = _
+  protected var channel: Channel = _
 
   def settings: AmqpConnectorSettings
   def connectionFactoryFrom(settings: AmqpConnectionSettings): ConnectionFactory
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
@@ -14,7 +14,8 @@ import com.rabbitmq.client.AMQP.BasicProperties
 import com.rabbitmq.client._
 
 import scala.collection.mutable
-import scala.concurrent.{ExecutionContextExecutor, Future, Promise}
+import scala.concurrent.{Future, Promise}
+import scala.util.Try
 
 object AmqpRpcFlowStage {
 
@@ -44,8 +45,6 @@ final class AmqpRpcFlowStage(settings: AmqpSinkSettings, bufferSize: Int, respon
   override def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, Future[String]) = {
     val promise = Promise[String]()
     (new GraphStageLogic(shape) with AmqpConnectorLogic {
-
-      implicit def ec: ExecutionContextExecutor = materializer.executionContext
 
       override val settings = stage.settings
       private val exchange = settings.exchange.getOrElse("")
@@ -79,9 +78,19 @@ final class AmqpRpcFlowStage(settings: AmqpSinkSettings, bufferSize: Int, respon
         channel.basicQos(bufferSize, true)
         val consumerCallback = getAsyncCallback(handleDelivery)
 
-        val commitCallback = getAsyncCallback[Unit] { (_) =>
-          unackedMessages.decrementAndGet()
-          if (unackedMessages.get() == 0) completeStage()
+        val commitCallback = getAsyncCallback[CommitCallback] {
+          case AckArguments(deliveryTag, multiple, promise) => {
+            channel.basicAck(deliveryTag, multiple)
+            unackedMessages.decrementAndGet()
+            if (unackedMessages.get() == 0) completeStage()
+            promise.complete(Try(Done))
+          }
+          case NackArguments(deliveryTag, multiple, requeue, promise) => {
+            channel.basicNack(deliveryTag, multiple, requeue)
+            unackedMessages.decrementAndGet()
+            if (unackedMessages.get() == 0) completeStage()
+            promise.complete(Try(Done))
+          }
         }
 
         val amqpSourceConsumer = new DefaultConsumer(channel) {
@@ -95,19 +104,17 @@ final class AmqpRpcFlowStage(settings: AmqpSinkSettings, bufferSize: Int, respon
               new CommittableIncomingMessage {
                 override val message = IncomingMessage(ByteString(body), envelope, properties)
 
-                override def ack(multiple: Boolean) =
-                  Future {
-                    channel.basicAck(message.envelope.getDeliveryTag, multiple)
-                    commitCallback.invoke(None)
-                    Done
-                  }
+                override def ack(multiple: Boolean) = {
+                  val promise = Promise[Done]()
+                  commitCallback.invoke(AckArguments(message.envelope.getDeliveryTag, multiple, promise))
+                  promise.future
+                }
 
-                override def nack(multiple: Boolean, requeue: Boolean) =
-                  Future {
-                    channel.basicNack(message.envelope.getDeliveryTag, multiple, requeue)
-                    commitCallback.invoke(None)
-                    Done
-                  }
+                override def nack(multiple: Boolean, requeue: Boolean) = {
+                  val promise = Promise[Done]()
+                  commitCallback.invoke(NackArguments(message.envelope.getDeliveryTag, multiple, requeue, promise))
+                  promise.future
+                }
               }
             )
 
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
@@ -3,6 +3,8 @@
  */
 package akka.stream.alpakka.amqp
 
+import java.util.concurrent.atomic.AtomicInteger
+
 import akka.stream._
 import akka.stream.stage._
 import akka.util.ByteString
@@ -25,15 +27,15 @@ object AmqpRpcFlowStage {
  *                            can be overridden per message by including `expectedReplies` in the the header of the [[OutgoingMessage]]
  */
 final class AmqpRpcFlowStage(settings: AmqpSinkSettings, bufferSize: Int, responsesPerMessage: Int = 1)
-    extends GraphStageWithMaterializedValue[FlowShape[OutgoingMessage, IncomingMessage], Future[String]]
+    extends GraphStageWithMaterializedValue[FlowShape[OutgoingMessage, CommittableIncomingMessage], Future[String]]
     with AmqpConnector { stage =>
 
   import AmqpRpcFlowStage._
 
   val in = Inlet[OutgoingMessage]("AmqpRpcFlow.in")
-  val out = Outlet[IncomingMessage]("AmqpRpcFlow.out")
+  val out = Outlet[CommittableIncomingMessage]("AmqpRpcFlow.out")
 
-  override def shape: FlowShape[OutgoingMessage, IncomingMessage] = FlowShape.of(in, out)
+  override def shape: FlowShape[OutgoingMessage, CommittableIncomingMessage] = FlowShape.of(in, out)
 
   override protected def initialAttributes: Attributes = defaultAttributes
 
@@ -44,10 +46,10 @@ final class AmqpRpcFlowStage(settings: AmqpSinkSettings, bufferSize: Int, respon
       override val settings = stage.settings
       private val exchange = settings.exchange.getOrElse("")
       private val routingKey = settings.routingKey.getOrElse("")
-      private val queue = mutable.Queue[IncomingMessage]()
+      private val queue = mutable.Queue[CommittableIncomingMessage]()
       private var queueName: String = _
+      private val unackedMessages = new AtomicInteger()
       private var outstandingMessages = 0
-      private val autoAck = settings.autoAck
 
       override def connectionFactoryFrom(settings: AmqpConnectionSettings) = stage.connectionFactoryFrom(settings)
       override def newConnection(factory: ConnectionFactory, settings: AmqpConnectionSettings): Connection =
@@ -73,6 +75,13 @@ final class AmqpRpcFlowStage(settings: AmqpSinkSettings, bufferSize: Int, respon
         channel.basicQos(bufferSize, true)
         val consumerCallback = getAsyncCallback(handleDelivery)
 
+        val commitCallback = getAsyncCallback[Unit] { (_) =>
+          unackedMessages.decrementAndGet()
+          if (outstandingMessages == 0 && unackedMessages.get() == 0 && isClosed(in)) {
+            completeStage()
+          }
+        }
+
         val amqpSourceConsumer = new DefaultConsumer(channel) {
           override def handleDelivery(
               consumerTag: String,
@@ -81,8 +90,10 @@ final class AmqpRpcFlowStage(settings: AmqpSinkSettings, bufferSize: Int, respon
               body: Array[Byte]
           ): Unit =
             consumerCallback.invoke(
-              if (autoAck) AckedIncomingMessage(ByteString(body), envelope, properties)
-              else UnackedIncomingMessage(ByteString(body), envelope, properties)
+              CommittableIncomingMessage(
+                IncomingMessage(ByteString(body), envelope, properties),
+                commitCallback
+              )
             )
 
           override def handleCancel(consumerTag: String): Unit =
@@ -112,9 +123,9 @@ final class AmqpRpcFlowStage(settings: AmqpSinkSettings, bufferSize: Int, respon
         promise.success(queueName)
       }
 
-      def handleDelivery(message: IncomingMessage): Unit =
+      def handleDelivery(message: CommittableIncomingMessage): Unit =
         if (isAvailable(out)) {
-          pushAndAckMessage(message)
+          pushMessage(message)
         } else {
           if (queue.size + 1 > bufferSize) {
             failStage(new RuntimeException(s"Reached maximum buffer size $bufferSize"))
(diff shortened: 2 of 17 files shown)


juanjoDiaz:
Does anyone has any clue of why is this failing the build?

`JavaConverters.seqAsJavaListConverter` is certainly a method. http://www.scala-lang.org/api/current/scala/collection/JavaConverters$.html#seqAsJavaListConverter[A](b:Seq[A]):scala.collection.convert.Decorators.AsJava[java.util.List[A]]

johanandren:
I think the build failed because it's been checked in with incorrect formatting. Run a compile of it before checking in and things should be alright.

juanjoDiaz:
Hi @johanandren ,

Thanks for the response. I fixed the styles but the issue persist.
```
[error] /home/travis/build/akka/alpakka/amqp/src/test/java/akka/stream/alpakka/amqp/javadsl/AmqpConnectorsTest.java:279: cannot find symbol
[error]   symbol:   method seqAsJavaList(scala.collection.immutable.Seq<akka.stream.alpakka.amqp.IncomingMessage>)
[error]   location: class scala.collection.JavaConverters
[error]     List<IncomingMessage> probeResult = JavaConverters.seqAsJavaList(probe.toStrict(Duration.create(5, TimeUnit.SECONDS)));
```

As mentioned, that's odd since the method definitely exist.

juanjoDiaz:
Hurray! It turns out I was living too much on the bleeding edge and using the latest Scala.
Now everything looks good.

Is there any chance to get and ETA of when this could be merged and released?

johanandren:
I'm not so sure this specific API is a good idea, would be better if #104 got anywhere and we could use the same committable interface across connectors than if each stage creates its own way to doing things. I also find it a bit weird that you can mix unacked and acked messages with the same materialised stage, deciding that upon creation would make more sense to me.

juanjoDiaz:
Hi Johan,

I agree that fixing #104 would be optimal. However, it is clearly stated that "Akka team is not working on this and we don't have it in plans for the next month."
Also, some connectors require acknowledging and some probably don't. Some take options and some don't. So I don't think that really possible to a have a general committable message that works for everything. But I'm open to suggestions.

Could you elaborate a bit on the issues you can see with the current approach?

You actually will always get only Acked or Unacked messages depending on whether `autoAck` is set to true or false.
They'll never be mixed. The reason to use a generic `IncomingMessage` stage is to avoid having to create 2 stages with 99% the same code. But I'm happy to do that if that will help to get this merged.

I'm happy to do any changes that are required to get this functionality in place. This issue is sort of jeopardizing the integrity of some of my system and I really wouldn't want to create my own fork for this.

johanandren:
Ok, fair point about #104 and nobody stepping up to do that.

Taking a step back I still think a better design is not to define committable per message but to look at reactive-kafka connector which is the most battle tested and worked through API, something along the lines of:

```
trait Committable {
  def message: IncomingMessage
  // I think it has to be future because commit leads to a blocking network call that needs to be managed
  def commit(): Future[Unit] 
}


AmqpSource.createCommittable(...): Source[Committable, ...] = ???
AmqpSource.create(...): Source[IncomingMessage, ...] = createComittable(...).mapAsync(1)(_.commit)
```

This also avoids having to implement two stages, while giving the same control to the user to do the batched commit instead of one message per time to get better throughput etc.

Wdyt?

juanjoDiaz:
Sounds good.

I took a look at akka-streams-kafka and took the same approach. Does it look better now?

amqp/src/main/scala/akka/stream/alpakka/amqp/javadsl/AmqpSource.scala
```diff
@@ -13,6 +13,31 @@ object AmqpSource {
...
   def create(settings: AmqpSourceSettings, bufferSize: Int): Source[IncomingMessage, NotUsed] =
+    atMostOnceSource(settings, bufferSize)
+
+  /**
+   * Java API: Convenience for "at-most once delivery" semantics. Each message is acked to Kafka
+   * before it is emitted downstream.
+   */
+  def atMostOnceSource(settings: AmqpSourceSettings, bufferSize: Int): Source[IncomingMessage, NotUsed] =
+    committableSource(settings, bufferSize)
+      .map(cm => {
```
juanjoDiaz: I tried:
```
      .mapAsync(1)(cm => {
        Future {
          println(s"Committing ${cm.message.bytes.utf8String}")
          cm.ack()
          cm.message
        } (ExecutionContexts.sameThreadExecutionContext)
      })
```
But that makes the `not ack messages unless they get consumed` test fail. For some reason it ack the second element in `subscriber` even though it doesn't fetch it, and then the element is missing on `subscriber2`.
Maybe you have some idea of why is this happening?

In any case, the `ack` method of RabbitMQ library is blocking so it should be pretty much the same either way.

amqp/src/main/scala/akka/stream/alpakka/amqp/javadsl/AmqpSource.scala
```diff
@@ -13,6 +13,31 @@ object AmqpSource {
...
   def create(settings: AmqpSourceSettings, bufferSize: Int): Source[IncomingMessage, NotUsed] =
+    atMostOnceSource(settings, bufferSize)
+
+  /**
+   * Java API: Convenience for "at-most once delivery" semantics. Each message is acked to Kafka
+   * before it is emitted downstream.
+   */
+  def atMostOnceSource(settings: AmqpSourceSettings, bufferSize: Int): Source[IncomingMessage, NotUsed] =
+    committableSource(settings, bufferSize)
+      .map((cm: CommittableIncomingMessage) => {
```
juanjoDiaz: I tried:
```
      .mapAsync(1)(cm => {
        Future {
          println(s"Committing ${cm.message.bytes.utf8String}")
          cm.ack()
          cm.message
        } (ExecutionContexts.sameThreadExecutionContext)
      })
```
But that makes the `not ack messages unless they get consumed` test fail. For some reason it ack the second element in `subscriber` even though it doesn't fetch it, and then the element is missing on `subscriber2`.
Maybe you have some idea of why is this happening?

In any case, the `ack` method of RabbitMQ library is blocking so it should be pretty much the same either way.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSourceStage.scala
```diff
@@ -12,6 +14,21 @@ import com.rabbitmq.client._
 import scala.collection.mutable
 
 final case class IncomingMessage(bytes: ByteString, envelope: Envelope, properties: BasicProperties)
+final case class CommittableIncomingMessage(
+    message: IncomingMessage,
+    callback: AsyncCallback[Unit] = new AsyncCallback[Unit] {
+      override def invoke(t: Unit): Unit = Unit
+    }
```
johanandren: The message shouldn't need to know about `AsyncCallback` just a regular function should do. After thinking about the things noted below, I think it should be callback: `(ackOrNack+data) => Future[Unit]` with the whole implementation of that function in the stage using `AsyncCallback`

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSourceStage.scala
```diff
@@ -12,6 +14,21 @@ import com.rabbitmq.client._
...
+)(implicit channel: Channel) {
+  def ack(multiple: Boolean = false): Unit = {
+    channel.basicAck(message.envelope.getDeliveryTag, multiple)
```
johanandren: This is tricky:

Channel is an internal resource of the stage, so it shouldn't be leaked outside, what if the stage was stopped or failed? Also, I think `basicAck` leads to a network call, is it blocking, if so what thread is this executed on?

On the other hand, if the logic is moved into the stage, then this ack needs to return a `Future[...]` to signal to the stream that the ack was completed or failed before emitting the message downstream.

amqp/src/main/scala/akka/stream/alpakka/amqp/javadsl/AmqpSource.scala
```diff
@@ -13,6 +13,31 @@ object AmqpSource {
    * Java API: Creates an [[AmqpSource]] with given settings and buffer size.
    */
   def create(settings: AmqpSourceSettings, bufferSize: Int): Source[IncomingMessage, NotUsed] =
```
johanandren: If we are just calling the new one, then maybe `@deprecate` it and tell people to use `atMostOnceSource` instead.

amqp/src/main/scala/akka/stream/alpakka/amqp/javadsl/AmqpRpcFlow.scala
```diff
@@ -59,4 +59,38 @@ object AmqpRpcFlow {
       .mapMaterializedValue(f => f.toJava)
       .asJava
 
+  /**
+   * Java API:
+   * Convenience for "at-most once delivery" semantics. Each message is acked to Kafka
+   * before it is emitted downstream.
```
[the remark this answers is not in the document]
juanjoDiaz: Oops! I went too much into kafka here :)

amqp/src/main/scala/akka/stream/alpakka/amqp/javadsl/AmqpSource.scala
juanjoDiaz: 100% agree on this one. I was going to ask about that but didn't want to modify much the API without taking first with core maintainers. I'll `@deprecate` the old constructor in favour of `atMostOnceSource `

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSourceStage.scala
juanjoDiaz: I see. So you are suggesting that I wrapped the Async callback in a plain function and pass that to the Committable message, right?

There is some logic in the Stage so it won't be completed until all the messages have been acked. If something really goes wrong I guess the messages won't be acked. I think it's the same in the akka-streams-kafka, we offer out of the box at-least-once delivery but not exactly-once delivery.

The ack method from rabbitMQ is blocking AFAIK. I tried wrapping it in a Future and using mapAsync but got the issue that I mentioned in a comment above. Maybe you have any suggestion?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSourceStage.scala
juanjoDiaz: I see. So you are suggesting that I wrapped the Async callback in a plain function and pass that to the Committable message, right?

There is some logic in the Stage so it won't be completed until all the messages have been acked. If something really goes wrong I guess the messages won't be acked. I think it's the same in the akka-streams-kafka, we offer out of the box at-least-once delivery but not exactly-once delivery.

The ack method from rabbitMQ is blocking AFAIK. I tried wrapping it in a Future and using mapAsync but got the issue that I mentioned in a comment above. Maybe you have any suggestion?

juanjoDiaz:
Alright, hope we are getting there.

I got a bunch of inspiration from Kafka and alpakka.ironMQ and I think I got all your concerns addressed. Let me know if there is anything I've missed.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
```diff
@@ -90,10 +92,25 @@ final class AmqpRpcFlowStage(settings: AmqpSinkSettings, bufferSize: Int, respon
...
-              CommittableIncomingMessage(
-                IncomingMessage(ByteString(body), envelope, properties),
-                commitCallback
-              )
+              new CommittableIncomingMessage {
+                override val message = IncomingMessage(ByteString(body), envelope, properties)
+
+                override def ack(multiple: Boolean) = {
+                  Future {
+                    channel.basicAck(message.envelope.getDeliveryTag, multiple)
```
johanandren: Maybe I was a bit unclear, I think this should go into the `commitCallback` logic, so that it is executed on the same thread as the stage itself and the future is not completed until both ack/nack and changing the counter is done, that means we need to pass a promise from here to the async callback, call the channel method, update the counter and then (if successful) complete the promise. 

There will still be some unknowns with that though, what if the stage already failed/stopped when a downstream stage calls ack/nack? What if it was alive when ack/nack was called but fails before the async callback has been executed. There is a solution for that coming up with https://github.com/akka/akka/pull/23185 though, so maybe complete this and then update to use that later.

juanjoDiaz: This future uses the `materializer.executionContext` from the stage. Which as far as I understand is exactly how Kafka and IronMQ connectors do it.

The call is blocking so the future won't complete until the call is done.

Adding the call to `channel.ack` or `channel.nack` to the `commitCallback` has the issue that we won't be able to use parameters (or we need a separate callback for each combination of parameters).

Can you elaborate a bit in what exactly you would like me to do here? I'm a bit lost :)

Regarding the fail/stop uncertainties. I think that we should complete this work and then update later as you suggest.

johanandren: The counting down inflight messages and the nack is now done in separate places/threads even though they both belong to the same "logical" operation. The future will can complete even though the `asyncCallback` has not executed yet

Handling the two different calls with a single callback can be done by just encoding the needed data plus the promise in a single or two case classes with a common supertype whose sole purpose is to be passed across the async boundary for the stage to do things with and then complete the promise. (something like how we do it in the `InputStreamStage` for example: https://github.com/akka/akka/blob/3ba093d27e830bea675e45d0cec56860de3d133e/akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSinkStage.scala#L56 )

johanandren: To say it in another way, rather than scheduling two blocks of logic to execute, the `Future` and the `asyncCallback` we can eliminate one, and execute everything in the `asyncCallback`.

juanjoDiaz:
Alright. Let's see if I'm getting any closer which my latest commit :)

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
```diff
@@ -66,7 +66,7 @@ private[amqp] trait AmqpConnector {
 private[amqp] trait AmqpConnectorLogic { this: GraphStageLogic =>
 
   private var connection: Connection = _
-  protected var channel: Channel = _
+  protected implicit var channel: Channel = _
```
johanandren: Shouldn't need to be implicit anymore, right?

amqp/src/test/scala/akka/stream/alpakka/amqp/scaladsl/AmqpConnectorsSpec.scala
```diff
@@ -284,20 +286,22 @@ class AmqpConnectorsSpec extends AmqpSpec {
...
 
       val subscriber2 = TestSubscriber.probe[IncomingMessage]()
       amqpSource.addAttributes(Attributes.inputBuffer(1, 1)).runWith(Sink.fromSubscriber(subscriber2))
 
       subscriber2.ensureSubscription()
-      subscriber2.request(4)
-      subscriber2.expectNext().bytes.utf8String shouldEqual "two"
+      subscriber2.request(3)
+      // MapAsync execute the provided function as it pull the elements regardless of whether they are passed
+      // downstream or not. Thus the second element is acked even though it didn't get pushed down by `subscriber`.
```
johanandren: This was true (and racy) before as well (although maybe it did not manifest itself in this test), an element could be fetched from the MQ and ack:ed but have downstream cancel/fail before it was emitted.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
juanjoDiaz: Yup, I'll revert this.

amqp/src/test/scala/akka/stream/alpakka/amqp/scaladsl/AmqpConnectorsSpec.scala
juanjoDiaz: I commented that line to fix the test in akka 2.4 but broke on akka 2.5.

I think that the issue is that mapAsync use to run the function as soon as it pulls it regardless of whether it was emitted or not. So that test could never work with mapAsync.

However, it seems to work as expected in 2.5. The only change I can see in the history is this: https://github.com/akka/akka/commit/a40826e1cbbcdc6a222f7a468efa3c00305db1fe. Any idea??

juanjoDiaz: To be honest, we just renamed this source as `atMostOnceSource` which means precisely this. IF things go south in unexpected ways, you might miss the element.

If you want to be 100% sure that you don't miss any element you should use `committableSource` which could also be `atLeastOnceSource`

juanjoDiaz: I take back what I just said 2 comments above. 2.5 Didn't fix mapAsync 🙂 

I guess the test it's just 100% indeterministic since whether the element is acked or not depends completely o how fast the probe close the connection... Still open to suggestions :)

juanjoDiaz:
Hi @johanandren 

Just checking up so this doesn't get stale.

Is there anything left apart from the racy test?
What should we do about it?
My opinion is that the test is not predictable enough and should be remove (it's inadequate).

johanandren:
I think it looks good apart from the failing test. I think you are right, the name of the test case already hints about it being not quite true. Lets remove the test.

Falmarri:
@juanjoDiaz thanks for cleaning up my version. I didn't like the duplication that i introduced, but didn't have the time to really make it good. so thanks =D

juanjoDiaz:
No problem.
It's been a good learning experience! :)

dwickern:
Thanks for implementing this! The code snippets don't seem to work in the published docs: https://developer.lightbend.com/docs/alpakka/current/amqp.html#acknowledging-messages-downstream

juanjoDiaz:
Hi @dwickern ,

The fix is already committed (see #507) but I guess it hasn't been published yet.

EmilDafinov:
@juanjoDiaz quick question about this PR: shouldn't the`CommittableIncomingMessage` be generic with respect to the type of the `message` field? That way the object could be transformed as it goes through the pipeline, while retaining the ability to ack/nack at the end. Or am I missing something when it comes to the way this is supposed to be used? Thanks in advance :)

Issue #456, opened 2017-09-02
EIP - Splitter


mpricope:
Hi everyone,

I remember that back in the days when I was learning Camel, a very good and useful source was the Enterprise Integration Patterns page, describing how you implement each of these with Camel. 

I think this would be a very useful piece of documentation for people new to Akka Streams (like myself) :).
 
So I started implement a few patterns (some can be done in pure Akka Streams). Maybe we can update:
http://developer.lightbend.com/docs/alpakka/current/patterns.html

So here's the first one. EIP Splitter. With and without an aggregator:

```
    //A source that nees splitting
    val source: Source[String, NotUsed] = Source(List("1-2-3", "2-3", "3-4"))

    //Some Sink
    val printSink = Sink.foreach[Any](f => {
      println(f)
    })

    //Simple Split
    source.map(s => s.split("-").toList)
    .mapConcat(identity)
      .map(s => s.toInt)
      .runWith(printSink)


    //Split - Aggregate
    source.map(s => s.split("-").toList)
      .splitWhen(a => true)
      .mapConcat(identity)
      .map(s => s.toInt)
      .reduce((a, b) => a + b)
      .mergeSubstreams

      .runWith(printSink)
```

Let me know how this looks

johanandren:
Good idea! Create a PR, that makes the review interaction much smoother.

PR #487, opened 2017-09-22
Eip split

mpricope:
Sample EIP Splitter and Splitter + Aggregator

Issue #456


build.sbt
```diff
@@ -261,7 +249,6 @@ lazy val docs = project
     inConfig(Compile)(defaultParadoxSettings),
     ParadoxPlugin.paradoxSettings(Local),
     inConfig(Local)(defaultParadoxSettings),
-    paradoxGroups := Map("Language" -> Seq("Scala", "Java")),
     paradoxProperties in Local ++= Map(
       // point API doc links to locally generated API docs
       "scaladoc.akka.stream.alpakka.base_url" -> rebase(
```
johanandren: Lots of unrelated changes here that should not be in the PR

build.sbt
```diff
@@ -24,6 +24,7 @@ lazy val alpakka = project
     sns,
     sqs,
     sse,
+    eip,
```
johanandren: I think we should try to put the sample code under `doc/src/test/{java/scala}`  rather than introducing a new module just for samples

docs/src/main/paradox/patterns.md
```diff
@@ -3,5 +3,32 @@
 Many [Enterprise Integration Patterns](http://www.eaipatterns.com/toc.html) can be implemented with Akka Streams 
 (see [Java documentation](http://doc.akka.io/docs/akka/2.4/java/stream/index.html) or [Scala documentation](http://doc.akka.io/docs/akka/2.4/scala/stream/index.html)).
 
+
+# Splitter
+
+You can achieve a [Splitter](http://www.enterpriseintegrationpatterns.com/patterns/messaging/Sequencer.html) as described in  [EIP](http://www.enterpriseintegrationpatterns.com) using out of the box Akka Streams dsl.
+
+![Splitter](http://www.enterpriseintegrationpatterns.com/img/Sequencer.gif)
```
johanandren: We cannot hotlink images from other servers like this, we need to create our own image or ask permission to copy and host that image.

eip/src/test/java/akka/stream/alpakka/eip/javadsl/SplitterExamples.java
```diff
@@ -0,0 +1,80 @@
...
+    public void splitAggregate() throws ExecutionException, InterruptedException {
+        //#Aggregate-Split
+        //Sample Source
+        Source<String, NotUsed> source = Source.from(Arrays.asList(new String[]{"1-2-3", "2-3", "3-4"}));
+
+
+        CompletionStage<List<Integer>> ret =
+                source.map(s -> Arrays.asList(s.split("-")))
+                        //split all messages into sub-streams
+                        .splitWhen(a -> true)
```
johanandren: Note that this has a high overhead because of the materialization of a substream for each element. Unless the individual aggregates are really large or the processing done for the sub stream is very cpu intense just materialising will likely dominate execution. 

In many if not most cases just putting the split-process-aggregate of each element in a single map. I wonder if it wouldn't be better to do that, with a `mapAsync` to achieve some parallelism in processing instead.

build.sbt
mpricope: I though about this. Then I started thinking how I would implement some other patterns (like the Enricher: http://www.enterpriseintegrationpatterns.com/patterns/messaging/DataEnricher.html) and realized it wont probably work out of the box. (I actually started working on this but it's not ready yet). So I started a new module as I needed to put the classes somewhere. Suggestions are welcome. (I can put these in docs and see afterwards when I finish the Enricher). Your call

docs/src/main/paradox/patterns.md
mpricope: I actually checked this and that Image is kindly provided under creative commons. (together with the pattern description). Camel links directly to it the docs and I did the same :). Your call guys here if Creative Commons is enough for you and if not we need to find another solution.

eip/src/test/java/akka/stream/alpakka/eip/javadsl/SplitterExamples.java
mpricope: Hmm interesting point, I never looked from this perspective.  My real life examples for using this pattern almost always involved some serious processing in the sub-streams. Either some heavy string parsing or I/O. I/O in sub-process  is actually good case for this pattern as usually involves more parallelism in the sub-process which adds the need of the back-pressure. (I so wished Camel has some elegant solution for this ... but that's why I am here :) )

Adding parallelism in the sub-stream is something that I think is out of scope of this pattern. 

Of course if your data is small and you just split for a very simple aggregation (like I did in the example) this pattern is actually overkill. You better just put your code in a map (sync or async) as you suggested and that's it. But using this pattern only for this is a little bit of going over yourself :). My usage intention was mostly on the examples that I described above (and that is how I use it in real life).

build.sbt
mpricope: sorry ... a myopic merge :(

eip/src/test/java/akka/stream/alpakka/eip/javadsl/SplitterExamples.java
johanandren: Fair enough, maybe make sure that it is clear from the accompanying text that this is just for example and when it would make sense in real life?

build.sbt
johanandren: In Akka core that is how we do it when there is none of the actual modules that naturally would host the sample code, so I think we should follow that here.

mpricope: ok moved everything to docs. (Travis seems to fail randomly :( )

eip/src/test/java/akka/stream/alpakka/eip/javadsl/SplitterExamples.java
mpricope: reworked the texts

docs/src/main/paradox/patterns.md
```diff
@@ -12,21 +12,28 @@ You can achieve a [Splitter](http://www.enterpriseintegrationpatterns.com/patter
...
+
 Scala
-: @@snip (../../../../eip/src/test/scala/akka/stream/alpakka/eip/scaladsl/SplitterExamples.scala) { #Aggregate-Split }
+: @@snip (../../test/scala/akka/stream/alpakka/eip/scaladsl/SplitterExamples.scala) { #Aggregate-Split }
 
 Java
-: @@snip (../../../../eip/src/test/java/akka/stream/alpakka/eip/javadsl/SplitterExamples.java) { #Aggregate-Split }
+: @@snip (../../test/java/akka/stream/alpakka/eip/javadsl/SplitterExamples.java) { #Aggregate-Split }
+
+While in real life this solution if overkill for such a simple problem (you can just do everything in a map), more complex scenarios, involving in particular I/O, will benefit from the fact that you can paralelize sub-streams and get back-pressure for "free".
```
johanandren: 👍

johanandren:
I think the failure is some code formatting still (weird though that just one of the builds failed with it)

mpricope:
fixed automagically

raboof:
(logged the test failure as https://github.com/akka/alpakka/issues/491)

docs/src/main/paradox/patterns.md
raboof: I think that means we can include the image, but we don't want to link to an external http image because that will cause 'mixed content' errors in browsers when browsing the documentation over https (https://developer.lightbend.com/docs/alpakka/current/)

johanandren: In addition to using someone elses server bandwitdth. :)

mpricope: ok, I added the image locally

Issue #491, opened 2017-09-25
CsvParserSpec: should parse Apple Numbers exported file


raboof:
https://travis-ci.org/akka/alpakka/jobs/279258623

```
[info] - should parse Apple Numbers exported file *** FAILED *** (264 milliseconds)
[info]   A timeout occurred waiting for a future to complete. Queried 6 times, sleeping 15 milliseconds between each query. (CsvParsingSpec.scala:130)
[info]   org.scalatest.concurrent.Futures$FutureConcept$$anon$1:
[info]   at org.scalatest.concurrent.Futures$FutureConcept.tryTryAgain$1(Futures.scala:538)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValueImpl(Futures.scala:550)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValueImpl$(Futures.scala:479)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:275)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue(Futures.scala:476)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue$(Futures.scala:475)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:275)
[info]   at akka.stream.alpakka.csv.scaladsl.CsvParsingSpec.$anonfun$new$15(CsvParsingSpec.scala:130)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.WordSpecLike$$anon$1.apply(WordSpecLike.scala:1078)
[info]   at org.scalatest.TestSuite.withFixture(TestSuite.scala:196)
[info]   at org.scalatest.TestSuite.withFixture$(TestSuite.scala:195)
[info]   at akka.stream.alpakka.csv.scaladsl.CsvParsingSpec.withFixture(CsvParsingSpec.scala:21)
[...]

PR #492, opened 2017-09-25
Widen scalatest future timeout

raboof:
Refs #491

Because it got hit on travis


raboof:
Agreed that makes sense, what do you think?

PR #493, opened 2017-09-26
Bump paradox and use the new dependency directive

jonas:


johanandren:
I'm not 100% sure about this, since it's hosted over at developer.lightbend.com rather than on akka.io it should probably follow the styling over there?

jonas:
Does the Paradox update and refactoring still make sense?

johanandren:
I think the other changes make sense! Sorry, should have been a bit more clear that the concern was just about the akka-theme.

jonas:
@johanandren Thanks for the clarification. Fixed and rebased.

johanandren:
Thanks @jonas !

PR #496, opened 2017-09-27
Add possibility of passing routing key per message

### LINKED ISSUE #364: Setting routing key per message in AMQP integration
RabbitMQ supports setting a routing key per message, but the AMQP Sink currently uses the routing key created when defining the sink (https://github.com/akka/alpakka/blob/master/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala#L84). Would be nice to do it on a per message level.

juanjoDiaz:
Allow setting a routing Key per message and resolves #364

### DIFF
--- a/amqp/src/test/java/akka/stream/alpakka/amqp/javadsl/AmqpConnectorsTest.java
+++ b/amqp/src/test/java/akka/stream/alpakka/amqp/javadsl/AmqpConnectorsTest.java
@@ -30,6 +30,7 @@ import akka.stream.testkit.TestSubscriber;
 import akka.stream.testkit.javadsl.TestSink;
 import akka.testkit.JavaTestKit;
 import akka.util.ByteString;
+import scala.Option;
 import scala.Some;
 import scala.collection.JavaConverters;
 import scala.concurrent.duration.Duration;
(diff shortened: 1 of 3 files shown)


juanjoDiaz:
Hi @johanandren, not meaning to be annoying but just checking if there is any reason for this PR to be stalled :)

johanandren:
Just no time to review yet, so many things to do, so little time. Sorry!

amqp/src/test/java/akka/stream/alpakka/amqp/javadsl/AmqpConnectorsTest.java
```diff
@@ -130,7 +131,7 @@ public void publishAndConsumeRpc() throws Exception {
     );
 
     amqpSource.map(b ->
-        new OutgoingMessage(b.bytes().concat(ByteString.fromString("a")), false, false, Some.apply(b.properties()))
+        new OutgoingMessage(b.bytes().concat(ByteString.fromString("a")), false, false, Some.apply(b.properties()), Option.apply(null))
```
johanandren: Can we provide a Java API here? I can see that it was already missing but a constructor taking `Optional` instead of `Option` would be nice.

juanjoDiaz:
Java API added for Outgoing messages.

Issue #498, opened 2017-09-28
XML Writer based on ParseEvent


Korbik:
I think the XML Writer as a stream can be as useful as the parser.

The idea is to use the StAX Writer implementation base on ParseEvent from xml-parser.

ennru:
Implementation merged.
Thank you for your contribution, @Korbik

PR #499, opened 2017-09-28
XML Writer #498

Korbik:
Implementation of the serialization of the ParseEvent from xml-parser

### DIFF
--- a/xml/src/main/scala/akka/stream/alpakka/xml/Xml.scala
+++ b/xml/src/main/scala/akka/stream/alpakka/xml/Xml.scala
@@ -4,11 +4,13 @@
 
 package akka.stream.alpakka.xml
 
+import java.nio.charset.Charset
 import java.util.Optional
+import javax.xml.stream.XMLOutputFactory
 
 import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
 import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
-import akka.util.ByteString
+import akka.util.{ByteString, ByteStringBuilder}
 import com.fasterxml.aalto.stax.InputFactoryImpl
 import com.fasterxml.aalto.{AsyncByteArrayFeeder, AsyncXMLInputFactory, AsyncXMLStreamReader}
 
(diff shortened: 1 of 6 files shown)


xml/src/main/scala/akka/stream/alpakka/xml/Xml.scala
```diff
@@ -178,6 +179,63 @@ object Xml {
   /**
    * Internal API
    */
+  private[xml] class StreamingXmlWriter(strEncoding: String) extends GraphStage[FlowShape[ParseEvent, ByteString]] {
```
ennru: Have you considered using `Charset` instead of String. Could be a better choice even though `XMLOutputFactory` accepts charset names only.

Korbik: I think I had changed my mind because of the charset name for XMLOutputFactory but you are right it would be better.

xml/src/test/scala/akka/stream/alpakka/xml/scaladsl/XmlWritingTest.scala
```diff
@@ -0,0 +1,125 @@
...
+import scala.concurrent.duration._
+
+class XmlWritingTest extends WordSpec with Matchers with BeforeAndAfterAll {
+  implicit val system = ActorSystem("Test")
+  implicit val mat = ActorMaterializer()
+
+  // #writer
+  val writer: Sink[ParseEvent, Future[String]] = Flow[ParseEvent]
+    .via(XmlWriting.writer)
+    .map[String](bs => new String(bs.toArray, StandardCharsets.UTF_8))
```
ennru: You'd might want to use `.map[String](_.utf8String)` here.
(Same in Java code)

ennru:
Hej @Korbik,
I like your XML writer and its a good addition to Alpakka. E.g. filtering really large XML documents is a great Alpakka  XML parser / XML writer use case.
Any comments from @johanandren or @jrudolph ?
Cheers,
Enno.

Korbik:
@ennru Thanks a lot for your feedbacks. I have updated the MR with them.

Issue #500, opened 2017-09-28
Add support for Oracle Bare Metal Cloud Storage


mradityagoyal:
Oracle Bare Metal Storage Cloud is a new offering similar to amazon S3. 

Overview - https://cloud.oracle.com/en_US/infrastructure/storage

Existing sdk  - https://docs.us-phoenix-1.oraclecloud.com/Content/API/Concepts/sdkconfig.htm

rest api - https://docs.us-phoenix-1.oraclecloud.com/Content/API/Concepts/apiref.htm. 

I have been working on creating a alpakka style connector at - https://github.com/mradityagoyal/alpakka/tree/master/bmcs/src

would like to contribute to this project to add the new connector.

PR #501, opened 2017-09-28
Add support for Oracle Bare Cloud Storage #500

mradityagoyal:
* Adds BmcsClient (similar to the S3Client) for upload and download of files to oracle BMCS


mradityagoyal:
Will look into what i need to (learn &) fix.

project/Dependencies.scala
```diff
@@ -163,6 +163,13 @@ object Dependencies {
     )
   )
 
+  val Bmcs = Seq(
+    libraryDependencies ++= Seq(
+      "org.bouncycastle" % "bcpkix-jdk15on" % "1.52",
```
johanandren: Comment with license

bmcs/src/test/resources/reference.conf
```diff
@@ -0,0 +1,32 @@
...
+      parsing.illegal-header-warnings = off
+    }
+  }
+  stream {
+    materializer {
+      subscription-timeout {
+        mode = warn
+        timeout = 10s
+      }
+    }
```
johanandren: You cannot override defaults in a `reference.conf`, because then if they are applied or not are dependent on class path ordering, put them in an `application.conf` instead.

bmcs/src/test/resources/oci_api_key.pem
```diff
@@ -0,0 +1,30 @@
...
+ASC9qHM9pqZz6g9ZEVUCgNRBSB4TZGH/dLHmuqJgQ+5/V7dStIaPJdv1oUJatn8l
+0dfPW64foXHu9HQJof9S1j8qAzXK0VJCv/RDtYRzI8cEY9sTf7EnAku1rwkKIozu
+ul4iIZnPlNzERuLBP7bq3u4QvnaIp71v41wmD9CMi61s7nHhr9G/lUCZB1JEKErQ
+KyxmjRMPIWTnkfHvM+/WNdpLYOb009vRG9Sn6pW7jNfduMcm/kr6eiVkxrfZvgvI
+Nipm4qbcT1eYKpbY/veSHtnXfQgv+28AnWiU+XTKqN+tCNuqT0hRGl4kBEk6nEw1
+BbhsgtAlGweVnKJywDi0ppJJL3ksyOfxaz3oA/VWoROSjzVSidFkxfb6fhOr+bZ5
+ALyG14gY2t1VK1SWFNFQtQxEqtQt4To8RTfHFaIN77tA8oT7+QMlo+Seg1fyAOcX
+DUsEU0RrWLImYGY3tzOLBLZFnSWUzZ9FDmUCa5lSHEKixpWTrw3jIvhOhNpecbsG
+6H+ichO/tjGy0vtzC2zLs/hF6tChvQbwZKkiEjwFFe2j4wUwRRE2KFQhw+ix2E14
+-----END RSA PRIVATE KEY-----
```
johanandren: This is just a dummy key, right?

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/auth/BmcsSigner.scala
```diff
@@ -0,0 +1,145 @@
...
+import java.util.Base64
+
+import akka.http.scaladsl.model.headers._
+import akka.http.scaladsl.model._
+import akka.stream.Materializer
+import akka.stream.alpakka.oracle.bmcs.BmcsException
+
+import scala.concurrent.Future
+
+object BmcsSigner {
```
johanandren: This isn't meant to be public api, right?

```
/**
 * INTERNAL API
 */
@InternalApi private[bmcs] BmcsSigner { ... 
```

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/auth/BmcsSigner.scala
```diff
@@ -0,0 +1,145 @@
...
+    val dateHdr = Date(DateTime(date.toEpochMilli))
+    request.method match {
+      case HttpMethods.GET | HttpMethods.DELETE | HttpMethods.PUT | HttpMethods.HEAD =>
+        Future.successful(Seq(dateHdr)) //for GET DELETE PUT and HEAD only Date header is added.
+      case HttpMethods.POST =>
+        request.entity.dataBytes
+          .runWith(digest())
+          .map(hash => encodeBase64(hash.toArray))
+          .map(bh => RawHeader(xContentSha256, bh))
+          .map(hd => Seq(hd, dateHdr))
```
johanandren: Does this mean there is an upload method that will have its stream materialized twice? That should be clearly documented on the user API if so, usually when you pass Source[ByteString, _] to some API you likely expect it to materialize that stream only once.

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/impl/BmcsStream.scala
```diff
@@ -0,0 +1,302 @@
...
+import akka.http.scaladsl.Http
+import akka.http.scaladsl.marshalling.Marshal
+import akka.http.scaladsl.model.headers.ByteRange
+import akka.http.scaladsl.model.{RequestEntity, _}
+import akka.http.scaladsl.unmarshalling.{Unmarshal, Unmarshaller}
+import akka.stream.alpakka.oracle.bmcs.{BmcsException, BmcsSettings, DiskBufferType, MemoryBufferType}
+import akka.stream.Materializer
+import akka.stream.alpakka.oracle.bmcs.auth.{BmcsCredentials, BmcsSigner}
+import akka.stream.alpakka.oracle.bmcs.scaladsl.ListObjectsResultContents
+import akka.stream.alpakka.s3.impl.{DiskBuffer, MemoryBuffer, SplitAfterSize}
```
johanandren: This isn't great, having one module depend on internals of another, but maybe it's worth it for this module.

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/impl/BmcsStream.scala
```diff
@@ -0,0 +1,302 @@
...
+  }
+
+  /**
+   *
+   * Returns a source of ByteString of the object in bmcs.
+   *
+   * @param bucket     the bucket
+   * @param objectName the object to download
+   * @param range      optional range of bytes.
+   * @return source built from the object in storage.
```
johanandren: "A source emitting the bytes of the requested object"?

Also good to document:
* Does the request happen when the returned source is materialized or right away?
* Can there returned source be materialized multiple times, 
* What happens on failure?

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/impl/BmcsStream.scala
```diff
@@ -0,0 +1,302 @@
...
+    })
+  }
+
+  /**
+   * Uploads a stream of ByteStrings to a specified location as a multipart upload.
+   */
+  def multipartUpload(bucket: String,
+                      objectName: String,
+                      chunkSize: Int = MinChunkSize,
+                      chunkingParallelism: Int = 4): Sink[ByteString, Future[CompleteMultipartUploadResult]] =
```
johanandren: Please document `chunkingParallelism` a bit here

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/impl/BmcsStream.scala
```diff
@@ -0,0 +1,302 @@
...
+  ): Flow[ByteString, PartUploadResponse, NotUsed] = {
+
+    // Multipart upload requests (except for the completion api) are created here.
+    //  The initial upload request gets executed within this function as well.
+    //  The individual upload part requests are created.
+    val requestFlow: Flow[ByteString, (HttpRequest, (MultipartUpload, Int)), NotUsed] =
+      createMultipartRequests(bucket, objectName, chunkSize, parallelism)
+
+    // The individual upload part requests are processed here
+    requestFlow.via(Http().superPool[(MultipartUpload, Int)]()).map {
```
johanandren: Is creating a new superpool for every `chunkAndRequest` call really the right thing to do?

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/impl/BmcsStream.scala
```diff
@@ -0,0 +1,302 @@
...
+      case (Success(r), (upload, index)) =>
+        r.entity.dataBytes.runWith(Sink.ignore)
+        val etag = r.headers.find(_.lowercaseName() == "etag").map(_.value)
+        etag
+          .map((t) => SuccessfulPartUpload(upload, index, t))
+          .getOrElse(FailedPartUpload(upload, index, BmcsException(r, "Cannot find etag")))
+
+      case (Failure(e), (upload, index)) => FailedPartUpload(upload, index, e)
```
johanandren: If one part fails to upload, should it really continue with the rest?

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/BmcsSettings.scala
```diff
@@ -0,0 +1,74 @@
...
+}
+
+object BmcsSettings {
+
+  /**
+   * Scala API: Creates [[BmcsSettings]] from the [[com.typesafe.config.Config]] attached to an [[akka.actor.ActorSystem]].
+   */
+  def apply()(implicit system: ActorSystem): BmcsSettings = apply(system.settings.config)
+
+  def apply(config: Config): BmcsSettings = {
```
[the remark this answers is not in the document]
mradityagoyal: Updated to not look for disk path. and added that to diskBufferType

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/auth/BmcsCredentials.scala
```diff
@@ -0,0 +1,70 @@
...
+        case pemEncryptedPair: PEMEncryptedKeyPair =>
+          val password = passphrase match {
+            case None => throw new RuntimeException(" This file needs a passphrase")
+            case Some(x) => x
+          }
+          val decProv = new JcePEMDecryptorProviderBuilder().build(password.toCharArray)
+          pemEncryptedPair.decryptKeyPair(decProv).getPrivateKeyInfo
+        case privateKeyInfo: PrivateKeyInfo => privateKeyInfo
+        case pemKeyPair: PEMKeyPair => pemKeyPair.getPrivateKeyInfo
+        case pubicKey: SubjectPublicKeyInfo =>
```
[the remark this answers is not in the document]
mradityagoyal: fixed and pushed.

bmcs/src/test/resources/oci_api_key.pem
mradityagoyal: Yes. this was used for testing. no longer active.

project/Dependencies.scala
mradityagoyal: MIT license. 
https://www.bouncycastle.org/licence.html

bmcs/src/test/resources/reference.conf
mradityagoyal: fixed and pushed.

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/auth/BmcsSigner.scala
mradityagoyal: I am not sure if we should mark this as internal api or internal use only. The signer can be used for signing any request for oracle BMCS. Might be usable in other oracle cloud services in future. 
I plan to add some utility methods to create bucket / delete objects.. so that this is a stand alone solution for most of the common tasks, which will also use this signer. 
I want to use this signer in those utility methods.. without having to add those to the connector?
What are your thoughts, do you think this belongs as internal API?

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/impl/BmcsStream.scala
mradityagoyal: yes. I agree. let's not repeat ourselves.

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/impl/BmcsStream.scala
mradityagoyal: I am not really sure. In theory, we can use the same pool for all the requests for all the uploads and downloads to the same region.. ?

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/impl/BmcsStream.scala
mradityagoyal: Personally, i would want to retry in this situation. But re-try will depend on the max materializations we have configured initially. I didn't implement the retry, cause i didn't want to leave the possibility of infinite materializations to get the hash and then fail. 

Instead of collecting failures in the sink stage.. what else can i do? can i just throw an exception?

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/auth/BmcsSigner.scala
mradityagoyal: documented in the documentation of signedRequest method.

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/impl/BmcsStream.scala
mradityagoyal: updated comment.

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/impl/BmcsStream.scala
mradityagoyal: updated comments.

project/Dependencies.scala
johanandren: As in add a comment to the right of dependency saying that, thanks :)

project/Dependencies.scala
```diff
@@ -163,6 +163,13 @@ object Dependencies {
...
+      "com.typesafe.akka" %% "akka-http-spray-json" % "10.0.10"
```
johanandren: There's a constant for Akka HTTP version to use instead of string here

bmcs/src/test/scala/akka/stream/alpakka/oracle/bmcs/scaladsl/BmcsNoMock.scala
```diff
@@ -0,0 +1,83 @@
...
+  implicit val defaultPatience: PatienceConfig =
+    PatienceConfig(timeout = Span(5, Seconds), interval = Span(30, Millis))
+
+  val userOcid = "ocid1.user.oc1..aaaaaaaaalwxriuznfhohggk7ejii6lpwo7mebuldxh455hiesnowaoaksyq"
+  val keyFingerprint = "cb:17:e0:45:d0:24:d3:ff:be:ee:1b:0e:f8:2c:58:27"
+  val tenancyOcid = "ocid1.tenancy.oc1..aaaaaaaa6gtmn46bketftho3sqcgrlvdfsenqemqy3urkbthlpkos54a6wsa"
+  val keyPath = "./bmcs/src/test/resources/oci_api_key.pem"
+  val passphrase: Option[String] = Some("adityag")
+  val cred = BasicCredentials(userOcid, tenancyOcid, keyPath, passphrase, keyFingerprint)
+  val bucket = "CEGBU_Prime"
```
johanandren: These are just dummy placeholders now, right?

bmcs/src/test/resources/oci_api_key.pem
johanandren: Please delete from the PR then

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/scaladsl/BmcsClient.scala
```diff
@@ -0,0 +1,89 @@
...
+  val MinChunkSize: Int = 5242880
+
+  def apply(cred: BmcsCredentials)(implicit system: ActorSystem, mat: Materializer): BmcsClient =
+    new BmcsClient(BmcsSettings(ConfigFactory.load()), cred)
+
+  def apply(cred: BmcsCredentials, settings: BmcsSettings)(implicit system: ActorSystem,
+                                                           mat: Materializer): BmcsClient =
+    new BmcsClient(settings, cred)
+
+}
```
johanandren: Java API convention is to provide `create()` methods as well.

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/javadsl/BmcsClient.scala
```diff
@@ -0,0 +1,100 @@
...
+                                  scalaContents.size,
+                                  scalaContents.md5,
+                                  scalaContents.timeCreated)
+      }
+      .asJava
+
+  def multipartUpload(bucket: String,
+                      objectName: String,
+                      chunkSize: Int = BmcsClient.MinChunkSize,
+                      chunkingParallelism: Int = 4): Sink[ByteString, CompletionStage[MultipartUploadResult]] =
```
johanandren: Java does not understand default parameters, so provide overloads instead.

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/javadsl/BmcsClient.scala
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /**
+   * Will return a source of object metadata for a given bucket with optional prefix.
+   * This will automatically page through all keys with the given parameters.
+   *
+   * @param prefix Prefix of the keys you want to list under passed bucket
+   * @return Source of object metadata
+   */
+  def listBucket(bucket: String, prefix: Option[String] = None): Source[ListObjectsResultContents, NotUsed] =
```
johanandren: `Option` is Scala, should be `Optional` from Java

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/javadsl/BmcsClient.scala
```diff
@@ -0,0 +1,100 @@
...
+  val MinChunkSize: Int = 5242880
+
+  def create(credentials: BmcsCredentials, system: ActorSystem, mat: Materializer): BmcsClient =
+    BmcsClient.create(BmcsSettings(ConfigFactory.load()), credentials, system, mat)
+
+  def create(settings: BmcsSettings, cred: BmcsCredentials, system: ActorSystem, mat: Materializer): BmcsClient =
+    new BmcsClient(settings, cred, system, mat)
+}
+
+final class BmcsClient(val settings: BmcsSettings, val cred: BmcsCredentials, system: ActorSystem, mat: Materializer) {
```
johanandren: Should really `settings` and `cred` be public?

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/javadsl/BmcsClient.scala
```diff
@@ -0,0 +1,100 @@
...
+
+final case class ListObjectsResultContents(
+    /** the name of the object in bmcs **/
+    name: String,
+    /** the name of the bucket in which this object is stored. **/
+    bucket: String,
+    /** the size of the object,  in bytes **/
+    size: Option[Long],
+    /** Base64 encoded MD5 hash of the object Data.  **/
+    md5: Option[String],
```
johanandren: `Optional` for these two if this is the `javadsl`

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/javadsl/BmcsClient.scala
```diff
@@ -0,0 +1,100 @@
...
+import akka.http.scaladsl.model.headers.{ByteRange => ScalaByteRange}
+import akka.stream.Materializer
+import akka.stream.alpakka.oracle.bmcs.BmcsSettings
+import akka.stream.alpakka.oracle.bmcs.auth.BmcsCredentials
+import akka.stream.alpakka.oracle.bmcs.impl.{BmcsStream, CompleteMultipartUploadResult, ObjectSummary}
+import akka.stream.javadsl.{Sink, Source}
+import akka.util.ByteString
+import com.typesafe.config.ConfigFactory
+
+final case class MultipartUploadResult(bucketName: String, objectName: String, etag: String)
```
johanandren: Often we have put model classes/settings that are shared between java and scala in a single package one level up, which would be `akka.stream.alpakka.oracle.bmcs` in this case. For those we have to make sure to provide both Scala and Java style constructors, methods and types used.

For types that are just returned it can make sense to represent the return type with a trait/interface instead as having a case class gives a lot of methods that then becomes public api.

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/impl/HttpRequests.scala
```diff
@@ -0,0 +1,87 @@
...
+import akka.http.scaladsl.model._
+import akka.http.scaladsl.model.Uri.{Authority, Query}
+import akka.http.scaladsl.model.headers.Host
+import akka.stream.alpakka.oracle.bmcs.BmcsSettings
+import akka.stream.scaladsl.Source
+import akka.util.ByteString
+
+import scala.collection.immutable.Seq
+
+object HttpRequests {
```
johanandren:
```
/** 
 * Internal API: 
 */
@InternalApi private[bmcs]
```

On all the things in this package

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/impl/BmcsStream.scala
johanandren: I think it should fail the upload and let the user handle retry. But maybe that is problematic and will leave garbage on bmcs?

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/impl/BmcsStream.scala
johanandren: How does the AWS S3 connector handle this? I think that has seen the most work/thought go into it so following along how that works is probably a good idea.

bmcs/src/main/scala/akka/stream/alpakka/oracle/bmcs/BmcsSettings.scala
```diff
@@ -0,0 +1,99 @@
...
+    apply(config).copy(region = region, namespace = namespace, bufferType = bufferType)
+
+  /**
+   * Scala API: Creates [[BmcsSettings]] from the [[com.typesafe.config.Config]] attached to an [[akka.actor.ActorSystem]].
+   */
+  def apply()(implicit system: ActorSystem): BmcsSettings = apply(system.settings.config)
+
+  def apply(config: Config): BmcsSettings = {
+
+    val bufferType = config.getString("akka.stream.alpakka.bmcs.buffer") match {
```
johanandren: this is the path that should not be full, it should look for `buffer`, `disk-buffer-path` etc. here  and extract the full path in `apply()(system)`. 

That makes it possible to have a custom config that does not have to live under the full path that you pass to this apply while still using the full path as default when you do not pass the config but use the implicit actor system factory method above.

Also, the config should not be inside the `akka` namespace as this is not Akka core (even if you saw some other connectors may be doing this). I think a good scope for the config would be `alpakka.bmcs`

raboof:
Looks like this PR gained some merge conflicts since, @mradityagoyal would you be interested in looking at @johanandren's review comments and fixing the merge conflicts?

mradityagoyal:
@raboof yes. i will look at the merge conflicts and also some of the remaining edits that @johanandren  had suggested.

mradityagoyal:
I lost access to oracle cloud storage when i switched jobs. I have some fixes for the merge conflicts and the comments in the Pull Request,  but can't run integration tests anymore to validate if everything works OK. 

I will try and see if I can get some temporary access to oracle cloud storage for evaluation purposes.

ennru:

PR #502, opened 2017-09-29
csv does not play well with framing

Voltir:
This test currently fails (hits the future timeout of the test suite) - without that timeout, this could otherwise consume all available memory, which seems not great.

### DIFF
--- a/csv/src/test/scala/akka/stream/alpakka/csv/scaladsl/CsvParsingSpec.scala
+++ b/csv/src/test/scala/akka/stream/alpakka/csv/scaladsl/CsvParsingSpec.scala
@@ -200,5 +200,17 @@ class CsvParsingSpec
         )
       )
     }
+
+    "compose well with other parts of the akka-streams ecosystem" in {
+      val fut = Source
+        .repeat(ByteString("a,b\n"))
+        .via(akka.stream.scaladsl.Framing.delimiter(ByteString("\n"), maximumFrameLength = 4096))
+        .via(CsvParsing.lineScanner())
+        .map(_.map(_.utf8String))
+        .take(1)
+        .runWith(Sink.seq)
+      val res = fut.futureValue
+      res(0) shouldBe List("a", "b")
+    }
   }
 }


lightbend-cla-validator:
Hi @Voltir,

ennru:
The CSV parser expects line breaks to be able to detect lines, it does the frameing itself. It does not require a line of CSV data to be contained in a single `ByteString`, simple frame detection with `Frameing.delimiter` is not valid for CSV as a column may contain line breaks (as long as it is quoted).

A worst case handle as #509 proposes could help.

raboof:
@Voltir I agree composition and separation of concerns are useful goals, however as @ennru correctly mentions above the current `Framing` implementation is not suitable for CSV data: CSV data might contain newlines inside the data cells, so naive `Framing` on newlines will split CSV lines where it shouldn't. It would indeed be interesting to see if it is possible to come up with a `Framing` that works for CSV without being CSV-specific.

As for when to decode the `ByteString`s into `String`s (as you mentioned in #509), this is a tricky trade-off:  the current approach, which postpones the decoding to later steps, is restrictive in the sense that it restricts the delimiters to single-byte characters and the supported encodings to encodings that keep those single-byte characters intact (such as latin1 and unicode). The advantage is when later in the pipeline you discard some lines or columns you haven't wasted time and memory decoding fields. It might make sense to support both, though I'm not sure that would be possible without too much duplication.

Do you have particular use cases where you're running into the limitations of the current model?

Would you be up for doing some experiments?

Voltir:
Right - I forgot about the end of line character possibly showing up in quoted text - so dealing with framing is a reasonable approach. I do think there is an alternative solution though: If we had something that was akin to `Flow[String,List[String]]`, we could check the state of the parser when it reaches the end of the input string - if its not "qouted", consider the line done and emit the list, if it is quoted, treat it as though the upstream framer split incorrectly, so pull the next string, appending both the end of line character and whatever else is remaining in the next string -- so possible, but something that would probably have to be documented well.

With regards to encoding, the current approach is morally equivalent to having a `Source[Byte]` as the input - parsing is done byte by byte, with all the encoding shenanigans that that entails. I think doing decoding before "csv-ing" would actually remove a lot of confusion and complexity. It would seem that that would require something along the lines of a transformation from a `Source[Byte]` to a `Source[Char]`, which I guess would be a stateful graph-stage that accumulated bytes until a representation of a Char was collected. Ie lets pretend we wanted to parse utf8, where up to 4 bytes can represent a single codepoint/glyph -- the proposed graph stage would pull bytes until a valid utf8 value was pulled, then emit the Char representation of it. This would move decoding into its own concern, and could be useful / composed with other flows, and would allow the csv code to only care about dealing with proper, in memory, `String`s.

The framing variant of csv parsing could then simply require a `Souce[Char]`, and the hacky row based one (which may or may not actually be useful, I concede) could still be done as the `Flow[String,List[String]]` proposed above.

ennru:
To support charsets properly we'd need to build something based on [`CharsetDecoder`](https://docs.oracle.com/javase/8/docs/api/java/nio/charset/CharsetDecoder.html) to provide a `Flow[ByteString, String]` or maybe a `Flow[ByteString, CharString]` where `CharString` would be some immutable wrapper of `nio.CharBuffer`.

ennru:
I fixed #509, but this example is still extreme as it sends very small elements through the framing to the CSV parser where they aggreate to a large `ByteString`. By default the parser fails the flow when it can't find a line break within 10 k bytes.
I can think of some optimisations in `CsvParser`, but I'd say this test case is nothing we should merge. 
Are you OK with that @Voltir ?

raboof:
I think there's some interesting ideas here, so if anyone wants to dig in and perform some experiments that could be interesting, but I guess it's time to close this PR itself as there's no specific clear problem or way forward here.

PR #504, opened 2017-09-30
Add an option to retry partial failed messages in ElasticsearchFlow

takezoe:
Current ElasticsearchFlow (and sink) has the ability to retry when a bulk request is failed entirely.

But Elasticsearch's bulk request might be partially for some reason. For instance, some shards in the cluster are inaccessible temporary. In this case, some messages in a bulk request are indexed, but some messages are responded as failure.

This pull request adds an option to retry these partial failed messages. This feature is enabled in default.


takezoe:
Reproduce the situation is difficult in a unit test, but I will try.

takezoe:
@raboof Added a unit test and fixed some problems I found during writing a unit test.

johanandren:
Fixes #558

PR #506, opened 2017-10-01
Amqp reusable connections

### LINKED ISSUE #420: Reusing connections in AMQP
Every `AmqpSource` or `AmqpSink` opens a new connection. It would be preferable if the API gave a possibility to reuse connections and create new channels on it instead.

juanjoDiaz:
Add Reusable Connections and closes #420


amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionSettings.scala
```diff
@@ -0,0 +1,231 @@
+/*
```
johanandren: Good that you moved it to a separate file.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionSettings.scala
```diff
@@ -0,0 +1,231 @@
...
+package akka.stream.alpakka.amqp
+
+import java.util.concurrent.atomic.AtomicInteger
+
+import com.rabbitmq.client.{Address, Connection, ConnectionFactory, ExceptionHandler}
+
+/**
+ * Only for internal implementations
+ */
+sealed trait AmqpConnectionSettings {
```
johanandren: It's rather a `...Provider` than a `...Settings` and then maybe the provider should be a field of settings rather than the other way around. I think that makes more sense than the current delegating model, "the settings used to create a stage can have a connection provider" vs "the settings is a provider that also has the actual settings".
That way we can have a default `object DefaultConnectionProvider` that creates a new connection every time `getConnection` is called.

Probably needs a step back though, and see if you can keep the current API for the default case, and have a way to have `Provider`s that makes sense. 

I'd put the `release` method in this trait and not have the stages know about the different implementations of acting on that. The stages never does a forced release, right? So make that a separate method in the provider where it makes sense (the cached one afaics)

Also, use the `@DoNotInherit` annotation on sealed traits.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionSettings.scala
```diff
@@ -0,0 +1,231 @@
...
+  def getConnection: Connection
+}
+
+final case class ReusableAmqpConnectionSettings(settings: AmqpConnectionSettings) extends AmqpConnectionSettings {
```
johanandren: I'd make this take a `Connection`, not have any vars and call it `ExternalConnectionProvider` or something along those lines. And then have it do nothing on `release` (because we just moved that responsibility to the user)

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionSettings.scala
```diff
@@ -0,0 +1,231 @@
...
+  def releaseConnection() = cachedConnection match {
+    case Some(connection) => {
+      if (connection.isOpen) connection.close()
+      cachedConnection = None
+    }
+    case None => Unit
+  }
+}
+
+final case class ReusableAmqpConnectionSettingsWithAutomaticRelease(settings: AmqpConnectionSettings)
```
johanandren: And then make this a `CachedConnectionProvider`

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionSettings.scala
```diff
@@ -0,0 +1,231 @@
...
+    extends AmqpConnectionSettings {
+  private val clients = new AtomicInteger(0)
+  private var cachedConnection: Option[Connection] = None
```
johanandren: This cached connection will be used concurrently by multiple threads so this is not safe. It needs to be thread safe, but not blocking, and we need to ensure only one ongoing connect is happening, so that two concurrent `getConnection` calls does not lead to two separate connections. The atomic counter counter is good thinking though.

One option to work around it would be to just let the user provide a `Connection` instead, and move the responsibility of doing it safely away from here.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionSettings.scala
juanjoDiaz: Yeah. It's a provider. I didn't want to change the name to don't break API compatibility.

The current `DefaultAmqpConnection` work exactly as you suggested `DefaultConnectionProvider `.

One option is to create all this as Providers and keep the Settings for backward compatibility marked as deprecated. Would that be ok?

The only reason I needed the stage to know about the provider type is so `ReusableConnection` instances are not closed. `ReusableAmqpConnectionWithAutomaticRelease` automatically take care of not closing if there are clients based on the atomic counter. But the `ReusableConnection` doens't do that so not sure of how to workaround that. Maybe we shoudl only have the type with automatic release and remove the other??

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionSettings.scala
juanjoDiaz: The whole idea of taking the setting is that you can release it and then reuse in a different stream and it will create another connection.

If you pass a connection and that connection is closed then the instance becomes unusable and unrecoverable.

Don't you think that's a valid case?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionSettings.scala
juanjoDiaz: The previous one is also cached. the only different is that this one release the connection when the last client disconnects and then open a new one if you reuse it again.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionSettings.scala
juanjoDiaz: True.

I'd rather the user providing the settings instead of the connection for the reasons stated on my previous comment (resiliency and reusability).

Would make the methods `getConnection` and `releaseConnection`synchronized on `this` help with this??

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionSettings.scala
johanandren: Not sure about the design one, requires further thinking. Would be nice if we didn't have to deprecate/force users to rewrite their code or get a warning, but if this could fit in without requiring that.

About the closing, the concrete implementation knows if a `release` should lead to a close or not, the stage just knows it needs to inform the provider about the release.

I think we should start out with one default which behaves like before, IE - new connection for every materialization based on the settings which gets closed on release, and then one that accepts whatever external connection and doesn't do anything on release. Even the reusable-cache-one we can maybe wait with, because it will be tricky to get right, and the requested feature is to allow sharing a connection and with an externally provided one we have achieved that.

Then maybe we can add the cache-magic-one in a separate PR to scope this one down a bit?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionSettings.scala
johanandren: No, no synchronized in async-land thanks :)

This is doable without resorting to blocking, just a bit tricky.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionSettings.scala
juanjoDiaz: Sorry for the radio silence. I don't have that much time lately.

I've been giving this some more thought to the current issues.

### Backward compatibility
The current implementation is 100% backwards compatible. Adding the `getConnection` method to the settings is just and implementation detail (since the settings are sealed/final). The settings vs provider thing is just a naming thing. I could rename the new ones to be `ReusableAmqpConnectionProvider` and leave the old one as `AmqpConnectionSettings` and even add an alias `AmqpConnectionProvider` for consistency if you think that's better.
So the options are:
* Leave it as it is. It works and it's backwards compatible but the name is maybe not the best.
* Create providers and pass them to the settings. I would work I think that's a bit counterintuitive to pass a factory to the config instead of the config to the factory.

Which one would you prefer?

### Stages not being aware of the provider
I don't think this is possible:
* For non-reused connection, every time you get a connection you get a new one. So when calling release, it's impossible to know which one to close. I can't think of a workaround for this.
* For the reusable connection without automatic release. There is no way to know if the release method is called by a stage (so ti should do nothing) or the user (so it should close the connection). A workaround for this is to offer only the connection with automatic closing.

Can you think of a good solution for this??

### Not using var and making the reusable connection thread-safe
The only solution I can think for this without blocking is to create the connection during the instantiation instead of when calling getConnection. The limitation here is that once the connection is closed the provider becomes unusable (which might be just fine). 
What do you think??

johanandren: Settings config/backwards compatibility: As I see it `ConnectorSettings`/`SourceSettings` is the big picture, that contains both details about how to connect and possibly settings that are more for the stage itself, while `AmqpConnectionSettings` is auth details, host/port etc, something you would feed to a `ConnectionProvider` to create a connection. 

Given that I think it makes most sense to have a design like this:

```
trait AmqpConnectorSettings {
  def connectionSettings: AmqpConnectionSettings
  def declarations: Seq[Declaration]
  def connectionProvider: ConnectionProvider
}

trait ConnectionProvider {
  def newConnection(s: AmqpConnectionSettings): Connection
  def release(c: Connection): Unit
}
```
I think this would allow us to keep the settings API compatible.

If the connection provider impl does anything in release or not is up to that stage, this makes it possible to have the stage only know about the trait and not the impl details. 

* the default one would always create a new connection, and close on release
* the cached would do some form of reference counting and only close on last release
* an external provider would just return the given connection and not do anything in release etc. and the user can do what they want with the connection, close it for example, since they provided it

If there is a specific provider we can think of where it makes sense to be able to close from the side (i'm not quite sure this is a good idea, closing the connection from under a running stage, canceling/completing the stream would be a better way to do that) that would be an implementation specific `close()` method. 

WDYT?

Let's start out with that bit and look at the concurrency issues after the design is in place.

juanjoDiaz: I like the approach.

However, maybe I misunderstood but I don't think that your approach is backward compatible since you will need to pass to the stage a `AmqpConnectorSettings` instead of a `AmqpConnectionSettings`.

Not so related questions, what are the `declarations` in the `AmqpConnectorSettings`??

johanandren: `AmqpSourceSettings` and `AmqpSinkSettings` which the stages take as a parameter are in fact already `AmqpConnectorSettings` so no problem there as far as I can see.

The declarations are something that creates queues/bindings/exchanges i RabbitMQ if they don't exist then the stages are materialized.

juanjoDiaz: I gave this another spin.
Your proposal 
```
trait ConnectionProvider {
  def newConnection(s: AmqpConnectionSettings): Connection
  def release(c: Connection): Unit
}
```
have the issue that passing the `AmqpConnectionSettings` to the newConnection method makes it impossible to cache the connection without using var and having the issue of not being thread-safe.
Also, it's a bit weird to pass the connection to the release method. What if the connection was generated by a different provider? Do we need to maintain a list of connections generated by a provider? This is not a problem for the DefaultProvider but might be for the cached one.

Any suggestion?

johanandren: I don't think it is a problem to require a connection provider that plans to cache/share the connections it returns to be thread safe. I'll help you out with that part when the rest is in place.

We are the only consumers of this API and those methods should be `private[amqp]`, being the only consumer we would never use two different providers in the same stage, so the create -> use -> release part I think isn't a problem (even if it would be in the general case).

juanjoDiaz:
@johanandren I've given this another shot and I really can't figure out how to implement your proposed approach.

I don't think that the proposed solution works.
```
sealed trait AmqpConnectorSettings {
  def connectionSettings: AmqpConnectionSettings
  def connectionProvider: AmqpConnectionProvider = DefaultAmqpConnectionProvider
  def declarations: Seq[Declaration]
}

[....]

trait ConnectionProvider {
  def newConnection(s: AmqpConnectionSettings): Connection
  def release(c: Connection): Unit
}
```
Maybe I didn't explained it very well before but for this to work we need to pass the provided already configured (we can't pass the settings and the provider separately in the `AmqpConnectorSettings`). Otherwise you can't get the same connection passing the same provider to two different connectors since the provider has to be separately configured for both (just imagine that you pass a provider that caches the connection to 2 stages with 2 different settings. It doesn't make any sense).
That's why I made the `connectionSettings` into a provider. So it's passed already configured.

Any suggestion? I'd really like to get this out in the next release if possible 🙂

juanjoDiaz:
No problem!

I agree. Currently our settings object is just really bloated.We should have separate settings for:
* The connections (host, port, auth, etc.)
* Amqp (declarations, subscriptions, etc.)

Could you defined what you would consider a good way?
For me it would be separating the settings as above and pass to the connector an already configured provided and all the AMQP configuration.
Is there anything else I'm missing here?

johanandren:
Sounds right to me.

johanandren:
One thought, should we perhaps close this PR and start a new one rather than piling up on top of the long discussions we already had here? I think that could make it a bit easier to follow (we can always link here from the PR description)

PR #508, opened 2017-10-02
the Frankfurt region has a different url

martinpallmann:
see http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region


martinpallmann:
forget it. I think I was too fast here.

PR #515, opened 2017-10-08
MQTT manual acks

### LINKED ISSUE #378: MQTT manual acknowledge
Hi there,

Is there a way to manually acknowledge messages ? I'm using the Java library. Paho supports it by using `void setManualAcks(boolean manualAcks)` of `IMqttAsyncClient` and then acknowledging each message using `void messageArrivedComplete(int messageId, int qos)`.

Or is it so that the message isn't acknowledged before it reaches the sink of the stream ? I don't think that's the case from what I understood of the code. According to Paho's documentation and Alpakka source code, a message is acknowledge as soon as `onMessage(message: MqttMessage): Unit` returns. Here's `onMessage` implementation:
```
private val onMessage = getAsyncCallback[MqttMessage] { message =>
    require(queue.size <= bufferSize)
    if (isAvailable(out)) {
      pushMessage(message)
    } else {
      queue.enqueue(message)
    }
  }
```
To me it looks like this method returns right away but I'm not sure.

Basically I don't want to acknowledge a message before it has reached my sink.

juanjoDiaz:
Add possibility to manually ack messages downstream

Resolves #378

### DIFF
--- a/docs/src/main/paradox/mqtt.md
+++ b/docs/src/main/paradox/mqtt.md
@@ -40,8 +40,8 @@ Java
 
 This source has a materialized value (@scaladoc[Future](scala.concurrent.Future) in Scala API and @extref[CompletionStage](java-api:java/util/concurrent/CompletionStage) in Java API) which is completed when the subscription to the MQTT broker has been completed.
 
-MQTT automatically acknowledges messages back to the server once they are passed downstream. The atLeastOnce source allow users to acknowledge the messages anywhere downstream.
-Please note that for manual acks to work CleanSession` should be set to false and MqttQoS should be AtLeastOnce.
+MQTT automatically acknowledges messages back to the server once they are passed downstream. The `atLeastOnce` source allow users to acknowledge the messages anywhere downstream.
+Please note that for manual acks to work `CleanSession` should be set to false and `MqttQoS` should be `AtLeastOnce`.
 
 Scala
 : @@snip (../../../../mqtt/src/test/scala/akka/stream/alpakka/mqtt/scaladsl/MqttSourceSpec.scala) { #create-source-with-manualacks }
@@ -49,7 +49,7 @@ Scala
 Java
 : @@snip (../../../../mqtt/src/test/java/akka/stream/alpakka/mqtt/javadsl/MqttSourceTest.java) { #create-source-with-manualacks }
 
-The atLeastOnce source returns CommittableMessage so you can acknowledge them by calling `messageArrivedComplete`.
+The `atLeastOnce` source returns @scaladoc[MqttCommittableMessage](akka.stream.alpakka.mqtt.scaladsl.MqttCommittableMessage) so you can acknowledge them by calling `messageArrivedComplete`.
 
 Scala
 : @@snip (../../../../mqtt/src/test/scala/akka/stream/alpakka/mqtt/scaladsl/MqttSourceSpec.scala) { #run-source-with-manualacks }
--- a/docs/src/main/paradox/mqtt.md
+++ b/docs/src/main/paradox/mqtt.md
@@ -40,6 +40,23 @@ Java
 
 This source has a materialized value (@scaladoc[Future](scala.concurrent.Future) in Scala API and @extref[CompletionStage](java-api:java/util/concurrent/CompletionStage) in Java API) which is completed when the subscription to the MQTT broker has been completed.
 
+MQTT automatically acknowledges messages back to the server once they are passed downstream. The atLeastOnce source allow users to acknowledge the messages anywhere downstream.
+Please note that for manual acks to work CleanSession` should be set to false and MqttQoS should be AtLeastOnce.
+
+Scala
+: @@snip (../../../../mqtt/src/test/scala/akka/stream/alpakka/mqtt/scaladsl/MqttSourceSpec.scala) { #create-source-with-manualacks }
+
+Java
+: @@snip (../../../../mqtt/src/test/java/akka/stream/alpakka/mqtt/javadsl/MqttSourceTest.java) { #create-source-with-manualacks }
+
+The atLeastOnce source returns CommittableMessage so you can acknowledge them by calling `messageArrivedComplete`.
+
+Scala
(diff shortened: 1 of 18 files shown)


johanandren:
Travis failure seems like it is related to the changes, please have a look.

juanjoDiaz:
I've been trying to reproduce it but it works on my machine 100% of the time.

The error seems to happen in different tests and be related to the connection closing before it should. That's odd since I haven't modify anything related to the connection closing.

Any hint is welcome 🙂

raboof:
Restarted the jobs, but it's suspicious they all failed...

ennru:
@juanjoDiaz Could you rebase, and incorparate the changes in master.
After that I'll try to understand where the problem is.

ennru:
It fails locally for me as well:
```
[info] mqtt source
[info] - should consume unacknowledged messages from previous sessions using manualAck *** FAILED *** (10 seconds, 138 milliseconds)
[info]   A timeout occurred waiting for a future to complete. Queried 98 times, sleeping 100000000 nanoseconds between each query. (MqttSourceSpec.scala:65)
```

juanjoDiaz:
Our last conclusion was that the connection was closing randomly at some point.
The error on travis looks a bit different than what you get and even is different scala versions in Travis throw different errors. But all of them seems to be the same. The connection closes before it should...

It's strange that I've run the tests at least a hundred times in my machine and haven't been able to replicate. Which setup do you have?

ennru:
I don't do anything more than starting the docker images and running the tests for the mqtt module.

ennru:

johanandren:
Now there were a few MQTT test failures

juanjoDiaz:
Yeah, I'm pretty confident at this point that the tests for MQTT are just not very reliable.

There mostly work on my machine but fail on Travis which make me think of timing issues.
Also the tests seems to pass or fail depending on whether you run the source before or after the sink which shouldn't be.
And also I suspect that the order and parallelism with which the test are run affect the results since all of them access the same topics.

johanandren:
One of the tests fails with a 10 s timeout, that sounds like it could be something more to it than just timing sensitivity, maybe there is a bug lurking here?

juanjoDiaz:
I suspect that the timeout is because other test is stealing the message, so the test just hang up forever waiting for the message.

I'll try to look into this.

juanjoDiaz:
There is also something definitely fishy about the stage...
This works:
```
      val elem = source.runWith(Sink.head)
      Source.single(MqttMessage(topic1, ByteString("ohi"))).runWith(MqttSink(sinkSettings, MqttQoS.atLeastOnce))
      elem.futureValue shouldBe MqttMessage(topic1, ByteString("ohi"))
```
but this doesn't:
```
      Source.single(MqttMessage(topic1, ByteString("ohi"))).runWith(MqttSink(sinkSettings, MqttQoS.atLeastOnce))
      val elem = source.runWith(Sink.head)
      elem.futureValue shouldBe MqttMessage(topic1, ByteString("ohi"))
```
and both should work just the same...

johanandren:
The source only works if the topic has been published to already?

juanjoDiaz:
It's weirder than that.
This works
```
    "support multiple materialization" in {
      val settings = MqttSourceSettings(sourceSettings, Map(topic1 -> MqttQoS.AtLeastOnce))
      val source = MqttSource.atMostOnce(settings, 8)

      val elem = source.runWith(Sink.head)
      Source.single(MqttMessage(topic1, ByteString("ohi"))).runWith(MqttSink(sinkSettings, MqttQoS.atLeastOnce))
      elem.futureValue shouldBe MqttMessage(topic1, ByteString("ohi"))

      val elem2 = source.runWith(Sink.head)
      Source.single(MqttMessage(topic1, ByteString("ohi"))).runWith(MqttSink(sinkSettings, MqttQoS.atLeastOnce))
      elem2.futureValue shouldBe MqttMessage(topic1, ByteString("ohi"))
    }
```

but this fails
```
    "support multiple materialization" in {
      val settings = MqttSourceSettings(sourceSettings, Map(topic1 -> MqttQoS.AtLeastOnce))
      val source = MqttSource.atMostOnce(settings, 8)

      val elem = source.runWith(Sink.head)
      Source.single(MqttMessage(topic1, ByteString("ohi"))).runWith(MqttSink(sinkSettings, MqttQoS.atLeastOnce))
      elem.futureValue shouldBe MqttMessage(topic1, ByteString("ohi"))

      val elem2 = source.runWith(Sink.head)
      Source.single(MqttMessage(topic1, ByteString("ohi"))).runWith(MqttSink(sinkSettings, MqttQoS.atLeastOnce))
      elem2.futureValue shouldBe MqttMessage(topic1, ByteString("ohi"))
    }
```
The weird thing is that it fails on the second materialization not the first one.

Also, based on Travis it seems that we randomly lose messages. I've noticed that I can reproduce it locally adding load to my machine and trying multiple times.

I have no idea of where those messages are going :/

johanandren:
Aren't those two code blocks identical?

juanjoDiaz:
Maybe, you saw it too quickly (I edited it a second after posting).
The second changes the order the source and sink are created.

juanjoDiaz:
Alright! Finally understood the issues.

New Mqtt subscribers only get messages published after they are created for the first time.
So the tests needed to wait until the subscriber is subscribed before start publishing messages.

This is ready now. @johanandren @ennru

ennru:
Is this something the user of this module needs to be aware of? Could it be ensured in the API? Should it be added to the documentation?

juanjoDiaz:
Nothing to do with this module and nothing we can do.
It's just the way MQTT works (which I had misunderstood).

I don't think that it should be documented here since:
- It's just an edge case where you want to initiate publisher and subscriber exactly at the same time one have zero messages lost. Normally it just make sense that you get messages from the time that you create the subscriber for the first time and not before. This doesn't affect a subscriber reconnecting; the state is kept in that case.
- Alpakka doesn't include documentation for how the other systems that there are connectors work

ennru:
Thank you for keeping it up.

Issue #518, opened 2017-10-12
Make the Spring MVC/Boot integration provide "Auto Configuration"


ktoso:
There's a way to "magically configure things if X is on classpath" in spring, though I'm not sure how to pull that off

https://docs.spring.io/spring-boot/docs/current/reference/html/using-boot-auto-configuration.html

If someone is interested in making it even easier to use Akka Streams in spring please comment here or provide a PR, thanks! 

Refs previous PR: https://github.com/akka/alpakka/pull/517

PR #521, opened 2017-10-12
+spr #518 enable autoconfiguration of akka streams in spring web

ktoso:
Solves #518 and I have a blog post about it coming up.

Please review


spring-web/src/test/java/akka/stream/alpakka/spring/web/MyAppConfiguration.java
```diff
@@ -1,42 +0,0 @@
-/*
```
ktoso: instead I made it "autoconfig" which is a magical thing they exec when they load classpath stuffs

PR #522, opened 2017-10-12
JMS :: set JMSType header

frederic-gendebien:
adds _type field in JmsTextMessage to be able to set the JMSType header


jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
```diff
@@ -18,25 +18,37 @@ sealed trait Destination
...
+  /**
+   * Java API: adds JMSProperty [[JmsTextMessage]]
+   */
+  def withProperty(name: String, value: Any) = copy(properties = properties + (name -> value))
+
+  /**
+   * Java API: add [[JmsTextMessage]]
+   */
+  @deprecated("Unclear method name, use withProperty instead")
+  def add(name: String, value: Any) = withProperty(name, value)
```
johanandren: 👍

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
```diff
@@ -18,25 +18,37 @@ sealed trait Destination
 final case class Topic(name: String) extends Destination
 final case class Queue(name: String) extends Destination
 
-final case class JmsTextMessage(body: String, properties: Map[String, Any] = Map.empty) {
+final case class JmsTextMessage(body: String, _type: Option[String] = None, properties: Map[String, Any] = Map.empty) {
```
johanandren: Make it `type` or `jmsType` instead. Underscore is used for private fields and this is a public field.

johanandren:
On second thought, what about adding a `JmsHeader` class and have a `Set[JmsHeader]` to make it easy to add other headers as well?

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
frederic-gendebien: I agree, but I can't call it type (type is reserved) and jmsType is redundant with the name of the class, don't you think? Let's say I'll call it that way

frederic-gendebien: I will also add some test, right!

frederic-gendebien:
Could be interesting, I gonna try that

johanandren:
One way would be to make it a `sealed trait JmsHeader` and then `foreach { case JmsType(t) => textMessage.setJMSType(t) }`

frederic-gendebien:
Funny, I already did it that way

johanandren:
Ah, travis failed on formatting, make sure to compile both sources and tests to get the code autoformatted and check in the result of that.

johanandren:
Fixes #522 520

frederic-gendebien:
It is good now

frederic-gendebien:
Thanks! Any idea when it will be released and available?

johanandren:
We just missed the release train with a few hours, we tend to release Alpakka fairly often but we don't have a planned date for the next one

Issue #527, opened 2017-10-13
CassandraSourceSpec instability


raboof:
```
akka.stream.alpakka.cassandra.scaladsl.CassandraSourceSpec *** ABORTED *** (2 seconds, 245 milliseconds)
[info]   com.datastax.driver.core.exceptions.NoHostAvailableException: All host(s) tried for query failed (tried: /127.0.0.1:9042 (com.datastax.driver.core.exceptions.DriverException: Timeout while trying to acquire available connection (you may want to increase the driver number of per-host connections)))
```

johanandren:
https://travis-ci.org/akka/alpakka/jobs/287566952

2m:
https://travis-ci.org/akka/alpakka/builds/288852191

PR #528, opened 2017-10-16
Remove javac options that confuse javadoc

raboof:
The javadoc tool returned an error (and failed the build) for me because of the
-Xlint:unchecked option which is passed to javadoc but not
recognized there.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -110,7 +110,9 @@ lazy val mqtt = alpakkaProject("mqtt", Dependencies.Mqtt)
 
 lazy val s3 = alpakkaProject("s3", Dependencies.S3)
 
-lazy val springWeb = alpakkaProject("spring-web", Dependencies.SpringWeb)
+lazy val springWeb = alpakkaProject("spring-web",
+  javacOptions := javacOptions.value.filter(_ != "-Xlint:unchecked"),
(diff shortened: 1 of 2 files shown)


2m:
Which command failed for you? AFAICS unidoc does not generate Javadoc in Alpakka.

raboof:
`sbt package` is fine, but `sbt publishLocal` gives the error for me

2m:
This is interesting. Since the spring-web alpakka module has only Java code, the `spring-web/compile:doc` task starts javadoc (in contrast this task for all other projects starts scaladoc).

The proper fix would be to scope the javacOptions to compile task, so it is not picked up by the doc task:

```
diff --git a/project/Common.scala b/project/Common.scala
index bcb767c..8e74141 100644
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -40,7 +40,7 @@ object Common extends AutoPlugin {
       "-Ywarn-dead-code",
       "-Xfuture"
     ),
-    javacOptions ++= Seq(
+    javacOptions in compile ++= Seq(
       "-Xlint:unchecked"
     ),
     autoAPIMappings := true,
```

2m:
Nice. Validation issue was #527

PR #532, opened 2017-10-17
Close remote file when stream is closed #530

### LINKED ISSUE #530: SFTP file is not closed when input/output stream terminates
So we're using alpakka to upload to an IBM websphere appliance that handles sftp. And we're being told that our client is terminating the connection instead of cleanly closing out the transfer. It seems like this is because the outputstream for the file upload is flushed, but the file handle isn't being closed properly.

Falmarri:


ktoso:
Thanks for the fix!
Please refer to ticket numbers in PR comment :) 
Resolves #530

ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpIOGraphStage.scala
```diff
@@ -138,33 +145,34 @@ private[ftp] trait FtpIOSinkStage[FtpClient, S <: RemoteFileSettings]
...
+              pull(in)
             } catch {
               case NonFatal(e) ⇒
                 matFailure(e)
-                try osOpt.foreach(_.close())
-                catch {
-                  case NonFatal(_) ⇒
+                // We have to do these here instead of waiting for postStop, because the connection
+                // is almost certainly dead at this point, and there's no reasonable way to let the
+                // postStop handler know that it should expect an exception from trying to disconnect
```
johanandren: Actually, this will then still leak if the stage is abruptly killed, so move this logic to postStop, you can keep a boolean flag or such in the stage logic to know it is stopping because of failure or not.

Falmarri: I figured if it was abruptly killed, it's not as big of a deal to make sure the actual output stream is closed cleanly, since you'll still get the disconnect. But I can add that check too

johanandren: Abrupt killed as in `ActorMaterializer.shutdown()`, not JVM kill. Can happen if you create a materializer and bind it to an actor lifecycle for example.

Falmarri: @johanandren I did some refactoring like you suggested with a boolean flag. This should handle all the cases (and even materialize to a failure if closing the stream fails. Since, as is the reason for the bug request, some upstream ftp servers don't consider that transfer to be "successful" in that case)

PR #538, opened 2017-10-20
#527 Increase acquire connection timeout

2m:
So previous #536 did not really work. Trying again. This time increasing [connection acquire](http://docs.datastax.com/en/drivers/java/2.1/com/datastax/driver/core/PoolingOptions.html#setPoolTimeoutMillis-int-) timeout from the same default of 5 seconds.

@chbatey are we doing this right? 

We are trying to prevent spurious errors like this while testing:
```
Cause: com.datastax.driver.core.exceptions.NoHostAvailableException: All host(s) tried for query failed (tried: /127.0.0.1:9042 (com.datastax.driver.core.exceptions.DriverException: Timeout while trying to acquire available connection (you may want to increase the driver number of per-host connections)))
```

Refs #527

### DIFF
--- a/cassandra/src/test/scala/akka/stream/alpakka/cassandra/scaladsl/CassandraSourceSpec.scala
+++ b/cassandra/src/test/scala/akka/stream/alpakka/cassandra/scaladsl/CassandraSourceSpec.scala
@@ -6,7 +6,7 @@ package akka.stream.alpakka.cassandra.scaladsl
 import akka.actor.ActorSystem
 import akka.stream.ActorMaterializer
 import akka.stream.scaladsl.{Sink, Source}
-import com.datastax.driver.core.{Cluster, PreparedStatement, SimpleStatement, SocketOptions}
+import com.datastax.driver.core.{Cluster, PoolingOptions, PreparedStatement, SimpleStatement}
 import org.scalatest._
 import org.scalatest.concurrent.ScalaFutures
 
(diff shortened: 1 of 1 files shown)


chbatey:
I can't think  of a good reason why this should be happening given that the failures aren't on the first test i.e. we're just using this timeout to let the docker cassandra get up and running.

I've raised https://github.com/akka/alpakka/pull/542 to get the driver logs which will be a bit spammy but should give us the lifecycle of connections

2m:
I am not completely sure how Cassandra connection pooling works, but my reasoning was that connections from the previous tests are not released quickly enough, so some later tests start to fail because of that. Therefore in this PR I have tried to increase the time it waits while trying to get a new connection. However from the tests logs (`** ABORTED *** (6 seconds, 921 milliseconds)`) it does not look like the test took longer that 7 seconds before it was aborted.

chbatey:
I think what could be happening is that we're dropping/creating keyspace between each test ( cassandra does not handle this very well) which causes errors on connections which means that all connections to our 1 cassandra node are gated.

Some of the time we're failing with:

```
[info] akka.stream.alpakka.cassandra.scaladsl.CassandraSourceSpec *** ABORTED *** (7 seconds, 966 milliseconds)
[info] com.datastax.driver.core.exceptions.ServerError: An unexpected error occurred server side on /127.0.0.1:9042: java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.NullPointerException
```

But we don't get the cassandra server logs as they are in the container.

I think we should either move to using an embedded C* so we get the logs or use a diff keyspace name for each test.

2m:
Original issue fixed in #543

PR #540, opened 2017-10-20
JMS - make acknowledge mode configurable

mmatloka:
Hi,
Referying to https://github.com/akka/alpakka/issues/104 and e.g. https://github.com/akka/alpakka/commit/d478a6168262e070588c0c83a9111ef6f6e3f04b I'd like to propose to allow for JMS acknowledge mode configuration.

### DIFF
--- a/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
@@ -102,7 +102,7 @@ object JmsTextMessage {
     JmsTextMessage(body = body, headers = headers.asScala.toSet, properties = properties.asScala.toMap)
 }
 
-sealed trait AcknowledgeMode {
+trait AcknowledgeMode {
   val jmsAcknowledgeMode: Int
 }
 
--- a/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
@@ -102,24 +102,13 @@ object JmsTextMessage {
     JmsTextMessage(body = body, headers = headers.asScala.toSet, properties = properties.asScala.toMap)
 }
 
-trait AcknowledgeMode {
-  val jmsAcknowledgeMode: Int
-}
+final class AcknowledgeMode(val mode: Int)
 
-case object AutoAcknowledge extends AcknowledgeMode {
-  override val jmsAcknowledgeMode: Int = jms.Session.AUTO_ACKNOWLEDGE
-}
-
-case object ClientAcknowledge extends AcknowledgeMode {
-  override val jmsAcknowledgeMode: Int = jms.Session.CLIENT_ACKNOWLEDGE
-}
-
-case object DupsOkAcknowledge extends AcknowledgeMode {
-  override val jmsAcknowledgeMode: Int = jms.Session.DUPS_OK_ACKNOWLEDGE
-}
-
-case object SessionTransacted extends AcknowledgeMode {
-  override val jmsAcknowledgeMode: Int = jms.Session.SESSION_TRANSACTED
+object AcknowledgeMode {
+  val AutoAcknowledge: AcknowledgeMode = new AcknowledgeMode(jms.Session.AUTO_ACKNOWLEDGE)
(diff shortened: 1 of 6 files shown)


nilsga:
Looking forward to seeing this merged, as we need to be able to ack messages manually. However, I'm wondering if it could be an idea to not seal `AcknowledgeMode`? There are messaging brokers that support additional acknowledge modes, like ActiveMQ which has an [INDIVIDUAL_ACKNOWLEDGE](http://activemq.apache.org/maven/apidocs/org/apache/activemq/ActiveMQSession.html#INDIVIDUAL_ACKNOWLEDGE) mode. So being able to extend `AcknowledgeMode` would be very useful.

mmatloka:
@nilsga you're right, this could be useful, I will make this possible :)

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
```diff
@@ -84,19 +87,39 @@ object JmsTextMessage {
...
   /**
    * Java API: create  [[JmsTextMessage]]
    */
   def create(body: String, headers: util.Set[JmsHeader], properties: util.Map[String, Any]) =
-    JmsTextMessage(body = body, headers = headers.toSet, properties = properties.toMap)
+    JmsTextMessage(body = body, headers = headers.asScala.toSet, properties = properties.asScala.toMap)
+}
+
+trait AcknowledgeMode {
+  val jmsAcknowledgeMode: Int
```
johanandren: I think using a `final class AcknowledgeMode(mode: Int)` and providing constants in the companion for the common modes would be cleaner than creating a type hierarchy just to provide different values (and will still allow custom ones as requested).

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
```diff
@@ -63,14 +63,7 @@ final class JmsSourceStage(settings: JmsSourceSettings) extends GraphStage[Sourc
             consumer.setMessageListener(new MessageListener {
               override def onMessage(message: Message): Unit = {
                 backpressure.acquire()
-                try {
-                  message.acknowledge()
-                  handleMessage.invoke(message)
-                } catch {
-                  case e: JMSException =>
-                    backpressure.release()
-                    handleError.invoke(e)
```
johanandren: I'm missing something here, why did the error handling go away?

jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsConnectorsTest.java
```diff
@@ -299,6 +301,51 @@ public void publishAndConsumeTopic() throws Exception {
...
+                msgsIn.add(JmsTextMessage.create(n.toString()));
+            }
+
+            Source.from(msgsIn).runWith(jmsSink, materializer);
+
+            //#create-jms-source-client-ack
+            Source<Message, NotUsed> jmsSource = JmsSource.create(JmsSourceSettings
+                    .create(connectionFactory)
+                    .withQueue("test")
+                    .withAcknowledgeMode(ClientAcknowledge$.MODULE$)
```
johanandren: This is not ok as a Java API, I think my proposed change will fix it though. If it doesn't, put the constants in a non-companion object called `AcknowledgeModes.ClientAck`

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
mmatloka: hm, I removed it because this was an exception thrown by `message.acknowledge()` which is not executed now here

jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsConnectorsTest.java
mmatloka: @johanandren I have implemented the final class + compation object variant, please take a look how java api looks now

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
johanandren: Ah, so `invoke` will never throw anything? Then it's ok.

johanandren:
Fixes #554

Issue #447, opened 2017-08-20
support for Dynamodb Streams


joost-de-vries:
I've implemented a first naïve support for Dynamodb Streams. You can find the implementation [here](https://github.com/joost-de-vries/alpakka/blob/dynamodb-streams-wip-1/dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/Streams.scala).
And [this](https://github.com/joost-de-vries/alpakka/blob/dynamodb-streams-wip-1/dynamodb/src/test/scala/akka/stream/alpakka/dynamodb/StreamsSpec.scala) is how to use it.

I'm saying that it's a naïve implementation because it just follows the [example](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.LowLevel.Walkthrough.html#Streams.LowLevel.Walkthrough.Step2) from the docs. And that example ignores new shards, expired shards and shards splitting. The AWS docs just advise to use the Kinesis Adapter instead of documenting how to deal with these features.

Given that it's not that explicitly documented how to deal with those cases it will take some reverse engineering to get those aspects right.
I guess we can study the [code](https://github.com/awslabs/dynamodb-streams-kinesis-adapter) of the Dynamodb - Kinesis adapter. And we can do load tests....

Any interest in further implementing this?

joost-de-vries:
@areyouspiffy I think you've implemented Dynamodb Streams once. Do you have any ideas on this?

PR #541, opened 2017-10-22
Support for Dynamodb Streams

joost-de-vries:
This PR provides an initial implementation for Dynamodb Streams. See issue #447 
I thought we can release this in 'alpha' status to get feedback. What do you think?

This should probably be labeled `p:dynamodb`

Next steps:
- add test that will use multiple shards
- add test that triggers resharding


dylan-raithel-lookout:
What happens in this code when a shard splits or merges? Are you relying on the Supervisor to handle restart and then re-describing the streams?

joost-de-vries:
@dylan-raithel-lookout the documentation really doesn't specify how resharding affects the api. It mentions that its a concern and advises to use the kinesis adapter. 
So I couldn't implement that just from the documentation. The next step would be, I think, to do some load tests with really good logging. And analyse the behaviour of the api. 
Do you have any ideas?

dayyan:
Here's what I see in [documentation for the Kinesis API].(http://docs.aws.amazon.com/streams/latest/dev/developing-consumers-with-sdk.html#kinesis-using-sdk-java-get-data-reshard)

>If `getRecordsResult.getNextShardIterator` returns `null`, it indicates the following: a shard split or merge has occurred that involved this shard, this shard is now in a `CLOSED` state, and you have read all available data records from this shard.
>
>In this scenario, you should re-enumerate the shards in the stream to pick up the new shards that were created by the split or merge.

joost-de-vries:
@dayyan good idea. That documentation definitely gives some direction in how to implement this. I remember some docs saying that the semantics are not 1-to-1 the same though.

Wrt the dynamodb adapter: are you suggesting to use that as an intermediary? I guess it's an option. There's [a Kinesis adapter of course ](https://developer.lightbend.com/docs/alpakka/current/kinesis.html) but that doesn't handle iterating over shards or handle resharding out of the box.

Unfortunately I'm rather busy currently. So I can't say I'll work on this very soon. If you want to take a crack at it: by all means.

Implementing those semantics is perhaps not the hardest thing: I'm wondering how to trigger resharding in a unittest in a predictable manner.

ennru:

Normal:
need this feature

PR #542, opened 2017-10-23
Add logback to cassandra tests

chbatey:
To get the driver logging to see why we're getting NoHostAvailable
exceptions during the tests.


chbatey:
:+1:

PR #543, opened 2017-10-23
Make test cassandra keyspace name dynamic

chbatey:
Cassandra does not handle the dropping/creating of a keyspcae with the
same name very well. This can lead to errors coming back and the driver
then refusing to hand out connections for that host.


patriknw:
wonder if we have a similar problem in akka-persistence-cassandra, e.g. many EventsByTagSpec using same keyspace, and I don't think we set unique keyspace for snapshots

chbatey:
I don't think we drop/create the keyspace each time in those tests do we?

patriknw:
right, I think we delete the data directory before launching

PR #545, opened 2017-10-24
S3: Make SigningKey non-constant to reflect possible token changes

tkroman:
AWS SDK credential provider supports token expiration, so the code needs to reflect that.

I was not sure how to go about scope of `def`: we can just make `akka.stream.alpakka.s3.auth.SigningKey#credentials` a `def` and leave `SigningKey` a `val`, but the key also depends on a current date. I'm not an expert in the field of security algorithms and to the extent of my understanding it it safer to recalculate the the key completely.

A better solution would be to try and use the signing mechanism provided by the AWS SDK, but I decided to keep the scope of changes smaller.


idzivinskyi:
@johanandren can you take a look on this PR?

johanandren:
Fixes #552

PR #546, opened 2017-10-25
IBM Eventstore support

olofwalker:
About IBM Eventstore: 
https://www.ibm.com/support/knowledgecenter/SSGNPV/eventstore/desktop/welcome.html

- This connector contains a flow and a sink make it possible to insert records into event store. 
- There are integration tests currently that requires you to install eventstore locally, these are set to be ignored.
- Documentation cover the basic usages and some caveats that may change in the future.

### DIFF
--- a/docs/src/main/paradox/ibm-eventstore.md
+++ b/docs/src/main/paradox/ibm-eventstore.md
@@ -8,32 +8,11 @@ IBM EventStore is an in-memory database built on top of Apache Spark and Apache 
 
 ## Artifacts
 
-sbt
-:   @@@vars
-    ```scala
-    libraryDependencies += "com.lightbend.akka" %% "akka-stream-alpakka-ibm-eventstore" % "$version$"
-    ```
-    @@@
-
-Maven
-:   @@@vars
-    ```xml
-    <dependency>
-      <groupId>com.lightbend.akka</groupId>
-      <artifactId>akka-stream-alpakka-ibm-eventstore_$scalaBinaryVersion$</artifactId>
-      <version>$version$</version>
-    </dependency>
-    ```
-    @@@
-
-Gradle
-:   @@@vars
-    ```gradle
-    dependencies {
-      compile group: "com.lightbend.akka", name: "akka-stream-alpakka-ibm-eventstore_$scalaBinaryVersion$", version: "$version$"
-    }
-    ```
-    @@@
+@@dependency [sbt,Maven,Gradle] {
+  group=com.lightbend.akka
+  artifact=akka-stream-alpakka-ibm-eventstore_$scalaBinaryVersion$
+  version=$version$
+}
 
 ## Usage
 
--- a/docs/src/main/paradox/ibm-eventstore.md
+++ b/docs/src/main/paradox/ibm-eventstore.md
@@ -1,0 +1,86 @@
+# IBM EventStore
+
+IBM EventStore is an in-memory database built on top of Apache Spark and Apache Parquet Data Format, the Alpakka IBM EventStore connector provides an Akka Stream flow and sink for insertion of records into a table.
+ 
+## External references 
+
+[IBM EventStore Documentation](https://www.ibm.com/support/knowledgecenter/SSGNPV/eventstore/desktop/welcome.html)
+
+## Artifacts
+
+sbt
+:   @@@vars
+    ```scala
+    libraryDependencies += "com.lightbend.akka" %% "akka-stream-alpakka-ibm-eventstore" % "$version$"
+    ```
+    @@@
+
+Maven
+:   @@@vars
+    ```xml
+    <dependency>
+      <groupId>com.lightbend.akka</groupId>
+      <artifactId>akka-stream-alpakka-ibm-eventstore_$scalaBinaryVersion$</artifactId>
+      <version>$version$</version>
+    </dependency>
+    ```
+    @@@
+
+Gradle
+:   @@@vars
+    ```gradle
+    dependencies {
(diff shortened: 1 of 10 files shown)


docs/src/main/paradox/ibm-eventstore.md
```diff
@@ -0,0 +1,86 @@
...
+    @@@
+
+Gradle
+:   @@@vars
+    ```gradle
+    dependencies {
+      compile group: "com.lightbend.akka", name: "akka-stream-alpakka-ibm-eventstore_$scalaBinaryVersion$", version: "$version$"
+    }
+    ```
+    @@@
```
johanandren: We have a new nifty paradox tag/fragment for these now, look at the other connector docs for a sample.

ibm-eventstore/src/main/scala/akka/stream/alpakka/ibm/eventstore/scaladsl/EventStore.scala
```diff
@@ -0,0 +1,60 @@
...
+import com.ibm.event.oltp.EventContext
+import com.ibm.event.oltp.InsertResult
+import org.apache.spark.sql.Row
+
+object EventStoreSink {
+
+  /**
+   * Creates a sink for insertion of records into EventStore.
+   * @param databaseName Name of the database, database has to exist before the call to this function.
+   * @param tableName    Name of the table, database has to exist before the call to this function.
```
johanandren: What happens if database/table does not, is the behavior undefined or does the stream fail upon materialization?

ibm-eventstore/src/main/scala/akka/stream/alpakka/ibm/eventstore/scaladsl/EventStore.scala
```diff
@@ -0,0 +1,60 @@
...
+   * @return             CompletionStage[Done]
```
johanandren: Remove, type is already in signature (and in addition to that, this is the wrong type)

ibm-eventstore/src/test/java/akka/stream/alpakka/ibm/eventstore/javadsl/EventStoreSpec.java
```diff
@@ -0,0 +1,190 @@
...
+@Ignore
+public class EventStoreSpec {
+    private static ActorSystem system;
+    private static Materializer materializer;
+    private static EventContext eventContext;
+    private static String databaseName = "TESTDB";
+    private static String tableName = "TESTTABLE";
+    private static void setEndpoint() {
+        // #configure-endpoint
+        ConfigurationReader.setConnectionEndpoints("127.0.0.1:5555");
```
johanandren: wat? So there is a global mutable context that the IBM thing uses? A single JVM can only connect to one of these at a time? No API for actually representing that context as an object and pass it around instead? I think that if that is possible, and it needs to be shared among stages, it would be much preferred to pass it around to the stream factories instead. If it doesn't require sharing and can be non-jvm-global then maybe even let the stage itself take the connection config and manage the lifecycle?

project/Dependencies.scala
```diff
@@ -125,6 +125,16 @@ object Dependencies {
     )
   }
 
+  val IbmEventStore = Seq(
+    libraryDependencies ++= Seq(
+      "org.apache.spark"    %% "spark-sql"        % "2.2.0",
+      "EventStore"          % "ibm-event-desktop" % "1.1.1",
```
johanandren: Licenses for these two

olofwalker: Asked IBM, waiting.

ibm-eventstore/src/main/scala/akka/stream/alpakka/ibm/eventstore/scaladsl/EventStore.scala
olofwalker: No, it throws before materialization when you try to lookup the table with ` EventContext.getEventContext(databaseName)/ context.getTable(tableName)

ibm-eventstore/src/test/java/akka/stream/alpakka/ibm/eventstore/javadsl/EventStoreSpec.java
olofwalker: To my knowledge this is the case, a global single endpoint.

https://www.ibm.com/support/knowledgecenter/SSGNPV/eventstore/desktop/dev-guide.html#insertData

ibm-eventstore/src/main/scala/akka/stream/alpakka/ibm/eventstore/scaladsl/EventStore.scala
johanandren: Ok, that is a likely a bit surprising, please add that to the scaladoc. 

I think you can skip the `@param`s as those are fairly self-explanatory and just have a paragraph saying that this will throw if the database or table does not exist when calling this factory method.

project/Dependencies.scala
olofwalker: Eventstore license:
https://github.com/IBMProjectEventStore/EventStore-DeveloperPreview/blob/master/EventStore/License/LA_en

Apache spark license:
https://github.com/apache/spark/blob/master/LICENSE

What steps do we need to take to evaluate the IBM one ?

johanandren: Please add a comment noting the licenses like on other external deps

olofwalker:
The license is providing challenges, for now we will close this PR until we can sort this out with IBM

PR #548, opened 2017-10-26
Specify request JSON encoding explicitly in Elasticsearch connector

### LINKED ISSUE #547: Elasticsearch connector - encoding bug
Hi,

Elasticsearch connector 0.14 fails to post json to elastic if json contains special characters.

Elastic is unable to parse the json and is reporting *Invalid UTF-8 middle byte 0x20*

The problem is located on this line:
https://github.com/akka/alpakka/blob/master/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala#L135

By changing it from:

     new StringEntity(json)

to

     new StringEntity(json, "utf-8")

the problem is fixed

takezoe:
Fixes #547

### DIFF
--- a/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
+++ b/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
@@ -4,6 +4,8 @@
 
 package akka.stream.alpakka.elasticsearch
 
+import java.nio.charset.StandardCharsets
+
 import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
 import akka.stream.stage._
 import org.apache.http.entity.StringEntity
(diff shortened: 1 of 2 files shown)


elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
```diff
@@ -132,7 +132,7 @@ class ElasticsearchFlowStage[T, R](
           "POST",
           "/_bulk",
           Map[String, String]().asJava,
-          new StringEntity(json),
+          new StringEntity(json, "UTF-8"),
```
johanandren: Use `StandardCharsets.UTF_8` instead.

takezoe:
@johanandren Thanks for your review. I fixed to use `StandardCharsets.UTF_8` instead of a string literal.

Issue #464, opened 2017-09-12
Provide support for Bluemix Cloud Object Storage S3 API


olofwalker:
Bluemix provides a service called Cloud Object Storage (COS), it has two API's, one of them is S3 compatible. 

With a minor change in the Unmarshalling of the of the inital response to a multipart upload the current S3 connector can be made compatible with COS.

The change allows the unmarshaller to accept either the AWS S3 "application/octet-stream content-type" or the COS required "application/xml".

PR #563, opened 2017-11-03
Adds S3 getObjectMetadata, deleteObject, putObject

### LINKED ISSUE #313: AWS S3 delete object missing
The alpakka client currently supports listing, adding, getting but not deleting. Is there a specific reason for that?

Something like this
```
s3Client.deleteObject(bucket, fileName)
```
(and deleteObjects())

I understand this is not a streaming call but still it's a bit annoying to have to depend on an other (blocking) client to do the delete.

What would the best return type be? Source[Unit, Unit] or Future[Unit]?

### LINKED ISSUE #314: AWS S3 putObject missing
The multipart upload is is rather slow (200-500ms) compared to a normal put operation (80-200ms) when used for small files.

### LINKED ISSUE #497: S3 download with metadata
The current return type for downloading from S3 is `Source[ByteString, NotUsed]`, it would be immensely useful if the source could materialize with the object metadata like last modified and content type (see `com.amazonaws.services.s3.model.ObjectMetadata`). So something like `Source[ByteString, ObjectMetadata]`.

Any interest in adding this?

francisdb:
fix #497 S3 download with metadata
fix #314 AWS S3 putObject missing
fix #313 AWS S3 delete object missing
adds S3 getObjectMetadata

Feedback is welcome

TODO

- [x] Introduce proper Metadata case class that gives access to underlying response headers
- [x] Fix metadata date parsing
- [x] Add more integrations tests
- [x] Add more java client mocked tests
- [x] Add more scala client mocked tests

#464 messed up the tests (they are not correctly testing any more), will split bluemix and aws tests (no longer parametrised)


ngbinh:
looking forward to this PR 👍

francisdb:
Planning to finish this today

francisdb:
For my part this is ready for merge, might however conflict with other recent pr's #645, #619
Not sure if there still is value in #631

francisdb:
@ ennru would you mind having a look at this?

francisdb:
I did only a partial implementation of `ObjectMetadata`, the original response headers are available so the user can always extend it with the needed functionality.

ennru:
Quite a mouthful of code. Looking good.
Can anybody with more S3 knowledge than me have a look, if we're good to merge this?

francisdb:
One more note, I restored the original `S3NoMock` spec as `AwsS3IntegrationSpec` as the modifications made to make it parametrized for Bluemix broke the validation of most tests. That test should probably just be de-parametrized and renamed to `BluemixIntegrationSpec` but I have no way to run it properly.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Headers.scala
```diff
@@ -51,14 +51,22 @@ object S3Headers {
...
   case object Standard extends StorageClass("STANDARD")
   case object InfrequentAccess extends StorageClass("STANDARD_IA")
+  case object Glacier extends StorageClass("GLACIER")
   case object ReducedRedundancy extends StorageClass("REDUCED_REDUNDANCY")
+
+  def apply(cls: String): Option[StorageClass] = cls match {
+    case "STANDARD" => Some(Standard)
+    case "STANDARD_IA" => Some(InfrequentAccess)
+    case "GLACIER" => Some(Glacier)
+    case "REDUCED_REDUNDANCY" => Some(ReducedRedundancy)
```
johanandren: Use the string constants already available through the objects when matching instead of having the constants twice.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -49,10 +54,18 @@ object S3Stream {
 
   def apply(settings: S3Settings)(implicit system: ActorSystem, mat: Materializer): S3Stream =
     new S3Stream(settings)
+
+  private final val CONTENT_LENGTH = "Content-Length"
+  private final val ETAG = "ETag"
+  private final val LAST_MODIFIED = "Last-Modified"
```
johanandren: Scala style for constants should be `CamelCase` (this is Java style)

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -263,4 +338,21 @@ private[alpakka] final class S3Stream(settings: S3Settings)(implicit system: Act
...
+  private def lastModifiedFromHeaders(headers: Seq[HttpHeader]): Option[Instant] =
+    headers
+      .find(_.lowercaseName() == LAST_MODIFIED.toLowerCase)
+      // FIXME we have problems reading the time format (Thu, 02 Nov 2017 23:47:39 GMT)
+      .flatMap(h => Try(Instant.parse(h.value())).toOption)
+
+  private def storageClassFromHeaders(headers: Seq[HttpHeader]): Option[StorageClass] =
+    headers.find(_.lowercaseName() == STORAGE_CLASS.toLowerCase).flatMap { header =>
+      StorageClass(header.value)
+    }
```
johanandren: I can't see that these are used anywhere, and I think that if they are they should use the existing header APIs of the Akka HTTP model instead. (somethingl like ``headers.collectFirst { case `Last-Modified(dateTime)` => dateTime }`` for example)

s3/src/main/scala/akka/stream/alpakka/s3/scaladsl/S3Client.scala
```diff
@@ -41,6 +44,104 @@ final case class ListBucketResultContents(
     storageClass: String
 )
 
+/**
+ * Modelled after com.amazonaws.services.s3.model.ObjectMetadata
+ * @param metadata the raw http headers
+ */
+final case class ObjectMetadata(
```
johanandren: Maybe not make it a case class as that adds a bunch of things to the api, `copy` for example which I think does not do make sense for this. Alternatively make a trait describing the public API and use a case class to implement it.

s3/src/main/scala/akka/stream/alpakka/s3/scaladsl/S3Client.scala
```diff
@@ -41,6 +44,104 @@ final case class ListBucketResultContents(
...
+   * associated object.
+   *
+   * @return The date and time at which Amazon S3 last recorded a modification
+   *         to the associated object.
+   */
+  lazy val lastModified: DateTime = metadata.collectFirst {
+    case ct: `Last-Modified` => ct.date
+  }.get
+
+}
```
johanandren: Ah, here they are done in the right way, so I guess the other methods were just left overs?

s3/src/main/scala/akka/stream/alpakka/s3/scaladsl/S3Client.scala
```diff
@@ -72,11 +173,31 @@ final class S3Client(val s3Settings: S3Settings)(implicit system: ActorSystem, m
...
   private[this] val impl = S3Stream(s3Settings)
 
-  def request(bucket: String, key: String): Future[HttpResponse] = impl.request(S3Location(bucket, key))
+  /**
+   * Use this to extend the library
+   */
+  def request(bucket: String, key: String, method: HttpMethod = HttpMethods.GET): Future[HttpResponse] =
+    impl.request(S3Location(bucket, key), method)
+
+  def getObjectMetadata(bucket: String, key: String): Future[Option[ObjectMetadata]] =
```
johanandren: Scaladoc what it means if it returns `None`,  (not found I'm guessing)

s3/src/main/scala/akka/stream/alpakka/s3/scaladsl/S3Client.scala
francisdb: yep, also the related constants have been removed

PR #567, opened 2017-11-04
Update sbt to 1.0.3.

lomigmegard:
Changelog: https://developer.lightbend.com/blog/2017-10-27-sbt-1-0-3-hotfix/


raboof:
Hmm note that sbt 1.0.3 doesn't work with jdk9 (https://github.com/sbt/sbt/issues/3684) - but perhaps we don't care just yet?

PR #575, opened 2017-11-08
SQS: More idiomatic MessageAction

### LINKED ISSUE #574: SQS: MessageAction are not idiomatic
The SQS `MessageAction` trait has two issues:

- the three implementations are directly in the `sqs` package,
- the `Delete` and `Ignore` are case classes but should be singletons.

It would be more idiomatic to put the classes in the companion object, and more efficient to use objects instead of case classes.

lomigmegard:
Fix #574.

Note that this is not backward compatible. It will break Java and ~~Scala~~ sources that use the `Ack` feature of SQS.

Edit: Only the Java sources are broken with the new PR.

### DIFF
--- a/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsAckFlowStage.scala
+++ b/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsAckFlowStage.scala
@@ -99,7 +99,7 @@ private[sqs] final class SqsAckFlowStage(queueUrl: String, sqsClient: AmazonSQSA
             val (message, action) = grab(in)
             val responsePromise = Promise[AckResult]
             action match {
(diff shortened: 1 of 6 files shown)


johanandren:
One way to keep compatibility would be to keep the case classes and make the "idiomatic" singletons instances of those. Might be worth it to not break source compatibility?

lomigmegard:
Would this version be more acceptable? 

- Add deprecation warnings for the old syntax in Scala (with performance benefits from the new one).
- It breaks only the Java sources (probably much less common?).

The first commit changes the API, while the second fixes the deprecation warnings only.

@johanandren I think it's still worth breaking sources compatibility for such young project as alpakka, the more we wait, the more difficult to do such changes.

PR #586, opened 2017-11-14
Elasticsearch streamable flow stage

mbknor:
The existing ElasticsearchFlowStage only forwards failed messages.
This makes it impossible to do many stream-related operations. Some examples:

- Chain posting to multiple Elastic indexes
- Commit to kafka AFTER messages has been posted to elastic.

In this PR I've created ElasticsearchStreamableFlowStage which (is not perfect at all), but since it returns both messages that have failed and succeeded, it is much more flexible.

This is how I use it in a Java-project (using Project Lombok) at work, where I stream from Kafka to Elastic:

```Java
        val elasticFlow = ElasticsearchStreamableFlow.<PojoAndKafkaOffset>create(elasticIndexName_currentState, elasticTypeName, elasticSettings, elasticClient, objectMapper);

        val thisActor = self();

        Consumer.<byte[], String>committableSource(consumerSettings, Subscriptions.topics(kafkaTopic))
                .map(msg -> { // deserialize json
                    final String json = msg.record().value();
                    //log().info(kafkaTopic+": received: " + json);
                    val pojo = objectMapper.readValue(json, CompanyPojoMetadataComplete.class);
                    val id = Long.toString(pojo.getPojoId());
                    return new PojoAndKafkaOffset( pojo, json, msg.committableOffset());
                })
                .via(elasticFlow) // Send to elastic
                .map( result -> {

                    if ( result.getFailedList().size() > 0) {
                        throw new Exception("Error posting to elastic");
                    }

                    // Build single batch-commit
                    ConsumerMessage.CommittableOffsetBatch batchCommit = ConsumerMessage.emptyCommittableOffsetBatch();
                    for ( val m : result.getSuccessList()) {
                        batchCommit = batchCommit.updated( m.getKafkaOffset() );
                    }

                    return batchCommit;
                })
                .mapAsync(3, c -> {
                    log().info("Committing to kafka: " + c);
                    return c.commitJavadsl();
                }) // Do the actual commit
                .runWith(Sink.ignore(), mat)
                .toCompletableFuture()
                .whenComplete( (done, cause) -> {

                    final Exception exception;
                    if ( cause == null ) {
                        exception = new Exception("The kafka->elastic stream stopped without exception");
                    } else {
                        // Tell actor that something went wrong
                        exception = new Exception("The kafka-elastic stream stopped with error: " + cause.toString(), cause);
                    }
                    // Send exception to actor so that it can log and restart
                    thisActor.tell( exception, null);
                });
```

The Java-pojo ("case class") I use to transport the message through ElasticsearchStreamableFlow looks like this:

```Java
@Value
class PojoAndKafkaOffset implements StreamableIncomingMessage {
    CompanyPojoMetadataComplete pojo;
    String json;
    ConsumerMessage.CommittableOffset kafkaOffset;

    @Override
    public Option<String> id() {
        return Option.apply(Long.toString(pojo.getPojoId()));
    }

    @Override
    public String json() {
        return json;
    }
}
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,5 @@
+scalaVersion in ThisBuild := "2.12.4"
+
 lazy val alpakka = project
   .in(file("."))
   .enablePlugins(PublishUnidoc)
--- a/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchStreamableFlowStage.scala
+++ b/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchStreamableFlowStage.scala
@@ -1,0 +1,214 @@
+/*
+ * Copyright (C) 2016-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.elasticsearch
+
+import java.nio.charset.StandardCharsets
+
+import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
+import akka.stream.stage._
+import org.apache.http.entity.StringEntity
+import org.elasticsearch.client.{Response, ResponseListener, RestClient}
+
+import scala.collection.mutable
+import scala.collection.JavaConverters._
+import spray.json._
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import ElasticsearchStreamableFlowStage._
+import akka.NotUsed
+import akka.stream.alpakka.elasticsearch.scaladsl.ElasticsearchSinkSettings
+import org.apache.http.message.BasicHeader
+import org.apache.http.util.EntityUtils
+import scala.collection.JavaConverters._
+import java.util.{List => JavaList}
+
+trait StreamableIncomingMessage {
+  def id: Option[String]
+  def json: String
+}
+
+final case class StreamableIncomingMessagesResult[T <: StreamableIncomingMessage](
+    success: Seq[T],
+    failed: Seq[T]
+) {
+  // Java-API
+  def getSuccessList(): JavaList[T] = success.toList.asJava
+  // Java-API
+  def getFailedList(): JavaList[T] = failed.toList.asJava
+}
+
+class ElasticsearchStreamableFlowStage[T <: StreamableIncomingMessage](
+    indexName: String,
+    typeName: String,
+    client: RestClient,
+    settings: ElasticsearchSinkSettings
+) extends GraphStage[FlowShape[T, Future[StreamableIncomingMessagesResult[T]]]] {
+
+  private val in = Inlet[T]("messages")
+  private val out = Outlet[Future[StreamableIncomingMessagesResult[T]]]("result")
+  override val shape = FlowShape(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new TimerGraphStageLogic(shape) with InHandler with OutHandler {
+
+      private var state: State = Idle
+      private val queue = new mutable.Queue[T]()
+      private val failureHandler = getAsyncCallback[(Seq[T], Throwable)](handleFailure)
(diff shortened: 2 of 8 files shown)


takezoe:
> Chain posting to multiple Elastic indexes

You can do it as following:

```scala
val f = ElasticsearchSource("source", "book", """{"match_all": {}}""", ElasticsearchSourceSettings(5))
  .map { message: OutgoingMessage[JsObject] =>
     IncomingMessage(Some(message.id), message.source)
  }
  .alsoTo(ElasticsearchSink("sink1", "book", ElasticsearchSinkSettings(5)))
  .runWith(ElasticsearchSink("sink2", "book", ElasticsearchSinkSettings(5)))
```

Anyway, it might be useful that  `ElasticsearchFlow` will become to propagate all messages with results to the downstream. 

I think updating `ElasticsearchFlow` is better than adding new flow implementation, how do you think?

mbknor:
Hi, and thanks for looking at this PR.

I know about alsoTo, but I do not think it is does what I need.
I need to do something (commit to kafka) AFTER the batch has been written to elastic.

I agree that updating existing ElasticsearchFlow is better, but I didn't know what you think about breaking backward compatibility.

Please also note https://github.com/akka/alpakka/pull/586/commits/c1aaa38453cde788b4486e5b328e4cd4c916659d, since the old ElasticsearchFlow probably have the same issue.

takezoe:
> I need to do something (commit to kafka) AFTER the batch has been written to elastic.

Yes, it requires changes to make possible to do it.

I'm not sure about backward compatibility policy in Alpakka, however Elasticsearch connector is young relatively, so I want to fix it as the correct interface if we have a chance to do it.

> Please also note c1aaa38, since the old ElasticsearchFlow probably have the same issue.

Thanks! Could you create an independent pull request to fix existing `ElasticsearchFlowStage`?

mbknor:
Yes, I will modify the existing one..

But There is one thing to discuss:

I use the trait *StreamableIncomingMessage* without [T], and you use  IncomingMessage[T].

By letting the "user" do the json-serialization BEFORE sending it to Elasticsearch*FlowStage, then Elasticsearch*FlowStage it self does not need to know about how to serialize it.

Jackson and Java can serialize anything of type Object, but spray-json cannot - it needs to know the type at compile-time.

This is possible to achieve, but the amount of generics needed when combining it with the trait (which is needed to get "unknown" stuff through the flow-stage) makes the java code using it really ugly.

What do you think about this?

build.sbt
```diff
@@ -1,3 +1,5 @@
+scalaVersion in ThisBuild := "2.12.4"
```
ennru: We want to continue supporting Scala 2.11, don't set the Scala version in the build files.

mbknor: ok

mbknor:
@takezoe - Can you comment on my last comment?

takezoe:
@mbknor I agree with you about generics complexity in the current implementation, but I prefer type safe interfaces more than string. In my opinion, we should to keep current interfaces. The string version might be good as an optinal solution to use easier.

In addition, your implementation uses a following case class to make possible to judge whether storing message was succeeded in the downstream:

```scala
final case class StreamableIncomingMessagesResult[T <: StreamableIncomingMessage](
  success: Seq[T],
  failed: Seq[T]
) {
  // Java-API
  def getSuccessList(): JavaList[T] = success.toList.asJava
  // Java-API
  def getFailedList(): JavaList[T] = failed.toList.asJava
}
```

I think it's better to keep the order of messages by a sequence of the case class like follows:

```scala
// emit Seq[MessagesResult[T]] to the downstream
case class MessagesResult[T](id: Option[String], source: T, succeeded: Boolean)
```

mbknor:
This PR has been replaced with https://github.com/akka/alpakka/pull/615

Issue #595, opened 2017-11-20
Connector for orientdb


sbcd90:
This issue is created to propose the addition of connector for [Orientdb](http://orientdb.com/orientdb/) in alpakka.

ennru:
OrientDB support was merged with #596

PR #596, opened 2017-11-20
Connector for orientdb #595

sbcd90:
This PR is created to propose the addition of connector for [OrientDB](http://orientdb.com/orientdb/) in alpakka.

Features:
- Provides source & sink for reading & writing to OrientDB Document & Graph databases.
- Supports custom Objects api in javadsl using OrientDB OObjectDatabaseTx.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -16,6 +16,7 @@ lazy val modules: Seq[ProjectReference] = Seq(
   kinesis,
   mongodb,
   mqtt,
+  orientdb,
   s3,
   springWeb,
   simpleCodecs,
@@ -109,6 +110,9 @@ lazy val kinesis = alpakkaProject("kinesis",
 lazy val mongodb = alpakkaProject("mongodb", Dependencies.MongoDb)
 
 lazy val mqtt = alpakkaProject("mqtt", Dependencies.Mqtt)
+
+lazy val orientdb =
+  alpakkaProject("orientdb", Dependencies.OrientDB, fork in Test := true, parallelExecution in Test := false)
 
 lazy val s3 = alpakkaProject("s3", Dependencies.S3)
 
--- a/docs/src/main/paradox/orientdb.md
+++ b/docs/src/main/paradox/orientdb.md
@@ -1,0 +1,134 @@
+# OrientDB Connector
+
+The OrientDB connector provides Akka Stream sources and sinks for OrientDB.
+
+For more information about OrientDB please visit the [official documentation](http://orientdb.com/orientdb/).
+
+## Artifacts
+
+@@dependency [sbt,Maven,Gradle] {
+  group=com.lightbend.akka
+  artifact=akka-stream-alpakka-orientdb_$scalaBinaryVersion$
+  version=$version$
+}
+
+## Usage
+
+Sources, Flows and Sinks provided by this connector need dbUrl & credentials to access to OrientDB.
+
+Scala
+: @@snip ($alpakka$/orientdb/src/test/scala/akka/stream/alpakka/orientdb/OrientDBSpec.scala) { #init-settings }
+
+Java
+: @@snip ($alpakka$/orientdb/src/test/java/akka/stream/alpakka/orientdb/OrientDBTest.java) { #init-settings }
+
+We will also need an @scaladoc[ActorSystem](akka.actor.ActorSystem) and an @scaladoc[ActorMaterializer](akka.stream.ActorMaterializer).
+
+Scala
+: @@snip ($alpakka$/orientdb/src/test/scala/akka/stream/alpakka/orientdb/OrientDBSpec.scala) { #init-mat }
+
+Java
+: @@snip ($alpakka$/orientdb/src/test/java/akka/stream/alpakka/orientdb/OrientDBTest.java) { #init-mat }
+
+This is all preparation that we are going to need.
+
+## ODocument message
+
+Now we can stream messages which contain OrientDB's `ODocument` (in Scala or Java)
+from or to OrientDB by providing the `ODatabaseDocumentTx` to the
+@scaladoc[OrientDBSource](akka.stream.alpakka.orientdb.scaladsl.OrientDBSource$) or the
+@scaladoc[OrientDBSink](akka.stream.alpakka.orientdb.scaladsl.OrientDBSink$).
+
+Scala
+: @@snip ($alpakka$/orientdb/src/test/scala/akka/stream/alpakka/orientdb/OrientDBSpec.scala) { #run-odocument }
+
+Java
+: @@snip ($alpakka$/orientdb/src/test/java/akka/stream/alpakka/orientdb/OrientDBTest.java) { #run-odocument }
+
+
+## Typed messages
+
+Also, it's possible to stream messages which contains any classes. 
+
+Java
+: @@snip ($alpakka$/orientdb/src/test/java/akka/stream/alpakka/orientdb/OrientDBTest.java) { #define-class }
+
+
+Use `OrientDBSource.typed` and `OrientDBSink.typed` to create source and sink instead.
+
+Java
+: @@snip ($alpakka$/orientdb/src/test/java/akka/stream/alpakka/orientdb/OrientDBTest.java) { #run-typed }
+
+
+## Configuration
+
+We can configure the source by `OrientDBSourceSettings`.
+
+Scala (source)
+: @@snip ($alpakka$/orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBSourceSettings.scala) { #source-settings }
+
+| Parameter        | Default | Description                                                                                                              |
+| ---------------- | ------- | ------------------------------------------------------------------------------------------------------------------------ |
+| maxPartitionSize |         | `OrientDBSource` and `OrientDBSink` uses this for initializing DB Connections. |
+| maxPoolSize      |    -1   | `OrientDBSource` and `OrientDBSink` uses this for initializing DB Connections. |
+| skip             |         | `OrientDBSource` uses this property to fetch data from the DB. |
+| limit            |         | `OrientDBSource` uses this property to fetch data from the DB. |
+| dbUrl            |         | url to the OrientDB database. |
+| username         |         | username to connect to OrientDB. |
+| password         |         | password to connect to OrientDB. | 
+
+Also, we can configure the sink by `OrientDBUpdateSettings`.
+
+Scala (sink)
+: @@snip ($alpakka$/orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBUpdateSettings.scala) { #sink-settings }
+
+| Parameter           | Default | Description                                                                                            |
+| ------------------- | ------- | ------------------------------------------------------------------------------------------------------ |
+| maxPartitionSize |         | `OrientDBSource` and `OrientDBSink` uses this for initializing DB Connections. |
+| maxPoolSize      |    -1   | `OrientDBSource` and `OrientDBSink` uses this for initializing DB Connections. |
+| maxRetry         |     1   | `OrientDBSink` uses this for retrying write operations to OrientDB. |
+| retryInterval    |  5000   | `OrientDBSink` uses this for retrying write operations to OrientDB. |
+| bufferSize       |         | `OrientDBSink` uses this for retrieving data from DB. |
+| dbUrl            |         | url to the OrientDB database. |
+| username         |         | username to connect to OrientDB. |
+| password         |         | password to connect to OrientDB. | 
+
+## Using OrientDB as a Flow
+
+You can also build flow stages. The API is similar to creating Sinks.
+
+Scala (flow)
+: @@snip ($alpakka$/orientdb/src/test/scala/akka/stream/alpakka/orientdb/OrientDBSpec.scala) { #run-flow }
+
+Java (flow)
+: @@snip ($alpakka$/orientdb/src/test/java/akka/stream/alpakka/orientdb/OrientDBTest.java) { #run-flow }
+
+### Passing data through OrientDBFlow
+
+When streaming documents from Kafka, you might want to commit to Kafka **AFTER** the document has been written to OrientDB.
+
+Scala
+: @@snip ($alpakka$/orientdb/src/test/scala/akka/stream/alpakka/orientdb/OrientDBSpec.scala) { #kafka-example }
+
+Java
+: @@snip ($alpakka$/orientdb/src/test/java/akka/stream/alpakka/orientdb/OrientDBTest.java) { #kafka-example } 
+
+## Running the example code
+
+The code in this guide is part of runnable tests of this project. You are welcome to edit the code and run it in sbt.
+
+  > Test code requires OrientDB server running in the background. You can start one quickly using docker:
+  >		  
+  > `docker run --rm -p 2424:2424 orientdb:latest`
+
+Scala
+:   ```
+    sbt
+    > orientdb/testOnly *.OrientDBSpec
+    ```
+
+Java
+:   ```
+    sbt
+    > orientdb/testOnly *.OrientDBTest
+    ```
--- a/orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBFlowStage.scala
+++ b/orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBFlowStage.scala
@@ -1,0 +1,215 @@
+/*
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.orientdb
+
+import akka.NotUsed
+import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
+import akka.stream.stage._
+import com.orientechnologies.orient.`object`.db.OObjectDatabaseTx
+import com.orientechnologies.orient.core.db.ODatabaseRecordThreadLocal
+import com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx
+import com.orientechnologies.orient.core.record.ORecord
+import com.orientechnologies.orient.core.record.impl.ODocument
+import com.orientechnologies.orient.core.tx.OTransaction
+
+import scala.collection.mutable
+import scala.concurrent.Future
+
+object OIncomingMessage {
+  // Apply method to use when not using passThrough
+  def apply[T](oDocument: T): OIncomingMessage[T, NotUsed] =
+    OIncomingMessage(oDocument, NotUsed)
+
+  // Java-api - without passThrough
+  def create[T](oDocument: T): OIncomingMessage[T, NotUsed] =
+    OIncomingMessage(oDocument, NotUsed)
+
+  // Java-api - with passThrough
+  def create[T, C](oDocument: T, passThrough: C) =
+    OIncomingMessage(oDocument, passThrough)
+}
+
+final case class OIncomingMessage[T, C](oDocument: T, passThrough: C)
+
+private[orientdb] class OrientDBFlowStage[T, C, R](
+    className: String,
+    settings: OrientDBUpdateSettings,
+    pusher: Seq[OIncomingMessage[T, C]] => R,
+    clazz: Option[Class[T]]
+) extends GraphStage[FlowShape[OIncomingMessage[T, C], Future[R]]] {
+
+  private val in = Inlet[OIncomingMessage[T, C]]("messages")
+  private val out = Outlet[Future[R]]("failed")
+  override val shape = FlowShape(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new TimerGraphStageLogic(shape) with InHandler with OutHandler {
+
+      private val queue = new mutable.Queue[OIncomingMessage[T, C]]()
+      private val failureHandler = getAsyncCallback[(Seq[OIncomingMessage[T, C]], Throwable)](handleFailure)
+      private val responseHandler = getAsyncCallback[(Seq[OIncomingMessage[T, C]], Option[String])](handleResponse)
+      private var failedMessages: Seq[OIncomingMessage[T, C]] = Nil
+      private var retryCount: Int = 0
+
+      private var client: ODatabaseDocumentTx = _
+      private var oObjectClient: OObjectDatabaseTx = _
+
+      override def preStart(): Unit = {
+        client = settings.oDatabasePool.acquire()
+        oObjectClient = new OObjectDatabaseTx(client)
+        pull(in)
+      }
+
+      override def postStop(): Unit = {
+        oObjectClient.close()
+        client.close()
+      }
+
+      private def tryPull(): Unit =
+        if (queue.size < settings.bufferSize && !isClosed(in) && !hasBeenPulled(in)) {
+          pull(in)
+        }
+
+      override def onTimer(timerKey: Any): Unit = {
+        sendOSQLBulkInsertRequest(failedMessages)
+        failedMessages = Nil
+      }
+
+      private def handleFailure(args: (Seq[OIncomingMessage[T, C]], Throwable)): Unit = {
+        val (messages, exception) = args
+        if (retryCount >= settings.maxRetry) {
+          failStage(exception)
+        } else {
+          retryCount = retryCount + 1
+          failedMessages = messages
+          scheduleOnce(NotUsed, settings.retryInterval)
+        }
+      }
+
+      private def handleSuccess(): Unit =
+        completeStage()
+
+      private def handleResponse(args: (Seq[OIncomingMessage[T, C]], Option[String])): Unit = {
+        retryCount = 0
+        val (messages, error) = args
+
+        val failedMessages = messages.flatMap {
+          case message =>
+            if (error.isEmpty) {
+              None
+            } else {
+              Some(message)
+            }
+        }
+
+        val nextMessages = (1 to settings.bufferSize).flatMap { _ =>
+          queue.dequeueFirst(_ => true)
+        }
+
+        if (nextMessages.isEmpty) {
+          handleSuccess()
+        } else {
+          sendOSQLBulkInsertRequest(nextMessages)
+        }
+
+        push(out, Future.successful(pusher(failedMessages)))
+      }
+
+      private def sendOSQLBulkInsertRequest(messages: Seq[OIncomingMessage[T, C]]): Unit =
+        try {
+          ODatabaseRecordThreadLocal.instance().set(client)
+          if (clazz.isEmpty) {
+            if (!client.getMetadata.getSchema.existsClass(className)) {
+              client.getMetadata.getSchema.createClass(className)
+            }
+            client.begin(OTransaction.TXTYPE.OPTIMISTIC)
+
+            var faultyMessages: List[OIncomingMessage[T, C]] = List()
+            var successfulMessages: List[OIncomingMessage[T, C]] = List()
+            messages.foreach {
+              case OIncomingMessage(oDocument: ODocument, passThrough: C) =>
+                val document = new ODocument()
+                oDocument
+                  .asInstanceOf[ODocument]
+                  .fieldNames()
+                  .zip(oDocument.asInstanceOf[ODocument].fieldValues())
+                  .foreach {
+                    case (fieldName, fieldVal) =>
+                      document.field(fieldName, fieldVal)
+                      ()
+                  }
+                document.setClassName(className)
+                client.save(document)
+                successfulMessages = successfulMessages ++ List(
+                  OIncomingMessage(oDocument.asInstanceOf[T], passThrough)
+                )
+                ()
+              case OIncomingMessage(oRecord: ORecord, passThrough: C) =>
+                client.save(oRecord)
+                successfulMessages = successfulMessages ++ List(OIncomingMessage(oRecord.asInstanceOf[T], passThrough))
+                ()
+              case OIncomingMessage(others: AnyRef, passThrough: C) =>
+                faultyMessages = faultyMessages ++ List(OIncomingMessage(others.asInstanceOf[T], passThrough))
+                ()
+            }
+            client.commit()
+            if (faultyMessages.nonEmpty) {
+              responseHandler.invoke((faultyMessages, Some("Records are invalid OrientDB Records")))
+            } else {
+              emit(out, Future.successful(pusher(successfulMessages)))
+              responseHandler.invoke((Seq(), None))
+            }
+          } else {
+            client.setDatabaseOwner(oObjectClient)
+            oObjectClient.getEntityManager.registerEntityClass(
+              clazz.getOrElse(throw new RuntimeException("Typed stream class is invalid"))
+            )
+
+            var faultyMessages: List[OIncomingMessage[T, C]] = List()
+            var successfulMessages: List[OIncomingMessage[T, C]] = List()
+            messages.foreach {
+              case OIncomingMessage(typeRecord: T, passThrough: C) =>
+                oObjectClient.save(typeRecord)
+                successfulMessages = successfulMessages ++ List(OIncomingMessage(typeRecord, passThrough))
+                ()
+              case OIncomingMessage(others: AnyRef, passThrough: C) =>
+                faultyMessages = faultyMessages ++ List(OIncomingMessage(others.asInstanceOf[T], passThrough))
+                ()
+            }
+
+            if (faultyMessages.nonEmpty) {
+              responseHandler.invoke((faultyMessages, Some("Records are invalid OrientDB Records")))
+            } else {
+              emit(out, Future.successful(pusher(successfulMessages)))
+              responseHandler.invoke((Seq(), None))
+            }
+          }
+        } catch {
+          case exception: Exception =>
+            failureHandler.invoke((messages, exception))
+        }
+
+      setHandlers(in, out, this)
+
+      override def onPull(): Unit = tryPull()
+
+      override def onPush(): Unit = {
+        val message = grab(in)
+        queue.enqueue(message)
+
+        val messages = (1 to settings.bufferSize).flatMap { _ =>
+          queue.dequeueFirst(_ => true)
+        }
+        sendOSQLBulkInsertRequest(messages)
+
+        tryPull()
+      }
+
+      override def onUpstreamFailure(exception: Throwable): Unit =
+        failStage(exception)
+
+      override def onUpstreamFinish(): Unit = handleSuccess()
+    }
+}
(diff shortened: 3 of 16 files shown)


lightbend-cla-validator:
Hi @sbcd90,

johanandren:
Fixes #595

lightbend-cla-validator:
Hi @sbcd90,

sbcd90:
I signed the CLA

sbcd90:
Hi @johanandren , some builds have tests for orientdb connector passed. some have not executed the tests & failed. Can you please let me know if the build results are correct?

ennru:
Scalaformat is complaining:

sbcd90:
Hello @ennru , 
I resolved the merge conflicts. The test cases for OrientDB Connector pass as well.

ennru:
The build results are inconsistent, it seems to work for Scala 2.11 with Akka 2.4, but fails for the others.
Could you add some documentation on how to use this connector, please.

orientdb/src/main/scala/akka/stream/alpakka/orientdb/javadsl/OrientDBSinkSettings.scala
```diff
@@ -0,0 +1,92 @@
...
+      this.maxRetry,
+      this.retryInterval,
+      this.bufferSize,
+      dbUrl,
+      username,
+      password
+    )
+
+  private[javadsl] def asScala: ScalaOrientDBSinkSettings =
+    new ScalaOrientDBSinkSettings(
```
ennru: Why did you implement settings for both Java and Scala? Could this be done in one case class containing `withXxx` methods?

project/Dependencies.scala
```diff
@@ -155,6 +155,13 @@ object Dependencies {
     )
   )
 
+  val OrientDB = Seq(
+    libraryDependencies ++= Seq(
+      "com.orientechnologies" % "orientdb-graphdb" % "2.2.30",
```
ennru: Could you annotate with the applicable license, please.

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBSinkSettings.scala
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright (C) 2016-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.orientdb
+
+//#sink-settings
+final class OrientDBSinkSettings(var maxPartitionSize: Int = Runtime.getRuntime.availableProcessors(),
```
ennru: Why are these fields vars? Would a case class be preferrable?

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBSourceSettings.scala
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright (C) 2016-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.orientdb
+
+//#source-settings
+final class OrientDBSourceSettings(var maxPartitionSize: Int,
```
ennru: Same questions as for Sink settings.

orientdb/src/main/scala/akka/stream/alpakka/orientdb/javadsl/OrientDBFlow.scala
```diff
@@ -0,0 +1,50 @@
...
+    akka.stream.scaladsl.Flow
+      .fromGraph(
+        new OrientDBFlowStage[ODocument, JavaList[IncomingMessage[ODocument]]](
+          className,
+          settings,
+          _.asJava,
+          None
+        )
+      )
+      .mapAsync(1)(identity)
```
ennru: What do you achieve with this?

orientdb/src/main/scala/akka/stream/alpakka/orientdb/javadsl/OrientDBSource.scala
```diff
@@ -0,0 +1,55 @@
...
+
+import akka.NotUsed
+import akka.stream.alpakka.orientdb._
+import akka.stream.javadsl.Source
+import com.orientechnologies.orient.core.record.impl.ODocument
+
+object OrientDBSource {
+
+  /**
+   * Java API: creates a [[OrientDBSourceStage]] that consumes as ODocument
```
ennru: Do sources *consume* documents?

orientdb/src/main/scala/akka/stream/alpakka/orientdb/javadsl/OrientDBSource.scala
```diff
@@ -0,0 +1,55 @@
...
+      )
+    )
+
+  private class ODocumentMessageReader[T] extends MessageReader[T] {
+
+    override def convert(oDocs: List[T]): OSQLResponse[T] =
+      try {
+        OSQLResponse(None, Some(OSQLResult(oDocs.map(OutgoingMessage(_)))))
+      } catch {
+        case exception: Exception => OSQLResponse(Some(exception.getMessage), None)
```
ennru: Maybe `exception.toString` would be more helpful. E.g. NPE has `null` as message.

docs/src/main/paradox/orientdb.md
```diff
@@ -0,0 +1,120 @@
...
+
+Scala
+: @@snip ($alpakka$/orientdb/src/test/scala/akka/stream/alpakka/orientdb/OrientDBSpec.scala) { #init-mat }
+
+Java
+: @@snip ($alpakka$/orientdb/src/test/java/akka/stream/alpakka/orientdb/OrientDBTest.java) { #init-mat }
+
+This is all preparation that we are going to need.
+
+### ODocument message
```
ennru: Put the sections on level '##' and they'll show in the right-hand side overview.

orientdb/src/main/scala/akka/stream/alpakka/orientdb/javadsl/OrientDBFlow.scala
sbcd90: This just resolves the future.
Like apply returns `Flow[IncomingMessage[ODocument], Seq[IncomingMessage[ODocument]], NotUsed]` but if the future is not resolved, then it would return `Flow[IncomingMessage[ODocument], Future[Seq[IncomingMessage[ODocument]]], NotUsed]`

sbcd90:
Hello @ennru, I refactored the code to handle all your remarks. Please have a look.

ennru:
Some builds keep failing (even though Scala 2.11 / Akka 2.4 goes Ok).
It might be timing-related.

E.g. this test failed:
Test akka.stream.alpakka.orientdb.OrientDBTest.oDocObjectStream failed: java.lang.AssertionError: expected:<[Akka Concurrency, Akka in Action, Effective Akka, Learning Scala, Programming in Scala, Scala Puzzlers, Scala for Spark in Production]> but was:<[]>, took 2.653 sec

sbcd90:
Hello @ennru ,
I have fixed the build failure issue & now all the builds ( Scala 2.11 / Akka 2.4, Scala 2.12 / Akka 2.4, Scala 2.12 / Akka 2.5 ) goes OK.
Can you please let me know if any further changes are required.

ennru:
I'm afraid all build stall. I can't see how this is related to your changes, but I won't merge them if we don't find the cause.

sbcd90:
Hello @ennru , 
I tried to check whether the stalling of builds is due to the changes in the PR. So, what I did is,
I commented out the other tests in `build.sbt` from running & ran only the OrientDB connector test cases.
https://github.com/sbcd90/alpakka/blob/bdb7ed942e07983b0ab47a98caa4fd2586ea2773/build.sbt#L5-#L30

Now all builds run successfully. I can further look exactly for which connector the builds are getting stalled. I feel the stalling of builds is not because of OrientDB connector. Please let me know your views.

raboof:
I re-triggered the builds on travis and they were all green - I suspect this might have been some travis/infrastructure glitch?

ennru:
Ok, good to see OrionDB tests go OK without the rest of Alpakka tests. 
Could there be any interaction by using the same ports? A few tests start servers from the test code.
Can you enable the tests again so we can see the whole thing succeed, please.

orientdb/src/main/scala/akka/stream/alpakka/orientdb/javadsl/OrientDBFlow.scala
ennru: Ok

ennru:
Right now both ftp and file modules are commented out. One build configuration fails, even after multiple re-triggering.
We need to understand what makes it so hard to get the build through with this new connector in the project.

ennru:
I tried this PR on my machine and get similar problems as Travis: some `Await`s time out when all tests are run, and the Cassandra tests fail for other reasons.

Please change the timeouts in `AmqpConnectorsSpec` and DyamoDb's `ExampleTest` and `ExampleSpec`. 

We might need to switch of parallel tests for all of Alpakka, or run all modules' tests in forked JVMs.

sbcd90:
Hello @ennru , I changed the timeouts like in `AmqpConnectorsSpec` and DyamoDb's `ExampleTest` and `ExampleSpec`.
But I see the builds stalled still. Any ideas you can suggest please?

ennru:
Hmm, I've no good idea. We haven't seen anything like this in other PRs. Spurious build failures happen quite often, but no stalling builds.
We could experiment with setting `fork in Test := true` and/or `parallelExecution in Test := false` for all projects.

ennru:
It failed for non Scalaformat formatted code in build.sbt. Try `test:compile`

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBSinkSettings.scala
```diff
@@ -0,0 +1,80 @@
...
+
+package akka.stream.alpakka.orientdb
+
+//#sink-settings
+final case class OrientDBSinkSettings(maxPartitionSize: Int = Runtime.getRuntime.availableProcessors(),
+                                      maxPoolSize: Int = -1,
+                                      maxRetry: Int = 1,
+                                      retryInterval: Int = 5000,
+                                      bufferSize: Int = 10,
+                                      dbUrl: String,
```
juanjoDiaz: Declaring non-optional parameters after optional parameters doesn't sound good to me.

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBSourceSettings.scala
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+package akka.stream.alpakka.orientdb
+
+//#source-settings
+final case class OrientDBSourceSettings(maxPartitionSize: Int = Runtime.getRuntime.availableProcessors(),
+                                        maxPoolSize: Int = -1,
+                                        skip: Int = 0,
+                                        limit: Int = 10,
+                                        dbUrl: String,
```
juanjoDiaz: Declaring non-optional parameters after optional parameters doesn't sound good to me.

build.sbt
```diff
@@ -47,30 +48,33 @@ lazy val alpakka = project
       """.stripMargin
   )
 
-lazy val amqp = alpakkaProject("amqp", Dependencies.Amqp)
+lazy val amqp = alpakkaProject("amqp", Dependencies.Amqp, fork in Test := true, parallelExecution in Test := false)
```
juanjoDiaz: Something is wrong if adding a new connector implies modifying the way we test all the existing ones :)

docs/src/main/paradox/orientdb.md
```diff
@@ -0,0 +1,120 @@
...
+Scala (flow)
+: @@snip ($alpakka$/orientdb/src/test/scala/akka/stream/alpakka/orientdb/OrientDBSpec.scala) { #run-flow }
+
+Java (flow)
+: @@snip ($alpakka$/orientdb/src/test/java/akka/stream/alpakka/orientdb/OrientDBTest.java) { #run-flow }
+
+## Running the example code
+
+The code in this guide is part of runnable tests of this project. You are welcome to edit the code and run it in sbt.
+
```
juanjoDiaz: Please add a couple liens here about docker dependencies as in #664

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBSinkSettings.scala
```diff
@@ -0,0 +1,80 @@
...
+                                      username: String,
+                                      password: String)
+//#sink-settings
+{
+  def this(dbUrl: String, username: String, password: String) =
+    this(Runtime.getRuntime.availableProcessors(), -1, -1, 5000, 10, dbUrl, username, password)
+
+  def withMaxPartitionSize(maxPartitionSize: Int): OrientDBSinkSettings =
+    new OrientDBSinkSettings(
```
juanjoDiaz: this is a case class. You can simply do `copy(maxPartitionSize = maxPartitionSize)` to do the same.

And the same to al the other withXXX methods.

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBFlowStage.scala
```diff
@@ -0,0 +1,211 @@
...
+          }
+        } else {
+          sendOSQLBulkInsertRequest(nextMessages)
+        }
+
+        push(out, Future.successful(pusher(failedMessages)))
+      }
+
+      private def sendOSQLBulkInsertRequest(messages: Seq[IncomingMessage[T]]): Unit = {
+        val oDatabase = new OPartitionedDatabasePool(settings.dbUrl, settings.username, settings.password)
```
juanjoDiaz: Shouldn't this (creating & getting the connection) be done in the `preStart` method?
And similar, having a `postStop` method to clean up stuff?

sbcd90:
Hello @ennru , @juanjoDiaz , i made the fixes as suggested. Reverted the changes that I made for running the other connectors. Unfortunately, the build still stalls & I have no clue.

ennru:
I don't know what is going on. Does the Travis VM run out of memory?
The sudo-enabled VM gets more memory: https://docs.travis-ci.com/user/reference/overview/#stq=&stp=0

juanjoDiaz:
It seems that OrientDB allocated memory in really big chunks (http://orientdb.com/memory-consumed-by-orientdb/).

Not sure if it will fix this but you might try to:
* Reduce that chunk (there should be some variable that you can override in docker compose).
* Limit the container memory to something reasonable.

build.sbt
```diff
@@ -54,13 +55,16 @@ lazy val awslambda = alpakkaProject("awslambda",
                                     // For mockito https://github.com/akka/alpakka/issues/390
                                     parallelExecution in Test := false)
 
-lazy val azureStorageQueue = alpakkaProject("azure-storage-queue", Dependencies.AzureStorageQueue)
+lazy val azureStorageQueue =
+  alpakkaProject("azure-storage-queue", Dependencies.AzureStorageQueue)
```
juanjoDiaz: Could you reformat this file as it was (except for the orientDB line obviously)?
There is no need to re-format the code if no changes are made

docs/src/main/paradox/orientdb.md
```diff
@@ -0,0 +1,124 @@
...
+  > Test code requires OrientDB server running in the background. You can start one quickly using docker:
+  
+  >		  > `docker run --rm -p 12424:2424 orientdb:latest`
```
juanjoDiaz: the whole `		  > ` shouldn't be there

docs/src/main/paradox/orientdb.md
```diff
@@ -0,0 +1,124 @@
...
+
+Java (flow)
+: @@snip ($alpakka$/orientdb/src/test/java/akka/stream/alpakka/orientdb/OrientDBTest.java) { #run-flow }
+
+## Running the example code
+
+The code in this guide is part of runnable tests of this project. You are welcome to edit the code and run it in sbt.
+
+  > Test code requires OrientDB server running in the background. You can start one quickly using docker:
+
```
juanjoDiaz: the is a `>` missing in this line to be consistent with the others

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBSinkSettings.scala
```diff
@@ -0,0 +1,71 @@
...
+                                      retryInterval: Int = 5000,
+                                      bufferSize: Int = 10)
+//#sink-settings
+{
+
+  def this(oDatabasePool: OPartitionedDatabasePool) =
+    this(oDatabasePool, Runtime.getRuntime.availableProcessors(), -1, -1, 5000, 10)
+
+  def withMaxPartitionSize(maxPartitionSize: Int): OrientDBSinkSettings =
+    copy(
```
juanjoDiaz: You can simply do `copy(maxPartitionSize = maxPartitionSize)` all the other fields are automatically copied (as the method name says).

The same applies to all the other `withXXX` methods

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBSourceStage.scala
```diff
@@ -0,0 +1,158 @@
...
+    override def end(): Unit = {
+      if (oDocs.nonEmpty) {
+        skipAndLimit.skip = skipAndLimit.skip + skipAndLimit.limit
+      }
+      onSuccess(oDocs)
+    }
+  }
+}
+
+private[orientdb] object SkipAndLimit {
```
juanjoDiaz: And object is static. This is dangerous because if you configure several connectors in the same application you might be innadvertedly overriding the config for one connector when creating the next.

I'd use a case class instead:
```
private[orientdb] case class SkipAndLimit (skip: Int, limit: Int)
```

if you can it to be use as `SkipAndLimit` without arguments you can add defaults
```
private[orientdb] case class SkipAndLimit (skip: Int = 0, limit: Int = 0)
```

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBFlowStage.scala
```diff
@@ -0,0 +1,209 @@
...
+import com.orientechnologies.orient.core.db.OPartitionedDatabasePool
+import com.orientechnologies.orient.core.record.ORecord
+import com.orientechnologies.orient.core.record.impl.ODocument
+import com.orientechnologies.orient.core.tx.OTransaction
+
+import scala.collection.mutable
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+final case class IncomingMessage[T](oDocument: T)
```
juanjoDiaz: `IncomingMessage ` and `OutgoingMessage` are exactly the same. Is there any reason to have both?

Also, I think that they are switched (at least not consistent with other connectors):
The Source should emit `IncomingMessage` and the sink should ingest `OutgoingMessage`.

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBSinkSettings.scala
```diff
@@ -0,0 +1,71 @@
...
+final case class OrientDBSinkSettings(oDatabasePool: OPartitionedDatabasePool,
+                                      maxPartitionSize: Int = Runtime.getRuntime.availableProcessors(),
+                                      maxPoolSize: Int = -1,
+                                      maxRetry: Int = 1,
+                                      retryInterval: Int = 5000,
+                                      bufferSize: Int = 10)
+//#sink-settings
+{
+
+  def this(oDatabasePool: OPartitionedDatabasePool) =
```
juanjoDiaz: This constructor will cause issues with the default constructor because scala won't be able tu figure out which one to use when having `OrientDBSinkSettings(myODatabasePool)`

It shouldn't be here.
If you need for the Java API, just add a method call `create` as other connectors do.
```
def create(oDatabasePool: OPartitionedDatabasePool): OrientDBSinkSettings = this(oDatabasePool)
```

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBFlowStage.scala
```diff
@@ -0,0 +1,209 @@
...
+          }
+        } else {
+          sendOSQLBulkInsertRequest(nextMessages)
+        }
+
+        push(out, Future.successful(pusher(failedMessages)))
+      }
+
+      private def sendOSQLBulkInsertRequest(messages: Seq[IncomingMessage[T]]): Unit = {
+        val client = settings.oDatabasePool.acquire()
```
juanjoDiaz: I still find weird that you acquire the connection on every insert instead of opening it on `preStart` and closing it on `postStop`

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBFlowStage.scala
```diff
@@ -0,0 +1,209 @@
...
+
+      override def onPush(): Unit = {
+        val message = grab(in)
+        queue.enqueue(message)
+
+        state match {
+          case Idle =>
+            state = Sending
+            val messages = (1 to settings.bufferSize).flatMap { _ =>
+              queue.dequeueFirst(_ => true)
```
juanjoDiaz: Why don't you use `queue.dequeue()`?

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBFlowStage.scala
```diff
@@ -0,0 +1,209 @@
...
+
+      setHandlers(in, out, this)
+
+      override def onPull(): Unit = tryPull()
+
+      override def onPush(): Unit = {
+        val message = grab(in)
+        queue.enqueue(message)
+
+        state match {
```
juanjoDiaz: Do you really need this state machine?
Can you simply call push after the previous insertion is done?

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBSourceSettings.scala
```diff
@@ -0,0 +1,32 @@
...
+//#source-settings
+final case class OrientDBSourceSettings(oDatabasePool: OPartitionedDatabasePool,
+                                        maxPartitionSize: Int = Runtime.getRuntime.availableProcessors(),
+                                        maxPoolSize: Int = -1,
+                                        skip: Int = 0,
+                                        limit: Int = 10)
+//#source-settings
+{
+
+  def this(oDatabasePool: OPartitionedDatabasePool) =
```
juanjoDiaz: Same as above. Use `create`

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBSourceSettings.scala
```diff
@@ -0,0 +1,32 @@
...
+    this(oDatabasePool, Runtime.getRuntime.availableProcessors(), -1, 0, 10)
+
+  def withMaxPartitionSize(maxPartitionSize: Int): OrientDBSourceSettings =
+    copy(this.oDatabasePool, maxPartitionSize, this.maxPoolSize, this.skip, this.limit)
```
juanjoDiaz: Same as above. No need to pass all the properties

orientdb/src/main/scala/akka/stream/alpakka/orientdb/scaladsl/OrientDBSink.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import scala.concurrent.Future
+
+object OrientDBSink {
+
+  /**
+   * Scala API: creates a sink based on [[OrientDBFlowStage]] that accepts as ODocument
+   */
+  def apply(className: String, settings: OrientDBSinkSettings): Sink[IncomingMessage[ODocument], Future[Done]] =
+    OrientDBFlow.apply(className, settings).toMat(Sink.ignore)(Keep.right)
```
juanjoDiaz: No need to use apply here

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBSourceStage.scala
```diff
@@ -0,0 +1,158 @@
...
+  override val shape = SourceShape(out)
+
+  override def createLogic(inheritedAttributes: Attributes) =
+    new OrientDBSourceLogic[T](className,
+                               query,
+                               settings,
+                               out,
+                               shape,
+                               reader,
+                               SkipAndLimit(settings.skip, settings.limit),
```
juanjoDiaz: Why using this wrapper object instead of simply reading this properties from the settings?

ennru:
@juanjoDiaz Thank you for looking into this. It must be the (at least) 1 GB of memory taken by OrionDB, as the Travis VM get 4 GB max.

I agree that a connection open/close per message is no good choice.

sbcd90:
Hi @juanjoDiaz , @ennru , I have fixed the connection open/close per message issue now. I'm using the
`preStart` and `postStop` methods now.

ennru:
The use of start and stop looks good.
The build still stalls, probably because of the memory consumption. Can you try to configure the OrientDB memory to a reasonable size?

sbcd90:
Hello @ennru , The builds pass finally. I reduced the memory. Can this PR be approved now?

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBSinkSettings.scala
```diff
@@ -0,0 +1,36 @@
...
+package akka.stream.alpakka.orientdb
+
+import com.orientechnologies.orient.core.db.OPartitionedDatabasePool
+
+//#sink-settings
+final case class OrientDBSinkSettings(oDatabasePool: OPartitionedDatabasePool,
+                                      maxPartitionSize: Int = Runtime.getRuntime.availableProcessors(),
+                                      maxPoolSize: Int = -1,
+                                      maxRetry: Int = 1,
+                                      retryInterval: Int = 5000,
```
ennru: Can you make `retryInterval` a `Duration`, please. ...and adopt the Java `withRetries` to take Long and TimeUnit.

docs/src/main/paradox/orientdb.md
```diff
@@ -0,0 +1,124 @@
...
+: @@snip ($alpakka$/orientdb/src/test/scala/akka/stream/alpakka/orientdb/OrientDBSpec.scala) { #init-mat }
+
+Java
+: @@snip ($alpakka$/orientdb/src/test/java/akka/stream/alpakka/orientdb/OrientDBTest.java) { #init-mat }
+
+This is all preparation that we are going to need.
+
+## ODocument message
+
+Now we can stream messages which contains OrientDB's `ODocument` (in Scala or Java)
```
ennru: "contain" without the s

docs/src/main/paradox/orientdb.md
```diff
@@ -0,0 +1,124 @@
...
+from or to OrientDB where we have access to by providing the `ODatabaseDocumentTx` to the
```
ennru: remove "where we have acccess to"

docs/src/main/paradox/orientdb.md
```diff
@@ -0,0 +1,124 @@
...
+@scaladoc[OrientDBSource](akka.stream.alpakka.orientdb.scaladsl.OrientDBSource$) or the
+@scaladoc[OrientDBSink](akka.stream.alpakka.orientdb.scaladsl.OrientDBSink$).
+
+Scala
+: @@snip ($alpakka$/orientdb/src/test/scala/akka/stream/alpakka/orientdb/OrientDBSpec.scala) { #run-odocument }
```
ennru: Do you think your `run-odocument` example moving data from OrientDB to OrientDB will be a common use case?
It might be a bit overwhelming for a user new to Akka Streams.

docs/src/main/paradox/orientdb.md
```diff
@@ -0,0 +1,124 @@
...
+
+Java
+: @@snip ($alpakka$/orientdb/src/test/java/akka/stream/alpakka/orientdb/OrientDBTest.java) { #run-odocument }
+
+
+## Typed messages
+
+Also, it's possible to stream messages which contains any classes. 
+
+Java
```
ennru: Is it intentional to just show Java code here?

orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBSourceStage.scala
```diff
@@ -0,0 +1,160 @@
...
+import akka.stream.stage.{GraphStage, GraphStageLogic, OutHandler}
+import com.orientechnologies.orient.`object`.db.OObjectDatabaseTx
+import com.orientechnologies.orient.core.command.OCommandResultListener
+import com.orientechnologies.orient.core.db.ODatabaseRecordThreadLocal
+import com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx
+import com.orientechnologies.orient.core.sql.query.{OSQLNonBlockingQuery, OSQLSynchQuery}
+
+import scala.collection.JavaConverters._
+
+final case class Message[T](oDocument: T)
```
ennru: I wonder if `Message` as a class name is a bit too generic? `OMessage`, `OData` ?
You may consider adding support for a `passThrough`, as well. E.g. the Elasticseach connector got support for it recently and other flows see the same need.

docs/src/main/paradox/orientdb.md
```diff
@@ -0,0 +1,124 @@
...
+| ---------------- | ------- | ------------------------------------------------------------------------------------------------------------------------ |
+| maxPartitionSize |         | `OrientDBSource` and `OrientDBSink` uses this for initializing DB Connections. |
+| maxPoolSize      |    -1   | `OrientDBSource` and `OrientDBSink` uses this for initializing DB Connections. |
+| skip             |         | `OrientDBSource` uses this property to fetch data from the DB. |
+| limit            |         | `OrientDBSource` uses this property to fetch data from the DB. |
+| dbUrl            |         | url to the OrientDB database. |
+| username         |         | username to connect to OrientDB. |
+| password         |         | password to connect to OrientDB. | 
+
+Also, we can configure the sink by `OrientDBSinkSettings`.
```
ennru: Is there a better name for the sink-and-flow-settings?

orientdb/src/main/scala/akka/stream/alpakka/orientdb/scaladsl/OrientDBFlow.scala
```diff
@@ -0,0 +1,31 @@
...
+import akka.stream.alpakka.orientdb._
+import akka.stream.scaladsl.Flow
+import com.orientechnologies.orient.core.record.impl.ODocument
+
+object OrientDBFlow {
+
+  /**
+   * Scala API: creates a [[OrientDBFlowStage]] that accepts as ODocument
+   */
+  def apply(
```
ennru: In my taste this should be `create` just as in the Java API. It makes it more natural in case we add other constructors.

orientdb/src/main/scala/akka/stream/alpakka/orientdb/scaladsl/OrientDBSink.scala
```diff
@@ -0,0 +1,22 @@
...
+import akka.stream.alpakka.orientdb._
+import akka.stream.scaladsl.{Keep, Sink}
+import com.orientechnologies.orient.core.record.impl.ODocument
+
+import scala.concurrent.Future
+
+object OrientDBSink {
+
+  /**
+   * Scala API: creates a sink based on [[OrientDBFlowStage]] that accepts as ODocument
```
ennru: The `OrientDBFlowStage` won't show in the Scaladoc as it is private. (I know other connectors do the same thing.)

orientdb/src/main/scala/akka/stream/alpakka/orientdb/scaladsl/OrientDBSource.scala
```diff
@@ -0,0 +1,38 @@
...
+import com.orientechnologies.orient.core.record.impl.ODocument
+
+object OrientDBSource {
+
+  /**
+   * Scala API: creates a [[OrientDBSourceStage]] that produces as ODocument
+   */
+  def apply(className: String,
+            settings: OrientDBSourceSettings,
+            query: String = null): Source[Message[ODocument], NotUsed] =
```
ennru: If query is optional, should it be an `Option`?

orientdb/src/test/java/akka/stream/alpakka/orientdb/OrientDBTest.java
```diff
@@ -0,0 +1,328 @@
...
+  private static String password = "root";
+  //#init-settings
+
+  private static String source = "source1";
+  private static String sink1 = "sink1";
+  private static String sink2 = "sink2";
+  private static String sink3 = "sink3";
+
+  //#define-class
+  public class source1 {
```
ennru: Is it intentional to have a non-capitalized class name?

orientdb/src/test/java/akka/stream/alpakka/orientdb/OrientDBTest.java
```diff
@@ -0,0 +1,328 @@
...
+      ))
+      .runWith(Sink.seq(), materializer);
+    //#run-flow
+
+    f1.toCompletableFuture().get();
+
+    // Assert docs in sink3
+    CompletionStage<List<String>> f2 = OrientDBSource.create(
+      sink3,
+      new OrientDBSourceSettings(oDatabase,
```
ennru: Please create a companion object so that even Java users can get away with just setting the database.

docs/src/main/paradox/orientdb.md
sbcd90: I just wanted to demonstrate how custom classes could be used for sending messages.

docs/src/main/paradox/orientdb.md
sbcd90: Can you please suggest a suitable name?

orientdb/src/test/java/akka/stream/alpakka/orientdb/OrientDBTest.java
sbcd90: Yes, the corresponding orientdb class name is `source1`.

sbcd90:
Hi @ennru,
The builds fail again due to `Amqp` connector test failure. Can you please let me know if this is an issue with my PR? I think its not the case.

ennru:
The Amqp failure seems to occur in master, as well. Not your fault.

docs/src/main/paradox/orientdb.md
ennru: How does `OrientDbUpdateSetting` sound?

raboof:
Hmm, the latest build (https://travis-ci.org/akka/alpakka/jobs/335496810) did fail with something that seems related to orientdb (search for OPartitionedDatabasePool)

sbcd90:
Hello @raboof ,
The builds pass again.

docs/src/main/paradox/orientdb.md
sbcd90: I pointed it to a simple OrientDBSource to console flow for Getting started.

ennru:
Thank you for this contribution. I'm glad you took it all the way!

Issue #603, opened 2017-11-20
Support UpdateTimeToLiveRequest in DynamoDB connector


gregbeech:
The DynamoDB connector doesn't currently support the `UpdateTimeToLiveRequest` which is necessary for creating expiring rows. It would be good to add this support.

ennru:
Thank you for making the DynamoDB connector more useful.
Would you have the chance to add something about the TTL in the docs?

ennru:
I reckon people needing it will find it even without text in the docs. Cheers.

gregbeech:
Hey, sorry, missed the notification about your previous comment regarding docs. As you say, I'm not sure there's that much value in putting anything specific in the Alpakka docs as it should be pretty easy to find once you've looked at the AWS docs (which I assume people will read in conjunction given the Alpakka lib is a fairly thin wrapper over it).

PR #604, opened 2017-11-20
Add DynamoDB TTL support #603

gregbeech:
Add DynamoDB support for operations:

* DescribeTimeToLive
* UpdateTimeToLive

Unfortunately cannot be tested locally as DynamoDB Local does not
support these operations.


johanandren:
Fixes #603

gregbeech:
@johanandren yeah, sorry, should have made that clear. I build a local version and tested both operations against a real instance where they work as expected.

Issue #523, opened 2017-10-13
GooglePubSub: emulator does not handle authentication


yanns:
When using the emulator in docker-compose:

```
pubsub:
  container_name: pubsub
  image: google/cloud-sdk
  command: gcloud beta emulators pubsub start --host-port 0.0.0.0:8042
  ports:
   - "8042:8042"
```

and setting the `PUBSUB_EMULATOR_HOST` to `localhost:8042`
(https://github.com/akka/alpakka/pull/299)

then I have the following error:
```
[info]   exception thrown spray.json.DeserializationException: Object is missing required member 'access_token'
[info]   	at spray.json.package$.deserializationError(package.scala:23)
[info]   	at spray.json.ProductFormats.fromField(ProductFormats.scala:60)
[info]   	at spray.json.ProductFormats.fromField$(ProductFormats.scala:50)
[info]   	at spray.json.DefaultJsonProtocol$.fromField(DefaultJsonProtocol.scala:30)
[info]   	at spray.json.ProductFormatsInstances$$anon$3.read(ProductFormatsInstances.scala:78)
[info]   	at spray.json.ProductFormatsInstances$$anon$3.read(ProductFormatsInstances.scala:68)
[info]   	at akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport.$anonfun$sprayJsonUnmarshaller$1(SprayJsonSupport.scala:30)
[info]   	at akka.http.scaladsl.util.FastFuture$.$anonfun$map$1(FastFuture.scala:23)
[info]   	at akka.http.scaladsl.util.FastFuture$.strictTransform$1(FastFuture.scala:41)
[info]   	at akka.http.scaladsl.util.FastFuture$.$anonfun$transformWith$3(FastFuture.scala:51)
[info]   	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:60)
[info]   	at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:55)
[info]   	at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:91)
[info]   	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
[info]   	at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:81)
[info]   	at akka.dispatch.BatchingExecutor$BlockableBatch.run(BatchingExecutor.scala:91)
[...]

PR #605, opened 2017-11-20
Do not use auth when publishing in emulation

lukasz-golebiewski:
Related to: https://github.com/akka/alpakka/issues/523

### DIFF
--- a/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/HttpApi.scala
+++ b/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/HttpApi.scala
@@ -35,6 +35,8 @@ private object HttpApi extends HttpApi {
   val PubSubGoogleApisHost: String = PubSubEmulatorHost.getOrElse(DefaultPubSubGoogleApisHost)
   val GoogleApisHost: String = PubSubEmulatorHost.getOrElse(DefaultGoogleApisHost)
 
+  override def isEmulated = PubSubEmulatorHost.nonEmpty
+
   private[pubsub] lazy val PubSubEmulatorHost: Option[String] = sys.props
     .get(PubSubEmulatorHostVarName)
     .orElse(sys.env.get(PubSubEmulatorHostVarName))
@@ -47,6 +49,8 @@ private object HttpApi extends HttpApi {
 private trait HttpApi {
   def PubSubGoogleApisHost: String
   def GoogleApisHost: String
+  def isEmulated: Boolean
+
   private implicit val pubSubMessageFormat = DefaultJsonProtocol.jsonFormat2(PubSubMessage)
   private implicit val pubSubRequestFormat = DefaultJsonProtocol.jsonFormat1(PublishRequest.apply)
   private implicit val gcePubSubResponseFormat = DefaultJsonProtocol.jsonFormat1(PublishResponse)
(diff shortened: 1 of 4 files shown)


google-cloud-pub-sub/src/test/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSpec.scala
```diff
@@ -67,6 +68,39 @@ class GooglePubSubSpec extends FlatSpec with MockitoSugar with ScalaFutures with
...
+      privateKey = TestCredentials.privateKey,
+      topic = "topic2"
+    )
+
+    val request = PublishRequest(Seq(PubSubMessage(messageId = "2", data = base64String("Hello Google!"))))
+
+    val source = Source(List(request))
+
+    // FIXME: not working, returns false
+    when(mockHttpApi.isEmulated).thenReturn(true)
```
lukasz-golebiewski: I would need help here, cannot force this mock to return `true`

lukasz-golebiewski: Looks to me like `Mockito` is not the best choice for Scala projects
https://github.com/mockito/mockito/issues/1102

vertexclique:
```
[pubsub] Dec 12, 2017 5:40:18 PM io.gapi.emulators.grpc.GrpcServer$3 operationComplete
[pubsub] INFO: Adding handler(s) to newly registered Channel.
[pubsub] Dec 12, 2017 5:40:18 PM io.gapi.emulators.netty.HttpVersionRoutingHandler channelRead
[pubsub] INFO: Detected non-HTTP/2 connection.
[pubsub] Dec 12, 2017 5:40:18 PM io.gapi.emulators.netty.NotFoundHandler handleRequest
[pubsub] INFO: Unknown request URI: /oauth2/v4/token
```

Still got this. Moreover, we don't need `apiKey` for service accounts. Are you going to merge this?

google-cloud-pub-sub/src/test/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSpec.scala
ennru: You can create a manual mock instance with 
```
    val mockHttpApi = new HttpApi {
      val PubSubGoogleApisHost: String = "..."
      val GoogleApisHost: String = "..."

      override def isEmulated: Boolean = true

      override def publish(
          project: String,
          topic: String,
          maybeAccessToken: Option[String],
          apiKey: String,
          request: PublishRequest
      )(implicit as: ActorSystem, materializer: Materializer): Future[Seq[String]] =
        Future.successful(Seq("id2"))
    }
```

If everything else in shape in this PR?

vertexclique: I can say no, just because pulling from emulator also hits to my comment stated below client tries to authenticate against emulator when pulling, acknowledging etc. So long story short all operations should be in `maybeAccessToken`.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/HttpApi.scala
```diff
@@ -97,7 +101,17 @@ private trait HttpApi {
     }
   }
 
-  def publish(project: String, topic: String, accessToken: String, apiKey: String, request: PublishRequest)(
+  private[this] def response(request: HttpRequest, maybeAccessToken: Option[String])(implicit as: ActorSystem,
```
ennru: This method is more like `doRequest` than `response`.

PR #611, opened 2017-11-22
logging to files in tests

### LINKED ISSUE #610: Direct logging to files in tests
The logging of the used libraries hides test failures in the build logs. 
By adding proper dependencies and configurations this logging can be removed.

ennru:
To direct log4j2 logging via slf4j to logback is harder than it looks. Didn't solve it in geode module, yet.
 
Fixes #610


ennru:
I'm quite happy with the effect in the Travis output. 
The same S3 test fails in all four variations, so I hope it's a timing issue caused by more logging.

geode/src/test/resources/log4j2.xml
```diff
@@ -5,9 +5,9 @@
         <Property name="geode-default">true</Property>
     </Properties>
     <Appenders>
-        <Console name="STDOUT" target="SYSTEM_OUT">
+        <File name="File" fileName="target/geode-log4j.log">
```
raboof: Couldn't we use a bridge jar to send log4j logging to slf4j/logback as well? Could be a separate pr though.

ennru: Yes, that sounds easy, but this far I failed to find the correct mixture of logging dependencies as Geode uses Log4j**2** which has its own bridge to slf4j.

2m:
The Travis output looks nice and neat when tests pass.

Wouldn't this make it harder to find the problem when a test fails? It will be harder to tell what logs where output from the failing test case.

On the other hand it was not easy to that before this change as well. So both before and after this change one might need to look at timestamps to determine what was going on.

raboof:
Since tests are run in parallel, having access to the output for each subproject separately might actually make it easier to tell what logs belong to a certain failing test.

Especially when individual tests are also ran in parallel it would be cool to collect the logs for each of those separately as well - but that might be something for another PR :).

ennru:
Travis doesn't like to handle this much output:
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).

The job has been terminated

2m:
Perhaps we should consider using Travis artifact upload feature for the logs: https://docs.travis-ci.com/user/deployment/s3/

We could also set the expiration time for the logs to automatically expire from the bucket in a week or so.

This could also make sense to have it for the generated docs.

geode/src/test/resources/log4j2.xml
ennru: The Geode `LogService` interacts directly with Log4j2 classes, not just the API.

PR #615, opened 2017-11-26
ElasticsearchFlow now returns info about both success and failed messages - with optional cargo

mbknor:
This a replacement-PR for https://github.com/akka/alpakka/pull/586.

This PR changes the original ElasticsearchFlowStage-implementation and adds the "returns info about both success and failed messages"-functionality to it.

The original ElasticsearchFlowStage-implementation only returns info about failed messages.
This makes it impossible to pass "info" through the elastic-flow.
Example of such info might be a Kafka-offset that should be committed AFTER the message has been written to elastic.

The fix in this PR makes this possible.

### DIFF
--- a/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
+++ b/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
@@ -23,21 +23,29 @@ import akka.stream.alpakka.elasticsearch.scaladsl.ElasticsearchSinkSettings
 import org.apache.http.message.BasicHeader
 import org.apache.http.util.EntityUtils
 
-trait IncomingMessageTrait[T] {
-  def id: Option[String]
-  def source: T
+object IncomingMessage {
+  // Apply method to use when not using cargo
+  def apply[T](id: Option[String], source: T): IncomingMessage[T, Any] =
+    IncomingMessage(id, source, null)
+
+  // Java-api - without cargo
+  def create[T](id: String, source: T): IncomingMessage[T, Any] =
+    IncomingMessage(Option(id), source)
+
+  // Java-api - with cargo
+  def create[T, C](id: String, source: T, cargo: C): IncomingMessage[T, C] =
+    IncomingMessage(Option(id), source, cargo)
 }
 
-final case class IncomingMessage[T](id: Option[String], source: T) extends IncomingMessageTrait[T]
+final case class IncomingMessage[T, C](id: Option[String], source: T, cargo: C)
 
-final case class IncomingMessageWithCargo[T, C](id: Option[String], source: T, cargo: C)
-    extends IncomingMessageTrait[T]
+object IncomingMessageResult {
+  // Apply method to use when not using cargo
+  def apply[T](source: T, success: Boolean): IncomingMessageResult[T, Any] =
+    IncomingMessageResult(source, null, success)
+}
 
-final case class MessageResult[T, X <: IncomingMessageTrait[T]](message: X, success: Boolean)
-
-// Using these special result-case-classes to reduce the generic-complexity when working the flows
-final case class IncomingMessageResult[T](source: T, success: Boolean)
-final case class IncomingMessageWithCargoResult[T, C](source: T, cargo: C, success: Boolean)
+final case class IncomingMessageResult[T, C](source: T, cargo: C, success: Boolean)
 
 trait MessageWriter[T] {
   def convert(message: T): String
@@ -47,26 +55,26 @@ trait MessageWriter[T] {
 // with and without cargo.
 // MessageResult is transformed into IncomingMessageResult or IncomingMessageWithCargoResult
 // in the javadsl- and scaladsl-ElasticsearchFlow-implementations
-class ElasticsearchFlowStage[T, X <: IncomingMessageTrait[T]](
+class ElasticsearchFlowStage[T, C](
     indexName: String,
     typeName: String,
     client: RestClient,
     settings: ElasticsearchSinkSettings,
     writer: MessageWriter[T]
-) extends GraphStage[FlowShape[X, Future[Seq[MessageResult[T, X]]]]] {
+) extends GraphStage[FlowShape[IncomingMessage[T, C], Future[Seq[IncomingMessageResult[T, C]]]]] {
 
-  private val in = Inlet[X]("messages")
-  private val out = Outlet[Future[Seq[MessageResult[T, X]]]]("result")
+  private val in = Inlet[IncomingMessage[T, C]]("messages")
+  private val out = Outlet[Future[Seq[IncomingMessageResult[T, C]]]]("result")
   override val shape = FlowShape(in, out)
 
   override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
     new TimerGraphStageLogic(shape) with InHandler with OutHandler {
 
       private var state: State = Idle
-      private val queue = new mutable.Queue[X]()
-      private val failureHandler = getAsyncCallback[(Seq[X], Throwable)](handleFailure)
-      private val responseHandler = getAsyncCallback[(Seq[X], Response)](handleResponse)
-      private var failedMessages: Seq[X] = Nil
+      private val queue = new mutable.Queue[IncomingMessage[T, C]]()
+      private val failureHandler = getAsyncCallback[(Seq[IncomingMessage[T, C]], Throwable)](handleFailure)
+      private val responseHandler = getAsyncCallback[(Seq[IncomingMessage[T, C]], Response)](handleResponse)
+      private var failedMessages: Seq[IncomingMessage[T, C]] = Nil
       private var retryCount: Int = 0
 
       override def preStart(): Unit =
@@ -82,7 +90,7 @@ class ElasticsearchFlowStage[T, X <: IncomingMessageTrait[T]](
         failedMessages = Nil
       }
 
-      private def handleFailure(args: (Seq[X], Throwable)): Unit = {
+      private def handleFailure(args: (Seq[IncomingMessage[T, C]], Throwable)): Unit = {
         val (messages, exception) = args
         if (retryCount >= settings.maxRetry) {
           failStage(exception)
@@ -96,7 +104,7 @@ class ElasticsearchFlowStage[T, X <: IncomingMessageTrait[T]](
       private def handleSuccess(): Unit =
         completeStage()
 
-      private def handleResponse(args: (Seq[X], Response)): Unit = {
+      private def handleResponse(args: (Seq[IncomingMessage[T, C]], Response)): Unit = {
         val (messages, response) = args
         val responseJson = EntityUtils.toString(response.getEntity).parseJson
 
@@ -122,7 +130,7 @@ class ElasticsearchFlowStage[T, X <: IncomingMessageTrait[T]](
           if (successMsgs.nonEmpty) {
             // push the messages that DID succeed
             val resultForSucceededMsgs = successMsgs.map { x =>
-              MessageResult[T, X](x, success = true)
+              IncomingMessageResult[T, C](x.source, x.cargo, success = true)
             }
             emit(out, Future.successful(resultForSucceededMsgs))
           }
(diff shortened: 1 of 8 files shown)


mbknor:
Example of how to use this new code with kafka can be found here:

https://github.com/mbknor/alpakka/blob/43c5f8f8bb95465b3aef2d69dcd3d14c1f853571/elasticsearch/src/test/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSpec.scala#L318

This code is also added to the documentation

takezoe:
I think that the cargo option is unnecessary because we can pass optional value to the downstream using tuple as source as below:

```scala
val f1 = Source(messagesFromKafka) // Assume we get this from Kafka
  .map { kafkaMessage: KafkaMessage =>

    val book = kafkaMessage.book
    val id = book.title
    println("title: " + book.title)

    // Transform message so that we can write to elastic
    IncomingMessage(Some(id), (book, kafkaMessage.offset))
  }
  .via( // write to elastic
    ElasticsearchFlow.create[(Book, KafkaOffset)](
      "sink3",
      "book",
      ElasticsearchSinkSettings(5)
    )
  ).map {
    messageResults =>
      messageResults.foreach {
        case IncomingMessageResult((_, kafkaOffset), success) =>
          if ( !success ) throw new Exception("Failed to write message to elastic")
          // Commit to kafka
          commitToKakfa(kafkaOffset)
      }
}.runWith(Sink.seq)
```

Anyway, this pull request is really great work! It will help to combine Elasticsearch connector with other connectors or processes as the added test case shows.

mbknor:
How will you extract the "source" from the tuple in your example? How do you know if you have tuple (with offset) or not?

takezoe:
We can also extract source by pattern matching.

```scala
messageResults.foreach {
  case IncomingMessageResult((source, kafkaOffset), success) =>
    ...
}
```

mbknor:
@takezoe: I don't see how you plan to do that with generics, with or without offset/cargo.

I also think it will be difficult to use from Java.

Maybe you could fork my fork and do some modifications to my branch?

Then I'll see what you mean.

takezoe:
@mbknor Sure, I will do it.

takezoe:
@mbknor Sorry, above my exampe doesn't work. See this commit: https://github.com/takezoe/alpakka/commit/851ce72719205febc7910f31fd31ccc8f7da2183

There is no modification from your pull request, modified only testcase. It shows that we can propagate additional information without cargo and IncomingMessage abstraction.

Scala version:
https://github.com/takezoe/alpakka/blob/851ce72719205febc7910f31fd31ccc8f7da2183/elasticsearch/src/test/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSpec.scala#L319

Java version:
https://github.com/takezoe/alpakka/blob/851ce72719205febc7910f31fd31ccc8f7da2183/elasticsearch/src/test/java/akka/stream/alpakka/elasticsearch/ElasticsearchTest.java#L282

mbknor:
I think your modified scala test code in previous comment have a significant problem:

```scala
        .flatMapConcat {
          case IncomingMessage(id, kafkaMessage) => // @mbknor: here kafkaMessage is one single message
            // Transform message so that we can write to elastic
            // @mbknor: You start a new stream that will only send one message in its life
            Source(List(IncomingMessage(id, kafkaMessage.book))).via { // write to elastic
              ElasticsearchFlow
                .create[Book](
                  "sink6",
                  "book",
                  ElasticsearchSinkSettings(5)
                )
                .map { messageResults =>
                  messageResults.map {
                    case IncomingMessageResult(_, success) =>
                      IncomingMessageResult(kafkaMessage, success) // @mbknor: You reference the single kafkaMessage from above
                  }
                }
}
```

When doing it like this, you shortcut the entire batch-logic in ElasticsearchFlow.

I guess the Java code does the same, so I have not looked at it.

mbknor:
I'll try one more time to combine **IncomingMessage** and **IncomingMessageWithCargo** (https://github.com/mbknor/alpakka/blob/43c5f8f8bb95465b3aef2d69dcd3d14c1f853571/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala#L31)
into one single *case class* able to hold both with and without cargo.

Maybe I can get it to work by using multiple apply-methods, some with default generic type for cargo.

takezoe:
> You reference the single kafkaMessage from above

Yes, but this isn't a problem because it passes a single element list of `IncomingMessage` to downstream.

```scala
// Create source form a single element list
Source(List(IncomingMessage(id, kafkaMessage.book))).via {
```

Therefore, following part is always called with single `IncomingMessageResult`:

```scala
.map { messageResults => // always has single element made by the above list
  messageResults.map {
    case IncomingMessageResult(_, success) =>
      IncomingMessageResult(kafkaMessage, success)
  }
}
```

In this case, there isn't any problem to refer same `kafkaMessage` instance to reassemble `IncomingMessageResult`. Of course, my testcase succeeds.

mbknor:
Your test creates a new stream inside **flatMapConcat** resulting in a single future waiting for one message to flow through the stream.

My test (https://github.com/mbknor/alpakka/blob/43c5f8f8bb95465b3aef2d69dcd3d14c1f853571/elasticsearch/src/test/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSpec.scala#L350) uses **via**.

the difference is that my test streams many single kafka-messages through the elastic-stream.
The elastic-stream batches them and sends many at once to elastic.
when a batch is done, then it returns info about all messages sent in batch.

Your test will never do this. it will always send one by one message.

mbknor:
As I said, I'll try one more time with only one IncommingMessage (that supports both with and without cargo in a decent way).

This is how it is solved int akka kafka stream: https://doc.akka.io/docs/akka-stream-kafka/current/producer.html#producer-as-a-flow

They call it passthrough (maybe better name) and must always declare it, the type for it.

takezoe:
> As I said, I'll try one more time with only one IncommingMessage (that supports both with and without cargo in a decent way).

Yes, but I was wonder how you will make it in my previous comment.

> This is how it is solved int akka kafka stream: https://doc.akka.io/docs/akka-stream-kafka/current/producer.html#producer-as-a-flow

Thanks. It looks like a cargo parameter. I understood that flows should have such parameter if we want to pass additional values to downstream after all.

mbknor:
@takezoe Please have a look at my latest commit.

I'm now only using **IncomingMessage** and **IncomingMessageResult**.

It can always hold cargo, but I've created convenient apply-/create-methods that can be used to prevent to overkill generics - both in Scala and Java.

Maybe we should rename *cargo* to *passThrough* - the name used in akka kafka stream.

mbknor:
I've also added Java-kafka-example to the docs.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
```diff
@@ -23,33 +23,58 @@ import akka.stream.alpakka.elasticsearch.scaladsl.ElasticsearchSinkSettings
...
 import org.apache.http.util.EntityUtils
 
-final case class IncomingMessage[T](id: Option[String], source: T)
+object IncomingMessage {
+  // Apply method to use when not using cargo
+  def apply[T](id: Option[String], source: T): IncomingMessage[T, Any] =
+    IncomingMessage(id, source, null)
+
+  // Java-api - without cargo
+  def create[T](id: String, source: T): IncomingMessage[T, Any] =
```
takezoe: It should be used `NotUsed` instead of `Any`.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
```diff
@@ -79,28 +104,52 @@ class ElasticsearchFlowStage[T, R](
...
-              message
+            item.asJsObject.fields("index").asJsObject.fields.get("error") match {
+              case Some(errorMessage) => (message, false)
+              case None => (message, true)
             }
         }
 
-        if (failed.nonEmpty && settings.retryPartialFailure && retryCount < settings.maxRetry) {
+        val successMsgs = messageResults.filter(_._2).map(_._1)
+        val failedMsgs = messageResults.filter(!_._2).map(_._1)
```
takezoe: Classifying successed messages and failed messages at once using `groupBy` might be better?

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
```diff
@@ -23,33 +23,58 @@ import akka.stream.alpakka.elasticsearch.scaladsl.ElasticsearchSinkSettings
...
+  // Java-api - with cargo
+  def create[T, C](id: String, source: T, cargo: C): IncomingMessage[T, C] =
+    IncomingMessage(Option(id), source, cargo)
+}
+
+final case class IncomingMessage[T, C](id: Option[String], source: T, cargo: C)
+
+object IncomingMessageResult {
+  // Apply method to use when not using cargo
+  def apply[T](source: T, success: Boolean): IncomingMessageResult[T, Any] =
```
takezoe: It should be used `NotUsed` instead of `Any`.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
```diff
@@ -23,33 +23,58 @@ import akka.stream.alpakka.elasticsearch.scaladsl.ElasticsearchSinkSettings
 import org.apache.http.message.BasicHeader
 import org.apache.http.util.EntityUtils
 
-final case class IncomingMessage[T](id: Option[String], source: T)
+object IncomingMessage {
+  // Apply method to use when not using cargo
+  def apply[T](id: Option[String], source: T): IncomingMessage[T, Any] =
```
takezoe: It should be used `NotUsed` instead of `Any`.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
```diff
@@ -23,33 +23,58 @@ import akka.stream.alpakka.elasticsearch.scaladsl.ElasticsearchSinkSettings
 import org.apache.http.message.BasicHeader
 import org.apache.http.util.EntityUtils
 
-final case class IncomingMessage[T](id: Option[String], source: T)
+object IncomingMessage {
```
takezoe: It might be useful if it also has factory methods to create an instance without id for Java.
```scala
def create[T](source: T): IncomingMessage[T, NotUsed] = IncomingMessage(None, source)
def create[T, C](source: T, cargo: C): IncomingMessage[T, C] =IncomingMessage(None, source, cargo)
```

takezoe:
@raboof This pull request looks very useful when we use `ElasticsearchFlow` with other connectors or resources. Could you have a look?

mbknor:
@takezoe I agree with all your comments. I'll fix it.

Do you have any comment on  **payload* vs **passThrough**?

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
mbknor: I'm not sure how to improve this one using groupBy

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
mbknor: done

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
mbknor: done

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
mbknor: done

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
mbknor: done

mbknor:
Added new commit replacing use of Any with NotUsed (pluss the other small improvements)

ennru:
Looks like a very useful improvement. 
Though I'd prefer the Kafka-connector name `passThrough` as many users may have used it and it describes the field's purpose.

ennru:
Great collaboration on this one @mbknor and @takezoe !

PR #633, opened 2017-12-07
Drop Akka 2.4 from Alpakka build

### LINKED ISSUE #624: Drop Akka 2.4 support
We have an Alpakka PR (https://github.com/akka/alpakka/issues/623) that utilizes functionality from Akka 2.5 series and 2.4 is approaching its EOL. We should consider dropping the support for Akka 2.4 in Alpakka.

Is there anyone that uses Alpakka with Akka 2.4 and can not migrate to Akka 2.5?

Otherwise we can declare Alpakka 0.14 as the last version having 2.4 support and continue with only 2.5 builds.

ennru:
and revise links to Akka documentation.

Fixes #624.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -3,8 +3,8 @@ import sbt._, Keys._
 object Dependencies {
 
   val AkkaVersion = sys.env.get("AKKA_SERIES") match {
-    case Some("2.5") => "2.5.7"
-    case _ => "2.4.19"
+    case Some("2.4") => sys.error("Akka 2.4 is not supported anymore")
+    case _ => "2.5.7"
   }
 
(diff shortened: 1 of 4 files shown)


project/Dependencies.scala
```diff
@@ -3,8 +3,8 @@ import sbt._, Keys._
 object Dependencies {
 
   val AkkaVersion = sys.env.get("AKKA_SERIES") match {
-    case Some("2.5") => "2.5.7"
-    case _ => "2.4.19"
+    case Some("2.4") => sys.error("Akka 2.4 is not supported anymore")
```
johanandren: Just FYI I logged this a while ago: #24030 to be able to do this at runtime as well.

PR #634, opened 2017-12-07
Ensure whitesource reports for stable versions

raboof:


RELEASING.md
```diff
@@ -1,5 +1,6 @@
 ## Releasing
 
+1. Rename the 'alpakka-xx-stable' report in [WhiteSource](https://saas.whitesourcesoftware.com) to reflect the latest version
```
ennru: This seems to require extra permission. Should it be phrased "Request Akka team to ..."?

raboof: 👍 good one will do!

PR #638, opened 2017-12-09
AMQP keep connection open if there are pending acks

juanjoDiaz:
I found some issues with the way the unacknowledged messages are tracked in the AMQP connector.

I also move around a couple things so it's easier to compare the SourceStage and the FlowStage to ensure that both use the same logic. In the future it might make sense to merge them in one stage (as the MQTT connector).

### DIFF
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
@@ -34,14 +34,12 @@ final class AmqpRpcFlowStage(settings: AmqpSinkSettings, bufferSize: Int, respon
     extends GraphStageWithMaterializedValue[FlowShape[OutgoingMessage, CommittableIncomingMessage], Future[String]]
     with AmqpConnector { stage =>
 
-  import AmqpRpcFlowStage._
-
   val in = Inlet[OutgoingMessage]("AmqpRpcFlow.in")
   val out = Outlet[CommittableIncomingMessage]("AmqpRpcFlow.out")
 
   override def shape: FlowShape[OutgoingMessage, CommittableIncomingMessage] = FlowShape.of(in, out)
 
-  override protected def initialAttributes: Attributes = defaultAttributes
+  override protected def initialAttributes: Attributes = AmqpRpcFlowStage.defaultAttributes
 
   override def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, Future[String]) = {
     val promise = Promise[String]()
@@ -53,7 +51,7 @@ final class AmqpRpcFlowStage(settings: AmqpSinkSettings, bufferSize: Int, respon
       private val queue = mutable.Queue[CommittableIncomingMessage]()
       private var queueName: String = _
       private val unackedMessages = new AtomicInteger()
-      private var outstandingMessages = 0
+      private val outstandingMessages = new AtomicInteger()
 
       override def connectionFactoryFrom(settings: AmqpConnectionSettings) = stage.connectionFactoryFrom(settings)
       override def newConnection(factory: ConnectionFactory, settings: AmqpConnectionSettings): Connection =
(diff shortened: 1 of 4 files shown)


ennru:
Is this something that could be verified in tests?

juanjoDiaz:
Some tests added.

In general I think that the tests of the AMQP connector need some review.
At least:
* Separate source tests, sink tests and flow test
* Ensure that coverage is good
* Ensure parity between Scala and Java tests

Same applies to the MQTT connector actually :)

juanjoDiaz:
Hei guys,

Any idea if the issue here might be related to the conversion from Scala's Future to Java's CompletableFuture??
It's really weird that exactly the same test (I just copy-pasted the code from Scala to Java) passes on Scala but fails on Java.
The same issue seems to be happening on #515 
Of course, I can't reproduce it on my computer :(

juanjoDiaz:
Nevermind, I think I found the cause.

I'll look into it tomorrow.

juanjoDiaz:
Alright, I'm seriously starting to think that I've just hit an issue not related to my changes. Maybe something to do with the way that streams are multithreaded??

It looks like in scala the connector does:
```
Receive msg1
Push msg1
Receive msg2
Push msg2
Receive msg3
Push msg3
Receive msg4
Push msg5
Receive msg5
Push msg5
```
while in java it does:
does:
```
Receive msg1
Push msg1
Receive msg2
Receive msg3
Receive msg4
Receive msg5
Push msg2
Push msg3
Push msg5
Push msg5
```


Also, it behaves differently if I do:
```
    // All test passing
     val result1 = amqpSource
        .mapAsync(1)(cm => cm.nack().map(_ => cm))
        .take(input.size)
        .runWith(Sink.seq)
```
or if I do:
```
    // Tests failing
     val result1 = amqpSource
        .take(input.size)
        .mapAsync(1)(cm => cm.nack().map(_ => cm))
        .runWith(Sink.seq)
```

Any hint of what might be the issue??

ennru:
Your current code works fine locally for me.

juanjoDiaz:
hallelujah! Got the issue 😃

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
```diff
@@ -174,18 +168,37 @@ final class AmqpRpcFlowStage(settings: AmqpSinkSettings, bufferSize: Int, respon
             if (queue.nonEmpty) {
               pushMessage(queue.dequeue())
             }
+
+          override def onDownstreamFinish(): Unit = {
+            setKeepGoing(true)
+            if (unackedMessages.get() == 0) super.onDownstreamFinish()
```
johanandren: This means that if a user forgot to ack/nack the stream will never complete, imagine doing a filter or having a buffer somewhere that has unconsumed elements in it. 

I'm not sure how to deal with this in a good way, I think a timeout of some kind may be the only solution and then either fail the stage or nack all those elements after the timeout. Wdyt?

juanjoDiaz: This is the way akka streams kafka does it. And I think that is a good way (I had exactly this problem last week).

If the user forgets to ack/nack (I did forget to nack bogus messages), he'll soon reach the inflight limit and the queue will block making it very obvious that messages are not flowing and not wasting messages or modifying the order of the queue.

Setting a timeout will essentially hide the issue.
Also, there are few thing:
* I'm not sure if nacking will preserve the order or if it's better to just fail and disconnect the client.
* How do we set the timeout? based on the oldest unacked message? Then we would need to store the time with the messages and store the list in the queue. Based on the latest message? This might be easier
* I think that the timeout will need to be configurable since we don't know how time-consuming the processing that the users are doing is.
* Is there a possibility that, if you are using some retry supervision like backoff, you might end up in a loop where everything seems to be running but nothing is actually happening??

Wdyt?

johanandren: There is a big difference with Kafka though, where there isn't really individual message-acks but  rather a highest-seen-committed counter. But maybe keep it as is and see if people have problems with it and solve that if so.

If we implement it, I think it should you are right it should be just failing the stage. The timeout could be from configuration, with a generous default, the timer would start when downstream or upstream completing, if the timeout hits and there still are outstanding non-acked (in counter) then we fail the stage and tear down the connection (I'm guessing the server will see that as nack:s then for the outstanding messages).

juanjoDiaz: I think that it's a good idea to keep it as it is and act if people find problems.
If you are happy with that I'll consider this done.

If you still consider the timeout necessary, just let me know :)

johanandren: Let's merge it like this and see if is surprising to people using it.

johanandren:
Fixes #665

johanandren:
Sorry, had to revert this as it consistently made master fail on 2.12 after I merged it, see https://travis-ci.org/akka/alpakka/jobs/319260495 and search for akka.stream.alpakka.amqp.javadsl.AmqpConnectorsTest

juanjoDiaz:
#669 add the fix for that, I believe :)

PR #640, opened 2017-12-09
[MQTT] Allow setting QoS & retained per message

### LINKED ISSUE #337: [MQTT] Retained Message
I'm trying to send retained messages to broker but I can't find a way to set retained flag for a regular message. Will class contains this option, but I can't see it for MqttMessage. Is there a way to accomplish this? Thank you :)

juanjoDiaz:
Closes #337

Allow setting QoS and retained flag per message.
Deprecate the old Will message since the new message already contains the necessary data.
Bring back the will test since I can't see why it was ignored.

No breaking changes in the existing APIs.

### DIFF
--- a/docs/src/main/paradox/mqtt.md
+++ b/docs/src/main/paradox/mqtt.md
@@ -65,6 +65,13 @@ Scala
 Java
 : @@snip ($alpakka$/mqtt/src/test/java/akka/stream/alpakka/mqtt/javadsl/MqttSourceTest.java) { #run-sink }
 
+The QoS and the retained flag can be configured on a per-message basis.
+Scala
+: @@snip ($alpakka$/mqtt/src/test/scala/akka/stream/alpakka/mqtt/scaladsl/MqttSourceSpec.scala) { #will-message }
+
+Java
+: @@snip ($alpakka$/mqtt/src/test/java/akka/stream/alpakka/mqtt/javadsl/MqttSourceTest.java) { #will-message }
+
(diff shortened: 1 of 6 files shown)


juanjoDiaz:
Ok. I remove what I just said. The will test seems flaky.

I think that at some point the whole test suite of the MQTT connector should be reviewed.
The Java tests are basically missing for example.

ennru:
Looks like you mismerged, I'm afraid.

juanjoDiaz:
Not even a merge problem.
My tests from the other PR where using the Scala API instead of the Java API apparently.

ennru:
It still looks like you merged changes from master into this branch instead of rebasing.

juanjoDiaz:
You're right. Damn! I'll fix it tomorrow.

juanjoDiaz:
So it turns out that github always keep the comparison with the point in time when you created the PR.

Changing the target branch to something else and back to the right one forces the refresh.

Does it look good now?

ennru:
Yes, looks alright now.

ennru:
The "retained" attribute is not set on a per-message basis, as far as I can see. Should the description be updated?

juanjoDiaz:
It is set per-message (https://github.com/akka/alpakka/pull/640/files#diff-9c15bc026f3b7b94ea7c3acd34408f1eR119)

You mean to add that to the docs?

ennru:
Ah, I see. 
Docs could be good, but I reckon people needing it will discover it in the API.

juanjoDiaz:
Docs added ;)

Issue #606, opened 2017-11-21
Streamlining connector settings


johanandren:
This may be a topic where there are different opinions on. Mine:

Only Akka-core configuration should live in the akka namespace. 

Alpakka should have its own top-level namespace `alpakka.connector-x` for default connector configs. 

Connectors providing config should provide one `ConnectorXSettings(ActorSystem)` factory which reads the default path and one `ConnectorXSettings(Config)` which just accepts the config block of the connector so that a user can have multiple configs for the same connector in her application.conf.

We should make a checklist for connectors, including a description of this as a requirement, and update existing connectors.

ennru:
Let's review `CONTRIBUTING.md` and come up with a good checklist there.

ennru:
I find these uses of the config namespaces
* akka.stream.alpakka.dynamodb
* akka.stream.alpakka.ironmq
* akka.stream.alpakka.s3

So it is not huge to migrate, I'd say.

2m:
Lets use `s"alpakka.$javaModuleName"` as the namespace for a given connector:

DynamoDB: `alpakka.aws.dynamodb`
IronMQ: `alpakka.ironmq`
S3: `alpakka.aws.s3`

ennru:
This migration happened a while back.

PR #645, opened 2017-12-13
S3: use default region provider

gipeshka:
Suggestion to continue with the approach used in https://github.com/akka/alpakka/pull/440 to use default AWS resolver for `region` as well.

Potentially simplifies configuration of an application when using AWS.

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/S3Settings.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/S3Settings.scala
@@ -64,12 +64,14 @@ object S3Settings {
    * Scala API: Creates [[S3Settings]] from a [[Config]] object.
    */
   def apply(config: Config, configurationPrefix: String = "akka.stream.alpakka.s3"): S3Settings = {
-    val bufferType = config.getString(s"$configurationPrefix.buffer") match {
+    val s3Config = config.getConfig(configurationPrefix)
+
+    val bufferType = s3Config.getString("buffer") match {
       case "memory" =>
         MemoryBufferType
 
       case "disk" =>
-        val diskBufferPath = config.getString(s"$configurationPrefix.disk-buffer-path")
+        val diskBufferPath = s3Config.getString("disk-buffer-path")
         DiskBufferType(Paths.get(diskBufferPath))
 
       case other =>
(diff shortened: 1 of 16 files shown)


ennru:
Looks legit, please have a look how it relates to #606, #388, and #90.

gipeshka:
hey @ennru 
checked the related issues, don't really think this PR affects/resolves them.

However I have a note related to https://github.com/akka/alpakka/issues/606. Imo it would be better to have a generic configuration for all AWS related services to authenticate a client. E.g.
```
alpakka.aws {
  credentials {
    provider = default
  }
  region {
    provider = default
  }

  s3 {
    # all other s3 related configs here as well
    # in case s3/dynamoDb/somethingElse has its own configuration different from the default one
    credentials {
      provider = static
      access-key-id = keyId
      secret-access-key = secretKey
    }
  }
}
```

Then provider would try to use `alpakka.aws.SERVICE_ID.credentials.provider`, otherwise fallback to `alpakka.aws.credentials.provider`. This credentials/region provider could be reused across the AWS connectors.

Not sure if this is what maintainers of other AWS connectors would like to have, e.g. SQS/SNS connector requires fully configured AmazonSQSAsync to be injected by a client rather than instantiating it. This looks a bit inconsistent, however I am not sure which approach suits alpakka philosophy better.

ennru:
Thank you for looking into these.
I'd prefer if the settings could be read from a Config (on a path prefix) instead of repeating the prefix everywhere. That would make it easier to support more than a single setup.

s3/src/main/scala/akka/stream/alpakka/s3/S3Settings.scala
```diff
@@ -51,47 +54,78 @@ object S3Settings {
...
+   */
+  def apply(configurationPrefix: String)(implicit system: ActorSystem): S3Settings =
+    apply(system.settings.config, configurationPrefix)
+
   /**
    * Scala API: Creates [[S3Settings]] from a [[Config]] object.
    */
-  def apply(config: Config): S3Settings = {
-    val bufferType = config.getString("akka.stream.alpakka.s3.buffer") match {
+  def apply(config: Config, configurationPrefix: String = "akka.stream.alpakka.s3"): S3Settings = {
```
gipeshka: @ennru added `configurationPrefix` for the discussed configuration purpose

gipeshka: Therefore pr affects https://github.com/akka/alpakka/issues/388 and partially https://github.com/akka/alpakka/issues/606

ennru: Great. 
You can make better use of Config by using `val config = configIn.getConfig(configurationPrefix)` and removing the prefix down the road.

gipeshka: @ennru updated the method and added a new method to simplify creation of S3Client with custom S3Settings https://github.com/akka/alpakka/pull/645/files#diff-edcf65bc8d22d487709bda00666179c1R83

gipeshka:
Please note that PR breaks BC on region configuration as it switches default behaviour from `static` to `default` like https://github.com/akka/alpakka/pull/644

Supposedly should be mentioned in release notes.

ennru:
Looks good to me, but I'd like another opinion from S3 users. 
@francisdb would you have the chance to have a look?

francisdb:
To be honest this is quite a big PR so hard to have a full view. By the way have you seen that the tests are currently totally messed up. They don't do real tests any more since the ibm stuff was added.

gipeshka:
@francisdb S3NoMock is ignored altogether, that's true. Not sure if I can provide something more than unittests to cover my case here.

francisdb:
Once this and #563 are merged in I will have another check if there are no regressions.

gipeshka:
Thanks a lot!

francisdb:
@ennru I know, I normally used that to see if everything still works but since #464 the test is mostly useless, that's why in #563 I went back to the original version and added more tests there.

Issue #646, opened 2017-12-14
XML Namespace Management


Korbik:
For some uses, the use of namespaces is mandatory to exchange file with other entities.

The management of the xml namespaces for elements and attributes could be a improvement for the library

ennru:
Solved with #647

PR #647, opened 2017-12-14
XML: Add the possibility to handle namespace (#646)

Korbik:


ennru:
Fixes #646

xml/src/main/scala/akka/stream/alpakka/xml/Xml.scala
```diff
@@ -42,9 +43,74 @@ case object EndDocument extends ParseEvent {
...
+  /**
+   * Java API
+   */
+  def create(localName: String,
+             attributesList: java.util.List[Attribute],
+             prefix: Optional[String],
+             namespace: Optional[String]): StartElement =
+    new StartElement(localName, attributesList.asScala.toList, prefix.asScala, namespace.asScala)
+
+  /**f
```
ennru: Extra `f`

xml/src/test/scala/akka/stream/alpakka/xml/scaladsl/XmlWritingTest.scala
```diff
@@ -114,6 +114,50 @@ class XmlWritingTest extends WordSpec with Matchers with BeforeAndAfterAll with
       resultFuture.futureValue(Timeout(3.seconds)) should ===(doc)
     }
```
ennru: Could you add Xml Writer tests with namespace use, please.

Korbik: I have improved the process because of missing cases. I have added few tests. Tell me if I still need to had some.

ennru:
Thank you for this contribution. We'll probably cut a release within two weeks time.

PR #652, opened 2017-12-15
HBase example refinement

bkowalik:
I think that example should look like that because it is self-explanatory. It does not suggest that `Put` has constructor that operates on `String` argument.


ennru:
Why is it even an implicit conversion? Wouldn't calling `getBytes` on the string make it even more obvious what is needed?

bkowalik:
I think that removing this implicit conversion will make this code harder to read. `Put` constructor and method fully rely on `Array[Byte]` so it is more convenient to have this conversion in scope. My point is to make it visible in the example rather than having to look into code.

ennru:
Thank you for your contribution. Keep them coming!

PR #653, opened 2017-12-15
Fix link to slick fromConfig docs

raboof:
Thanks to https://github.com/lightbend/paradox/issues/166#issuecomment-351898341

### DIFF
--- a/docs/src/main/paradox/slick.md
+++ b/docs/src/main/paradox/slick.md
@@ -70,7 +70,7 @@ Oracle
 SQL Server
 : @@snip ($alpakka$/slick/src/test/resources/application.conf) { #config-sqlserver }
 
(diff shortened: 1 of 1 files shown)


docs/src/main/paradox/slick.md
```diff
@@ -125,3 +125,5 @@ Scala
 
 Java
 : @@snip ($alpakka$/slick/src/test/java/akka/stream/alpakka/slick/javadsl/DocSnippetFlow.java) { #flow-example }
+
+ [jdbcbackend-api]: http://slick.lightbend.com/doc/3.2.1/api/index.html#slick.jdbc.JdbcBackend$DatabaseFactoryDef@forConfig(String,Config,Driver,ClassLoader):Database
```
ennru: You could define Slick Version in `Dependencies` and use it to keep it pointing to the version in use.
Looks like more API links are version dependent...

raboof: That doesn't appear to work inside URL's (https://github.com/akka/alpakka/tree/slickVersionFromDependencies)

PR #655, opened 2017-12-15
AWS S3: Support for customer key server side encryption

### LINKED ISSUE #654: S3 Support Server Sside Encryption - Customer Keys
Currently, 
Server side encryption - customer keys (SSE-C) is not supported by S3 alpakka

Reference: http://docs.aws.amazon.com/AmazonS3/latest/dev/ServerSideEncryptionCustomerKeys.html

mlischetti:
fix #654 S3 support server side encryption - customer keys (SSE-C)

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Headers.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Headers.scala
@@ -75,12 +75,19 @@ object StorageClass {
  */
 sealed abstract class ServerSideEncryption {
   def headers: immutable.Seq[HttpHeader]
+
+  def headers(request: S3Request): immutable.Seq[HttpHeader]
 }
 
 object ServerSideEncryption {
 
   case object AES256 extends ServerSideEncryption {
     override def headers: immutable.Seq[HttpHeader] = RawHeader("x-amz-server-side-encryption", "AES256") :: Nil
+
+    override def headers(request: S3Request): immutable.Seq[HttpHeader] = request match {
+      case PutObject | InitiateMultipartUpload => headers
+      case _ => Nil
+    }
   }
 
   /**
@@ -99,6 +106,11 @@ object ServerSideEncryption {
       val headers = baseHeaders ++ context.map(ctx => RawHeader("x-amz-server-side-encryption-context", ctx)).toSeq
       headers
     }
+
+    override def headers(request: S3Request): immutable.Seq[HttpHeader] = request match {
+      case PutObject | InitiateMultipartUpload => headers
+      case _ => Nil
+    }
   }
 
   /**
@@ -116,5 +128,11 @@ object ServerSideEncryption {
         val md5 = Md5Utils.md5AsBase64(decodedKey)
         md5
       })) :: Nil
+
+    override def headers(request: S3Request): immutable.Seq[HttpHeader] = request match {
+      case GetObject | HeadObject | PutObject | InitiateMultipartUpload | UploadPart =>
+        headers
+      case _ => Nil
+    }
   }
 }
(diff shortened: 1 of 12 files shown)


ennru:
Could you add the corresponding methods to the Java API as well, please.

ennru:
We merged a rather large PR for S3, could you re-base this PR, please.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -63,9 +63,12 @@ private[alpakka] final class S3Stream(settings: S3Settings)(implicit system: Act
   // def because tokens can expire
   def signingKey = SigningKey(settings.credentialsProvider, CredentialScope(LocalDate.now(), settings.s3Region, "s3"))
 
-  def download(s3Location: S3Location, range: Option[ByteRange] = None): Source[ByteString, Future[ObjectMetadata]] = {
+  def download(s3Location: S3Location,
+               range: Option[ByteRange] = None,
+               sse: Option[ServerSideEncryption] = None): Source[ByteString, Future[ObjectMetadata]] = {
```
ennru: As this is internal API I'd prefer to not use default parameters (applies to several methods).

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Headers.scala
```diff
@@ -71,30 +72,67 @@ object StorageClass {
...
-      RawHeader("x-amz-server-side-encryption", "aws:kms") ::
-      RawHeader("x-amz-server-side-encryption-aws-kms-key-id", kmsKeyId.get) ::
-      RawHeader("x-amz-server-side-encryption-context", context.get) ::
-      Nil
-    case _ => throw new IllegalArgumentException("Unsupported encryption algorithm.")
-  }
+sealed abstract class ServerSideEncryption {
+  def headers: immutable.Seq[HttpHeader]
+
+  def headers(request: S3Request): immutable.Seq[HttpHeader]
```
ennru: It is not obvious what this `headers` method is supposed to do. Is  `applicableFor` a better name?

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -63,9 +63,12 @@ private[alpakka] final class S3Stream(settings: S3Settings)(implicit system: Act
...
     import mat.executionContext
-    val future = request(s3Location, rangeOption = range)
+    val s3Headers = S3Headers(sse.map(_.headers(GetObject)).getOrElse(Seq.empty))
```
ennru: a `fold` might be simpler

johanandren: or a `match` which is even more readable IMO (and does not allocate any extra lambdas)

mlischetti: LGTM I will changes _map + getOrElse_ by _fold_

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
mlischetti: @ennru Are you proposing simply remove the default parameters or split this method in 4 methods?

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Headers.scala
mlischetti: Before created this issue, **_ServerSideEncryption_** class had this method:
	_**def headers: immutable.Seq[HttpHeader]**_
and this method return all headers according with the specified server side encryption
but actually this headers are definied by the current request.
Maybe I should change the other method:
	**_def headers(request: S3Request):  immutable.Seq[HttpHeader]_**
by 
	**_def headersByRequest(request: S3Request):  immutable.Seq[HttpHeader]_**

ennru: Yes, or `headersFor`?

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
ennru: All parameters (without defaults) in a single method to keep the internal API smaller.

PR #668, opened 2017-12-22
Elasticsearch: revised documentation

ennru:
I tried to make the ElasticSearch connector documentation more comprehensive.

### DIFF
--- a/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/scaladsl/ElasticsearchSource.scala
+++ b/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/scaladsl/ElasticsearchSource.scala
@@ -9,14 +9,32 @@ import akka.stream.alpakka.elasticsearch._
 import akka.stream.scaladsl.Source
 import org.elasticsearch.client.RestClient
 import spray.json._
-import DefaultJsonProtocol._
 
+/**
+ * Scala API to create Elasticsearch sources.
+ */
 object ElasticsearchSource {
 
   /**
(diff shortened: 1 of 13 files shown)


ennru:
@mbknor Can you have a look?

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/scaladsl/ElasticsearchSource.scala
```diff
@@ -9,14 +9,32 @@ import akka.stream.alpakka.elasticsearch._
...
+
+  /**
+   * Creates a [[akka.stream.scaladsl.Source]] from Elasticsearch that streams [[OutgoingMessage]]s
+   * of Spray's [[spray.json.JsObject]].
    */
-  def apply(indexName: String, typeName: String, query: String, settings: ElasticsearchSourceSettings)(
+  def create(indexName: String,
+             typeName: String,
+             query: String,
+             settings: ElasticsearchSourceSettings = ElasticsearchSourceSettings())(
```
johanandren: Not sure what we have said in Alpakka in general, but default parameter is in general a real headache when it comes to maintaining binary compatibility, so that could be the reason we didn't have it here.

johanandren: For the future I mean, right now we don't guarantee bincomp here.

ennru: Ok, I'll try to keep that in mind.

PR #669, opened 2017-12-22
Amqp keep connection if pending acks

juanjoDiaz:
I haven't been able to reproduce the error happening in prod but I've noticed a small difference between the Java and the Scala test and correct it. I suspect that to be the issue.

### DIFF
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
@@ -4,8 +4,6 @@
 
 package akka.stream.alpakka.amqp
 
-import java.util.concurrent.atomic.AtomicInteger
-
 import akka.Done
 import akka.stream._
 import akka.stream.alpakka.amqp.scaladsl.CommittableIncomingMessage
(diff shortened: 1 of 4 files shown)


juanjoDiaz:
Fixes #665

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
```diff
@@ -53,7 +51,7 @@ final class AmqpRpcFlowStage(settings: AmqpSinkSettings, bufferSize: Int, respon
       private val queue = mutable.Queue[CommittableIncomingMessage]()
       private var queueName: String = _
       private val unackedMessages = new AtomicInteger()
-      private var outstandingMessages = 0
+      private val outstandingMessages = new AtomicInteger()
```
johanandren: As this was actually only touched in async callbacks (and therefore always on the stage thread) it can safely be a "normal" var. I think this is also true for `unackedMessages`, must have been some rework of this stage from an implementation that did modify them concurrently.

juanjoDiaz: I've changed them to var.

amqp/src/test/java/akka/stream/alpakka/amqp/javadsl/AmqpConnectorsTest.java
```diff
@@ -248,128 +245,163 @@ public void publishAndConsumeRpcWithoutAutoAck() throws Exception {
...
-                    .map(bytes -> new OutgoingMessage(bytes, false, false, Optional.empty(), Optional.empty()))
-                    .via(ampqRpcFlow)
-                    .runWith(TestSink.probe(system), materializer);
+    result.first().toCompletableFuture().get(5, TimeUnit.SECONDS);
 
-    Sink<OutgoingMessage, CompletionStage<Done>> amqpSink = AmqpSink.createReplyTo(
-            AmqpReplyToSinkSettings.create(DefaultAmqpConnection.getInstance())
+    Sink<OutgoingMessage, CompletionStage<Done>> amqpSink = AmqpSink.createReplyTo(AmqpReplyToSinkSettings.create(DefaultAmqpConnection.getInstance()));
+
+    final Source<IncomingMessage, NotUsed> amqpSource = AmqpSource.atMostOnceSource(
```
juanjoDiaz: The actual issue wasn't on the stage but on the Java test, I think.

This line was a `committableSource` instead of an `atMostOnceSource` leaving unacked messages in the queue and the buffer size was 10 instead of 1 which might be making the stage pulling one element too many from the source and causing the timeout.
These were the only differences I could find between the scala test and the failing java test.

PR #670, opened 2017-12-24
Support reusable connections

juanjoDiaz:
Second attempt to add reusable connections for AMQP.

This breaks the API in a minimal way. The fix is to wrap the `AmqpConnectionSettings` with a `DefaultAmqpConnectionProvider`.
Or even use an implicit conversion from `AmqpConnectionSettings => DefaultAmqpConnectionProvider(AmqpConnectionSettings)`.

The get/release methods should be thread-safe now. Not sure if that's good enough or you have something else in .


amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -0,0 +1,105 @@
...
+import java.util.concurrent.atomic.AtomicReference
+
+import com.rabbitmq.client.Connection
+
+import scala.compat.java8.FunctionConverters._
+
+/**
+ * Only for internal implementations
+ */
+sealed trait AmqpConnectionProvider {
```
johanandren: `@DoNotInherit`

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  /**
+   * Java API
+   */
+  def create(settings: AmqpConnectionSettings): DefaultAmqpConnectionProvider =
+    DefaultAmqpConnectionProvider(settings: AmqpConnectionSettings)
+}
+
+final case class DefaultAmqpConnectionProvider(settings: AmqpConnectionSettings) extends AmqpConnectionProvider {
+  override def get = settings.getConnection
```
johanandren: Hmmm, shouldn't the connection logic go here rather than live in the settings now that we have a `provider`?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * Java API
+   */
+  def create(settings: AmqpConnectionSettings, automaticRelease: Boolean): ReusableAmqpConnectionProvider =
+    ReusableAmqpConnectionProvider(settings: AmqpConnectionSettings, automaticRelease)
+}
+
+final case class ReusableAmqpConnectionProvider(settings: AmqpConnectionSettings, automaticRelease: Boolean = true)
+    extends AmqpConnectionProvider {
+  var clients = 0
```
johanandren: I think we discussed this before, this is not threadsafe. Make an internal state ADT like so:

```
private sealed trait State
private case object Empty extends State
privat case object Connecting extends State
private final case class Connected(connection: Connection, clients: Int)
private case object Closing extends State
private val state = new AtomicReference[State](Empty)
```

And do CAS on it for all updates like so: 
```
@tailrec
override def get: Connection = {
  state.get match {
     case Empty =>
        if (state.compareAndSet(Empty, Connecting)) {
          // only one thread will get here, so now we can safely connect
          ... cas swap Connected in ...
       } else get // another thread is connecting, try again
    case Connecting => get // another thread is connecting, try again
    case c @ Connected(connection, clients) => 
       if (state.compareAndSet(c, Connected(connection, clients + 1)) connection
       else get // try again
    ... and so on...
```

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -0,0 +1,105 @@
...
+    ReusableAmqpConnectionProvider(settings: AmqpConnectionSettings)
+
+  /**
+   * Java API
+   */
+  def create(settings: AmqpConnectionSettings, automaticRelease: Boolean): ReusableAmqpConnectionProvider =
+    ReusableAmqpConnectionProvider(settings: AmqpConnectionSettings, automaticRelease)
+}
+
+final case class ReusableAmqpConnectionProvider(settings: AmqpConnectionSettings, automaticRelease: Boolean = true)
```
johanandren: Add scaladoc for what `automaticRelease` does

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionSettings.scala
```diff
@@ -0,0 +1,166 @@
...
+sealed trait AmqpConnectionSettings {
+  private[amqp] def getConnection: Connection
+}
+
+/**
+ * Connects to a local AMQP broker at the default port with no password.
+ */
+case class AmqpConnectionLocal() extends AmqpConnectionSettings {
+  override def getConnection: Connection = new ConnectionFactory().newConnection
+}
```
johanandren: No need for a class, use a single `object` extending `AmqpConnectionSettings` instead.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionSettings.scala
```diff
@@ -0,0 +1,166 @@
...
+
+package akka.stream.alpakka.amqp
+
+import com.rabbitmq.client.{Address, Connection, ConnectionFactory, ExceptionHandler}
+
+/**
+ * Only for internal implementations
+ */
+sealed trait AmqpConnectionSettings {
+  private[amqp] def getConnection: Connection
```
johanandren: Shouldn't this be `createConnection` or `newConnection` rather, it should never do any caching, right?

amqp/src/test/scala/akka/stream/alpakka/amqp/scaladsl/AmqpConnectionProvidersSpec.scala
```diff
@@ -0,0 +1,129 @@
...
+
+class AmqpConnectionProvidersSpec extends AmqpSpec {
+  "The AMQP Default Connection Provider" should {
+    "create a new connection per invocation of DefaultAmqpConnection" in {
+      val connectionProvider = DefaultAmqpConnectionProvider.create(DefaultAmqpConnection)
+      val connection1 = connectionProvider.get
+      val connection2 = connectionProvider.get
+      connection1 should not equal connection2
+      connectionProvider.release(connection1)
+      connectionProvider.release(connection2)
```
johanandren: As these calls to get and release can happen concurrently, I think it would be valuable if the tests were concurrent as well. Will not be deterministic but would hopefully have caught the concurrency bug I noted above.

docs/src/main/paradox/amqp.md
```diff
@@ -12,6 +12,19 @@ The AMQP connector provides Akka Stream sources and sinks to connect to AMQP ser
 
 ## Usage
 
+### Connecting to AMQP server
+
+All the AMQP connectors are configured using a @scaladoc[AmqpConnectionProvider](akka.stream.alpakka.amqp.AmqpConnectionProvider) and a list of @scaladoc[Declaration](akka.stream.alpakka.amqp.Declaration)
+
+There are two types of @scaladoc[AmqpConnectionProvider](akka.stream.alpakka.amqp.AmqpConnectionProvider):
+* @scaladoc[DefaultAmqpConnectionProvider](akka.stream.alpakka.amqp.DefaultAmqpConnectionProvider) which create a new connection for each connector and releases it when the stage stops.
+* @scaladoc[ReusableAmqpConnectionProvider](akka.stream.alpakka.amqp.ReusableAmqpConnectionProvider) which reuses the same connections for all stages. By default it closes the connection whenever the last stage using the provider stops. Optionally, it takes `automaticRelease` boolean parameter so the connection is not automatically release and the user have to release it explicitly.
```
johanandren: How could the user release it explicitly when `release` checks the boolean flag?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
juanjoDiaz: It could but then each provider will need to match against all possible settings.
It's more reliable if each settings can provide its own connection so all providers work with all settings.

Other option is what I did in my original PR. Get rid of settings and consider them as providers and the reusable provider just a provider that wrap other providers.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionSettings.scala
juanjoDiaz: I can rename it. It doesn't do caching.

docs/src/main/paradox/amqp.md
juanjoDiaz: Release never check a boolean flag any more since that was specific for the cached connection and we needed a generic release method.
The reusable connection offers a client counter so you can do `provider.clients.forEach(_ => provider.release)` or something like that

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -0,0 +1,105 @@
...
+    extends AmqpConnectionProvider {
+  var clients = 0
+  private val cachedConnectionRef = new AtomicReference[Option[Connection]](None)
+
+  override def get =
+    cachedConnectionRef
+      .updateAndGet(
```
juanjoDiaz: Doesn't this atomic way of updating guarantee that only a thread is running at once?
Similarly to your proposed solution.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
johanandren: Hmm, maybe it's more of a naming thing, but, now we have a provider that does not provide anything but has a special case where it caches connections, and a `Settings` that creates/provides connections, this makes me think we are missing something with the modeling ;)

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
johanandren: The problem is that you side effect inside of the function (see the java doc of `updateAndGet` explicitly saying you cannot do that)

johanandren: The clients update could interleave or even not be seen ever between other threads, meaning that the number is pretty much arbitrary. In addition to that the connect call could be called multiple times in case of contention leading to connections being created and leaked.

juanjoDiaz:
I've completely removed the connection settings and left only connection providers. As you noted, it made no sense having both. I've also renamed the providers to a more consistent naming schema and updated the docs.

I've implemented the non-blocking logic as you suggested.

The only thing left from your requests is to make the test concurrent.
Are you suggesting that I wrap all the creation/release of connections in Futures so they are run in parallel?
Some tests `release -> check -> release -> check`. So if they are concurrent I can't make the checks. Any idea is welcome.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -6,100 +6,210 @@ package akka.stream.alpakka.amqp
...
+
+  import akka.stream.alpakka.amqp.ReusableAmqpConnectionProvider._
+  private val state = new AtomicReference[State](Empty)
+
+  @tailrec
+  override def get: Connection = state.get match {
+    case Empty =>
+      if (state.compareAndSet(Empty, Connecting)) {
+        val connection = provider.get
+        state.compareAndSet(Connecting, Connected(connection, 0))
```
johanandren: Client count should be 1 here

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -6,100 +6,210 @@ package akka.stream.alpakka.amqp
...
+    case c @ Connected(connection, clients) =>
+      if (state.compareAndSet(c, Connected(connection, clients + 1))) connection
+      else get
+    case _ => get
+  }
+
+  @tailrec
+  override def release(connection: Connection): Unit = state.get match {
+    case c @ Connected(cachedConnection, clients) =>
+      if (state.compareAndSet(c, Closing)) {
```
johanandren: Check the cached connection as the first thing in `release` instead of modifying the state and then checking.

Don't change to closing unless the clients are 0 and automatic release is true, for other cases it should just subtract one from the client count.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  @tailrec
+  override def release(connection: Connection): Unit = state.get match {
+    case c @ Connected(cachedConnection, clients) =>
+      if (cachedConnection != connection)
+        throw new IllegalArgumentException("Can't release a connection that's not owned by this provider")
+
+      if (state.compareAndSet(c, Closing)) {
+        if (clients == 0 || !automaticRelease) {
```
johanandren: What thread will see it get 0 if we are the last client and only subtract in the other branch? It should also not do the Closing swap unless it wants to close, which should only be if clients - 1 becomes 0 and automaticRelease, this you can check before doing the cas.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -0,0 +1,216 @@
...
+          provider.release(connection)
+          state.compareAndSet(Closing, Empty)
+        } else {
+          state.compareAndSet(Closing, Connected(cachedConnection, clients - 1))
```
johanandren: So what happens if we just lost this cas because another thread also got here?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
juanjoDiaz: I've updated this to be 1 instead of 0. So when there is only 1 client left, the connection is closed.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
juanjoDiaz: How could that happen?
Doesn't `state.compareAndSet(c, Closing)` in line 206 ensures that only 1 thread can go into this code?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -0,0 +1,216 @@
...
+
+      if (state.compareAndSet(c, Closing)) {
+        if (clients == 1 || !automaticRelease) {
+          provider.release(connection)
+          state.compareAndSet(Closing, Empty)
+        } else {
+          state.compareAndSet(Closing, Connected(cachedConnection, clients - 1))
+        }
+      } else release(connection)
+    case _ => release(connection)
```
johanandren: I think it would be good to not match-all-recurse here but match explicitly, and for the unexpected cases throw an IllegalStateException, so that we get to know if there is a case here that we didn't think of.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -0,0 +1,216 @@
...
+    case _ => get
+  }
+
+  @tailrec
+  override def release(connection: Connection): Unit = state.get match {
+    case c @ Connected(cachedConnection, clients) =>
+      if (cachedConnection != connection)
+        throw new IllegalArgumentException("Can't release a connection that's not owned by this provider")
+
+      if (state.compareAndSet(c, Closing)) {
```
johanandren: It should not always switch to `Closing`, only if it is the last client and automatic release is enabled it should try to win the cas for `Closing`, in the other case it should just update state.compareAndSet(c, c.copy(clients = c.clients - 1))` and it must check if that succeeded and recurse if it failed (because another client may have `released` or called `get` concurrently)

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -191,26 +191,31 @@ final case class AmqpCachedConnectionProvider(provider: AmqpConnectionProvider,
...
+        if (state.compareAndSet(c, Closing)) {
           provider.release(connection)
           state.compareAndSet(Closing, Empty)
-        } else {
-          state.compareAndSet(Closing, Connected(cachedConnection, clients - 1))
         }
-      } else release(connection)
-    case _ => release(connection)
+      } else {
+        if (!state.compareAndSet(Closing, Connected(cachedConnection, clients - 1))) release(connection)
```
johanandren: The cas should not be from `Closing` here.  We observed `c` and we want to update the value based on the assumption that the state didn't change, if we see that assumption was wrong (cas failed) we recurse.

This is what makes closing work if there are two concurrent clients releasing concurrently, both sees `clients == 2` and comes down here, tries to cas, one wins, while the other recurses and sees `clients == 1` going for the close-branch.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -191,26 +191,31 @@ final case class AmqpCachedConnectionProvider(provider: AmqpConnectionProvider,
...
     case c @ Connected(cachedConnection, clients) =>
       if (cachedConnection != connection)
         throw new IllegalArgumentException("Can't release a connection that's not owned by this provider")
 
-      if (state.compareAndSet(c, Closing)) {
-        if (clients == 1 || !automaticRelease) {
+      if (clients == 1 || !automaticRelease) {
+        if (state.compareAndSet(c, Closing)) {
           provider.release(connection)
           state.compareAndSet(Closing, Empty)
```
johanandren: This cas should not fail, but if it did we'd have a bug, so please check the result and throw illegal state if it failed here.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -191,26 +191,31 @@ final case class AmqpCachedConnectionProvider(provider: AmqpConnectionProvider,
...
-        } else {
-          state.compareAndSet(Closing, Connected(cachedConnection, clients - 1))
         }
-      } else release(connection)
-    case _ => release(connection)
+      } else {
+        if (!state.compareAndSet(Closing, Connected(cachedConnection, clients - 1))) release(connection)
+      }
+    case Closing => release(connection)
+    case _ => throw new RuntimeException("Unknown provider state.")
```
johanandren: Throw `IllegalState` and include the _unexpected_ state.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -191,26 +191,31 @@ final case class AmqpCachedConnectionProvider(provider: AmqpConnectionProvider,
...
 
-      if (state.compareAndSet(c, Closing)) {
-        if (clients == 1 || !automaticRelease) {
+      if (clients == 1 || !automaticRelease) {
+        if (state.compareAndSet(c, Closing)) {
           provider.release(connection)
           state.compareAndSet(Closing, Empty)
-        } else {
-          state.compareAndSet(Closing, Connected(cachedConnection, clients - 1))
         }
```
johanandren: If we lost this cas to get to close, we need to recurse

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -191,26 +191,31 @@ final case class AmqpCachedConnectionProvider(provider: AmqpConnectionProvider,
...
       if (state.compareAndSet(c, Connected(connection, clients + 1))) connection
       else get
-    case _ => get
+    case Closing => get
+    case _ => throw new RuntimeException("Unknown provider state.")
   }
 
   @tailrec
   override def release(connection: Connection): Unit = state.get match {
+    case Empty => ()
```
johanandren: Isn't this an illegal-state one as well, how could it be `Empty` when we release unless something is wrong?

juanjoDiaz: If the user call release before the connection is even opened.

We could:
* Throw an exception stating that you can't release if the connection hasn't been opened.
* Just let it be. It's already released so there is nothing to do

I chose option B. Would you prefer option A?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
juanjoDiaz: You mean `state.compareAndSet(Closing, Empty)`?

Didn't you say above that you rather throwing an exception since we should never lose it?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
juanjoDiaz: true!

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
johanandren: Yes, absolutely option A, because it would either be a bug in our concurrency logic and be _us_ calling it or somehow a bug in the user logic, either way we'd not want it to silently behave as everything is ok but to detect that something was not what we thought it is.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
johanandren: I think GitHub made something weird with the offset, because now I'm not sure which one it relates to. I'll take another look at the complete source.

johanandren: In general for *every* cas you do, you should look at the return value and act in some way. I think you are correct here, if it was changed from `Closing` to something else by another thread that is a bug, so throw IllegalState for that.

juanjoDiaz:
Should be all good now.

Issue #673, opened 2017-12-25
JMS: Allow clean shutdown and abort of the stream


akara:
In `ClientAcknowledge` or `SessionTransacted` mode, we should be able to cleanly shutdown the steam or even abort the stream without message loss.

Issue #672, opened 2017-12-25
JMS: Support transactional mode consumption


akara:
Allowing commit and rollback for the message consumption allows consumers to make decision whether the message is properly processed or not.

PR #674, opened 2017-12-25
Solidify JMS adapter for transactional use

### LINKED ISSUE #671: JMS: Support acknowledge on session thread
Calling `message.acknowledge()` from the stream and not from the session thread may not be compliant to JMS spec and can cause issues for some message brokers. `message.acknowledge()` in many cases acknowledges the session and not the message itself, contrary to what the API makes you believe to be. For instance, please see:

https://stackoverflow.com/questions/2490692/using-a-jms-session-from-different-threads

akara:
- Separate stream source stages for tx and ack modes, making the ack or commit go through the session thread and no other thread
- Materializes to KillSwitch for proper shutdown without message loss
- Customizable AcknowledgeModes with reasonable defaults


jms/src/main/scala/akka/stream/alpakka/jms/javadsl/JmsSource.scala
```diff
@@ -6,8 +6,8 @@ package akka.stream.alpakka.jms.javadsl
```
akara: I could have created a separate entry point with all same signatures, like `JmsSource`, `JmsAckSource`, and `JmsTxSource` with the same signatures inside. Just wanted to wait for some opinions from the code review before making these changes.

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
```diff
@@ -6,13 +6,29 @@ package akka.stream.alpakka.jms
...
+}
+
+case class TxEnvelope private[jms] (message: Message, private val jmsSession: JmsTxSession) {
+
+  @volatile var committed = false
+
+  def commit(): Unit = if (!committed) jmsSession.commit()
+
+  def rollback(): Unit = if (!committed) jmsSession.rollback()
+}
```
akara: Also, please let me know for any better ideas than `TxEnvelope` and `AckEnvelope` as type names for the message wrappers that provide safe ack and tx functionality from inside the stream.

jms/src/test/resources/logback-test.xml
```diff
@@ -7,6 +7,8 @@
         </encoder>
     </appender>
 
+    <logger name="org.apache.activemq" level="info" />
```
akara: ActiveMQ with additional tests becomes very noisy at "debug" log level. The sheer volume of the output causes the tests to fail from hitting Travis output limits. Setting ActiveMQ logging to "info" instead.

ennru:
fixes #671
fixes #672
fixes #673

docs/src/main/paradox/jms.md
```diff
@@ -305,7 +309,66 @@ Scala
...
+Scala
+: @@snip ($alpakka$/jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsBufferedAckConnectorsSpec.scala) { #run-jms-source }
+
+Java
+: @@snip ($alpakka$/jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsBufferedAckConnectorsTest.java) { #run-jms-source }
+
+**Notes:**
+
+*  Using multiple sessions increases throughput, especially if a acknowledging message by message is desired.
+*  Messages will arrive out of order if `sessionCount` is larger than 1.
```
johanandren: `may/can` rather than `will`, right?

docs/src/main/paradox/jms.md
```diff
@@ -305,7 +309,66 @@ Scala
...
+*  Message-by-message acknowledgement can be achieved by setting `bufferSize` to 0, thus disabling buffering. The outstanding messages before backpressure will be the `sessionCount`.
+*  The default `AcknowledgeMode` is `ClientAcknowledge` but can be overridden to custom `AcknowledgeMode`s, even implementation-specific ones by setting the `AcknowledgeMode` in the `JmsSourceSettings` when creating the stream.
```
johanandren: Isn't the default `AutoAcknowledge`?

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
```diff
@@ -97,3 +122,41 @@ private[jms] case class JmsSession(connection: jms.Connection, session: jms.Sess
...
+private[jms] class JmsAckSession(override val connection: jms.Connection,
+                                 override val session: jms.Session,
+                                 override val destination: jms.Destination,
+                                 val maxPendingAcks: Int)
+    extends JmsSession(connection, session, destination) {
+
+  private[jms] var pendingAck = 0
+  private[jms] val ackQueue = new ArrayBlockingQueue[() => Unit](maxPendingAcks + 1)
+
+  // def ack(message: jms.Message): Unit = ackQueue.put(message.acknowledge _)
```
johanandren: Why was this replaced with the same thing and commented out?

jms/src/main/scala/akka/stream/alpakka/jms/JmsSinkStage.scala
```diff
@@ -18,15 +19,25 @@ final class JmsSinkStage(settings: JmsSinkSettings) extends GraphStage[SinkShape
   override protected def initialAttributes: Attributes =
     ActorAttributes.dispatcher("akka.stream.default-blocking-io-dispatcher")
 
-  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
-    new GraphStageLogic(shape) with JmsConnector {
+  def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) with JmsConnector with StageLogging {
```
johanandren: StageLogging added but not used

jms/src/main/scala/akka/stream/alpakka/jms/JmsMessages.scala
```diff
@@ -263,12 +263,12 @@ final case class JmsObjectMessage(serializable: java.io.Serializable,
...
    * Java API: adds a Jms header e.g. JMSType [[JmsObjectMessage]]
    */
-  def withHeader(jmsHeader: JmsHeader) = copy(headers = headers + jmsHeader)
+  def withHeader(jmsHeader: JmsHeader): JmsObjectMessage = copy(headers = headers + jmsHeader)
 
   /**
    * Java API: adds JMSProperty [[JmsObjectMessage]]
    */
-  def withProperty(name: String, value: Any) = copy(properties = properties + (name -> value))
+  def withProperty(name: String, value: Any): JmsObjectMessage = copy(properties = properties + (name -> value))
```
johanandren: good

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
```diff
@@ -5,80 +5,319 @@
...
 
+import scala.annotation.tailrec
 import scala.collection.mutable
+import scala.concurrent.Future
+import scala.util.control.NonFatal
 import scala.util.{Failure, Success}
 
-final class JmsSourceStage(settings: JmsSourceSettings) extends GraphStage[SourceShape[Message]] {
+final class JmsSourceStage(settings: JmsSourceSettings)
+    extends GraphStageWithMaterializedValue[SourceShape[Message], KillSwitch] {
```
johanandren: Do we really need to materialize that killswitch? Can't the stage easily keep track of outstanding non-ack:ed `Message`s and keep going until those are done?

akara:
@johanandren I'll address the specific review comments later.

Just on your comment: I'm not sure I understand what the JMS docs say or not, but if they say the session can only be accessed by a single (same) thread then the IO-dispatcher is not enough as that is a thread pool, in that case the stage will need a pinned dispatcher.

This is not quite true. The session runs on a thread created by ActiveMQ. This is the thread that executes the `onMessage` loop. All the commits/acks should essentially be executed from inside `onMessage` - that's what this PR is mainly about. A pinned dispatcher should not be needed.

johanandren:
Ok, thanks for clarifying.

docs/src/main/paradox/jms.md
akara: Almost a guarantee, unless the additional sessions are idle. But I'll change as commented.

docs/src/main/paradox/jms.md
akara: No. I've changed the acknowledgeMode field in the settings to be `Option[AcknowledgeMode]` allowing each source type to define their own default. The default for `ackSource` is `ClientAcknowledge`. The default for the standard no-ack mode is `AutoAcknowledge`, and the default for `txSource` is `SessionTransacted`. We still allow overriding this as seen in one test case of `ackSource` with buffering. To really prevent message loss on abort, you need to use an ActiveMQ-specific mode called `INDIVIDUAL_ACKNOWLEDGE`.

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
akara: Ahhh, thanks for the catch. This is actually the same as below. The below allowed me to add logging while debugging this work. Will reset it back.

jms/src/main/scala/akka/stream/alpakka/jms/JmsSinkStage.scala
akara: Thanks! Another remnant of debugging.

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
akara: Only if you use the proprietary `INDIVIDUAL_ACKNOWLEDGE`. Also, what is the formal way to stop the stream? Sticking an extra `KillSwitch` behind the source allows for messages to be fetched and not processed. This will cause re-delivery in the appropriate ack modes and transactional mode, and message loss in `AUTO_ACKNOWLEDGE`. Actually, IMO, any message consumer sources (Kafka included) should have a proper `KillSwitch` at materialization time. Kindly review the shutdown process in JmsSourceStage.

akara:
Addressed initial review comments with fixes and/or clarifications. Please do keep comments coming. Thank you very much!

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
```diff
@@ -0,0 +1,371 @@
...
+          env.acknowledge()
+        }
+        .to(Sink.ignore)
+        .run()
+
+      Thread.sleep(2000)
+      killSwitch.shutdown()
+
+      Thread.sleep(500)
+      println("Elements in resultQueue before first shutdown: " + resultQueue.size)
```
johanandren: Remove printlns or turn into debug-logs, or assertions if there is something to assert, `resultQueue` being non-empty perhaps?

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
```diff
@@ -0,0 +1,371 @@
...
+      val url: String = ctx.url
+      val connectionFactory = new ActiveMQConnectionFactory(url)
+
+      val jmsSink: Sink[JmsTextMessage, NotUsed] = JmsSink(
+        JmsSinkSettings(connectionFactory).withQueue("numbers")
+      )
+
+      val numsIn = 1 to 100
+      val msgsIn = numsIn map { n =>
+        JmsTextMessage(n.toString).add("Number", n)
```
johanandren: `add` is deprecated, use `withProperty` instead

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
```diff
@@ -0,0 +1,371 @@
...
+        .take(msgsIn.size)
+        .throttle(10, 1.second, 2, ThrottleMode.shaping)
+        .map { env =>
+          resultQueue.add(env.message.asInstanceOf[TextMessage].getText)
+          env.acknowledge()
+        }
+        .to(Sink.ignore)
+        .run()
+
+      Thread.sleep(2000)
```
johanandren: Please add a comment in front of the sleeps providing some rationale showing that the sleep time isn't arbitrary.

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
```diff
@@ -0,0 +1,371 @@
...
+
+      val killSwitch = jmsSource
+        .take(msgsIn.size)
+        .throttle(10, 1.second, 2, ThrottleMode.shaping)
+        .map { env =>
+          resultQueue.add(env.message.asInstanceOf[TextMessage].getText)
+          env.acknowledge()
+        }
+        .to(Sink.ignore)
+        .run()
```
johanandren: Entire part from `map` to `run` can be a single `runForeach` as all it does is side effect (same fora few more identical below)

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
```diff
@@ -0,0 +1,371 @@
...
+
+      Thread.sleep(2000)
+      killSwitch.shutdown()
+
+      Thread.sleep(500)
```
johanandren: Isn't what we actually want here to wait for the stream to complete? Blocking on `Future[Done]` from `Sink.ignore` would let us do that instead of arbitrary sleep.

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
```diff
@@ -0,0 +1,371 @@
...
+
+      resultList should contain theSameElementsAs numsIn.map(_.toString)
+    }
+
+    "ensure no message loss when aborting a stream" in withServer() { ctx =>
+      val url: String = ctx.url
+      val connectionFactory = new ActiveMQConnectionFactory(url)
+
+      val jmsSink: Sink[JmsTextMessage, NotUsed] = JmsSink(
+        JmsSinkSettings(connectionFactory).withQueue("numbers")
```
johanandren: Isn't there a risk that the tests interfere with each other if we use the same queue in many test cases?

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
```diff
@@ -0,0 +1,371 @@
...
+      //#run-topic-sink
+      Source(inNumbers).runWith(jmsTopicSink2)
+
+      val expectedList: List[String] = in ++ inNumbers
+      result1.futureValue should contain theSameElementsAs expectedList
+      result2.futureValue should contain theSameElementsAs expectedList
+    }
+
+    "ensure no message loss when stopping a stream" in withServer() { ctx =>
+      val url: String = ctx.url
```
johanandren: only used in the row below, so no need to put in an explicitly typed val here

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
```diff
@@ -0,0 +1,371 @@
...
+
+      keepPolling()
+
+      println("Elements in resultList now: " + resultList.size)
+      killSwitch2.shutdown()
+
+      implicit val _ = Ordering.by { s: String =>
+        s.toInt
+      }
+      resultList.to[SortedSet] should contain theSameElementsAs numsIn.map(_.toString)
```
johanandren: No need to do ordering and use sorted set, `toSet should contain theSameElementsAs ...` is enough.

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
```diff
@@ -0,0 +1,371 @@
+/*
+ * Copyright (C) 2016-2017 Lightbend Inc. <http://www.lightbend.com>
```
johanandren: Should be `2018` now :)

and no start year because that is the year when the code was created.

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
```diff
@@ -5,80 +5,319 @@
...
 
   private val out = Outlet[Message]("JmsSource.out")
 
   override def shape: SourceShape[Message] = SourceShape[Message](out)
 
-  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
-    new GraphStageLogic(shape) with JmsConnector with StageLogging {
+  override def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, KillSwitch) = {
+    val logic = new NoAckLogic(shape, out, settings, inheritedAttributes)
```
johanandren: I think putting the graph stage logic inline here and accessing the killswitch with reflective access instead of having an explicit class is fine. Following the logic around here gets a bit convoluted for no obvious reason like this.

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
```diff
@@ -5,80 +5,319 @@
...
+  }
+
+  override private[jms] def onSessionOpened(jmsSession: JmsSession): Unit =
+    jmsSession
+      .createConsumer(settings.selector)
+      .onComplete {
+        case Success(consumer) =>
+          consumer.setMessageListener(new MessageListener {
+            def onMessage(message: Message): Unit = {
+              backpressure.acquire()
```
johanandren: So here we are blocking on a thread that belongs to the JMS provider/client?

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
```diff
@@ -5,80 +5,319 @@
...
+        case _ =>
+          val closeSessionFutures = jmsSessions.map { s =>
+            val f = s.closeSessionAsync()
+            f.onFailure { case e => log.error(e, "Error closing jms session") }
+            f
+          }
+          Future.sequence(closeSessionFutures).onComplete { _ =>
+            try {
+              jmsConnection.close()
+              log.info("JMS connection {} closed", jmsConnection)
```
johanandren: What does the toString on jmsConnection return? No secrets or arbitrary long strings?

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
```diff
@@ -5,80 +5,319 @@
...
+    jmsSession match {
+      case session: JmsAckSession =>
+        session.createConsumer(settings.selector).onComplete {
           case Success(consumer) =>
             consumer.setMessageListener(new MessageListener {
-              override def onMessage(message: Message): Unit = {
-                backpressure.acquire()
-                handleMessage.invoke(message)
+
+              var listenerStopped = false
```
johanandren: Is `onMessage` guaranteed to be called by the same thread? If not this is not thread safe, even if they are not read and written concurrently the visibility may be a problem. (same question goes for `session.pendingAck` mutated below)

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
```diff
@@ -5,80 +5,319 @@
...
+
+              def onMessage(message: Message): Unit = {
+
+                @tailrec
+                def ackQueued(): Unit =
+                  Option(session.ackQueue.poll()) match {
```
johanandren: Minor performance thing, if you can use `akka.util.OptionVal` or just a null check here it saves us an allocation (and subsequent gc) per queued ack.

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
```diff
@@ -5,80 +5,319 @@
...
+                    handleMessage.invoke(AckEnvelope(message, session))
+                    session.pendingAck += 1
+                    if (session.pendingAck > maxPendingAck) {
+                      val action = session.ackQueue.take()
+                      action()
+                      session.pendingAck -= 1
+                    }
+                    ackQueued()
+                  } catch {
+                    case e: java.lang.IllegalStateException if e.getMessage == "Shutdown" =>
```
johanandren: just `IllegalStateException` is fine

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
```diff
@@ -5,80 +5,319 @@
...
+                       out: Outlet[AckEnvelope],
+                       settings: JmsSourceSettings,
+                       attributes: Attributes)
+    extends SourceStageLogic[AckEnvelope](shape, out, settings, attributes) {
 
-      override private[jms] def onSessionOpened(): Unit =
-        jmsSession.createConsumer(settings.selector).onComplete {
+  private val maxPendingAck = settings.bufferSize
+
+  private[jms] def createSession(connection: Connection, createDestination: Session => javax.jms.Destination) = {
```
johanandren: I think a bunch of these can either be entirely `private` or `protected` (may be a left over of previous code rather than introduced in this PR but please look those through and adjust as needed)

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
akara: That's a very good point. There are no secrets in case of ActiveMQ, but the same cannot be said for arbitrary messaging systems. It is better to remove this.

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
akara: Let me move these in. You're right, it is better to have it inline.

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
akara: Yes, this is the same code as before. We do need to block the `onMessage` to pause consumption on that session.

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
akara: Will do. Thanks!

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
akara: This is a very good observation. To the best of my knowledge, it is always called from the same session thread. At least it is in all our tests. Also the merit of the spec also intended each session to be single-threaded.

Having that said, I need to find language in the spec (or lack thereof) that could open this to multiple threads.

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
akara: Nice! Let me try that. Thanks for the advice.

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
akara: I'm no Lightbend lawyer. This is how it was (in the other files). I can change it as you wish provided it does not cause any other effect. Please confirm.

ennru: Just leave it, the sbt header plugin will replace it with 2018 before we release anyway.

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
akara: That is correct, until you have a mismatch. And since re-delivery is always out of order, it is extremely hard to figure out what is missing. This is why I prefer to keep it sorted, just for the sanity of looking at the failed results when it does fail.

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
akara: As there is no evidence of yet there is sensitive information or excessively long strings, I'll leave the error cases in, though. Removing it is easy if anybody finds that is an issue. Please let me know whether you agree.

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
akara: Checked them. All implement or override the trait `JmsConnector` which I defined as `private[jms]`. For this reason they cannot be private. The call now is to make them `protected` or `private[jms]` which has the same effect but is just package private. Can you please confirm you really want to open this up for everyone to inherit/override? With confirmation I can change to `protected`. Thanks!

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
akara: That also means all files I've touched as part of this PR need to be updated to `2016-2018`. Making the changes now.

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
akara: Yup, this was copied from `JmsConnectorsSpec`. I'll clean it all up (including the old code).

akara: While cleaning up, I also see we don't even have to create the val. We can use `ctx.url` directly. There is just one use anyway. Change across the board?

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
akara: For all the tests we have, each test spawns its separate broker. There is nothing persisted so each test is really separate. For what I do for our own use cases of this, I start one broker per test suite. In that case using the same queue may have side-effects (depending on how clean you leave it.

The barebones ActiveMQ broker is light enough to keep going with this scheme. Fixing this would need restructuring the tests - even the ones already there. I'm happy to help, but I feel this needs a separate issue.

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
akara: I'm fixing this where I can. Since `runForeach` does not give me the `Source`s materialized value, I'll have to stick to this for the tests I need the `KillSwitch` back out. But if there are better suggestions, please do let me know.

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
akara: Sorry, I've made all these changes. All it takes is one build to undo my changes. `The sbt-header` plugin is doing it and the configuration is here: https://github.com/akka/alpakka/blob/master/project/Common.scala#L53

That change should not be in scope for this PR.

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
akara: This comment has not yet been addressed for this round, for sanity of the reviewers 😄 since it will be a restructure/refactor that hides the other previous comments. I've addressed pretty much every other comment.

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
akara: Sorry, @ennru. Somehow I did not see your comments during daytime today.

akara:
Last commit did the inlining of graph stage logic.

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
johanandren: Ah, didn't think of that, then I think `.to(Sink.foreach { side-effecting })` would make it more clear than mapping elements to `Unit`

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsAckConnectorsSpec.scala
johanandren: I forgot that we (will) have the plugin release plugin, you can just ignore this.

johanandren:
Incorrect assertion?

```
- should ensure no message loss when stopping a stream *** FAILED *** (49 seconds, 898 milliseconds)
[info]   500 was not less than 500 (JmsConnectorsSpec.scala:479)
[info]   at org.scalatest.MatchersHelper$.indicateFailure(MatchersHelper.scala:340)
```

ennru:
I just merge a PR which makes JmsSinks materialize with `Future[Done]` a few conflicts arise. Please rebase on master.

akara:
@johanandren The assertion is correct. That assertion will make sure we kill the stream while we have not yet exhausted the input so we can indeed test for no message drop. But sorry, any assumption on timing goes out of vac by Travis - that's a lesson I should have learnt long time ago. I just updated the message loss tests to ensure there is no timing assumption at all. Those are the only changes. Please review (as soon as I rebase).

akara:
@ennru Thanks for the info. Let me rebase.

akara:
Should be all good now. Also took care of one review comment I missed the second round - to not map elements to `Unit` and use `Sink.foreach` instead.

johanandren:
Great. Thanks!

russellyou:
May I know the next release date please? I have got exact the same issue in production. Each time I close the stream using kill switch, some messages were lost. Even after I changing to bufferSize to 1, 1 msg was lost. 

Appreciate if a quick minor release is possible.   Thanks a lot.

ennru:
We plan to release 0.16 this week.

viktorklang:
@ennru Have the 0.16 bits been pushed to Maven Central? (I.e. are we waiting for it to sync or are we waiting to push the bits?)

ennru:
I'm having trouble with the deploy from Travis and the Akka team is too busy to help me, I'm afraid.

viktorklang:
@ennru What seems to be the problem? Do you have a conversation open already?

jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala
```diff
@@ -5,80 +5,304 @@
...
+                      }
+
+                    if (!listenerStopped)
+                      try {
+                        handleMessage.invoke(AckEnvelope(message, session))
+                        session.pendingAck += 1
+                        if (session.pendingAck > maxPendingAck) {
+                          val action = session.ackQueue.take()
+                          action()
+                          session.pendingAck -= 1
```
matthedude: @akara will this work for `CLIENT_ACKNOWLEDGE`? As it acknowledges all the messages on the session shouldn't this be reset to 0?

matthedude: @akara Or thinking about it for `CLIENT_ACKNOWLEDGE` should it only be possible to set `maxPendingAck` to 0?

johanandren: This PR is already merged, please open a new ticket for issues or use some other channel (gitter or akka-user mailing list) if it is more of a discussion/question, thanks!

matthedude: Ah sorry was just looking at the history and didn't think it's probably a strange communication channel! I'll get on gitter or create an issue if I find something wrong. Thanks.

Issue #181, opened 2017-02-11
S3: custom s3 endpoint


l15k4:
Hey,

I'm using https://github.com/jubos/fake-s3 extensively in many projects and I'm very satisfied with it so far. 
It requires the aws s3 client to be set with : 
```
s3.setEndpoint(s"http://localhost:1234")
s3.setS3ClientOptions(S3ClientOptions.builder().setPathStyleAccess(true).disableChunkedEncoding().build())
```
for it to work. At least the endpoint, the second setting is not necessary. Is there a way to change endpoint without changing code? It is not possible with the `proxy`

francisdb:
Interesting, I wonder if there is a similar project for the jvm

carstenlenz:
We use https://github.com/findify/s3mock 
Sadly there are some issues with Alpakka S3 and S3Mock.

l15k4:
Yeah, I've been using both. But I'm way more productive with fake-s3, s3mock had too many issues, fake-s3 is almost flawless for testing purposes.

ennru:
The work of #619 is now continued in #677

easel:
For reference, there is a (passing) unit test in #619 that uses custom endpoints to connect to minio, including a one-liner to set up and configure minio for testing using docker. From what I can tell it's a fairly complete s3 implementation and worked as expected after I made a tweak to the url encoding of paths.

ennru:
Fixed with #677

PR #619, opened 2017-12-02
S3: support custom endpoint - WIP

takashima0411:
#181 
WIP.
get unexpected exceptions.
at use fakes3 on localhost.

at fileupload
```
[info]   akka.stream.alpakka.s3.impl.FailedUpload: Cannot find etag
[info]   at akka.stream.alpakka.s3.impl.S3Stream.$anonfun$completionSink$2(S3Stream.scala:237)
[info]   at scala.concurrent.Future.$anonfun$flatMap$1(Future.scala:302)
[info]   at scala.concurrent.impl.Promise.$anonfun$transformWith$1(Promise.scala:37)
[info]   at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:60)
[info]   at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:55)
[info]   at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:91)
[info]   at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
[info]   at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:81)
[info]   at akka.dispatch.BatchingExecutor$BlockableBatch.run(BatchingExecutor.scala:91)
[info]   at akka.dispatch.TaskInvocation.run(AbstractDispatcher.scala:39)
[info]   at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(AbstractDispatcher.scala:415)
[info]   at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
[info]   at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
[info]   at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
[info]   at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)
``` 

at download
```
[info] - should support custom endpoint configured by `baseUrl`2 *** FAILED *** (842 milliseconds)
[info]   akka.stream.alpakka.s3.S3Exception: The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.
[info]   at akka.stream.alpakka.s3.impl.S3Stream.$anonfun$entityForSuccess$1(S3Stream.scala:263)
[...]

PR #677, opened 2017-12-30
Support custom s3 api endpoints

easel:
Rebased #619 against latest master.

### DIFF
--- a/s3/src/main/resources/reference.conf
+++ b/s3/src/main/resources/reference.conf
@@ -75,5 +75,5 @@ akka.stream.alpakka.s3 {
   path-style-access = false
 
   # Custom endpoint url, used for alternate s3 implementations
-  endpoint-url = null
+  # endpoint-url = null
 }
--- a/s3/src/main/resources/reference.conf
+++ b/s3/src/main/resources/reference.conf
@@ -73,4 +73,7 @@ akka.stream.alpakka.s3 {
   # When using virtual hosted–style buckets with SSL, the S3 wild card certificate only matches buckets that do not contain periods.
   # Buckets containing periods will lead to certificate errors. In those cases it's useful to enable path-style access.
   path-style-access = false
+
+  # Custom endpoint url, used for alternate s3 implementations
+  endpoint-url = null
 }
--- a/s3/src/test/java/akka/stream/alpakka/s3/javadsl/DocSnippets.java
+++ b/s3/src/test/java/akka/stream/alpakka/s3/javadsl/DocSnippets.java
@@ -48,7 +48,7 @@ public class DocSnippets extends S3WireMockBase {
                 credentials,
                 regionProvider,
                 true,
-                Option.<String>none().asScala() //FIXME
+                scala.Option.empty()
         );
         final S3Client s3Client = new S3Client(settings,system(), mat);
         // #java-bluemix-example
--- a/s3/src/test/java/akka/stream/alpakka/s3/javadsl/DocSnippets.java
+++ b/s3/src/test/java/akka/stream/alpakka/s3/javadsl/DocSnippets.java
@@ -4,6 +4,7 @@
 
 package akka.stream.alpakka.s3.javadsl;
 
+import akka.japi.Option;
 import akka.stream.ActorMaterializer;
 import akka.stream.Materializer;
 import akka.stream.alpakka.s3.MemoryBufferType;
@@ -46,7 +47,8 @@ public class DocSnippets extends S3WireMockBase {
                 Some.apply(proxy),
                 credentials,
                 regionProvider,
-                true
+                true,
+                Option.<String>none().asScala() //FIXME
         );
         final S3Client s3Client = new S3Client(settings,system(), mat);
         // #java-bluemix-example
--- a/s3/src/test/java/akka/stream/alpakka/s3/javadsl/S3ClientTest.java
+++ b/s3/src/test/java/akka/stream/alpakka/s3/javadsl/S3ClientTest.java
@@ -31,10 +31,6 @@ import org.junit.Test;
 import scala.Option;
 import scala.Some;
 
-import java.util.Arrays;
-import java.util.concurrent.CompletionStage;
-import java.util.concurrent.TimeUnit;
-
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
--- a/s3/src/test/java/akka/stream/alpakka/s3/javadsl/S3ClientTest.java
+++ b/s3/src/test/java/akka/stream/alpakka/s3/javadsl/S3ClientTest.java
@@ -60,7 +60,7 @@ public class S3ClientTest extends S3WireMockBase {
             credentials,
             regionProvider("us-east-1"),
             false,
-            akka.japi.Option.<String>none().asScala() //FIXME
+            scala.Option.empty()
     );
     private final S3Client client = new S3Client(settings, system(), materializer);
     //#client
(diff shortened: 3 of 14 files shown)


easel:
I've added some documentation and adjusted the s3 integration tests to work against Minio. Aside from this patch, getting minio to pass the tests requires falling back to url encoding the responses from the upload calls. 

The regular test suite still passes for me, and unless I'm missing something I don't believe any of the changes should change the code path unless an endpoint url is specified. Happy to address any other feedback.

@takashima0411 @ktoso any thoughts?

ktoso:
Sorry for the delay, family time back home until New Years :)
Will review then!

easel:
Thanks @ktoso, happy New Year!

ennru:
Does this replace #619 ? Please close it if so.

easel:
Yes, it's includes all of #619 but includes fixes, additional capabilities and tests. I didn't create #619 so I don't believe I can update it or delete it, hence the new PR.

francisdb:
Please keep the `AwsS3IntegrationSpec` as the same previous refactor to `S3IntegrationSpec` was no longer doing proper validation of all AWS S3 cases.

I personally think it's better to have a specific test for each service as they each have their edge cases and url path issues.

easel:
The AWSS3IntegrationSpec is still there, I just lifted the test methods up to a trait so I could inject the differing config for this patch. The tests are all identical to before and they all pass against minio -- as far as I am concerned if a provider claims to be s3 compatible it should pass the same integration suite.

francisdb:
Seems I was a bit too fast, it's the `S3NoMock` test that is totally broken and that was why I re-introduced an old version as the AWSS3IntegrationSpec.

s3/src/main/resources/reference.conf
```diff
@@ -73,4 +73,7 @@ akka.stream.alpakka.s3 {
   # When using virtual hosted–style buckets with SSL, the S3 wild card certificate only matches buckets that do not contain periods.
   # Buckets containing periods will lead to certificate errors. In those cases it's useful to enable path-style access.
   path-style-access = false
+
+  # Custom endpoint url, used for alternate s3 implementations
+  endpoint-url = null
```
francisdb: I wonder if this is discouraged

> If you do need a setting to be optional: checking hasPath() in advance should be the same amount of code (in Java) as checking for null afterward, without the risk of NullPointerException when you forget. In Scala, you could write an enrichment class like this to use the idiomatic Option syntax:
https://github.com/lightbend/config#how-to-handle-defaults

s3/src/test/java/akka/stream/alpakka/s3/javadsl/DocSnippets.java
```diff
@@ -46,7 +47,8 @@ public String getRegion() {
                 Some.apply(proxy),
                 credentials,
                 regionProvider,
-                true
+                true,
+                Option.<String>none().asScala() //FIXME
```
francisdb: Still needs fix to allow a java Optional?

easel:
Yeah, I couldn't figure that one out at all =p The AWSS3IntegrationSpec worked ok, I just added some delete calls to make it idempotent.

s3/src/test/java/akka/stream/alpakka/s3/javadsl/S3ClientTest.java
```diff
@@ -58,7 +59,8 @@ public String getRegion() throws SdkClientException {
             Some.apply(proxy),
             credentials,
             regionProvider("us-east-1"),
-            false
+            false,
+            akka.japi.Option.<String>none().asScala() //FIXME
```
francisdb: Still needs fix to allow a java Optional?

s3/src/main/resources/reference.conf
easel: Perhaps? I guess the the alternate solution would be to include the key, commented out. It seems like it should be documented in the reference.conf file in some capacity or another.

s3/src/test/java/akka/stream/alpakka/s3/javadsl/DocSnippets.java
easel: I'm not sure I'm following. Are you suggesting we could simplify that last line?

francisdb: I saw the FIXME so my guess was you wanted something cleaner?

easel: Ah, no, sorry, I carried that over from master when I merged this branch with master. I'm not sure what the intent was... off-hand I'd guess they'd prefer to provide something smarter than a hardcoded none but weren't sure what.

ennru:
@easel and @francisdb can you get this one ready for merge and release? I'd like to get 0.16 out on Thursday.

francisdb:
> francisdb approved these changes 5 days ago 
just added some small remarks

s3/src/test/java/akka/stream/alpakka/s3/javadsl/S3ClientTest.java
ennru: What about `scala.Option.empty()`?
We should add Java API to S3Settings to bridge the case class. (see e.g. JmsSourceSettings), but it doesn't need to be done in the PR.

takashima0411:
@easel sorry for delay, and thank you for the patches. 
Minio seems good  :+1:

easel:
Thanks all for the feedback. I've rebased against master and adjusted things per earlier comments.

- Removed the remaining references to baseUrl
- Default to a commented out endpoint-url key, and added tests that ensure missing keys and resetting to null both are handled as None
- Switched java options to scala.Option.empty(), which compiles. I don't use java api's anymore so I can't say if it's a better solution personally but the code is cleaner!

ennru:
Fixes #181 

Thank you for you contribution! Keep going!

easel:
Thanks @ennru !

ennru:
Cheers. We 'll release 0.16 on Thursday.

l15k4:
Thanks guys, it seems to be working with https://github.com/findify/s3mock but I still cannot figure out why multipart upload doesn't work on https://github.com/jubos/fake-s3 ...

I set it this way : 
```
      endpointUrl = Some(s"http://localhost:$randomPort"),
      pathStyleAccess = true
```

```
akka.http.scaladsl.unmarshalling.Unmarshaller$UnsupportedContentTypeException: Unsupported Content-Type, supported: application/xml, application/octet-stream
	at akka.http.scaladsl.unmarshalling.Unmarshaller$UnsupportedContentTypeException$.apply(Unmarshaller.scala:158)
	at akka.http.scaladsl.unmarshalling.Unmarshaller$EnhancedFromEntityUnmarshaller$.$anonfun$forContentTypes$3(Unmarshaller.scala:114)
	at akka.http.scaladsl.unmarshalling.Unmarshaller$$anon$1.apply(Unmarshaller.scala:58)
	at akka.http.scaladsl.unmarshalling.Unmarshaller$EnhancedUnmarshaller$.$anonfun$mapWithInput$3(Unmarshaller.scala:91)
	at akka.http.scaladsl.unmarshalling.Unmarshaller$$anon$1.apply(Unmarshaller.scala:58)
	at akka.http.scaladsl.unmarshalling.Unmarshaller.$anonfun$transform$3(Unmarshaller.scala:23)
	at akka.http.scaladsl.unmarshalling.Unmarshaller$$anon$1.apply(Unmarshaller.scala:58)
	at akka.http.scaladsl.unmarshalling.Unmarshal.to(Unmarshal.scala:25)
	at akka.stream.alpakka.s3.impl.S3Stream.$anonfun$initiateMultipartUpload$3(S3Stream.scala:216)
```

bandrzejczak:
@l15k4 Yes it does. While using standard S3 client I had to disable chunked encoding to make it work with fake-s3. It still allows for sending your document in multiple chunks though, right?

PR #679, opened 2017-12-31
S3: Format x-amz-date with year-of-era

guersam:
"YYYY" in `DateTimeFormatter.ofPattern("YYYYMMdd")` means
week-based-year, which formats `LocalDate.of(2017, 12, 31)` as
"20181231" and results in unexpected rejections from S3 as below:

> The difference between the request time and the current time is too large.

Replacing it with "yyyy" fixes the bug.


ktoso:
Resolves https://github.com/akka/alpakka/pull/679

PR #688, opened 2018-01-04
MQTT: Make actor system names unique per spec

ennru:
Introduce unique names for different tests / specs.


mqtt/src/test/scala/akka/stream/alpakka/mqtt/scaladsl/MqttSourceSpec.scala
```diff
@@ -22,7 +22,7 @@ import scala.concurrent.Await
 import scala.concurrent.duration._
 
 class MqttSourceSpec
-    extends TestKit(ActorSystem("MqttSinkSpec"))
+    extends TestKit(ActorSystem("MqttSourceSpec"))
```
raboof: ;)

Issue #686, opened 2018-01-03
Elasticsearch: Add the ability to create an upsert sink


jmengoni:
Elasticsearch offers the possiblity to upsert documents through the Update API:
[https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html#_literal_doc_as_upsert_literal](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html#_literal_doc_as_upsert_literal)

This fonctionnality is compatible with the Bulk API:
[https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-update](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-update)

Would it be possible to create a sink to upsert documents instead of inserting them ?

PR #689, opened 2018-01-04
Elasticsearch: Implement upsert functionnality

jmengoni:
Fixes [#686](https://github.com/akka/alpakka/issues/686)

Added a new sink setting **docAsUpsert**. If _true_, the sink uses Elasticsearch upsert method.
Added a test:

- Insert using upsert setting
- Partial update using upsert setting


ennru:
ScalaFormat forces another format. Run `test:compile` to run ScalaFormat locally, than push the re-formatted files.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
```diff
@@ -203,6 +203,21 @@ class ElasticsearchFlowStage[T, C](
...
+          writer.convert(message.source)
+        } else {
+          JsObject(
+            "doc" -> writer.convert(message.source).parseJson,
+            "doc_as_upsert" -> JsTrue
+          ).toString
+        }
+      }
+
+      private def getInsertKeyword(): String = {
```
ennru: You could initiate a `val` with the correct keyword instead of a method.

ennru:
Thank you for this contribution. Keep going!

jmengoni:
Thank you for your help. It was a pleasure !

PR #690, opened 2018-01-04
Self-contained examples

ennru:
Finally found a way to integrate what I did in https://scalasthlm.github.io/alpakka-integration-patterns/
I had to create another module for the example code as there would be a circular depenency between `alpakka` and `docs` if the code would be in `docs`. (The code in there was not compiled as `docs` is not aggregated in `alpakka`.)


jboner:
Really nice job @ennru.

ennru:
Tack Jonas, 
will get it out on the net on Thursday with the 0.16 release.

jboner:
Excellent. Thanks a lot for your contributions. Good stuff. Keep it coming ;)

Issue #695, opened 2018-01-07
MqttFlowStage does not close the underlaying client on stop


mwegrz:
`MqttFlowStage` [instantiates](https://github.com/akka/alpakka/blob/052757f009ef79a6bc428aceab93ffb6f487faa2/mqtt/src/main/scala/akka/stream/alpakka/mqtt/MqttFlowStage.scala#L222) a new `MqttAsyncClient` on every materialization. When stopping, the underlaying client gets [disconnected](https://github.com/akka/alpakka/blob/052757f009ef79a6bc428aceab93ffb6f487faa2/mqtt/src/main/scala/akka/stream/alpakka/mqtt/MqttFlowStage.scala#L222) but never properly [closed](https://www.eclipse.org/paho/files/javadoc/org/eclipse/paho/client/mqttv3/IMqttAsyncClient.html#close--). This leaks resources and eventually leads to an`OutOfMemoryException`.

PR #696, opened 2018-01-07
Close the underlying MQTT client when stopping MqttFlowStage

mwegrz:
Resolves https://github.com/akka/alpakka/issues/695


mqtt/src/main/scala/akka/stream/alpakka/mqtt/MqttFlowStage.scala
```diff
@@ -219,8 +219,13 @@ final class MqttFlowStage(sourceSettings: MqttSourceSettings,
             .tryFailure(
               new IllegalStateException("Cannot complete subscription because the stage is about to stop or fail")
             )
-        try mqttClient.disconnect()
-        catch { case _: Throwable => () }
+
+        // Try to disconnect. In case disconnect attempt failed try again with force and throw here if it does not work
+        Try(mqttClient.disconnect().waitForCompletion())
```
ennru: I'm not sure if blocking with `waitForCompletion` is a good choice here. 
Should it be wrapped in Futures, @johanandren ?

johanandren: Yeah, the stage does not run on the blocking-io-dispatcher, so this should be done in futures, on the blocking-io-dispatcher.

mwegrz:
@ennru, @johanandren, thx for comments. Have addressed the problem differently, with an `IMqttActionListener` so blocking/waiting is done on MQTT client's pool.

ennru:
Great, sounds even better than picking the blocking io dispatcher in Akka.
But in the case of falling back to forced disconnect, there is the same issue. The MQTT docs state "It will wait for a maximum of 30 seconds for work to quiesce before disconnecting and wait for a maximum of 10 seconds for sending the disconnect packet to server."

mwegrz:
@ennru, I think the forced disconnect doesn't happen on Akka's dispatcher but on MQTT client's executor service so we might be OK here. Is my understanding right?

I also added two more settings, `disconnectQuiesceTimeout` and `disconnectTimeout` to give full control over the process.

johanandren:
Ach, missed the last conversation about the timeouts. Will look again when that is in place.

johanandren:
And that was already pushed as well, was looking at a stale page. Dang. NVM original LGTM stands then :)

ennru:
Thank you for fixing this "client leak".

PR #703, opened 2018-01-08
AMQP: add support for native ConnectionFactory as a configuration option

agboom:
This PR adds a new AmqpConnectionSettings type that accepts a native `ConnectionFactory` from the underlying RabbitMQ client.

The motivation is that I needed to give an SSL truststore with an imported mutual certificate. This required me to inject a `SocketFactory` with that truststore directly into the AMQP `ConnectionFactory`. This isn't supported by the `AmqpConnectionDetails` class. I thought it might be useful in some more complex use cases (like this one) to give the underlying `ConnectionFactory` directly to the Alpakka AMQP graphstage, so you can use the more advanced features of the Java RabbitMQ client.

The new settings type (`AmqpConnectionFactory`) can be constructed with a single `ConnectionFactory`. I maintained support for multiple host/port combinations with an extra `hostAndPorts` parameter. If `hostAndPorts` is empty, the host/port from the `ConnectionFactory` is used.

Please let me know your feedback.

PR #706, opened 2018-01-09
Amqp reusable connections 3

juanjoDiaz:
The third time lucky!

Continuation of #670

Everything as it was. And also includes the functionality wanted by #703 (connection provider based on a raw ConnectionFactory) but well documented and tested.


docs/src/main/paradox/amqp.md
```diff
@@ -2,12 +2,7 @@
 
 The AMQP connector provides Akka Stream sources and sinks to connect to AMQP servers.
 
-### Reported issues
-
-[Tagged issues at Github](https://github.com/akka/alpakka/labels/p%3Aamqp)
```
ennru: Please keep the reported issues section.

juanjoDiaz: Done!

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -0,0 +1,259 @@
...
+          if (!state.compareAndSet(Closing, Empty))
+            throw new ConcurrentModificationException(
+              "Unexpected concurrent modification while closing the connection."
+            )
+        }
+      } else {
+        if (!state.compareAndSet(c, Connected(cachedConnection, clients - 1))) release(connection)
+      }
+    case Closing => release(connection)
+    case unknownState => throw new IllegalStateException(s"Unknown provider state ($unknownState).")
```
johanandren: Ah, now I see why you wrote "unknown" here, as state is sealed we don't actually have any unknown states, we only need to cover all the possible ones in the ADT, so you can remove the `catch whatever =>` part from both open and release

juanjoDiaz:
Should be good now.

Just for my own understanding @johanandren, could you briefly explain what's the advantage of non-blocking CAS vs synchronized, for example? Aren't we blocking the thread anyway while the CAS is recursing?

ennru:
Compilation errors.
And you'll need to rebase and run `test:compile` to get this year's header edition.

ennru:
@juanjoDiaz There is more to the compilation errors. Classes define twice.
I'll wait for you to fix it before I cut 0.16 (hopefully today).

juanjoDiaz:
Alright. It should be good now!

IntelliJ IDE must do some weird caching and refuses to use the 2018 headers :)

ennru:
Cheers @juanjoDiaz you really kept it rolling here!

PR #715, opened 2018-01-12
jms: fix NPE when connection failed to initialize

dwickern:
When `javax.jms.ConnectionFactory#createConnection()` throws an exception (e.g. failed to connect), we are left with a null connection and throw a NPE when closing the stage.


ennru:
Thank you for fixing this. Looking forward to your other contributions!

PR #716, opened 2018-01-12
jms: add failed test for buffered ack

dwickern:
Failed test for an issue introduced in #674 . When using `ackSource` with `bufferSize > 0`, some messages are left unacknowledged.

I believe it's because `JmsAckSourceStage` only polls the `ackQueue` when a message arrives in the `MessageListener`. So after the last message, if there are still some unprocessed messages in the `ackQueue` they will never be acknowledged.

/cc @akara


ennru:
The Buffered Ack Connector test you changed fails in the Scala 2.12 config, but worked in the Scala 2.11 config. 
```
The JMS Ack Connectors
[info] - should publish and consume strings through a queue *** FAILED *** (5 seconds, 152 milliseconds)
[info]   Vector(AckEnvelope(ActiveMQTextMessage {commandId = 27, responseRequired = true, messageId = ID:travis-job-77292fee-2dd7-4e8d-93e8-710e9b168c47-32814-1516106565046-136:1:1:1:23, originalDestination = null, originalTransactionId = null, producerId = ID:travis-job-77292fee-2dd7-4e8d-93e8-710e9b168c47-32814-1516106565046-136:1:1:1, destination = queue://test, transactionId = null, expiration = 0, timestamp = 1516106814377, arrival = 0, brokerInTime = 1516106814377, brokerOutTime = 1516106814401, correlationId = null, replyTo = null, persistent = true, type = null, priority = 4, groupID = null, groupSequence = 0, targetConsumerId = null, compressed = false, userID = null, content = org.apache.activemq.util.ByteSequence@4d53fbef, marshalledProperties = null, dataStructure = null, redeliveryCounter = 1, size = 0, properties = null, readOnlyProperties = true, readOnlyBody = true, droppable = false, jmsXGroupFirstForConsumer = false, text = w},akka.stream.alpakka.jms.JmsAckSession@3327b461)) was not empty (JmsBufferedAckConnectorsSpec.scala:54)
```

dwickern:
I found some time to take a closer look. I can see 2 problems:

1. `ackQueue` is only polled when the next message arrives to the `onMessage` callback. After the last message arrives, there are usually a few unacknowledged messages still in the queue.

    The test passes when `bufferSize == 0` because the [ackQueue.take()](https://github.com/akka/alpakka/blob/master/jms/src/main/scala/akka/stream/alpakka/jms/JmsSourceStage.scala#L118) blocks for an acknowledgement after each message is consumed.

    Maybe the best solution here is a periodic scheduled timer to poll the ack queue?

1. When `sessionCount > 1`, there is a race condition incrementing/decrementing `pendingAck`. The `onMessage` callback is called from a JMS thread, not from a graph stage thread.

    We can use `AtomicInteger` instead, or move the callback to the graph stage thread using `getAsyncCallback`.

akara:
1. Correct. The only user code that can run on that thread is inside `onMessage`. The problem here is even if you have a scheduler, you won't be able to poll and send the acks from that target thread, which again violates the JMS spec (and even the ActiveMQ impl). The `onMessage` will pick up those acks on its next invocation. This next invocation may not happen only because connector is shutdown/aborted or the stream collapsed by other means. Only if ack mode is `INDIVIDUAL_ACKNOWLEDGE`, those messages won't be acknowledged. They will not be lost but be re-delivered (outside the scope of the test) which meets the at-least-once delivery requirements.
If your ack mode is `CLIENT_ACKNOWLEDGE` you'll have message losses. This is because the ack is on the session and every message sent over the session (and read in the buffer) is acknowledged. I know it is an odd semantic, but that's what it is. Please check whether you can reproduce any loss with `INDIVIDUAL_ACKNOWLEDGE`. If not, it may just be the mis-interpretation of the ack modes.

2. `pendingAck` is actually `session.pendingAck`. The session is supposed to be accessed by only the thread running `onMessage` for that session and is not shared between threads (except for shutting it down). It should not even move threads.  Therefore there should never be a race condition on `session.pendingAck`.

akara:
I actually fail to see the issue for this PR. Given the at-least-once semantics of message delivery in certain ack modes, we should not try to guarantee all messages are acknowledged. These will be re-delivered, anyway. We should rather try to guarantee no messages are lost. And we do have tests for those cases.

ennru:
I think @akara 's comment is very relevant. Verifying all messages are acknowledged is easy, but not as important as one initially might think.

dwickern:
I agree that this behavior is _correct_ as we still have at-least-once delivery guarantees (depending on ack mode). I don't think it's _desirable_ though to have redelivery in even the most trivial case, i.e. write 1 message to a queue, then consume it. I think it will be surprising to users of this library.

I understand now there's nothing we can do to solve this within the constraints of JMS. What we can do is document the behavior and perhaps use a default bufferSize=0 for the ackSource.

Issue #708, opened 2018-01-10
Failed: JmsAckConnectorsSpec


johanandren:
https://travis-ci.org/akka/alpakka/jobs/327144130

```
[info] - should ensure no message loss when stopping a stream *** FAILED *** (2 minutes, 2 seconds)
[info]   A timeout occurred waiting for a future to complete. Queried 1995 times, sleeping 60 milliseconds between each query. (JmsAckConnectorsSpec.scala:279)
[info]   org.scalatest.concurrent.Futures$FutureConcept$$anon$1:
[info]   at org.scalatest.concurrent.Futures$FutureConcept$class.tryTryAgain$1(Futures.scala:538)
[info]   at org.scalatest.concurrent.Futures$FutureConcept$class.futureValueImpl(Futures.scala:550)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:275)
[info]   at org.scalatest.concurrent.Futures$FutureConcept$class.futureValue(Futures.scala:476)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:275)
[info]   at akka.stream.alpakka.jms.scaladsl.JmsAckConnectorsSpec$$anonfun$1$$anonfun$apply$mcV$sp$7$$anonfun$apply$mcV$sp$15.apply(JmsAckConnectorsSpec.scala:279)
[info]   at akka.stream.alpakka.jms.scaladsl.JmsAckConnectorsSpec$$anonfun$1$$anonfun$apply$mcV$sp$7$$anonfun$apply$mcV$sp$15.apply(JmsAckConnectorsSpec.scala:243)
```

akara:
The ActiveMQ client stop time can be a bit indeterministic. It blocks and waits for all sessions to stop. Most of the time we see it stop within 1-2 seconds. In some cases, on my laptop, I can see this test running into the 30+ seconds. On travis I've seen it go close to 1 minute in some cases, but yes, it can go to 2 minutes or more. The patience config is set to 2 minutes at this time. As we can see, it is sometimes not enough, especially with Travis.

This has more to do with active messages being processed in all sessions than with messages in the queue. Not sure the #717 has to do with this.

ennru:
I've seen long shutdown times locally, as well.
It would be an idea to start a single ActiveMQ per spec/test-class instead of one for every single test. We'd need to ensure unique queue/topic names per test of course.

[...]

PR #717, opened 2018-01-12
jms: add queue browser

dwickern:
Exposes the JMS queue browser API which iterates messages on the queue without consuming them.


dwickern:
The failing test is the same transient failure as #708

ennru:
Can you add a section to the documentation in `jms.md` as well, please

jms/src/main/scala/akka/stream/alpakka/jms/JmsBrowseStage.scala
```diff
@@ -0,0 +1,64 @@
...
+    new GraphStageLogic(shape) with OutHandler {
+      setHandler(out, this)
+
+      var connection: Connection = _
+      var session: Session = _
+      var browser: QueueBrowser = _
+      var messages: JEnumeration[Message] = _
+
+      override def preStart(): Unit = {
+        val queue = settings.destination.getOrElse { throw new IllegalArgumentException("Destination is missing") }
```
johanandren: Check this on construction instead of materialization to fail as fast as possible

ennru:
Thank you for your contribution!

Issue #705, opened 2018-01-09
Support for Parquet (file format)


ktoso:
Was proposed by Nils-Helge Garli Hegvik @nilsga on Akka gitter.
Hoping for a contribution :)

ennru:
This is now available in https://doc.akka.io/docs/alpakka/current/avroparquet.html

PR #720, opened 2018-01-14
WIP - Parquet sink for Alpakka

nilsga:
Parquet sinks for producing files in parquet format (#705)


juanjoDiaz:
Hi @nilsga ,

I don't think that we should create connectors for specific file formats.
This should be a flow to format the content and then use the `File` sink to write to disk.

You can see other existing data transformations: https://developer.lightbend.com/docs/alpakka/current/data-transformations/index.html

nilsga:
Thanks for the feedback. That makes sense, but I'm not sure how well that would work with Parquet though? I am certainly no Parquet expert, but the format itself seems very tightly coupled to the concept of a file system, with Hadoop providing the implementation of the different file systems (local files, hdfs, s3 etc). I will certainly look into it though. I'm very open to any suggestions 👍

juanjoDiaz:
I'm not an expert either :)
My assumption is that `AvroParquetWriter` is creating some sort of stream that then is flushing to disk. Probably we want to get that stream and pass it downstream.
This will probably require to go one level deeper into Parquet and I'm not sure if that's even possible.

nilsga:
The easiest seems to be to write an "in memory" filesystem implementation that the `AvroParquetWriter` can use by providing a custom uri scheme. However, it feels very hackish, and like re-inventing the wheel. Maybe the parquet connector should not be part of alpakka, but rather a component of it's own.

nilsga:
It seems like the next version of the parquet libraries for Java has some crucial changes for decoupling from the Hadoop file APIs: https://issues.apache.org/jira/browse/PARQUET-1142

juanjoDiaz:
Yeah. That's exactly what I was thinking.
Sounds very wrong to my that you need to rely on hadoop to use parquet.

I guess we just need to wait for the next parquet release :)

raboof:
Should we pause/close this PR until a new version of Parquet is released?

raboof:
Would appreciate it! Closing for now in anticipation of the parquet release then...

juanjoDiaz:
@nilsga I'd be interested in checking you future PR about this if you remember to add me :)

nilsga:
Sure thing, @juanjoDiaz 👍

ennru:
An Avro Parquet module just got merged with https://github.com/akka/alpakka/pull/1131

PR #722, opened 2018-01-15
JMS: Removed unused (and duplicate) snippet tags

### LINKED ISSUE #721: JMS snippet tags duplicated
Perhaps this is a Paradox issue. But for example: https://developer.lightbend.com/docs/alpakka/current/jms.html#transactionally-receiving-s-from-a-jms-provider

You'll see duplicate Java code samples. Checking the source snips/references, I cannot find anything wrong there. There is only one snip and there are no duplicates on the references.

ennru:
There was a lot of copy-and-paste going on to create the JMS tests...

Fixes #721


akara:
Sorry for missing this on the PR, and even after the search. And thanks for fixing it. From what I see much more than the duplicates are removed.

Issue #723, opened 2018-01-16
Support SQS Message Requests rather than simple string messages


randomcoder:
The SQS send flows and Sinks should support sending of `SqsMessageRequest`s as well as requests generated from Strings.

This will allow the send messages to have message attributes added as well as allowing the API to be more flexible in keeping up with the AWS API and usage changes.

PR #724, opened 2018-01-16
Add support for send message request to sqs

randomcoder:
FIxes akka/alpakka#723

### DIFF
--- a/sqs/src/main/scala/akka/stream/alpakka/sqs/javadsl/SqsFlow.scala
+++ b/sqs/src/main/scala/akka/stream/alpakka/sqs/javadsl/SqsFlow.scala
@@ -11,6 +11,9 @@ import akka.stream.javadsl.Flow
 import akka.stream.scaladsl.{Flow => SFlow}
 import com.amazonaws.services.sqs.AmazonSQSAsync
 import java.lang.{Iterable => JIterable}
+
+import com.amazonaws.services.sqs.model.SendMessageRequest
+
 import scala.collection.JavaConverters._
 
 object SqsFlow {
@@ -18,7 +21,9 @@ object SqsFlow {
   /**
    * Java API: creates a flow based on [[SqsFlowStage]] for a SQS queue using an [[AmazonSQSAsync]]
    */
-  def create(queueUrl: String, settings: SqsSinkSettings, sqsClient: AmazonSQSAsync): Flow[String, Result, NotUsed] =
+  def create(queueUrl: String,
+             settings: SqsSinkSettings,
+             sqsClient: AmazonSQSAsync): Flow[SendMessageRequest, Result, NotUsed] =
     scaladsl.SqsFlow.apply(queueUrl, settings)(sqsClient).asJava
 
   /**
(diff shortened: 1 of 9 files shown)


sqs/src/main/scala/akka/stream/alpakka/sqs/javadsl/SqsFlow.scala
```diff
@@ -11,49 +11,55 @@ import akka.stream.javadsl.Flow
...
 
 object SqsFlow {
 
   /**
    * Java API: creates a flow based on [[SqsFlowStage]] for a SQS queue using an [[AmazonSQSAsync]]
    */
-  def create(queueUrl: String, settings: SqsSinkSettings, sqsClient: AmazonSQSAsync): Flow[String, Result, NotUsed] =
+  def create(queueUrl: String,
+             settings: SqsSinkSettings,
+             sqsClient: AmazonSQSAsync): Flow[SendMessageRequest, Result, NotUsed] =
```
johanandren: Can we keep one signature that accepts strings perhaps, and just map into this one?

randomcoder: The String based flow would require additional data to add the attributes so it would need to add those to the flow somehow. This would be possible by adding a second flow stage that maps the request to request with provided attributes, but the attribute set would need to be added to the signature, adding another 6 methods to the java api to support attributes with and without settings.

randomcoder: I could create a `simple(queueUrl: String, settings: SqsSinkSettings, sqsClient: AmazonSQSAsync): Flow[String, Result, Future[Done]]` maybe?

johanandren: Why couldn't it just do `.map(new SendMessageRequest(queueUrl, _)).via(...)`, isn't that exactly what the original implementation did?

randomcoder: The original created and sent the `SendMessageRequest` inside the `SqsFlowStage` which was why I couldn't just add a new part to do the conversion.

It would be simple to add for the `ScalaDsl` but as I mentioned the `JavaDsl` gets very unwieldy if we want to support the various parameter combinations.

johanandren: Ah, ok, fair enough skip it then, it is easy enough for users to create themselves even if it had been nice not to break existing code

randomcoder: I agree, I couldn't find a sane way to not break old code without totally reimplementing the internals of the flow, so thought this was the more suitable and future flexible solution.

ennru:
Looks good so far. Can you review the documentation to reflect your changes and additions, please.

gabfssilva:
Waiting for this feature.

randomcoder:
The main documentation is generated from the test cases which have been changed for the flow/sink cases in order to work. The Ack cases do not change as they are `Sources`. I'll rebuild and check again though to be sure.

randomcoder:
I've updated the docs to give both `String` and `SendMessageRequest` examples for the Sinks for clarity of both parts of functionality.

ennru:
Thank you for your contribution. Keep going!

Issue #468, opened 2017-09-15
S3: Support for v1 list-bucket


smarkey:
The S3 v1 list-bucket is not currently supported by Alpakka. I would like to introduce it so that I can use [S3Proxy](https://github.com/andrewgaul/s3proxy)

johanandren:
Duplicate of #192 ?

smarkey:
I don't believe so, this is for v1 support specifically rather than the v2 support already implemented.

johanandren:
Ah, thanks for clarifying

francisdb:
you might rather want to push for S3Proxy to support v2 listings?
https://github.com/andrewgaul/s3proxy/issues/168

PR #729, opened 2018-01-19
Add support for s3 list buckets v1

randomcoder:
This PR adds the option to `listBucket` using the legacy version 1 api or the current version 2 api. The main driver for this is to support legacy requirements (e.g. third party clients that use version 1, or testing using mock S3 implementations that do not yet support the version 2 api).

The default behaviour is unchanged, using version 2. The Scala implementation makes use of default parameter values to trigger the different versions, whereas the Java implementation adds a new `listBucketV1` method which keep the v1 usage explicit and does not break any existing code that uses the version 2 api.

Looking through the response processing code the `ListBucketResult` only uses response elements that are the same in both version 1 and 2 so I have made no changes there.

Fixes akka/alpakka#468

### DIFF
--- a/s3/src/main/resources/reference.conf
+++ b/s3/src/main/resources/reference.conf
@@ -76,4 +76,8 @@ akka.stream.alpakka.s3 {
 
   # Custom endpoint url, used for alternate s3 implementations
   # endpoint-url = null
+
+  # Which version of the list bucket api to use. Set to 1 to use the old style version 1 API.
+  # By default the newer version 2 api is used.
+  list-bucket-api-version = 2
 }
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
@@ -21,11 +21,10 @@ private[alpakka] object HttpRequests {
   def listBucket(
       bucket: String,
       prefix: Option[String] = None,
-      continuationToken: Option[String] = None,
-      apiVersion: ApiVersion = ListBucketVersion2
+      continuationToken: Option[String] = None
   )(implicit conf: S3Settings): HttpRequest = {
 
-    val (listType, continuationTokenName) = apiVersion match {
+    val (listType, continuationTokenName) = conf.listBucketApiVersion match {
       case ListBucketVersion1 => (None, "marker")
       case ListBucketVersion2 => (Some("2"), "continuation-token")
     }
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
@@ -21,14 +21,16 @@ private[alpakka] object HttpRequests {
   def listBucket(
       bucket: String,
       prefix: Option[String] = None,
-      continuationToken: Option[String] = None
+      continuationToken: Option[String] = None,
+      useApiVersion2: Boolean = true
   )(implicit conf: S3Settings): HttpRequest = {
 
+    val (listType, continuationTokenName) = if (useApiVersion2) (Some("2"), "continuation-token") else (None, "marker")
     val query = Query(
       Seq(
-        "list-type" -> Some("2"),
+        "list-type" -> listType,
         "prefix" -> prefix,
-        "continuation-token" -> continuationToken
+        continuationTokenName -> continuationToken
       ).collect { case (k, Some(v)) => k -> v }.toMap
     )
 
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
@@ -22,10 +22,14 @@ private[alpakka] object HttpRequests {
       bucket: String,
       prefix: Option[String] = None,
       continuationToken: Option[String] = None,
-      useApiVersion2: Boolean = true
+      apiVersion: ApiVersion = ListBucketVersion2
   )(implicit conf: S3Settings): HttpRequest = {
 
-    val (listType, continuationTokenName) = if (useApiVersion2) (Some("2"), "continuation-token") else (None, "marker")
+    val (listType, continuationTokenName) = apiVersion match {
+      case ListBucketVersion1 => (None, "marker")
+      case ListBucketVersion2 => (Some("2"), "continuation-token")
+    }
+
     val query = Query(
       Seq(
         "list-type" -> listType,
--- a/s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3Client.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3Client.scala
@@ -486,15 +486,36 @@ final class S3Client(s3Settings: S3Settings, system: ActorSystem, mat: Materiali
   }
 
   /**
-   * Will return a source of object metadata for a given bucket with optional prefix.
+   * Will return a source of object metadata for a given bucket with optional prefix using version 2 of the LIst Bucket API.
    * This will automatically page through all keys with the given parameters.
+   *
+   * @see https://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html
+   *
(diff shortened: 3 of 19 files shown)


s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3WireMockBase.scala
```diff
@@ -165,6 +165,31 @@ abstract class S3WireMockBase(_system: ActorSystem, _wireMockServer: WireMockSer
...
+                            <MaxKeys>1000</MaxKeys>
+                            <IsTruncated>false</IsTruncated>
+                            <Contents>
+                                <Key>$listKey</Key>
+                                <LastModified>2009-10-12T17:50:30.000Z</LastModified>
+                                <ETag>&quot;fba9dede5f27731c9771645a39863328&quot;</ETag>
+                                <Size>434234</Size>
+                                <StorageClass>STANDARD</StorageClass>
+                            </Contents>
+                        </ListBucketResult>""".stripMargin)
```
ennru: There is no margin to be stripped!?

s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
```diff
@@ -21,14 +21,16 @@ private[alpakka] object HttpRequests {
   def listBucket(
       bucket: String,
       prefix: Option[String] = None,
-      continuationToken: Option[String] = None
+      continuationToken: Option[String] = None,
+      useApiVersion2: Boolean = true
```
ennru: What do we do when v3 comes along? An argument `apiVersion` is preferrable.

randomcoder:
I'm happy to switch to a version variable, that was my original
implementation. Would a simple Int suffice or should I add a typed
class/object (my preference)?



On 19 Jan 2018 21:36, "Enno" <notifications@github.com> wrote:

*@ennru* commented on this pull request.

I'm not convinced we should add support for a legacy API. It will make it
very much harder to support newer features.
------------------------------

In s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3WireMockBase.scala
<https://github.com/akka/alpakka/pull/729#discussion_r162734647>:

> +            .withHeader("Content-Type", "application/xml")
+            .withBody(s"""<?xml version="1.0" encoding="UTF-8"?>
+                        <ListBucketResult
xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
+                            <Name>bucket</Name>
+                            <Prefix>$listPrefix</Prefix>
+                            <Marker/>
+                            <MaxKeys>1000</MaxKeys>
+                            <IsTruncated>false</IsTruncated>
+                            <Contents>
+                                <Key>$listKey</Key>
+
<LastModified>2009-10-12T17:50:30.000Z</LastModified>
+
<ETag>&quot;fba9dede5f27731c9771645a39863328&quot;</ETag>
+                                <Size>434234</Size>
+                                <StorageClass>STANDARD</StorageClass>
+                            </Contents>
+                        </ListBucketResult>""".stripMargin)

There is no margin to be stripped!?
------------------------------

In s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
<https://github.com/akka/alpakka/pull/729#discussion_r162735629>:

> @@ -21,14 +21,16 @@ private[alpakka] object HttpRequests {
   def listBucket(
       bucket: String,
       prefix: Option[String] = None,
-      continuationToken: Option[String] = None
+      continuationToken: Option[String] = None,
+      useApiVersion2: Boolean = true

What do we do when v3 comes along? An argument apiVersion is preferrable.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/akka/alpakka/pull/729#pullrequestreview-90248759>, or mute
the thread

s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3WireMockBase.scala
randomcoder: True enough, I didn't spot that when I copied the original. I'll add one in.

randomcoder:
From our (personal) perspective yes, the mock we use doesn't yet support V2. However as a wider capability, V1 has not yet been officially deprecated, simply revised so will remain active for some time. From the changes made, supporting V1 is not a significant overhead at this time.

I'm happy to have some other way to provide support, but to me the commonality outweighs the differences at this time.

johanandren:
I think this should go under a config flag rather, if you need to use v1 you always need to use v1, not mix the two, right?

randomcoder:
So you have the two implementations, but force the selection based on config, defaulting to V2? Happy to make those changes.

I had done something like this in early work however, as this is only for the list bucket request wouldn't it potentially open up requiring config for versions of different API calls as well. However
a _quick_ look over the API docs suggest that this is currently the only case where there are two active versions so it may not be much of an issue.

s3/src/main/resources/reference.conf
```diff
@@ -76,4 +76,8 @@ akka.stream.alpakka.s3 {
 
   # Custom endpoint url, used for alternate s3 implementations
   # endpoint-url = null
+
+  # Which version of the list bucket api to use. Set to 1 to use the old style version 1 API.
+  # By default the newer version 2 api is used.
+  list-bucket-api-version = 2
```
johanandren: Maybe just call it `api-version` but document that it is only effective for `list-bucket` (if additional methods provide multi-versions in the future they'll have to amend the comment)

s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3Client.scala
```diff
@@ -486,8 +486,11 @@ final class S3Client(s3Settings: S3Settings, system: ActorSystem, mat: Materiali
   }
 
   /**
-   * Will return a source of object metadata for a given bucket with optional prefix.
+   * Will return a source of object metadata for a given bucket with optional prefix using version 2 of the LIst Bucket API.
```
johanandren: weird uppercase I

Isn't this the public api where you should note that the behavior depends on the setting?

francisdb:
Looks ok to me, have you run the integration tests with v1/v2?

s3/src/main/resources/reference.conf
randomcoder: Isn't it clearer, and contained in a single place to leave it as is?

s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3Client.scala
randomcoder: Removed typo and updated both Javadoc and Scaladoc

s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3IntegrationSpec.scala
```diff
@@ -263,7 +273,7 @@ trait S3IntegrationSpec extends FlatSpecLike with BeforeAndAfterAll with Matcher
  * (tests that do listing counts might need some tweaking)
  *
  */
-@Ignore
+//@Ignore
```
ennru: You left the Integration tests un-ignored, so the builds fail.

s3/src/main/resources/reference.conf
johanandren: My thinking is that it isn't unlikely that the same scenario happens in the future for next version of AWS apis, or for other operations on S3, and then I'm guessing that you would want to apply the version setting to all such operations, not on a method-per-method basis, no?

randomcoder: How would you then support (for example) v2  of list bucket and v3 of upload with a single setting?

randomcoder:
As I don't have an AWS account yet,, I have successfully run the integration tests against the Minio server.

s3/src/main/resources/reference.conf
johanandren: What is the scenario where you would want to have such fine grained control? I'm thinking that if this turns out to be useful for more operations you'd want all of them to use the other version, no?

This isn't that important though so if you have strong feelings for the current setting name, just keep it.

johanandren:
Would be good if someone could run it against actual AWS before we merge this.

ennru:
Thank you for your contribution!

PR #730, opened 2018-01-19
Elasticsearch: enable optimistic locking/concurrency controll (and more)

mbknor:
This PR contains multiple improvements to the elasticsearch-connector:

- Added support for **optimistic locking/concurrency control** (https://www.elastic.co/guide/en/elasticsearch/guide/current/optimistic-concurrency-control.html)
- The flow now **guaranties** that message-results are returned in the **same order** as input.
- Changed the **default** value of the setting **retryPartialFailure to false**. (If retryPartialFailure is true, failed messages will retry, but succeeded messages will be emitted downstream. This breaks the order of messages. Also: I cannot see a situation where single failed documents in a bulk update  can start succeeding when retrying)
- source now supports **custom ObjectMapper**
- **Improved Jackson json code** in java source.

(Please see individual commits)


mbknor:
@ennru It would be great if you could have a look at these changes

mbknor:
Thanks for good feedback, @ennru. I've modified the code accordingly.

PR #732, opened 2018-01-19
Cryptography

HarrisonBaxter:
Providing support for encryption and decryption without any dependencies by making use of underlying Java cryptographic libraries. Provide are four different flows to encrypt and decrypt bytestring to bytestring Using symmetric and asymmetric algorithms.


lightbend-cla-validator:
Hi @HarrisonBaxter,

HarrisonBaxter:
I have signed the CLA

ennru:
I don't understand who is the author of this PR.
Isn't this the work @fmsbeekmans started off in #702 ?

fmsbeekmans:
It is, we worked together on this.

ennru:
Ok, so you may close #702 than.
Put the documentation in the Data Transformations section, please.
I'll have a more thorough look another day.

ennru:
I realised this simple approach won't work.
If you e.g. stream the encrypted ByteStrings to a file and than read it again, you won't get the same bytes per ByteString and won't be able to decrypt. There needs to be some frameing so that the correct part of the byte stream can be used for decryption.
This is related to https://github.com/akka/akka/issues/24361

fmsbeekmans:
Will try and make a failing test case for this over the weekend. We are testing arbitrarily chunked strings, not bytes.

ennru:
To test this, you need to encrypt a couple of ByteStrings, concat those, split them differently and decrypt.

fmsbeekmans:
Tested with bytestrings chunked per byte, still works. The encryption and decryption works on `Array[Byte]`, not on `String`.

cryptography/src/test/scala/akka/stream/alpakka/cryptography/scaladsl/CryptographicFlowSpec.scala
```diff
@@ -41,11 +41,27 @@ class CryptographicFlowSpec extends WordSpec with Matchers with ScalaFutures wit
...
+        val keyGenerator = KeyGenerator.getInstance("AES")
+        val secretKey = keyGenerator.generateKey()
+
+        val byteString = ByteString("byte by byte")
+
+        val res = Source.single(byteString)
+          .mapConcat(bs => bs.toArray.toList)
+          .map(b => ByteString(b))
+          .via(symmetricEncryption(secretKey))
+          .via(symmetricDecryption(secretKey))
```
ennru: Add a concat between the encryption and decryption and you'll see what I mean. If we - for example - send the encrypted data via a socket or http, we'll get other ByteStrings on the other side. We'd need some frameing to cut out the same ByteStrings for decryption.

fmsbeekmans: Oooh you're right, will have a look before the weekend is over!

fmsbeekmans:
Added a chunking stage to the symmetric encryption and decryption. Still have to figure out the chunk size based on the key.

Realized that for the asymmetric encryption it should encrypt or decrypt the complete data at once. Not sure whether it makes sense for a flow or not.

fmsbeekmans:
Should be able to make some time to look into this again soon.

git-josip:
@fmsbeekmans  What is the status of this PR? Do we know is it going to be merged anytime soon , so client side encryption for Amazon S3 can be used.

fmsbeekmans:
Still quite a bit of work to be done sadly.

git-josip:
@fmsbeekmans Thanks, appreciate your work!

ennru:

PR #737, opened 2018-01-23
Cancel SSE EventSource when downstream cancels (closes #598)

### LINKED ISSUE #598: SSE clients seem not to disconnect when downstream terminates
The feedback loop for the last-event-ID in the SSE client (v 0.14) keeps the upstream connection alive indefinitely.

Would it be possible to add `eagerCancel=true` to the broadcast, or to make it user-configurable?

hseeberger:


ennru:
Thank you for this efficient turn-around.

ionos:
Thank you!

PR #727, opened 2018-01-18
Added PubSubMessage attributes support for Google Cloud PubSub.

vladimir-malyk:
Issue #510

PR #738, opened 2018-01-23
add PubSubMessage publishTime support

mertkavi:
add publishTime optional support

### DIFF
--- a/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/Model.scala
+++ b/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/Model.scala
@@ -7,7 +7,7 @@ package akka.stream.alpakka.googlecloud.pubsub
 import scala.collection.immutable
 import scala.collection.JavaConverters._
 
-final case class PubSubMessage(data: String, messageId: String)
+final case class PubSubMessage(data: String, messageId: String, publishTime: Option[String])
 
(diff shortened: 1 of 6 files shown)


ennru:
This PR complements #727

mertkavi:
@ennru No, this is not complements: https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage

ennru:
It does: #727 adds `attributes` this one adds `publishTime`.

mertkavi:
what can we do @ennru ? this feature is must.

ennru:
No worries, we'll sort this out before the next release.
Regarding `publishTime`: wouldn't it be approriate to use a proper `ZonedDateTime` and have a formatter for the expected format?

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/Model.scala
```diff
@@ -7,7 +7,7 @@ package akka.stream.alpakka.googlecloud.pubsub
 import scala.collection.immutable
 import scala.collection.JavaConverters._
 
-final case class PubSubMessage(data: String, messageId: String)
+final case class PubSubMessage(data: String, messageId: String, publishTime: Option[String])
```
huntc: [Instant](https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html] instead of a string me thinks...

PR #739, opened 2018-01-23
Apache Solr connector

burakkose:
The API is very similar to Elasticsearch implementation.


ennru:
Looks good so far.
You'll nedd to switch off debug logging for `o.a.s.m.r.jmx.JmxMetricsReporter` in `logback-test.xml` to cut down Travis' logg size.

burakkose:
@ennru, thank you for reviewing. I have fixed the issue with the test zookeeper server and the log size. Moreover, I have rewritten the Git history.

solr/src/main/scala/akka/stream/alpakka/solr/javadsl/SolrSinkSettings.scala
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.solr.javadsl
+
+import akka.stream.alpakka.solr._
+import scaladsl.{SolrSinkSettings => ScalaSolrSinkSettings}
+
+final class SolrSinkSettings(val bufferSize: Int, val retryInterval: Int, val maxRetry: Int, val commitWithin: Int) {
```
ennru: 1. Other modules than Elasticsearch keep their settings classes in the main module package instead of having a Java and Scala variant. I think that's a better solution.
1. For the Sink settings, you might want to consider `SolrUpdateSettings` as they are used for both Flow and Sink.
1. `retryInterval` should be a `FiniteDuration` instead

solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
```diff
@@ -0,0 +1,184 @@
...
+  def create[T, C](source: T, passThrough: C): IncomingMessage[T, C] =
+    IncomingMessage(source, passThrough)
+
+}
+
+final case class IncomingMessage[T, C](source: T, passThrough: C)
+
+final case class IncomingMessageResult[T, C](source: T, passThrough: C, status: Int)
+
+class SolrFlowStage[T, C](
```
ennru: Please tighten the visibility with `private[solr]` to keep the internal APIs from being used.

docs/src/main/paradox/solr.md
```diff
@@ -0,0 +1,169 @@
...
+@@dependency [sbt,Maven,Gradle] {
+  group=com.lightbend.akka
+  artifact=akka-stream-alpakka-solr_$scalaBinaryVersion$
+  version=$version$
+}
+
+## Set up client
+
+Sources, Flows and Sinks provided by this connector need a prepared `org.apache.solr.client.solrj.SolrClient` to
+access to Solr.
```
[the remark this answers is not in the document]
burakkose: Thank you for your reviewing, I don't think this is a strict requirement. Especially, `client.close()` based on exceptions and it can be used out of the actor system in the application.

ennru:
Just noticed: You need to add Solr to `connectors.md`, as well.

burakkose:
Yes, my bad. It is done.

Issue #740, opened 2018-01-24
Object metadata available before stream materializtion in s3 download


kag0:
So I messed up.
in https://github.com/akka/alpakka/issues/497 I suggested a return type like this `Source[ByteString, Future[ObjectMetadata]]` for s3 downloads. 
I'm coming to realize that that was not well thought out, and that a return type more like `Future[(Source[ByteString, ObjectMetadata], ObjectMetadata)]` or `(Source[ByteString, Future[ObjectMetadata]], Future[ObjectMetadata])` would be better.
The reasoning for this is that it's desirable to have access to the metadata before the stream materializes. 
One simple use case would be an akka http route that simply downloads a file from s3 and streams it back to the client. We would most likely want to use the metadata in the headers of our response, but won't have access to the metadata since the stream materialization is done by akka http.

I'm happy to take a shot at this if you're open to it.

ennru:
@francisdb Any comment?

francisdb:
@kag0 makes sense, wondering if the `Future[(Source[ByteString, xxx], ObjectMetadata)]` could be something else that brings value or maybe just `Done`

ktoso:
Yeah it makes sense. 
I wonder if for convenience the current API can be kept in case people don't need the metadata for some reason? Or do you think it should be dropped?

kag0:
I think a `(Source[ByteString, xxx], Future[ObjectMetadata])` return could replace the current api with little inconvenience to users since their existing `val source = download(...)` could be easily replaced with `val (source, _) = download(...)`

francisdb:
Should indeed not be to hard to switch: `Source.formFutureSource(download(...).map(_._1))`

francisdb:
I see you went for the other option of returning a tuple.
Any the akka/alpakka api design best practices regarding these return types?
 * `(Source[ByteString, NotUsed], Future[ObjectMetadata])`
 * `Future[(Source[ByteString, NotUsed], ObjectMetadata)]`

kag0:
Yes, I found the convenience of being able to use `val (source, _) = download(...)` tipped me in that direction.
But indeed if anyone knows if there are any best practices or conventions here then I'll switch it to whatever that is.

ennru:
Fixed by #742

PR #742, opened 2018-01-24
Make s3 object metadata available on downloads

kag0:
independent from stream materialization ( #740)

* download method in s3 client now returns a tuple with stream and metadata
* stream from download now materializes into NotUsed

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
@@ -72,13 +72,12 @@ private[alpakka] final class S3Stream(settings: S3Settings)(implicit system: Act
     import mat.executionContext
     val s3Headers = S3Headers(sse.fold[Seq[HttpHeader]](Seq.empty) { _.headersFor(GetObject) })
     val future = request(s3Location, rangeOption = range, s3Headers = s3Headers)
-    (
-      Source
-        .fromFuture(future.flatMap(entityForSuccess))
-        .map(_.dataBytes)
-        .flatMapConcat(identity),
-      future.map(resp ⇒ ObjectMetadata(resp.headers))
-    )
+    val source = Source
+      .fromFuture(future.flatMap(entityForSuccess))
+      .map(_.dataBytes)
+      .flatMapConcat(identity)
+    val meta = future.map(resp ⇒ ObjectMetadata(resp.headers))
+    (source, meta)
   }
 
   def listBucket(bucket: String, prefix: Option[String] = None): Source[ListBucketResultContents, NotUsed] = {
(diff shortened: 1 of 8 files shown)


s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -68,19 +68,17 @@ private[alpakka] final class S3Stream(settings: S3Settings)(implicit system: Act
...
-          ObjectMetadata(resp.headers)
-        }
-      }
+    (
+      Source
+        .fromFuture(future.flatMap(entityForSuccess))
+        .map(_.dataBytes)
+        .flatMapConcat(identity),
+      future.map(resp ⇒ ObjectMetadata(resp.headers))
+    )
```
ennru: Took me a moment to see the tuple, wouldn't it be good to use intermediate vals?

s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3Client.scala
```diff
@@ -421,49 +423,49 @@ final class S3Client(s3Settings: S3Settings, system: ActorSystem, mat: Materiali
...
+    JPair(stream.asJava, meta.map(metaDataToJava)(mat.executionContext).toJava)
+  }
+
   /**
    * Downloads a S3 Object
    *
    * @param bucket the s3 bucket name
    * @param key the s3 object key
-   * @return A [[Source]] of [[ByteString]] that materializes into a [[CompletionStage]] containing the [[ObjectMetadata]]
+   * @return A [[Pair]] with a [[Source]] of [[ByteString]], and a [[CompletionStage]] containing the [[ObjectMetadata]]
```
ennru: Links to other classes almost always need to be fully qualified in Scaladoc. (But it's wront in many places in Alpakka.)

kag0: added fully qualified `Pair`, let me know if you want to update `Source`, `CompletionStage`, etc as well.

ennru: Yes, please.

kag0:
@johanandren here's one with akka http as it might look if one were streaming a file from s3 back to a client
```scala
val (data, eventualMeta) = s3Client.download(???, ???)
complete( eventualMeta.map ( meta ⇒
    HttpResponse(
      entity = HttpEntity(
        meta.contentType.flatMap(ContentType.parse(_).toOption).getOrElse(`application/octet-stream`),
        meta.contentLength,
        data
      )
    )
) )
```

ennru:
It would be good to see this kind of example in the documentation.

kag0:
I'm not familiar with paradox, so you'll have to let me know if I did that right.

ennru:
Thank you for your contribution! The S3 connector seems to be a popular way to get started using Alpakka.

Issue #726, opened 2018-01-18
Make Slick Flow variable in return type


diversit:
Currently the Slick.flow signature fixes the return type to Int.
`Slick.flow(..): Flow[T, Int, NotUsed]`

However sometimes it might be useful to be able to return a different type.
Case: when consuming from a Kafka topic and you want to commit the message once the message has been persisted.
In the current situation this is not possible since you lose the CommittableMessage or Committable after using the Slick Flow.
It might be possible to split the flow and rejoin again after persisting via Slick, but this greatly complicates a very simple flow.

Proposed solution:
Make the Slick.flow return type dynamic. This would allow to do something like:
```
Slick.flow { committableMessage =>
  val myRow = msgToRow(committableMessage)
  (Tables.MyTable += row).map(insertCount => (committableMessage, insertCount))
}
```
The committable message is now retained and can be committed in a next flow step.

Alternative:
Add a additional 'flow' function which does not fix the return type. But what would then be a good name for this function?

If this change is/gets approved, I will make a pull request.

ennru:
Good idea.
We see this need in other connectors, as well.
The Elasticsearch flow got support in the latest release and mimics the Reactive Kafka naming of a `passThrough`.

diversit:
Created a proposal for this feature in pull request #743

ennru:
Fixed by #743

PR #743, opened 2018-01-25
Make Slick Flow variable in return type

diversit:
Proposal for additional 'Slick.flow' function to be flexible in the return type as suggested in [Make Slick Flow variable in return type](https://github.com/akka/alpakka/issues/726) issue.

Based on @ennru's suggestion called the new functions 'flowWithPassThrough', similar to ElasticsearchFlow.
The Scala implementation was quite forward. 
The Java implementation was a bit tricky. Added an additional 'mapper' argument to the method signature to map the DBIO result into something else.

Added test cases including an example with Kafka (also copied from Elasticsearch test case).
Updated documentation including 'with kafka' example.

Let me know what you think of this solution.

### DIFF
--- a/slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
+++ b/slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
@@ -4,7 +4,7 @@
 
 package akka.stream.alpakka.slick.javadsl
 
-import java.util.concurrent.{CompletionStage, ExecutorService}
+import java.util.concurrent.CompletionStage
 import java.util.function.{BiFunction => JBiFunction, Function => JFunction}
 
 import scala.compat.java8.FunctionConverters._
@@ -96,6 +96,8 @@ object Slick {
    *           and allows to combine the statement result and element into a result type R.
    *
    * @param session The database session to use.
+   * @param executionContext ExecutionContext used to run mapper function in.
+   *                         E.g. the dispatcher of the ActorSystem.
    * @param toStatement A function that creeates the SQL statement to
    *                    execute from the current element. Any DML or
    *                    DDL statement is acceptable.
@@ -104,11 +106,11 @@ object Slick {
    */
   def flowWithPassThrough[T, R](
       session: SlickSession,
-      executorService: ExecutorService,
+      executionContext: ExecutionContext,
       toStatement: JFunction[T, String],
       mapper: JBiFunction[T, java.lang.Integer, R]
   ): Flow[T, R, NotUsed] =
-    flowWithPassThrough(session, executorService, 1, toStatement, mapper)
+    flowWithPassThrough(session, executionContext, 1, toStatement, mapper)
 
   /**
    * Java API: creates a Flow that takes a stream of elements of
@@ -118,10 +120,12 @@ object Slick {
    *           and allows to combine the statement result and element into a result type R.
    *
    * @param session The database session to use.
+   * @param executionContext ExecutionContext used to run mapper function in.
+   *                         E.g. the dispatcher of the ActorSystem.
    * @param parallelism How many parallel asynchronous streams should be
    *                    used to send statements to the database. Use a
    *                    value of 1 for sequential execution.
-   * @param toStatement A function that creeates the SQL statement to
+   * @param toStatement A function that creates the SQL statement to
    *                    execute from the current element. Any DML or
    *                    DDL statement is acceptable.
    * @param mapper A function to create a result from the incoming element T
@@ -129,7 +133,7 @@ object Slick {
    */
   def flowWithPassThrough[T, R](
       session: SlickSession,
-      executorService: ExecutorService,
+      executionContext: ExecutionContext,
       parallelism: Int,
       toStatement: JFunction[T, String],
       mapper: JBiFunction[T, java.lang.Integer, R]
@@ -138,7 +142,7 @@ object Slick {
       .flowWithPassThrough[T, R](parallelism, (t: T) => {
         toDBIO(toStatement)
           .apply(t)
-          .map(count => mapper.apply(t, count))(ExecutionContext.fromExecutorService(executorService))
+          .map(count => mapper.apply(t, count))(executionContext)
       })(session)
       .asJava
 
--- a/slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
+++ b/slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
@@ -4,22 +4,21 @@
 
 package akka.stream.alpakka.slick.javadsl
 
-import java.util.concurrent.CompletionStage
-import java.util.function.{Function => JFunction, BiFunction => JBiFunction}
+import java.util.concurrent.{CompletionStage, ExecutorService}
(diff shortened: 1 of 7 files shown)


ennru:
Can you let Scalaformat do its thing by running `test:compile` and than push the reformatted code, please.

diversit:
I have committed the reformatted code.
Tests now successful.

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
```diff
@@ -89,6 +89,57 @@ object Slick {
...
+   */
+  def flowWithPassThrough[T, R](
+      session: SlickSession,
+      parallelism: Int,
+      toStatement: JFunction[T, String],
+      mapper: JBiFunction[T, java.lang.Integer, R]
+  ): Flow[T, R, NotUsed] =
+    ScalaSlick
+      .flowWithPassThrough[T, R](parallelism, (t: T) => {
+        import concurrent.ExecutionContext.Implicits.global
```
johanandren: Don't use the global execution context, have the user pass one as a parameter instead (use implicit param for the Scala side)

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
```diff
@@ -105,10 +104,11 @@ object Slick {
    */
   def flowWithPassThrough[T, R](
       session: SlickSession,
+      executorService: ExecutorService,
```
johanandren: I think it is fine to use the Scala API here, if Java users wants to provide an ´ExecutorService` they can use `akka.dispatch.ExecutionContexts.fromExecutorService`. Most common use case will be to use the one the dispatcher.

slick/src/test/java/akka/stream/alpakka/slick/javadsl/DocSnippetFlowWithPassThrough.java
```diff
@@ -72,6 +75,7 @@ public static void main(String[] args) throws Exception {
         .via(
           Slick.flowWithPassThrough(
             session,
+            executorService,
```
johanandren: Just use the dispatcher one instead here. No reason to introduce an additional, manually lifecycle, thread pool here when the stream is tied to the actor system lifecycle

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
diversit: Ok. I'm not familiar with using Akka with Java.
What about replacing it with `dispatcher: ExecutionContextExecutor`? Then the `ActorSystem.dispatcher` can be used. Although the type is `ExecutionContextExecutor`, I thought it might be better to name it `dispatcher` just like in `ActorSystem`.

johanandren: Oh I see I was maybe a tad unclear there, accept an `ExecutionContext` is fine is what I meant. Usually we just call such params `executionContext` (or `ec` when it's an implicit) even if it will most likely be the dispatcher.

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
```diff
@@ -164,4 +221,8 @@ object Slick {
   private def toDBIO[T](javaDml: JFunction[T, String]): T => DBIO[Int] = { t =>
     SQLActionBuilder(javaDml.asScala(t), SetParameter.SetUnit).asUpdate
   }
+
+  class JDBIO[T] {
+    def map[B](f: T => B): JDBIO[B] = ???
+  }
```
johanandren: What is this, and why does it throw, Isn't DBIO a type in slick?

slick/src/main/scala/akka/stream/alpakka/slick/scaladsl/Slick.scala
```diff
@@ -76,6 +76,44 @@ object Slick {
...
+   * @param session The database session to use.
+   */
+  def flowWithPassThrough[T, R](
+      parallelism: Int,
+      toStatement: T => DBIO[R]
+  )(implicit session: SlickSession): Flow[T, R, NotUsed] =
+    Flow[T]
+      .mapAsync(parallelism) { t =>
+        session.db.run(toStatement(t))
+      }
```
johanandren: Isn't this identical to `flow` now, so this is not needed. 
Alternatively we provide a signature just like the Java one which takes a mapper.

Also makes me wonder, do we really need the Java one, can't that also be done by closing over state in the `toStatement` function?

slick/src/test/scala/akka/stream/alpakka/slick/scaladsl/DocSnippets.scala
```diff
@@ -184,3 +184,72 @@ object SlickFlowExample extends App {
...
+      }
+      .runWith(Sink.ignore)
+
+  done.onComplete {
+    case _ =>
+      session.close()
+      system.terminate()
+  }
+}
+//#flowWithPassThrough-example
```
johanandren: I'd rather see this in a separate file as well just like the Java sample.

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
diversit: This was an attempt to create a Java version of Slick's `DBIO` so the additional `mapper` would not be necessary in the Java API, but gave up on this solution after trying it out for a while. 
Removed this code.

slick/src/main/scala/akka/stream/alpakka/slick/scaladsl/Slick.scala
diversit: `flowWithPassThrough` is indeed the same implementation as `flow`, but yes it is needed since `flow` is fixed on the return type and `flowWithPassThrough` isn't. But since implementation is the same, changed `flow` implementation to call `flowWithPassThrough` function.

The Java method is also needed, IMHO, since the 'problem' with the Java implementation is that the `toStatement` function is a function from the input `T` into a Slick `String` to not expose the Slick Scala `DBIO` type in the Java API I suppose. So the `toStatement` cannot be a function from `T` to `R` and the additional `mapper` function is needed there.

slick/src/test/scala/akka/stream/alpakka/slick/scaladsl/DocSnippets.scala
diversit: Looking at the Alpakka sources, it is quite common to have multiple Scala examples in the same file. So since this is according to the current 'standard' I'd like to keep it here. Having them in the same file is convenient IMHO since it allows you easily to compare examples.

Fine by me if you'd like to change this in the future, but I do not want to shave that yak now.

johanandren: Sure, not super important.

slick/src/main/scala/akka/stream/alpakka/slick/scaladsl/Slick.scala
johanandren: Ah, I see, thanks for the explanation.

PR #755, opened 2018-01-30
AMQP: use host & port from underlying connection factory

agboom:
Due to conflicts in https://github.com/akka/alpakka/pull/703 with https://github.com/akka/alpakka/pull/670 a new PR was made for re-usable connections in https://github.com/akka/alpakka/pull/706. This resulted in loss of some functionality in the underlying connection factory PR (https://github.com/akka/alpakka/pull/703)

This PR restores that functionality. The idea is that when a native ConnectionFactory is used, it may contain a host/port. This is ignored in the current implementation of `AmqpConnectionFactoryConnectionProvider`. As a result, setting the host/port in the underlying ConnectionFactory and using this constructor with an empty `hostAndPortList` leads to unexpected behavior (i.e. you set the host and port, but the RabbitMQ client throws an IO Exception saying "failed to connect").

To keep the interface consistent with other providers and retain support for multiple connections, the optional host and port list is used if it's not empty.

I also restored some Scaladoc that was lost in the process.

Another possible implementation would be to always construct the connection with a union of both the host/port in the ConnectionFactory and the host and port list in the Provider constructor.

### DIFF
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
@@ -156,13 +156,30 @@ object AmqpCredentials {
     AmqpCredentials(username, password)
 }
 
+/**
+ * Uses a native [[com.rabbitmq.client.ConnectionFactory]] to configure an AMQP connection factory.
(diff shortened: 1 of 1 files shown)


ennru:
Thank you for keeping after this.

PR #758, opened 2018-01-31
Elasticsearch improvements: Use jackson from scala + fix logging issue

mbknor:


ennru:
Thank you for this API improvement.

Issue #556, opened 2017-10-31
Add a GRPC connector


queimadus:
Add support for [grpc](https://grpc.io/docs/tutorials/basic/java.html).

The java library exposes a blocking client. I've been using and contributing to an akka stream non-blocking implementation [grpc-akka-stream-subrepo](https://github.com/eiennohito/grpc-akka-stream-subrepo) but I think it would benefit from more exposure and contributions.

There is [scalaPB](https://github.com/scalapb/ScalaPB) which creates all the boilerplate code from the `.protobuf` definitions but can be apart from this project.

I've been thinking for a while about migrating this to an alpakka connector. Do you guys think this would be a good fit?

raboof:
Hi @queimadus - thanks for this request! We have in fact been working on gRPC support in a separate project, https://github.com/akka/akka-grpc - right now we're also wrapping grpc-stub for the client, but the server is already implemented based on akka-http directly for a more consistent experience. It's still fairly early days, but I'd be very interested in your feedback!

ennru:
[Akka gRPC](https://doc.akka.io/docs/akka-grpc/current/) is around and solves this request.

PR #759, opened 2018-01-31
Introduce secondary API utilising gRPC instead of HTTP and more configuration options

### LINKED ISSUE #741: Google Pub/Sub gRPC API Support
Are there any plans to add support for the gRPC API exposed by Google in the pub/sub connector?

CremboC:
Currently, the Google Cloud PubSub source uses HTTP requests for pulling messages and acknowledging them.

Google provides a faster and more efficient API for doing the same via [gRPC](https://cloud.google.com/pubsub/docs/reference/rpc/). While it is still in beta, this still can be used if better performance is required _now_.

What I've done was basically copy the existing logic for the source, but modified it to use the gRPC API as well as made it slightly more configurable.

Most crucially I've added the **parallelism** option. In the current PubSub source, there can only be a single in-flight HTTP request. This means that if downstream is faster than the amount of time it takes for the request to finish, the source will always be the bottleneck.

The parallelism parameter allows the user to configure the number of in-flight requests, ensuring downstream is never waiting for a request to finish.

Furthermore, I've made the number of max messages per request configurable.

Finally, this API no longer requires storing the private key and API key of Google Cloud in plaintext and passing them as arguments. The gRPC API uses the built-in credentials that the Google Cloud uses, namely as defined by [GoogleCredentials.getApplicationDefault](https://github.com/google/google-auth-library-java/blob/a8929df490d3751f2349986264290a074dee35e1/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java#L99).

We're already using this code in production at [Permutive](https://github.com/permutive) and it seems to work better the current HTTP API.

Apparently fixes #741 #556


CremboC:
Some final notes: if the community thinks this is merge-worthy, I can add some example usage and hopefully tests, if possible!

CremboC:
One thing that I have noticed running this code over the past couple of days is that the "parallelism" option I have added doesn't _really_ work as I expect it to. The counter never increments past 1. This is probably due to my lack of understanding of how backpressure and how the downstream pulls the upstream.

The goal of parallelism factor was to avoid doing something like:
```
source.merge(source).via(processingFlow).runWith(Sink.ignore)
```

Instead, I wanted to have a single source that can output **n** times, i.e. the above code would be equivalent to:
```
source(parallelism = 2).via(processingFlow).runWith(Sink.ignore)
```

Unfortunately my solution doesn't seem to work that way.

I was hoping I could get some feedback why it doesn't work that way and how, if possible, I could fix it?

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSourceGrpc.scala
```diff
@@ -0,0 +1,102 @@
...
+    implicit as: ActorSystem
+) extends GraphStage[SourceShape[v1.ReceivedMessage]] {
+
+  val out: Outlet[v1.ReceivedMessage] = Outlet("GooglePubSubSourceGrpc.out")
+  override val shape: SourceShape[v1.ReceivedMessage] = SourceShape(out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) {
+      private var state: State = Pending
+      private val fetches: AtomicInteger = new AtomicInteger(0)
```
ennru: No need to make `fetches` an AtomicInteger, the graph stage code won't be executed by more than one thread at a time.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSourceGrpc.scala
```diff
@@ -0,0 +1,102 @@
...
+
+      private val callback = getAsyncCallback[(Try[v1.PullResponse], Materializer)] {
+        case (Failure(tr), _) =>
+          failStage(tr)
+        case (Success(response), mat) =>
+          response.getReceivedMessagesList.asScala.toList match {
+            case head :: tail =>
+              state match {
+                case HoldingMessages(oldHead :: oldTail) =>
+                  state = HoldingMessages(oldTail ++ Seq(head) ++ tail)
```
ennru: Is this buffer bounded in some way?

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GuavaConversions.scala
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.googlecloud.pubsub
+
+import com.google.common.util.concurrent.{FutureCallback, Futures, ListenableFuture}
+
+import scala.concurrent.{Future, Promise}
+
+object GuavaConversions {
```
ennru: This object should be `private`, as well.

project/Dependencies.scala
```diff
@@ -122,6 +122,7 @@ object Dependencies {
     libraryDependencies ++= Seq(
       "com.typesafe.akka" %% "akka-http" % AkkaHttpVersion,
       "com.typesafe.akka" %% "akka-http-spray-json" % AkkaHttpVersion,
+      "com.google.cloud" % "google-cloud-pubsub" % "0.33.0-beta",
```
ennru: What license does this have?

ennru:
Regarding your question on parallelism in `mapAsync`: The parallelism specifies how many elements can enter the map step, the function is evaluated for each element (in your case calling `publisherStub.publish(pr)`. Which counter did you expect to show something else?

project/Dependencies.scala
CremboC: https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/LICENSE

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSourceGrpc.scala
```diff
@@ -0,0 +1,102 @@
...
+              }
+            case Nil =>
+              state = Fetching
+              fetch(mat)
+          }
+      }
+
+      setHandler(
+        out,
+        new OutHandler {
```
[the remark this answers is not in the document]
CremboC: I basically copy pasted the structure of the non-GRPC variant of this! Of course I can change it as you say.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSourceGrpc.scala
CremboC: I guess only by the limitations of the `List` type (if there are any?). Not to excuse it, but once again, this is copied over from the non-GRPC variant.

ennru:
Look what the Akka team has been working on lately: https://github.com/akka/akka-grpc

CremboC:
I feel like using that could complicate the logic. Google's library handles the authentication and everything.

ennru:
Yes, I didn't want to suggest rewriting on Akka gRPC. 
We'll try to release 0.18 the next week. Would be great if we can ship this one, as well.

CremboC:
I've pushed some of the changes you asked as well as example usage. I messed up your nickname though in the commit message, sorry about that!

One thing I'd love to do, but is not necessary per se for this PR is to properly handle when GRPC closes the connection that we establish to it. I swear I remember somewhere in the docs that the connection can get terminated at any time, meaning the whole graph fails (as per https://github.com/CremboC/alpakka/blob/pub-sub-grpc/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSourceGrpc.scala#L43).

I think it would be better to add an option to retry with exponential backoff or fail, as is now.

CremboC:
I think the build failure is transient, seems like something to do with AMPQ (which this PR doesn't touch). Just need to restart.

ennru:
The build went through.
We'd still need documentation in written text in `google-cloud-pub-sub.md`.

ennru:
We've serious issues with the Travis builds. I've restarted lots of times today, but it doesn't get through.

ennru:
This PR build finally went through. I didn't dare to merge it to master before releasing 0.18 though. We'll have to improve the CI cycle to get reliable builds.
I expect the next release to happen in May.

CremboC:
No worries. I hope to test it better in production before that so I'll report if there's any issues and obviously push more commits if need be.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSourceGrpc.scala
```diff
@@ -0,0 +1,98 @@
...
+
+  val out: Outlet[v1.ReceivedMessage] = Outlet("GooglePubSubSourceGrpc.out")
+  override val shape: SourceShape[v1.ReceivedMessage] = SourceShape(out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) with OutHandler {
+      private var state: State = Pending
+      private var fetches: Int = 0
+
+      def fetch(implicit mat: Materializer): Unit = {
```
lustefaniak: Maybe I don't understand that but what is the point of passing Materializer here if it can be retrieved inside GraphStageLogic in same way as you do in onPull?

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GrpcApi.scala
```diff
@@ -0,0 +1,75 @@
...
+
+  private val subscriptionFqrn = s"projects/$project/subscriptions/$subscription"
+
+  private val grpcChannel: ManagedChannel = {
+    val builder = ManagedChannelBuilder.forAddress(config.host, config.port)
+    if (config.usePlaintext) builder.usePlaintext(true)
+    builder.build()
+  }
+
+  private val callCredentials: CallCredentials =
```
lustefaniak: It would be extremaly usefull to allow providing Credentials from the outside, eg. `NoCredentialsProvider.create` for the Emulator.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSourceGrpc.scala
lustefaniak: I understand it is async, but I think code running inside async callback has access to materializer.

lustefaniak:
I would also consider publishing separate artifact, as pulling `"com.google.cloud" % "google-cloud-pubsub" % "0.41.0-beta",` might be not great for people using only http interface.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GrpcApi.scala
lustefaniak: You are not using builders which accept `CredentialsProvider` but `CallCredentials` directly, is there any reason for that? 
When using API that way to add support for NoCredentialsProvider you would have to skip `withCallCredentials` call when provided credentials are `null`/`None`.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSourceGrpc.scala
```diff
@@ -0,0 +1,98 @@
...
+        import mat.executionContext
+        fetches += 1
+
+        grpcApi.read.onComplete { tr =>
+          fetches -= 1
+          callback.invoke(tr -> mat)
+        }
+
+        state = Fetching
```
lustefaniak: This is probably not needed, or other `state = Fetching` should be removed

2m:
Thank you @CremboC for pushing this PR forward! We agree that an alternative connector for Google Cloud PubSub using gRPC is useful. However we would like the implementation to use the upcoming https://github.com/akka/akka-grpc library instead.

akka-grpc provides Akka Stream connectors for gRPC services, which will simplify this connector a bit, such as replacing the `GooglePubSubSourceGrpc` custom graph stage. The authentication will still be handled by Google libraries, so we will not have to re-implement it.

WDYT? Would you be keen on helping to migrate this PR to use akka-grpc?

Since akka-grpc does not have a release yet (one is coming soon), I think we could release a milestone version of akka-grpc and use that for this connector until a public release becomes available. I could kick-start the migration by pushing a commit to this PR where it starts using akka-grpc.

CremboC:
Hey @2m, to be honest, it is completely up to you how you think is best for the community! I agree the implementation I'm providing is not very nice, but at the same time it's quite simple.

As I mentioned before, the version you're see here (with all the latest commits) is what we use in production right now and it seems to be very performant and fault tolerant (esp. after implementing the retry-on-failure).

Regarding `akka-grpc` - I'll have a look at it once it gets a release and let you know! Though obviously some examples would be needed for me to efficiently transform this implementation.

2m:
Superseded by #1162

@CremboC, sorry it took a while. #1162 now implements Google Pub Sub connector via gRPC using Akka Grpc. The integration code between Akka Streams and gRPC client now comes from Akka Grpc. So the new connector is a bit lighter on the code. I would greatly appreciate if you could take a look at it.

Issue #173, opened 2017-02-05
Add Java examples to SSE docs


patriknw:
see https://github.com/akka/alpakka/issues/47

ennru:
This got fixed earlier in #762 and is online since 0.17.

PR #762, opened 2018-02-02
add missing Java docs for SSE Source

ktoso:
The SSE connector has existing JavaDSL but it was not shown in docs, misleading users to think does not exist.

I added a docs sample, compile only spec.


ktoso:
cc @myfear thanks for pointing it out

ennru:
Fixes #173

raboof:
(travis complains about whitespace below the header in `EventSourceTest.java`)

ktoso:
Oh, good that it had a ticket.

PR #764, opened 2018-02-03
google-fcm

### LINKED ISSUE #750: Google FCM push notification sender flow
The basic idea is to send push notifications thought streams.

Starting point could be this project: https://github.com/Ceratech/fcm-scala but changing playWs and DI and play-json to the more akka-feel; akkaHttp-core and spray-json.

I will start to working on this in the next week! If sb interested feel free to comment below!

tg44:
fixes #750 

Things I did:
 - inited a new module
 - wrote the necessary models + jsconverters
 - wrote the communication
 - can authenticate and communicate with google
 - wrote test for things
 - can send notifications \o/

Things missing but I will working on after get some responses:
 - write some docs 
 - write the missing tests (after the api freezes)
 - write the java api
 - solve the leftover todos in the code

Questions:
 1. The google-pubsub connector do the jwt encoding by hand, I used the jwt-core. By hand coding, or use a lib?
 2. The google-pubsub use alpakka.googlecloud.pubsub, I used alpakka.google.firebase.fcm. More intermed package or not?
 3. I have "common" codes with the google-pubsub. If you want I can modify both projects to use the same classes, or we could merge these two modul (not so much dependency, not large codebase). (We will have an other google connector (bigquery)  soon, basically the same auth process, same deps the answer to this point can depends on this new info :D If I get close to merge with this, I will port our bigquery code to alpakka ready and PR it too.)
 4. Do I really need to check the configs (get a token) at prestart?
 5. In the FcmNotification class I have an isSendable def which can check if exactly one of the mandatory fields is present. Is it ok to let the users be responsible if they try to send a not sendable message or not? (There are other things I can't check and can cause problems so calling this function is not saving you just help you.)
 6. Is it better to try to parse the error to caseclasses or let it in a jsValue format and let the users decide if they parse it (and handle it) or not? (or try to parse, and if we fail give back the JsValue?)
 7. How can I integration test this? Any idea?
 8. As I asked before how bad using shared HttpExts?

### DIFF
--- a/docs/src/main/paradox/google-fcm.md
+++ b/docs/src/main/paradox/google-fcm.md
@@ -1,0 +1,108 @@
+# Google Firebase Cloud Messaging
+
+The google firebase cloud messaging connector provides a way to send notifications https://firebase.google.com/docs/cloud-messaging/ .
+
+@@@ warning { title='Early state' }
+The whole FCM server implementation doc is a bit unclear. 
+This connector is build from scratch following the documentation.
+So the error parsing, the condition builder, the apns object and some other object/case class
+could (and possibly will) be improved.
+@@@
+
+
+### Reported issues
+
+[Tagged issues at Github](https://github.com/akka/alpakka/labels/p%3Agoogle-fcm)
+
+
+## Artifacts
+
+@@dependency [sbt,Maven,Gradle] {
+  group=com.lightbend.akka
+  artifact=akka-stream-alpakka-google-fcm_$scalaBinaryVersion$
+  version=$version$
+}
+
+## Usage
+
+Possibly needed imports for the following codes.
+
+Scala
+: @@snip ($alpakka$/google-fcm/src/test/scala/akka/stream/alpakka/google/firebase/fcm/scaladsl/Examples.scala) { #imports }
+
+Java
+: @@snip ($alpakka$/google-fcm/src/test/java/akka/stream/alpakka/google/firebase/fcm/javadsl/Examples.java) { #imports }
+
+
+Prepare the actor system and materializer.
+
+Scala
+: @@snip ($alpakka$/google-fcm/src/test/scala/akka/stream/alpakka/google/firebase/fcm/scaladsl/Examples.scala) { #init-mat }
+
+Java
+: @@snip ($alpakka$/google-fcm/src/test/java/akka/stream/alpakka/google/firebase/fcm/javadsl/Examples.java) { #init-mat }
+
+
+Prepare your credentials for access to FCM.
+
+Scala
+: @@snip ($alpakka$/google-fcm/src/test/scala/akka/stream/alpakka/google/firebase/fcm/scaladsl/Examples.scala) { #init-credentials }
+
+Java
+: @@snip ($alpakka$/google-fcm/src/test/java/akka/stream/alpakka/google/firebase/fcm/javadsl/Examples.java) { #init-credentials }
+
+The last two parameters in the above example are the predefined values. 
+You can send test notifications [(so called validate only).](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages/send)
+And you can set the number of maximum concurrent connections.
+There is a limitation in the docs; from one IP you can have maximum 1k pending connections, 
+and you may need to configure `akka.http.host-connection-pool.max-open-requests` in your application.conf.
+
+To send a notification message create your notification object, and send it!
+
+Scala
+: @@snip ($alpakka$/google-fcm/src/test/scala/akka/stream/alpakka/google/firebase/fcm/scaladsl/Examples.scala) { #simple-send }
+
+Java
+: @@snip ($alpakka$/google-fcm/src/test/java/akka/stream/alpakka/google/firebase/fcm/javadsl/Examples.java) { #simple-send }
+
+With fire and forget you will just send messages and ignore all the errors.
+This is not so healthy in general so you can use the send instead.
+
+Scala
+: @@snip ($alpakka$/google-fcm/src/test/scala/akka/stream/alpakka/google/firebase/fcm/scaladsl/Examples.scala) { #asFlow-send }
+
+Java
+: @@snip ($alpakka$/google-fcm/src/test/java/akka/stream/alpakka/google/firebase/fcm/javadsl/Examples.java) { #asFlow-send }
+
+With this type of send you can get responses from the server.
+These responses can be @scaladoc[positive](akka.stream.alpakka.google.firebase.fcm.FcmFlowModels.FcmSuccessResponse) or @scaladoc[negative](akka.stream.alpakka.google.firebase.fcm.FcmFlowModels.FcmErrorResponse). 
+You can choose what you want to do with this information, but keep in mind
+if you try to resend the failed messages you will need to implement exponential backoff too!
+
+To help the integration and error handling or logging, there is a variation of the flow where you can send data beside your notification.
+
+Scala
+: @@snip ($alpakka$/google-fcm/src/test/scala/akka/stream/alpakka/google/firebase/fcm/scaladsl/Examples.scala) { #withData-send }
+
+Java
+: @@snip ($alpakka$/google-fcm/src/test/java/akka/stream/alpakka/google/firebase/fcm/javadsl/Examples.java) { #withData-send }
+
+Here I send a simple string, but you could use any type.
+
+## Scala only
+
+You can build any notification described in the original documentation.
+It can be done by hand, or using some builder method.
+If you build your notification from scratch with options (and not with the provided builders), worth to check isSendable before sending.
+
+Scala
+: @@snip ($alpakka$/google-fcm/src/test/scala/akka/stream/alpakka/google/firebase/fcm/scaladsl/Examples.scala) { #noti-create }
+
+There is a condition builder too.
+
+Scala
+: @@snip ($alpakka$/google-fcm/src/test/scala/akka/stream/alpakka/google/firebase/fcm/scaladsl/Examples.scala) { #condition-builder }
+
+## Running the examples
+
+To run the example code you will need to configure a project and notifications in google firebase and provide your own credentials.
--- a/google-fcm/src/main/scala/akka/stream/alpakka/google/firebase/fcm/javadsl/GoogleFcmFlow.scala
+++ b/google-fcm/src/main/scala/akka/stream/alpakka/google/firebase/fcm/javadsl/GoogleFcmFlow.scala
@@ -1,0 +1,33 @@
+/*
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.google.firebase.fcm.javadsl
+
+import akka.NotUsed
+import akka.actor.ActorSystem
+import akka.http.scaladsl.Http
+import akka.japi.Pair
+import akka.stream.alpakka.google.firebase.fcm.FcmFlowModels.{FcmFlowConfig, FcmResponse}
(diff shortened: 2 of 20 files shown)


tg44:
I will have some time to work on this in the weekend so:
@stephennancekivell what do you think about the first four questions?

stephennancekivell:
Hi @tg44. That looks pretty cool. I've got a few thoughts but Im not expert.

> The google-pubsub connector do the jwt encoding by hand, I used the jwt-core. By hand coding, or use a lib?

jwt-core looks good. I wish i new about some it before.

> The google-pubsub use alpakka.googlecloud.pubsub, I used alpakka.google.firebase.fcm. More intermed package or not?

`alpakka.google.firebase.fcm` looks good to me. firebase looks like a standalone google product than pubsub

> I have "common" codes with the google-pubsub. If you want I can modify both projects to use the same classes, or we could merge these two modul (not so much dependency, not large 
codebase). (We will have an other google connector (bigquery) soon, basically the same auth process, same deps the answer to this point can depends on this new info :D If I get close to merge with this, I will port our bigquery code to alpakka ready and PR it too.)

A shared module is a nice idea, but it sees like a bit of a yack shave. Could be done later.

> Do I really need to check the configs (get a token) at prestart?

> In the FcmNotification class I have an isSendable def which can check if exactly one of the mandatory fields is present. Is it ok to let the users be responsible if they try to send a not sendable message or not? (There are other things I can't check and can cause problems so calling this function is not saving you just help you.)

It would be nicer if only valid states were representable. Maybe `FcmNotification` could be a `sealed trait` of some kind.

> Is it better to try to parse the error to caseclasses or let it in a jsValue format and let the users decide if they parse it (and handle it) or not? (or try to parse, and if we fail 
give back the JsValue?)

I think exposting spray's JsValue is kinda bad. It means the library cant change the internal json lib without breaking compatability. It could just be the json string or a bool.

> How can I integration test this? Any idea?

I dont think its a requirement. Testing it against mocks (wire-mock) is good. If you really want it looks like there is a local-emulator, I even saw some docker images for it. https://firebase.google.com/docs/functions/local-emulator

> As I asked before how bad using shared HttpExts?

Dont know. I wouldnt think so.


A few other tips
---
* It would be good to have a test and example docs showing how to use it from a Stream and/or Sink.
* Be very restrictive with accessors `private` and package private. Only expose what end users what they really need.
* add a javadsl/scaladsl api. I think thats wanted for alpakka modules. Its surprisingly little work to add java support using the converters. A java test will be very helpful.

tg44:
Thanks for the tips! Amended. I think this is not work in progress anymore.

The things I not really like:
 - No good java api for the whole notification forging. (maps and seqs in the caseclasses... but can send basic notifications)
 - Apns class (raw string, could be some caseclass instead, but I have no time to red the whole apns doc too.)
 - Error handling (raw string, not find the format in the docs so need more digging or some reverse engineering)
 - Condition builder (builds the strings, but a bit underdocumented, can't test in real life)

If we can live with these things I think the code is mergable. PR-s can solve these "issues" later.

docs/src/main/paradox/google-fcm.md
```diff
@@ -0,0 +1,99 @@
+# Google Firebase Cloud Messaging
+
+The google firebase cloud messaging connector provides a way to send notifications https://firebase.google.com/docs/cloud-messaging/ .
```
ennru: Make it a link instead.

docs/src/main/paradox/google-fcm.md
```diff
@@ -0,0 +1,99 @@
...
+: @@snip ($alpakka$/google-fcm/src/test/scala/akka/stream/alpakka/google/firebase/fcm/scaladsl/Examples.scala) { #noti-create }
+
+There is a condition builder too.
+
+Scala
+: @@snip ($alpakka$/google-fcm/src/test/scala/akka/stream/alpakka/google/firebase/fcm/scaladsl/Examples.scala) { #condition-builder }
+
+## This is an early state
+
+The whole FCM server implementation doc is a bit unclear.
```
ennru: Wouldn't it be fair to tell the reader this in the beginning? Maybe as a [Paradox note](https://developer.lightbend.com/docs/paradox/latest/features/callouts.html#warning-callout)?

project/Dependencies.scala
```diff
@@ -127,6 +127,15 @@ object Dependencies {
     )
   )
 
+  val GoogleFcm = Seq(
+    libraryDependencies ++= Seq(
+      "com.typesafe.akka" %% "akka-http" % AkkaHttpVersion,
+      "com.typesafe.akka" %% "akka-http-spray-json" % AkkaHttpVersion,
+      "com.pauldijou" %% "jwt-core" % "0.14.1",
```
ennru: For completeness please state the license here.

docs/src/main/paradox/google-fcm.md
tg44: https://developer.lightbend.com/docs/alpakka/current/google-cloud-pub-sub.html
https://raw.githubusercontent.com/akka/alpakka/master/docs/src/main/paradox/google-cloud-pub-sub.md

It will be a link. (It's a link in local too.)

project/Dependencies.scala
tg44: fixed

docs/src/main/paradox/google-fcm.md
tg44: Nice! Fixed!

tg44:
I just resolved the conflict in the build.sbt. I broke down the google-cloud to google.cloud in the pubsub too.

google-fcm/src/main/scala/akka/stream/alpakka/google/firebase/fcm/javadsl/GoogleFcm.scala
```diff
@@ -0,0 +1,39 @@
...
+import akka.actor.ActorSystem
+import akka.http.scaladsl.Http
+import akka.japi.Pair
+import akka.stream.alpakka.google.firebase.fcm.FcmFlowModels.{FcmFlowConfig, FcmResponse}
+import akka.stream.alpakka.google.firebase.fcm.impl.FcmSender
+import akka.stream.alpakka.google.firebase.fcm.{FcmFlowModels, FcmNotification}
+import akka.stream.scaladsl.{Flow, Sink}
+import akka.stream.{javadsl, Materializer}
+
+object GoogleFcm {
```
ennru: Most other connectors have the factories split into Source, Flow, Sink.
Might be an idea to adhere to that scheme and make it `GoogleFcmFlow` and `GoogleFcmSink`.

google-fcm/src/test/java/akka/stream/alpakka/google/firebase/fcm/javadsl/Examples.java
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.google.firebase.fcm.javadsl;
+
+import akka.actor.ActorSystem;
+import akka.japi.Pair;
```
ennru: Make the imports show in the documentation, please.

google-fcm/src/test/scala/akka/stream/alpakka/google/firebase/fcm/impl/FcmSenderSpec.scala
```diff
@@ -0,0 +1,106 @@
...
+        http.singleRequest(any[HttpRequest](),
+                           any[HttpsConnectionContext](),
+                           any[ConnectionPoolSettings](),
+                           any[LoggingAdapter]())
+      ).thenReturn(
+        Future.successful(HttpResponse(entity = HttpEntity(ContentTypes.`application/json`, """{"name": "test"}""")))
+      )
+
+      val result: FcmFlowModels.FcmResponse =
+        Await.result(sender.send("projectId", "token", http, FcmSend(false, FcmNotification.empty)), timeout)
```
ennru: You might want to use Scalatest's ScalaFutures trait instead of Await.

tg44:
I'll fix these today!

google-fcm/src/test/scala/akka/stream/alpakka/google/firebase/fcm/impl/FcmSenderSpec.scala
tg44: This was not as successful as I want bcs of the multiple assertions.

tg44:
MQTT connection lost :( It works here.

docs/src/main/paradox/google-fcm.md
```diff
@@ -0,0 +1,108 @@
...
+
+Prepare your credentials for access to FCM.
+
+Scala
+: @@snip ($alpakka$/google-fcm/src/test/scala/akka/stream/alpakka/google/firebase/fcm/scaladsl/Examples.scala) { #init-credentials }
+
+Java
+: @@snip ($alpakka$/google-fcm/src/test/java/akka/stream/alpakka/google/firebase/fcm/javadsl/Examples.java) { #init-credentials }
+
+The last two parameter in the above example are the predefined values.
```
ennru: parameter*s*

docs/src/main/paradox/google-fcm.md
```diff
@@ -0,0 +1,108 @@
...
+Java
+: @@snip ($alpakka$/google-fcm/src/test/java/akka/stream/alpakka/google/firebase/fcm/javadsl/Examples.java) { #init-credentials }
+
+The last two parameter in the above example are the predefined values. 
+You can send test notifications [(so called validate only).](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages/send)
+And you can set the number of maximum concurrent connections.
+There is a limitation in the docs; from one IP you can have maximum 1k pending connections, 
+and you may need to configure `akka.http.host-connection-pool.max-open-requests` in your application.conf.
+
+To send notification message create your notification object, and send it!
```
ennru: To send *a* notification...

google-fcm/src/main/scala/akka/stream/alpakka/google/firebase/fcm/javadsl/GoogleFcmFlow.scala
```diff
@@ -0,0 +1,33 @@
...
+import akka.http.scaladsl.Http
+import akka.japi.Pair
+import akka.stream.alpakka.google.firebase.fcm.FcmFlowModels.{FcmFlowConfig, FcmResponse}
+import akka.stream.alpakka.google.firebase.fcm.impl.FcmSender
+import akka.stream.alpakka.google.firebase.fcm.{FcmFlowModels, FcmNotification}
+import akka.stream.scaladsl.{Flow, Sink}
+import akka.stream.{javadsl, Materializer}
+
+object GoogleFcmFlow {
+  def sendWithData[T](
```
ennru: In other connectors this is called `...WithPassThrough`. (And the typical use case is to commit to Kafka after data is pushed somewhere else.)

tg44: In this case, when you fail a send maybe you want to resend/reschedule. And the returning error message tells you nearly nothing about what was the input data. This was the main reason I created this function. (And I read it in the "good practices" section too ;) )

ennru:
The builds fail for a lovely formatting issue.

tg44:
Fml. I do a paradox build for this reason only (and actually chacked the html too)... Now I know that just builds and not format...

ennru:
The Paradox build doesn't even compile. It runs Scaladoc and Paradox. I managed to push compile errors before...

ennru:
Thank you, for this new connector!

Issue #731, opened 2018-01-19
S3: S3Exception has no relevant toString


ennru:
Just noticed that the Alpakka S3Exception has useful information, but doesn't tell about it in its `toString` method.

```
[error] Caused by: akka.stream.alpakka.s3.S3Exception: 
[error]     at akka.stream.alpakka.s3.impl.S3Stream$$anonfun$akka$stream$alpakka$s3$impl$S3Stream$$entityForSuccess$1.apply(S3Stream.scala:362)
```

ennru:
Fixed by #765

PR #765, opened 2018-02-04
adding toString override to S3Exception #731

svezfaz:
Refs #731


ennru:
Thank you for fixing this.

PR #768, opened 2018-02-09
Contributor advice

### LINKED ISSUE #752: Contributor advice proposal
I see the need to collect some advice how to build good Alpakka modules following the style that has evolved.
Please comment to get a common view of this. I'll create a PR with it when we get close to the next release.

Here is goes:

# Some advice for Alpakka contributors

## API design

Depending on the technology you integrate with Akka Streams and Alpakka you'll create Sources, Flows and Sinks.

When designing Sinks, please check if a Flow might be a better fit. Most other systems will give some kind of response which
could continue in a Flow. Sinks can easily created be attaching a Sink.ignore to the Flow. In case you want to offer 
Sinks based on Flows, focus on the most important cases and leave the more intricate variants just in the Flow API.

When designing Flows, consider adding an extra field to the in- and out-messages which is passed through. A common
use case we see, is committing a Kafka offset after passing data to another system.

### Settings

Most technologies will have a couple of configuration settings that will be needed for all Flows. Create case classes
collecting these settings instead of passing them in every method.

Create a default instance in the companion object with good defaults which can be updated via `copy` or `withXxxx` methods.

As Java can't use the `copy` methods, add `withXxxx` methods to specify certain fields in the settings instance. 

In case you see the need to support reading the settings from `Config`, offer a method taking the `Config` instance so
that the user can apply a proper namespace.
Refrain from using `akka.stream` as prefix, prefer `alpakka` as root namespace.

### Packages & Scoping

Use `private`, `private[connector]` and `final` extensively to limit the API surface. 

| Package                                  | Purpose
| -----------------------------------------|------------------------
| `akka.stream.alpakka.connector.javadsl`  | Java-only part of the API, normally factories for Sources, Flows and Sinks
| `akka.stream.alpakka.connector.scaladsl` | Scala-only part of the API, normally factories for Sources, Flows and Sinks
| `akka.stream.alpakka.connector`          | Shared API, eg. settings classes, implementation if thorowly made private
| `akka.stream.alpakka.connector.impl`     | Optional, implementation in separate package

### Regarding Java APIs

* Provide factory methods for Sources, Flows and Sinks in the `javadsl` package
* The Akka Stream Scala instances have a `.asJava` method to convert to the `akka.stream.javadsl` counterparts
* Let the Java API use Java standard types (`java.util.Optional`, `java.util.List`, ...) and convert using 
`JavaConverters` where needed
* Use the `akka.japi.Pair` class to return tuples
* Use `scala.compat.java8.FutureConverters` to translate Futures to `CompletionStage`s

## Graph stage checklist

* Keep mutable state within the `GraphStageLogic` only
* Open/close connections in `preStart` 
* Release resources in `postStop`
* Fail early on configuration errors

## Documentation

Create or complement the documentation in the `docs` module using [Paradox](https://github.com/lightbend/paradox) syntax.
Prepare code snippets to be integrated by Paradox in the tests.

Use ScalaDoc if you see the need to describe the API usage better than the naming does.

ennru:
Fixes #752

### DIFF
--- a/contributor-advice.md
+++ b/contributor-advice.md
@@ -2,31 +2,20 @@
 
 ## API design
 
-### Flows
-
-When designing Flows, consider adding an extra field to the in- and out-messages which is passed through. A common
-use case we see, is committing a Kafka offset after passing data to another system.
-
-### Settings
-
-Most technologies will have a couple of configuration settings that will be needed for all Flows. Create case classes 
-collecting these settings instead of passing them in every method.
-
-Create a default instance in the companion object with good defaults which can be updated via `copy` or `withXxxx` methods.
-
-As Java can't use the `copy` methods, add `withXxxx` methods to specify certain fields in the settings instance.
-
-In case you see the need to support reading the settings from `Config`, offer a method taking the `Config` instance so
-that the user can apply a proper namespace.
-Refrain from using `akka.stream` as prefix, prefer `alpakka` as root namespace.
-
 ### Public factory methods
 
 Depending on the technology you integrate with Akka Streams and Alpakka you'll create Sources, Flows and Sinks.
 Regardless on how they are implemented make sure that you create the relevant Sources, Sinks and Flows APIs so they are
 simple and easy to use.
 
+### Flows
+
+When designing Flows, consider adding an extra field to the in- and out-messages which is passed through. A common
+use case we see, is committing a Kafka offset after passing data to another system.
+
 ### Java APIs
+
+Alpakka, same as Akka, aims to keep 100% feature parity between the various language DSLs.
 
 * Provide factory methods for Sources, Flows and Sinks in the `javadsl` package wrapping all the methods in the Scala API
 * The Akka Stream Scala instances have a `.asJava` method to convert to the `akka.stream.javadsl` counterparts
@@ -34,6 +23,19 @@ simple and easy to use.
 `JavaConverters` where needed
 * Use the `akka.japi.Pair` class to return tuples
 * Use `scala.compat.java8.FutureConverters` to translate Futures to `CompletionStage`s
+
+### Settings
+
+Most technologies will have a couple of configuration settings that will be needed for several Sinks, Flows, or Sinks. 
+Create case classes collecting these settings instead of passing them in every method.
+
+Create a default instance in the companion object with good defaults which can be updated via `copy` or `withXxxx` methods.
+
+As Java can't use the `copy` methods, add `withXxxx` methods to specify certain fields in the settings instance.
+
+In case you see the need to support reading the settings from `Config`, offer a method taking the `Config` instance so
+that the user can apply a proper namespace.
+Refrain from using `akka.stream.alpakka` as Config prefix, prefer `alpakka` as root namespace.
 
 
 ## Implementation details
@@ -72,15 +74,17 @@ Use `private`, `private[connector]` and `final` extensively to limit the API sur
 * Open connections in `preStart`
 * Release resources in `postStop`
 * Fail early on configuration errors
-* No Blocking At Any Time 
-* Make sure the code is thread-safe
+* Make sure the code is thread-safe; if in doubt, please ask!
+* No Blocking At Any Time -- in other words, avoid blocking whenever possible and replace it with asynchronous 
+programming (async callbacks, stage actors)
 
 ### Keep the code DRY
 
 Avoid duplication of code between different Sources, Sinks and Flows. Extract the common logic to a common abstract
 base `GraphStageLogic` that is inherited by the `GraphStage`s.
 
-In some cases it may be useful to implement the Sources or the Sinks by means of Flows:
+Sometimes it may be useful to provide a Sink or Source for a connector, even if the main concept is implemented
+as a Flow. This can be easily done by reusing the Flow implementation:
 * Source: `Source.maybe.viaMat(MyFlow)(Keep.right)`
 * Sink: `MyFlow.toMat(Sink.ignore)(Keep.right)`
 
@@ -97,8 +101,9 @@ Please ensure that you limit the amount of resources used by the containers.
 
 ## Documentation
 
-Using [Paradox](https://github.com/lightbend/paradox) syntax, create or complement the documentation in the `docs` module.
-Prepare code snippets to be integrated by Paradox in the tests. Such example should be part of real tests and no in
+Using [Paradox](https://github.com/lightbend/paradox) syntax (which is very close to markdown), create or complement
+the documentation in the `docs` module.
+Prepare code snippets to be integrated by Paradox in the tests. Such example should be part of real tests and not in
 unused methods.
 
 Use ScalaDoc if you see the need to describe the API usage better than the naming does.
--- a/contributor-advice.md
+++ b/contributor-advice.md
@@ -2,33 +2,62 @@
 
 ## API design
 
-Depending on the technology you integrate with Akka Streams and Alpakka you'll create Sources, Flows and Sinks.
-
-When designing Sinks, please check if a Flow might be a better fit. Most other systems will give some kind of response which
-could continue in a Flow. Sinks can easily created be attaching a Sink.ignore to the Flow. In case you want to offer 
-Sinks based on Flows, focus on the most important cases and leave the more intricate variants just in the Flow API.
+### Flows
 
 When designing Flows, consider adding an extra field to the in- and out-messages which is passed through. A common
 use case we see, is committing a Kafka offset after passing data to another system.
 
-
 ### Settings
 
-Most technologies will have a couple of configuration settings that will be needed for all Flows. Create case classes
+Most technologies will have a couple of configuration settings that will be needed for all Flows. Create case classes 
 collecting these settings instead of passing them in every method.
 
 Create a default instance in the companion object with good defaults which can be updated via `copy` or `withXxxx` methods.
 
-As Java can't use the `copy` methods, add `withXxxx` methods to specify certain fields in the settings instance. 
+As Java can't use the `copy` methods, add `withXxxx` methods to specify certain fields in the settings instance.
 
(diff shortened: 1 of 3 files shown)


contributor-advice.md
```diff
@@ -0,0 +1,107 @@
...
+
+Write tests for all public methods and possible settings.
+
+Use Docker containers for any testing needed by your application. The container's configuration can be added to the
+`docker-compose.yml` file.
+Please ensure that you limit the amount of resources used by the containers.
+
+## Documentation
+
+Using [Paradox](https://github.com/lightbend/paradox) syntax, create or complement the documentation in the `docs` module.
```
ktoso: syntax (which is very close to markdown)

contributor-advice.md
```diff
@@ -0,0 +1,107 @@
...
+Prepare code snippets to be integrated by Paradox in the tests. Such example should be part of real tests and no in
```
ktoso: no => not

contributor-advice.md
```diff
@@ -0,0 +1,107 @@
...
+* Make sure the code is thread-safe
+
+### Keep the code DRY
+
+Avoid duplication of code between different Sources, Sinks and Flows. Extract the common logic to a common abstract
+base `GraphStageLogic` that is inherited by the `GraphStage`s.
+
+In some cases it may be useful to implement the Sources or the Sinks by means of Flows:
+* Source: `Source.maybe.viaMat(MyFlow)(Keep.right)`
+* Sink: `MyFlow.toMat(Sink.ignore)(Keep.right)`
```
ktoso: I'd reword a bit: Sometimes it may be useful for users to provide them with Sink or Source versions of an operation, even if the main concept is implemented (and most useful) as a Flow. This can be easily implemented as (reusing the core Flow implementation):

contributor-advice.md
```diff
@@ -0,0 +1,107 @@
...
+| `akka.stream.alpakka.connector.scaladsl` | Scala-only part of the API, normally factories for Sources, Flows and Sinks
+| `akka.stream.alpakka.connector`          | Shared API, eg. settings classes, implementation if thorowly made private
+| `akka.stream.alpakka.connector.impl`     | Optional, implementation in separate package
+
+### Graph stage checklist
+
+* Keep mutable state within the `GraphStageLogic` only 
+* Open connections in `preStart`
+* Release resources in `postStop`
+* Fail early on configuration errors
```
ktoso: fail using `failStage` instead of throwing exceptions?

cc @patriknw

contributor-advice.md
```diff
@@ -0,0 +1,107 @@
...
+* No Blocking At Any Time
```
ktoso: yeap, classic motto, but let's explain it too, for new people:

"No Blocking At Any Time" -- in other words, avoid blocking whenever possible and replace it with asynchronous programming (async callbacks, stage actors)

contributor-advice.md
```diff
@@ -0,0 +1,107 @@
...
+| `akka.stream.alpakka.connector.impl`     | Optional, implementation in separate package
+
+### Graph stage checklist
+
+* Keep mutable state within the `GraphStageLogic` only 
+* Open connections in `preStart`
+* Release resources in `postStop`
+* Fail early on configuration errors
+* No Blocking At Any Time 
+* Make sure the code is thread-safe
```
ktoso: add: "; if in doubt, please ask!"

contributor-advice.md
```diff
@@ -0,0 +1,107 @@
...
+Refrain from using `akka.stream` as prefix, prefer `alpakka` as root namespace.
+
+### Public factory methods
+
+Depending on the technology you integrate with Akka Streams and Alpakka you'll create Sources, Flows and Sinks.
+Regardless on how they are implemented make sure that you create the relevant Sources, Sinks and Flows APIs so they are
+simple and easy to use.
+
+### Java APIs
+
```
ktoso: Alpakka, same as Akka, aims to keep 100% feature parity between the various language DSLs.

contributor-advice.md
```diff
@@ -0,0 +1,107 @@
...
+* Provide factory methods for Sources, Flows and Sinks in the `javadsl` package wrapping all the methods in the Scala API
+* The Akka Stream Scala instances have a `.asJava` method to convert to the `akka.stream.javadsl` counterparts
+* Let the Java API use Java standard types (`java.util.Optional`, `java.util.List`, ...) and convert using
+`JavaConverters` where needed
+* Use the `akka.japi.Pair` class to return tuples
+* Use `scala.compat.java8.FutureConverters` to translate Futures to `CompletionStage`s
```
ktoso: also mention collection converters? Suggest using List in Java

contributor-advice.md
```diff
@@ -0,0 +1,107 @@
+# Some advice for Alpakka contributors
+
+## API design
+
+### Flows
+
+When designing Flows, consider adding an extra field to the in- and out-messages which is passed through. A common
+use case we see, is committing a Kafka offset after passing data to another system.
+
+### Settings
```
ktoso: move settings under javadsl since we talk about it here

contributor-advice.md
```diff
@@ -0,0 +1,107 @@
...
+
+In case you see the need to support reading the settings from `Config`, offer a method taking the `Config` instance so
+that the user can apply a proper namespace.
+Refrain from using `akka.stream` as prefix, prefer `alpakka` as root namespace.
+
+### Public factory methods
+
+Depending on the technology you integrate with Akka Streams and Alpakka you'll create Sources, Flows and Sinks.
+Regardless on how they are implemented make sure that you create the relevant Sources, Sinks and Flows APIs so they are
+simple and easy to use.
```
ktoso: good

contributor-advice.md
```diff
@@ -0,0 +1,107 @@
...
+### Packages & Scoping
+
+Use `private`, `private[connector]` and `final` extensively to limit the API surface.
+
+| Package                                  | Purpose
+| -----------------------------------------|------------------------
+| `akka.stream.alpakka.connector.javadsl`  | Java-only part of the API, normally factories for Sources, Flows and Sinks
+| `akka.stream.alpakka.connector.scaladsl` | Scala-only part of the API, normally factories for Sources, Flows and Sinks
+| `akka.stream.alpakka.connector`          | Shared API, eg. settings classes, implementation if thorowly made private
+| `akka.stream.alpakka.connector.impl`     | Optional, implementation in separate package
```
ktoso: great table

contributor-advice.md
```diff
@@ -0,0 +1,107 @@
+# Some advice for Alpakka contributors
+
+## API design
+
+### Flows
```
juanjoDiaz: I'd move this whole `Flows` section below the `Public factory methods` where we mention the Sources, Sinks and Flows.

contributor-advice.md
```diff
@@ -0,0 +1,107 @@
...
+
+When designing Flows, consider adding an extra field to the in- and out-messages which is passed through. A common
+use case we see, is committing a Kafka offset after passing data to another system.
+
+### Settings
+
+Most technologies will have a couple of configuration settings that will be needed for all Flows. Create case classes
```
juanjoDiaz: Most technologies will have a couple of configuration settings that will be needed for all *connector*

We do more than flows here :)

contributor-advice.md
```diff
@@ -0,0 +1,107 @@
...
+Most technologies will have a couple of configuration settings that will be needed for all Flows. Create case classes 
+collecting these settings instead of passing them in every method.
+
+Create a default instance in the companion object with good defaults which can be updated via `copy` or `withXxxx` methods.
+
+As Java can't use the `copy` methods, add `withXxxx` methods to specify certain fields in the settings instance.
+
+In case you see the need to support reading the settings from `Config`, offer a method taking the `Config` instance so
+that the user can apply a proper namespace.
+Refrain from using `akka.stream` as prefix, prefer `alpakka` as root namespace.
```
juanjoDiaz: This sentence is very true but has little to do with Settings. I'd move it to a separate section for this kind of general stuff

contributor-advice.md
juanjoDiaz: I'd move the thread-safe bullet point to be before the no-blocking one. Since blocking tends to be the first approach that people think to tackle thread safety.

contributor-advice.md
```diff
@@ -0,0 +1,107 @@
...
+* Sink: `MyFlow.toMat(Sink.ignore)(Keep.right)`
+
+You do not need to expose every configuration a Flow offers this way -- Focus on the most important ones.
+
+
+## Test
+
+Write tests for all public methods and possible settings.
+
+Use Docker containers for any testing needed by your application. The container's configuration can be added to the
```
burakkose: In my view, before suggesting to use docker, we should encourage to use internal testing API like MiniSolrCloudCluster for Solr, ElasticsearchClusterRunner for Elasticsearch, if it is possible.

juanjoDiaz: Using Docker make the testing environment more isolated than if we configure stuff just for the test. It allows us to limit, memory, disk or whatever we want to limit. And it's trivial to update regardless of the connector version or dependencies.

Also, I'd rather pulling a container for testing than adding an extra dependency.

contributor-advice.md
ennru: Anything more than it says in "Let the Java API use Java standard types (`java.util.Optional`, `java.util.List`, ...) and convert using `JavaConverters` where needed" ?

contributor-advice.md
ennru: Ah, it talks about the Config namespace. No the package names. I make it more specific.

contributor-advice.md
ktoso: Sounds good

ktoso: Good as is then

juanjoDiaz:
Does that last commit about Slick belong to this PR or was it accidental?

ennru:
It does not belong to it, but was intentional. The builds failed with too long log output, so I need to cut it down.

PR #781, opened 2018-02-15
S3: Rewrite DiskBuffer using plain streams to not leak handles

### LINKED ISSUE #780: S3: File handles leak when using DiskBuffer
The S3 client needs to calculate hashes of 5MB-sized fragments for the amazon API. It can be configured to buffer this to disk, using the `DiskBuffer` class.

The current implementation seems to leak file handles. That's probably related to it using memory-mapped files, and [FileChannel not being unmappable](http://stackoverflow.com/questions/2972986/how-to-unmap-a-file-from-memory-mapped-using-filechannel-in-java) in Java. However, even a GC seems to not clean up the file handles currently. (the files themselves are deleted though).

It should probably be fine to just use ordinary streams instead of memory-mapped.

jypma:
The current DiskBuffer implementation leaks a file handle, because it
doesn't unmap the file channel (and the JVM doesn't seem to unmap it
on GC, even though it promises to do so).

This commit rewrites the DiskBuffer using ordinary streams.

Fixes #780.

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/DiskBuffer.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/DiskBuffer.scala
@@ -4,8 +4,8 @@
 
 package akka.stream.alpakka.s3.impl
 
-import java.io.{File, FileOutputStream, RandomAccessFile}
-import java.nio.channels.FileChannel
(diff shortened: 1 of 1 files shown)


ennru:
Thank you.

jypma:
thanks for merging :)

PR #783, opened 2018-02-15
elasticsearch - improvements

### LINKED ISSUE #774: Elasticsearch: no way to query all the indexes at once
It seems there is no way to query not a particular ES index but all the indexes at once.
In the ES client one can construct `org.elasticsearch.action.search.SearchRequest` with no indexes specified and it would mean that all indexes are going to searched.
But in `ElasticsearchSource` index name is a required attribute.

### LINKED ISSUE #775: Elasticsearch: Searching an index without specifying a type is not supported
I have to access an index that has no types defined. 
Currently, this is not possible because I have to specify a type in order to create an ElasticSearchSource.

mbknor:
Improvements:
*  It is now possible to search without specifying typeName - Fixes #775
* you can now search in all indexes by specifying empy indexName - Fixes #774

### DIFF
--- a/elasticsearch/src/test/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSpec.scala
+++ b/elasticsearch/src/test/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSpec.scala
@@ -38,6 +38,7 @@ class ElasticsearchSpec extends WordSpec with Matchers with BeforeAndAfterAll {
   import DefaultJsonProtocol._
 
   case class Book(title: String)
+
   implicit val format: JsonFormat[Book] = jsonFormat1(Book)
   //#define-class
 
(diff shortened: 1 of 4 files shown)


elasticsearch/src/test/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSpec.scala
```diff
@@ -720,11 +721,11 @@ class ElasticsearchSpec extends WordSpec with Matchers with BeforeAndAfterAll {
...
-    "should prefetch next scroll" in {
+    "should let you search without specifying typeName" in {
       val f1 = ElasticsearchSource
         .typed[Book](
           indexName = "source",
-          typeName = "book",
+          typeName = None,
           query = """{"match_all": {}}""",
           settings = ElasticsearchSourceSettings(bufferSize = 5)
         )
```
johanandren: Maybe add a test instead of changing an existing one?

mbknor: I've also renamed the test.

The "old test" was added by me in my previous PR.

I used it to verify that prefetching of next scroll was working, but as you see, it does not assert anything other than that it receives all docs.

So since it did not actually assert anything special about "should prefetch next scroll", I just re-purposed it instead of removing it and adding a new one..

johanandren: Ok, thanks for clarifying!

PR #787, opened 2018-02-18
Cassandra: Support passThrough in Flow

beikern:
* Added 'CassandraFlow' to Cassandra, for both Java and Scala.
* Added tests and updated documentation including examples for both Java and Scala.


beikern:
This allows to passthrough the upstream elements for futher processing  instead of finishing in a Sink.

beikern:
@ennru Done :+1: tell me if anything more needs to be fixed :)

ennru:
Thank you for your contribution! Keep them coming.

PR #796, opened 2018-02-20
Update mongo-scala-driver dependency to 2.2.1 and use generics instead of Document

mikela:
Hi! First PR here, 
1) Update the mongo-scala-driver to the latest version 2.2.1.
2) Mongo: Replace the Document datatype by generics (T) so we can use case classes to represent documents in a collection. Commented [here](https://github.com/akka/alpakka/issues/115#issuecomment-290978042). More info about the feature [here](http://mongodb.github.io/mongo-scala-driver/2.2/getting-started/quick-tour-case-classes/)

Feedback always welcome! 
Thanks

### DIFF
--- a/mongodb/src/test/scala/akka/stream/alpakka/mongodb/MongoSourceSpec.scala
+++ b/mongodb/src/test/scala/akka/stream/alpakka/mongodb/MongoSourceSpec.scala
@@ -10,7 +10,7 @@ import akka.stream.ActorMaterializer
 import akka.stream.alpakka.mongodb.scaladsl.MongoSource
 import akka.stream.scaladsl.{Sink, Source}
 import org.bson.codecs.configuration.CodecRegistries.{fromProviders, fromRegistries}
-import org.mongodb.scala.MongoClient
+import org.mongodb.scala.{MongoClient, MongoCollection}
 import org.mongodb.scala.bson.codecs.DEFAULT_CODEC_REGISTRY
 import org.mongodb.scala.bson.codecs.Macros._
 import org.mongodb.scala.bson.collection.immutable.Document
@@ -28,11 +28,6 @@ class MongoSourceSpec
     with BeforeAndAfterAll
     with MustMatchers {
 
-  // case class and codec for mongodb macros
-  case class Number(_id: Int)
-
-  val codecRegistry = fromRegistries(fromProviders(classOf[Number]), DEFAULT_CODEC_REGISTRY)
-
   //#init-mat
   implicit val system = ActorSystem()
   implicit val mat = ActorMaterializer()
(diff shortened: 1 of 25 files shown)


mikela:
It certainly does, using [classOf](https://github.com/scala/scala/blob/2.11.x/src/library/scala/Predef.scala#L83)
The MongoDB Scala Driver related [doc](http://mongodb.github.io/mongo-scala-driver/2.2/getting-started/quick-tour-case-classes/#configuring-case-classes)

johanandren:
The MongoCollection picks up a type class when created, I think that's fine.

Is it still possible to work with the BSON datatypes or this requires applying a Codec?

mikela:
Codec not required, you can work with Document as usual, [ref](http://mongodb.github.io/mongo-scala-driver/2.2/getting-started/quick-tour/#insert-a-document)

johanandren:
Thanks for clarifying.

vgkowski:
I would add some tests in spec like the one I added in MongoSourceSpec in this partially duplicated PR  #807

mikela:
@vgkowski I've manually merged your test case since the full commit wouldn't do (the new apply method didn't seem necessary but I can be mistaken). Is that OK?

mongodb/src/test/scala/akka/stream/alpakka/mongodb/MongoSourceSpec.scala
```diff
@@ -25,6 +28,11 @@ class MongoSourceSpec
     with BeforeAndAfterAll
     with MustMatchers {
 
+  // case class and codec for mongodb macros
+  case class Number(_id: Int)
+
+  val codecRegistry = fromRegistries(fromProviders(classOf[Number]), DEFAULT_CODEC_REGISTRY)
```
johanandren: Shouldn't this also be shown in the sample?

vgkowski: can you point me to this sample ?

vgkowski:
Let me add some tests for writing this evening

mongodb/src/test/scala/akka/stream/alpakka/mongodb/MongoSourceSpec.scala
mikela: Ah, right. I think we should leave MongoSourceSpec as it was (using Document in all test cases) and create a new MongoSourceWithCodecSpec using the Number case class in all test cases.
This way we can easily refer to this new optional feature in the docs and leave the existing basic one as is. Sounds good @vgkowski ?

mikela: This is the example @jochenschneider is talking about: https://developer.lightbend.com/docs/alpakka/latest/mongodb.html#usage

The docs will show the code inside the #init-connection tags in MongoSourceSpec. Since we've added ".withCodecRegistry(codecRegistry)", readers will be left wondering where is the _codecRegistry_ coming from.

vgkowski: Let's try this ? https://github.com/vgkowski/alpakka/tree/pr796

mikela:
@johanandren any idea why the last commit broke the build only for Scala 2.12? I'm running "sbt +test" locally and they seem to pass fine :thinking: 

@vgkowski Here are some things I've learned that could be useful for you too:

A. (Just a reminder) Don't forget to make sure the tests pass before committing:
1. Open a couple consoles at the alpakka project level.
2. Use one to run "docker-compose up".
3. Use the other one to run "sbt +test".
4. When the tests pass, just ctrl+c to stop 1.

B. Check the documentation is shown correctly by running "sbt docs/local:paradox" and going to "./docs/target/paradox/site/local/mongodb.html".

This and more info can be read at [CONTRIBUTING](https://github.com/akka/alpakka/blob/master/CONTRIBUTING.md) They seem to have removed the "docker-compose up" part but I read it there [before](https://github.com/akka/alpakka/blob/v0.16/CONTRIBUTING.md)

PR #691, opened 2018-01-05
Make JmsSink materialize to Future[Done]

nilsga:
Sometimes it's useful to know when the stream has completed. Materializing `JmsSink` to a `Future` makes this easier.

PR #799, opened 2018-02-20
jms: add flow producer

dwickern:
This allows the JMS producer (`JmsSink`) to also act as a flow, similar to what's offered by the [kafka producer](https://doc.akka.io/docs/akka-stream-kafka/current/producer.html). For example, you can ensure that a message is persisted to the queue before subsequent processing.

This actually eliminates the need for #691 since we can pipe the flow to `Sink.ignore` to get a sink with materialized `Future[Done]`. As a bonus, you get the actual exception instead of a generic `AbruptStageTerminationException`.

`JmsProducer` might be a better name than `JmsSink` at this point. However I didn't want to go and rename things without getting feedback first.

### DIFF
--- a/docs/src/main/paradox/jms.md
+++ b/docs/src/main/paradox/jms.md
@@ -63,7 +63,7 @@ The created @javadoc[ConnectionFactory](javax.jms.ConnectionFactory) is then use
 ## Sending messages to a JMS provider
 
 Use a case class with the subtype of @scaladoc[JmsMessage](akka.stream.alpakka.jms.JmsMessage$) to wrap the messages you want to send and optionally set their properties.
-@java[@scaladoc[JmsSink](akka.stream.alpakka.jms.javadsl.JmsSink$)]@scala[@scaladoc[JmsSink](akka.stream.alpakka.jms.scaladsl.JmsSink$)] contains factory methods to facilitate
+@java[@scaladoc[JmsProducer](akka.stream.alpakka.jms.javadsl.JmsProducer$)]@scala[@scaladoc[JmsProducer](akka.stream.alpakka.jms.scaladsl.JmsProducer$)] contains factory methods to facilitate
 the creation of sinks according to the message type (see below for an example).
 
 
@@ -197,7 +197,7 @@ Java
 
 ## Receiving messages from a JMS provider
 
-@java[@scaladoc[JmsSource](akka.stream.alpakka.jms.javadsl.JmsSource$)]@scala[@scaladoc[JmsSource](akka.stream.alpakka.jms.scaladsl.JmsSource$)] contains factory methods to facilitate
+@java[@scaladoc[JmsConsumer](akka.stream.alpakka.jms.javadsl.JmsConsumer$)]@scala[@scaladoc[JmsConsumer](akka.stream.alpakka.jms.scaladsl.JmsConsumer$)] contains factory methods to facilitate
 the creation of sinks according to the message type (see below for an example).
 
 
@@ -317,7 +317,7 @@ Java
 
 **Notes:**
 
-*  The default `AcknowledgeMode` is `AutoAcknowledge` but can be overridden to custom `AcknowledgeMode`s, even implementation-specific ones by setting the `AcknowledgeMode` in the `JmsSourceSettings` when creating the stream.
+*  The default `AcknowledgeMode` is `AutoAcknowledge` but can be overridden to custom `AcknowledgeMode`s, even implementation-specific ones by setting the `AcknowledgeMode` in the `JmsConsumerSettings` when creating the stream.
 
 ### Receiving @javadoc[javax.jms.Message](javax.jms.Message)s messages from a JMS provider with Client Acknowledgement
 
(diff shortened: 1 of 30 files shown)


jms/src/main/scala/akka/stream/alpakka/jms/JmsSinkStage.scala
```diff
@@ -7,25 +7,21 @@ package akka.stream.alpakka.jms
...
-import akka.Done
-import akka.stream.stage.{GraphStageLogic, GraphStageWithMaterializedValue, InHandler}
 import akka.stream._
+import akka.stream.stage._
 
-import scala.concurrent.{Future, Promise}
-
-final class JmsSinkStage(settings: JmsSinkSettings)
-    extends GraphStageWithMaterializedValue[SinkShape[JmsMessage], Future[Done]] {
+final class JmsSinkStage(settings: JmsSinkSettings) extends GraphStage[FlowShape[JmsMessage, JmsMessage]] {
```
johanandren: This isn't a `JmsSinkStage` anymore.

docs/src/main/paradox/jms.md
```diff
@@ -172,6 +172,29 @@ Scala
 Java
 : @@snip ($alpakka$/jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsConnectorsTest.java) { #create-messages-with-headers }
 
+
+### Sending messages as a Flow
+
+The producer can also act as a flow, in order to publish messages in the middle of stream processing.
+For example, you can ensure that a message is persisted to the queue before subsequent processing.
```
johanandren: "producer" is kafka terminology and I don't think it means anything in general for people using JMS. What about something along the lines of "Sending messages to a queue can also be done in a flow where the elements are passed downstream after they have been persisted allowing for further processing"

dwickern:
Re: naming, JMS also calls them [producer and consumer](https://docs.oracle.com/javaee/7/api/javax/jms/Session.html#createProducer-javax.jms.Destination-). I think `JmsProducer` would be a better name given it can act as both a sink and a flow. If you think that's too breaking we can call it `JmsFlow` as you said.

johanandren:
Ah, never saw that before :) 

What about keeping the existing Sink/Source factories then, maybe annotating with `@deprecated`, and introducing new Producer/Consumer factory objects plus renaming the actual stage?

dwickern:
What about `JmsSourceSettings` / `JmsSinkSettings`? Is there a way to shim those for source compat?

johanandren:
Yeah, that's tricky given that they are case classes. For Scala a type alias + alias-reference to the companion object could do the trick but for Java that's not possible.

ktoso:
Alpakka has the benefit of not being bound by source compat (yet), so I would not worry too much about it. Document it somewhere though such that upgrading is easy?

dwickern:
I added a package object with some aliases. Should be good enough.

jms/src/main/scala/akka/stream/alpakka/jms/javadsl/JmsSink.scala
```diff
@@ -16,66 +16,40 @@ import scala.compat.java8.FutureConverters
...
    * Java API: Creates an [[JmsSink]] for [[JmsMessage]]s
    */
   def create[R <: JmsMessage](
       jmsSinkSettings: JmsProducerSettings
   ): akka.stream.javadsl.Sink[R, CompletionStage[Done]] =
-    akka.stream.alpakka.jms.scaladsl.JmsSink
-      .apply(jmsSinkSettings)
-      .mapMaterializedValue(FutureConverters.toJava)
-      .asJava
+    JmsProducer.create(jmsSinkSettings)
```
johanandren: Rename the parameters as well, looks a bit silly that they just repeat the (old) class name, so maybe just call them `settings`?

jms/src/main/scala/akka/stream/alpakka/jms/JmsProducerStage.scala
```diff
@@ -7,25 +7,21 @@ package akka.stream.alpakka.jms
...
-import akka.Done
-import akka.stream.stage.{GraphStageLogic, GraphStageWithMaterializedValue, InHandler}
 import akka.stream._
+import akka.stream.stage._
 
-import scala.concurrent.{Future, Promise}
-
-final class JmsSinkStage(settings: JmsSinkSettings)
-    extends GraphStageWithMaterializedValue[SinkShape[JmsMessage], Future[Done]] {
+final class JmsProducerStage(settings: JmsProducerSettings) extends GraphStage[FlowShape[JmsMessage, JmsMessage]] {
```
johanandren: Couldn't we use a bounded type for this like you do in the java method, `[T <: JmsMessage]` and have the type flow through nicely so that you get a `Flow[JmsTextMessage, JmsTextMessage, _]` rather than loose the type when you use it?

dwickern: Sure. For my purposes I didn't need the original message any more, so `Unit` would have been good enough.

johanandren: Good, same for the flow factories then. I'd imagine that if you do indeed use it you'd like the type kept.

jms/src/main/scala/akka/stream/alpakka/jms/JmsConsumerStage.scala
```diff
@@ -19,7 +19,7 @@ import scala.concurrent.Future
 import scala.util.control.NonFatal
 import scala.util.{Failure, Success}
 
-final class JmsSourceStage(settings: JmsConsumerSettings)
+final class JmsConsumerStage(settings: JmsConsumerSettings)
```
ennru: Ah, could you make it `private[jms]`, please.

dwickern: For all the graph stage implementations I assume

ennru:
Wow, impressive turn-around. 
When the build succeeds, I'll merge this.

ennru:
Scalaformat wants your attention, please re-format.

ennru:
Thank you for this contribution!

Issue #122, opened 2016-12-19
Add Amazon Kinesis Firehose support


devsprint:
Integrate Amazon Kinesis Firehose such that data from streams can be pushed to S3/Redshift to be used for analytic processes.

utgarda:
@devsprint  I believe it's already there:
https://github.com/akka/alpakka/tree/v0.14/kinesis/src

ennru:
This has been around for a while now. https://doc.akka.io/docs/alpakka/current/kinesis.html#kinesis-firehose-streams

PR #802, opened 2018-02-21
Kinesis Firehose support #122

gregbeech:
This direct port of the relevant Kinesis functionality to use the Kinesis Firehose types for #122. I put it in the `kinesis` project rather than creating a new one as this matches the way AWS ship their own libraries. Similarly the `kinesisfirehose` package is used to match AWS.

It's not exactly an elegant solution to have significant portions of the Kinesis code duplicated with just package and type names changed. However, I spent a good while looking at some way to generalise the code but there isn't really that much that can be shared. The graph logic could be extracted out for the flow stage, but it starts to get a little messy and would have an impact on the existing interface, such as requiring exceptions to be generic and adding more methods to `KinesisFlowSettings` with sensible defaults for additional stream types.

My feeling was that the deduplication internally wasn't worth the (mildly) detrimental effect it had on the external interface, and given the connector doesn't currently support either Video or Analytics streams it's possibly too early to try and generalise the implementation (rule of threes and all that).

Thoughts on this approach are welcome. If people are happy with it I'll squash the commits.

### DIFF
--- a/docs/src/main/paradox/kinesis.md
+++ b/docs/src/main/paradox/kinesis.md
@@ -1,16 +1,12 @@
 # AWS Kinesis Connector
 
-Amazon Kinesis supports collecting, processing, and analyzing video and data streams in real time.
-
-The AWS Kinesis connects to Akka Streams to Kinesis streams.
+The AWS Kinesis connector provides flows for streaming data to and from Kinesis Data streams and to Kinesis Firehose streams.
 
 For more information about Kinesis please visit the [Kinesis documentation](https://aws.amazon.com/documentation/kinesis/).
-
 
 ### Reported issues
 
 [Tagged issues at Github](https://github.com/akka/alpakka/labels/p%3Akinesis)
-
 
 ## Artifacts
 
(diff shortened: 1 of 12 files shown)


ennru:
Hmm, a shame there is no obvious way a abstract about both Kinesis and Firehose.

gregbeech:
You can extract the `GraphStageLogic` so it's shared between the two, and although it's a little messy it's not too awful. The compromises you need to make are that it would change the interfaces a bit in terms of the exceptions thrown on failure, and the `KinesisFlowSettings` class methods would need a bit of a name change to make them less ambiguous as to which Kinesis variant they're for (as settings for one are totally unsuitable for the other).

Happy to put up an additional commit which shows this so that it's easier to make a call between one and the other. I wasn't convinced it actually made things better overall. But additional opinions on this would be welcome.

ennru:
Does anybody else have an opinion, if we're happy with this logic duplication within the same connector?

ennru:
Thank you for this contribution! Please keep an eye on it to see if we can simplify the Kinesis / Kineses Firehose split.

PR #807, opened 2018-02-22
Support Mongodb codec and macros for MongoSource

vgkowski:
This PR allows to use Mongodb codecs and macros in MongoSource to directly read case class objects from Mongodb.
It introduces a new apply function with a classtag parameter and corresponding test.

For reference on mongodb and case class integration:
[mongodb <-> case class](https://github.com/mongodb/mongo-scala-driver/blob/master/docs/reference/content/getting-started/quick-tour-case-classes.md)


johanandren:
There is a recent PR dealing with that open already #796, is there any reason that is not enough?

vgkowski:
you are right it should be enough, maybe some tests missing ?

johanandren:
Please review that one and add feedback as needed, thanks!

PR #809, opened 2018-02-22
Handles failure on disconnect for MQTT

### LINKED ISSUE #805: MqttFlowStage does not tolerate already having been disconnected
It appears as though the `postStop` method of `MqttFlowStage` doesn't tolerate already being disconnected, which then causes an error:

```
[ERROR] [02/22/2018 15:18:33.374] [lora-server-akka.actor.default-dispatcher-5] [akka://lora-server/user/StreamSupervisor-0/flow-13-0-ignoreSink] Error during postStop in [akka.stream.alpakka.mqtt.MqttFlowStage@2cae551d]: Client is disconnected
Client is disconnected (32101)
	at org.eclipse.paho.client.mqttv3.internal.ExceptionHelper.createMqttException(ExceptionHelper.java:31)
	at org.eclipse.paho.client.mqttv3.internal.ClientComms.disconnect(ClientComms.java:489)
	at org.eclipse.paho.client.mqttv3.MqttAsyncClient.disconnect(MqttAsyncClient.java:826)
	at akka.stream.alpakka.mqtt.MqttFlowStage$$anon$2.postStop(MqttFlowStage.scala:244)
```

This is a particular problem because it occurs when an MQTT client loses connectivity with its broker - this happens often in my case.

I suggest that the following line in `postStop` is surrounded by a try/catch for `MqttException`:

```
mqttClient.disconnect(
```

huntc:
Fixes #805 

Also, I think I've learnt that the `automaticReconnect` setting doesn't appear to re-connect, and that I must use `RestartFlow.withBackoff` and friends to reconnect - would this be correct?

### DIFF
--- a/mqtt/src/main/scala/akka/stream/alpakka/mqtt/MqttFlowStage.scala
+++ b/mqtt/src/main/scala/akka/stream/alpakka/mqtt/MqttFlowStage.scala
@@ -21,6 +21,7 @@ import org.eclipse.paho.client.mqttv3.{
   MqttAsyncClient,
   MqttCallbackExtended,
   MqttConnectOptions,
+  MqttException,
   MqttMessage => PahoMqttMessage
 }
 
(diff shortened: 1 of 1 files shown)


huntc:
The CI build fails on the following:

```
2018-02-22 10:40:00,204 ERROR [MqttSourceSpec-akka.actor.default-dispatcher-7] [akka.io.TcpListener]  Bind failed for TCP channel on endpoint [localhost/127.0.0.1:1337]
java.net.BindException: Address already in use
```

I'm unsure why my changes would affect this. Is this just a flaky test?

huntc:
An issue to remind us: https://github.com/akka/alpakka/issues/812

Issue #812, opened 2018-02-23
MQTT reconnect doc and setting


huntc:
An issue to remind us that we should remove the auto reconnect option from the MQTT settings and provide an example of using the retry/back off stages.

huntc:
Closing given that #814 addresses this problem

PR #814, opened 2018-02-25
Automatic reconnect was not working for MQTT

huntc:
This PR fixes auto-reconnection and counteracts the need for #812 by avoiding the failing of the stage when we know that an automatic reconnect is planned.

The changes here are difficult to test without stopping/starting the docker container. However, I've tested locally by observing that I can stop the MQTT server and then start it again, both with and without messages pending upstream. All appears well.

All existing tests are working of course and I've engineered my changes to be quite benign without the automatic reconnection feature being enabled.

### DIFF
--- a/mqtt/src/main/scala/akka/stream/alpakka/mqtt/MqttFlowStage.scala
+++ b/mqtt/src/main/scala/akka/stream/alpakka/mqtt/MqttFlowStage.scala
@@ -27,6 +27,7 @@ import org.eclipse.paho.client.mqttv3.{
 
 import scala.collection.mutable
 import scala.concurrent.{Future, Promise}
+import scala.util.control.NonFatal
 import scala.util.{Failure, Success, Try}
 
 final class MqttFlowStage(sourceSettings: MqttSourceSettings,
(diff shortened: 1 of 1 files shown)


mqtt/src/main/scala/akka/stream/alpakka/mqtt/MqttFlowStage.scala
```diff
@@ -59,9 +61,9 @@ final class MqttFlowStage(sourceSettings: MqttSourceSettings,
       private def onConnect =
         getAsyncCallback[IMqttAsyncClient]((client: IMqttAsyncClient) => {
           if (manualAcks) client.setManualAcks(true)
-          val (topics, qos) = sourceSettings.subscriptions.unzip
+          val (topics, qoses) = sourceSettings.subscriptions.unzip
```
huntc: This was hiding the outer scope - not a biggy, but I thought I'd address it.

mqtt/src/main/scala/akka/stream/alpakka/mqtt/MqttFlowStage.scala
```diff
@@ -90,7 +92,7 @@ final class MqttFlowStage(sourceSettings: MqttSourceSettings,
       }
 
       private val onPublished = getAsyncCallback[Try[IMqttToken]] {
-        case Success(_) => pull(in)
+        case Success(_) => if (!hasBeenPulled(in)) pull(in)
```
huntc: An additional check as I noticed a scenario where we had pulled already given auto-reconnect.

huntc:
Flaky test again? The test certainly worked locally...

raboof:
Seems like it, filed #817

huntc:
OK to merge?

PR #819, opened 2018-03-01
Add Java module names

### LINKED ISSUE #769: Add automatic module names (for JDK9) to all Alpakka modules
In order to be usable as JDK9 modules, Alpakka modules should add module names.
We have a mini helper for it in akka: https://github.com/akka/akka/blob/master/build.sbt#L56 

Impl: https://github.com/akka/akka/blob/master/project/AutomaticModuleName.scala 

Though be aware that it seems to break when combined with the osgi plugin?
https://github.com/akka/akka/issues/24520

Once https://github.com/akka/akka/issues/24520 is fixed, apply the module names to alpakka

ennru:
For some Alpakka modules the Java module name is not straightforward.
* I prefixed connectors to AWS technologies with `aws`. (Which should have been added to the Alpakka module names to begin with, IMHO.)
* Others use the main package name approach.

Fixes #769

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -52,22 +52,24 @@ lazy val alpakka = project
       """.stripMargin
   )
 
-lazy val amqp = alpakkaProject("amqp", Dependencies.Amqp)
+lazy val amqp = alpakkaProject("amqp", "amqp", Dependencies.Amqp)
 
 lazy val awslambda = alpakkaProject("awslambda",
+                                    "aws.lambda",
                                     Dependencies.AwsLambda,
                                     // For mockito https://github.com/akka/alpakka/issues/390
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -52,32 +52,35 @@ lazy val alpakka = project
       """.stripMargin
   )
 
-lazy val amqp = alpakkaProject("amqp", Dependencies.Amqp)
+lazy val amqp = alpakkaProject("amqp", "amqp", Dependencies.Amqp)
 
 lazy val awslambda = alpakkaProject("awslambda",
+                                    "aws.lambda",
```
ktoso: good

PR #821, opened 2018-03-02
Reset pendingAck variable if session has been aborted/shutdown.

### LINKED ISSUE #820: JmsConsumer.ackSource fails to shutdown (or abort) when messages are not acknowledged
To be more specific it appears that if the number of messages consumed by the `JmsConsumer.ackSource` and not acknowledged is greater than the `bufferSize` the stream never completes when the `KillSwitch` is used to `shutdown` or `abort`.
An easy way to see this is to go to the `JmsConnectorsSpec` and comment out `env.acknowledge()` in the `lose some elements when aborting a stream` test. The `streamDone` `Future` never completes.

I can provide a condensed example if this is not clear, I have observed streams completing where the number of messages not acknowledged is less than or slightly over the `bufferSize`, I guess the problem occurs when backpressure kicks in but I don't have any greater insights yet.

matthedude:
Closes #820 

I believe when the `ackSource` is shutdown or aborted the `listenerStopped` variable is never set to `true` (this is probably wrong but I'm not sure how to make it be set). It would be ok but the `ackQueue` is emptied and the number of `pendingAck`s not reset leading to an indefinite `take` operation on the `ackQueue`. By resetting `pendingAck` we avoid trying to process any more acks and the stream completes.

### DIFF
--- a/jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
@@ -134,7 +134,13 @@ private[jms] class JmsAckSession(override val connection: jms.Connection,
 
   def ack(message: jms.Message): Unit = ackQueue.put(message.acknowledge _)
 
+  override def closeSession(): Unit = {
+    pendingAck = 0
+    session.close()
+  }
+
   override def abortSession(): Unit = {
+    pendingAck = 0
     ackQueue.put(() => throw new java.lang.IllegalStateException("Shutdown"))
     session.close()
   }


lightbend-cla-validator:
Hi @matthedude,

ennru:
@matthedude Thank you for investigating this issue.
Do I understand your closing of the PR correctly that you can't agree to the Lightbend CLA?

matthedude:
No just was a bit premature saying I had fixed it! Will try and sort next week

ennru:
Ah, OK. Great. Looking forward to it.

Issue #833, opened 2018-03-12
Slick/JDBC: create SlickSession from DatabaseConfig instance


ennru:
Make it possible to share Slick's `DatabaseConfig`, when Slick is used directly and via Alpakka.

ennru:
Fixed by #835

PR #835, opened 2018-03-13
Slick: create SlickSession from DatabaseConfig instance (#833)

tuleism:
Solve #833.


ennru:
Thank you for fixing this right after I noted this missing functionality.
Keep going!

Issue #635, opened 2017-12-07
[mqtt] report connection error


aitorhh:
Hi,

I am really missing the error reporting in the mqtt connector. In my application, I have no clue if the connector is in "connected" state or not. Is there any way that the streams stop working and sends an exception to its supervisor? For example, I current handle other exception in my stream with, by defining the following:

```
  val decider: Supervision.Decider = {
    case e =>
      logger.error(s"Error processing message: $e. Stopping")
      Supervision.Stop
  }
  implicit val system = ActorSystem()
  implicit val materializer = ActorMaterializer(
    ActorMaterializerSettings(system)
      .withSupervisionStrategy(decider)
    )
```

Currently, I am completely blind on error cases with mqtt. I hope there is already a solution that I haven't figured it out yet.

Thanks!

juanjoDiaz:
Hi Aitor,
Can you add code example of what you would expect?
AFAIK the MQTT connector fails its stage if the connection drops.

juanjoDiaz:
The error missing on two consumers with the same clientId is an issues with most MQTT libraries (they don't report when a client is kicked out). I'm following up the issue in https://github.com/eclipse/paho.mqtt.java/issues/461 and I'll fix it in alpakka as soon as it is fixed in Paho.

Automatic reconnect is not enabled by default. It has to be enabled.

Otherwise, maybe I'm misunderstanding but I don't think that an exception should be thrown from the stream. It just fails the stage and all the stages upstream or downstream are failed. Are you expecting something else?

aitorhh:
Thanks for the information.

When you say that the stream fails, how can the user/system be notified of
the failure? That's actually what I'm after. Seems that I am missing
something that already exists.

In the previous examples, the application just remains running without any
message, log or anything that indicates the error.

aitorhh:
Great to see these features added. I close the issue. Thanks!

PR #849, opened 2018-03-19
Ensure that an exception is materialized on MQTT connection error

juanjoDiaz:
Similar to what #848 does for AMQP

Might close some of the issues raised in #635

### DIFF
--- a/mqtt/src/main/scala/akka/stream/alpakka/mqtt/MqttFlowStage.scala
+++ b/mqtt/src/main/scala/akka/stream/alpakka/mqtt/MqttFlowStage.scala
@@ -70,11 +70,7 @@ final class MqttFlowStage(sourceSettings: MqttSourceSettings,
           }
         })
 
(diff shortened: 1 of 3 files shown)


huntc:
Without reading through the referenced issues, I'm curious. Have you observed an actual issue? I ask because I'm quite able to manage MQTT connection management. For example:

```
upstream
  .via(
    RestartFlow
      .withBackoff(
        minBackoff = minBackoff,
        maxBackoff = maxBackoff,
        randomFactor = randomFactor
      )(() => mqttFlow(mqttTopicTypesUpdated.mqttCommandTopic))
  )
  .via(downstream)
```

juanjoDiaz:
You are right! I digged into the GraphInterpreter and it already catches exceptions in the preStart method and fail the stage automatically.

So actually what this PR actually solve is that there were 2 error conditions that fail the stage but never materialize the exception in the materialized value.

PR #837, opened 2018-03-14
Spill the beans on Paho's connection management

huntc:
Unfortunately the Paho library [doesn't support the management of initial connects failing, by design](https://github.com/eclipse/paho.mqtt.golang/issues/77). This PR warns the reader.

PR #850, opened 2018-03-19
Add MQTT automatic reconnect test

juanjoDiaz:
#635 reported that MQTT automatic reconnect doesn't work.

It does work, but is not documented. The trick is that it needs `cleanSession` set to true.
This PR add a test proving it.

Whenever I have some time I'd also update the documentation.


huntc:
Nice. I also found that it does indeed work, but it is quite useless to me. The Paho library distinguishes between initial connection and subsequent connection. Reconnect only kicks in when an initial connection has been established. [I recently added a note in the doc](https://github.com/akka/alpakka/pull/837) regarding this.

juanjoDiaz:
I definitely don't like it.
It might be a good opinion to remove completely the automatic reconnect from the MQTT and the AMQP connectors and leave the Restart stages as the only alternative.

Wdyt?

ennru:
Seems like a reasonable idea. We could add the Restart code in the examples to free the users from finding out themselves.

PR #851, opened 2018-03-19
Prevent the AMQP cached connection provider from entering an invalid …

juanjoDiaz:
…state

Just noticed that, if there is an exception getting the connection, the cached connection provider gets permanently in the `Connecting` state. Any any further attempt of getting a connection will result in an endless loop. This should fix that.


ennru:
Thank you for your efforts to get this stuff fixed!

Issue #514, opened 2017-10-08
STOMP support


ktoso:
Being able to expose STOMP streams  https://stomp.github.io/stomp-specification-1.2.html

I've heard this request and agree could be pretty useful, but we can't dedicate someone to work on it now - if someone would like to pick it up that would be excellent :)

ktoso:
We could pay special attention to ack modes: https://stomp.github.io/stomp-specification-1.2.html#SUBSCRIBE_ack_Header

This seems to be something camel is not able to expose

gosubpl:
I can try to take a stab at it, but cannot promise I will be done before New Year. Would that be fine for you?

And yes, ACK modes are important for STOMP :)

ktoso:
Sure, please give it a shot :)

nachinius:
@gosubpl, what's the state of this?

gosubpl:
@nachinius haven't yet started. Quite busy recently. If you want to give it a shot - please do, I can help reviewing.

nachinius:
Ok, let's see if we are in the same page.

I can imagine several ways to use stomp in akka streams, the two most basic I think of are StompSink and StompSource. The descriptions of what they do are:
- StompClientSink. This is a sink that acts as a stomp client _producer_ that connected to a given stomp server, it will deliver the receiving messages to such server.
- StomClientSource. This is a source that acts as a stomp client _consumer_ that connected to a given stomp server, it subscribes to a _topic_ and receives messages.

On start both StompSink and StompSource, requires a Connection to a stomp server. The StompSource also requires, at start, a _topic_ to subscribe to.

### backpropagation
In the _consumer_ (source), we can use stomp's ACK to acknowledge processing, as a way to implement backpropagating. For the _producer_ (sink) we can request stomp header RECEIPT to achieve the same purpose. Both of these, are set by our end, so they could be controlled by settings. I suggest ON by default.

### heartbeats
I would leave this, for a following iteration, or even after implementing the stomp server endpoints (see below).

### protocol
For the protocol implemenation, I found vertx-stomp which is an eclipse foundation project, licensed under both eclipse license and apache license. No need to repeat the protocol here. It's the only one I found to be complete and able to handle 1.2.

### second version, StompServerSource and StompServerSink
[...]

PR #856, opened 2018-03-20
Add STOMP connectors: StompClientSource and StompClientSink #514

nachinius:
Add connectors to a STOMP server as a stream client or source.

Ref #514 

Before finishing the documentation and test for java, I wish this to be reviewed.

As stated in https://github.com/akka/alpakka/issues/514#issuecomment-368509908 here we have
the
- StompClientSource
- StompClientSink
and both support backpropagation

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -263,6 +263,13 @@ object Dependencies {
     )
   )
 
+  val Stomp = Seq(
+    libraryDependencies ++= Seq(
+      // https://mvnrepository.com/artifact/io.vertx/vertx-stomp
+      "io.vertx" % "vertx-stomp" % "3.5.1" // ApacheV2/EPL1
+    )
+  )
+
   val UnixDomainSocket = Seq(
     libraryDependencies ++= Seq(
       "com.github.jnr" % "jnr-unixsocket" % "0.18" // BSD/ApacheV2/CPL/MIT as per https://github.com/akka/alpakka/issues/620#issuecomment-348727265
--- a/stomp/src/main/scala/akka/stream/alpakka/stomp/client/SinkStage.scala
+++ b/stomp/src/main/scala/akka/stream/alpakka/stomp/client/SinkStage.scala
@@ -62,7 +62,10 @@ final class SinkStage(settings: ConnectorSettings)
               pull(in)
             }
             if (settings.withAck) {
-              connection.send(vertxFrame, continue.invoke(_))
+              import VertxStompConversions._
+              connection.send(vertxFrame, { frame: Frame =>
+                continue.invoke(frame)
+              })
             } else {
               connection.send(vertxFrame)
               pull(in)
--- a/stomp/src/main/scala/akka/stream/alpakka/stomp/client/SinkStage.scala
+++ b/stomp/src/main/scala/akka/stream/alpakka/stomp/client/SinkStage.scala
@@ -62,10 +62,7 @@ final class SinkStage(settings: ConnectorSettings)
               pull(in)
             }
             if (settings.withAck) {
-              connection.send(vertxFrame, (frame: Frame) => {
-                // server receive the message, continue pulling
-                continue.invoke(frame)
-              })
+              connection.send(vertxFrame, continue.invoke(_))
             } else {
               connection.send(vertxFrame)
               pull(in)
--- a/stomp/src/main/scala/akka/stream/alpakka/stomp/client/SinkStage.scala
+++ b/stomp/src/main/scala/akka/stream/alpakka/stomp/client/SinkStage.scala
@@ -1,0 +1,94 @@
+/*
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.stomp.client
+
+import akka.Done
+import akka.stream.stage.{GraphStageLogic, GraphStageWithMaterializedValue, InHandler}
+import akka.stream.{ActorAttributes, Attributes, Inlet, SinkShape}
+import io.vertx.ext.stomp.{Frame, StompClientConnection}
+
+import scala.concurrent.{Future, Promise}
+
+/**
+ * Connects to a STOMP server upon materialization and sends incoming messages to the server.
+ * Each materialized sink will create one connection to the server.
+ */
+final class SinkStage(settings: ConnectorSettings)
+    extends GraphStageWithMaterializedValue[SinkShape[SendingFrame], Future[Done]] {
+  stage =>
+
+  override def shape: SinkShape[SendingFrame] = SinkShape.of(in)
+
+  override def toString: String = "StompClientSink"
+
+  override protected def initialAttributes: Attributes = SinkStage.defaultAttributes
+
+  val in = Inlet[SendingFrame]("StompClientSink.in")
+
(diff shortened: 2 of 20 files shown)


project/Dependencies.scala
```diff
@@ -263,6 +263,13 @@ object Dependencies {
     )
   )
 
+  val Stomp = Seq(
+    libraryDependencies ++= Seq(
+      // https://mvnrepository.com/artifact/io.vertx/vertx-stomp
+      "io.vertx" % "vertx-stomp" % "3.5.1" // ApacheV2/EPL1
```
huntc: Do we have a sense of the dependencies here ie is it bringing in many?

huntc: What’s the thread usage of this dependency in terms of the number of threads required?

huntc: I’ve implemented STOMP before - the protocol is easy. Should we even need this dependency?

stomp/src/main/scala/akka/stream/alpakka/stomp/client/SinkStage.scala
```diff
@@ -0,0 +1,94 @@
...
+import akka.stream.{ActorAttributes, Attributes, Inlet, SinkShape}
+import io.vertx.ext.stomp.{Frame, StompClientConnection}
+
+import scala.concurrent.{Future, Promise}
+
+/**
+ * Connects to a STOMP server upon materialization and sends incoming messages to the server.
+ * Each materialized sink will create one connection to the server.
+ */
+final class SinkStage(settings: ConnectorSettings)
```
huntc: At a min, we should have a Source and Flow. Sink is a convenience on Flow. Please see the guidelines for more info.

huntc: Flows should also permit an opaque element to be carried through them. I think this is also in the guidelines.

project/Dependencies.scala
nachinius: > Do we have a sense of the dependencies here ie is it bringing in many?

It brings stuff from vertx like vertx-core

> What’s the thread usage of this dependency in terms of the number of threads required?

How would you check it?

> I’ve implemented STOMP before - the protocol is easy. Should we even need this dependency?

Good! Congrats! is it in scala or java? will you open source it?
This alpakka connector basically sits on top of the particular stomp protocol. Our main objective is able to expose Stomp streams. Nevertheless, we could implement the stomp protocol in following versions of this connector to drop the depencency on vertx-stomp. The advantage of vertx-stomp is that is complete, and can handle out of the box several versions of stomp and all features.

In short: eventually we don't need this dependency, but for the moment works.

stomp/src/main/scala/akka/stream/alpakka/stomp/client/SinkStage.scala
nachinius: @huntc we do have a Source in SourceStage.scala. 
Regarding the Flow, at this stage, it's not clear to me what do you mean. Could you point me the place in the guidelines that talk about it, since I'm unable to find it. Furthermore, and an example of a flow with the opaque element. Neither I was able to find it in the guidelines. Thanks

project/Dependencies.scala
huntc: I have no code for you unfortunately.

Profiling this would be great to get a sense of resource usage along with thread usage.

stomp/src/main/scala/akka/stream/alpakka/stomp/client/SinkStage.scala
huntc: Here are the guidelines I referenced: https://github.com/akka/alpakka/blob/master/contributor-advice.md

nachinius: Yes, I had read those. I can't find what you mention. Nevertheless, the basic blocks to construct anything, AFAIU is a Source and a Sink. Later we could add flows, that connects those sources and sink, depending in the most common use cases. 

Anyways, what kind of Flow are you imaging? A flow in a stream, that has a side effect of sending the whatever is being passed to a stomp server? That is what you meant by opaque element?

project/Dependencies.scala
nachinius: Profiling this is not in my tool belt yet. For the sake of awareness, could you point me in the direction on where to start (blog, book, etc)? thks

stomp/src/main/scala/akka/stream/alpakka/stomp/client/SinkStage.scala
huntc: Hopefully, these links will help:

https://github.com/akka/alpakka/blob/master/contributor-advice.md#public-factory-methods
https://github.com/akka/alpakka/blob/master/contributor-advice.md#flows

You generally construct sinks from flows:

https://github.com/akka/alpakka/blob/master/contributor-advice.md#keep-the-code-dry

Some examples, also illustrating the passing through of an element:

```
  /**
    * Append a command request to a queue
    * @tparam A The type of data to carry through
    * @return a flow that takes a command and returns with an acknowledgement
    */
  def flow[A]: Flow[CommandRequest[A], CommandReply[A], NotUsed]

  /**
    * Append a command request to a queue via a sink for convenience
    * @tparam A The type of data to carry through
    * @return a sink that takes a command and returns with an acknowledgement
    */
  def sink[A]: Sink[CommandRequest[A], Future[Done]] =
    flow.toMat(Sink.ignore)(Keep.right)
```

project/Dependencies.scala
huntc: Perhaps start with [VisualVM](https://docs.oracle.com/javase/8/docs/technotes/guides/visualvm/profiler.html).

ennru:
Thank you for this effort.
Just as @huntc I'm a bit reluctant to pull in Vert.x as a dependency to implement the STOMP protocol.

stomp/src/main/scala/akka/stream/alpakka/stomp/client/SinkStage.scala
nachinius: Oh, I see what you say. For me, the example is key. I had already read the guidelines links a month ago and didn't get this idea. The example is crystal, can it be added to the contributor-advice? (is there any copyright restriction on the example you wrote?)

nachinius: I also see the approach you propose is cleaner and more powerful. I would love to have hear this when **I outlined the shape and plan for this contribution on feb 26th** ! A month ago! (https://github.com/akka/alpakka/issues/514#issuecomment-368509908) 

For the moment I'm reluctant to switch approaches. Since this works for several use cases that I find useful.

ennru:
I learned a bit more about vert.x and understand now that only a few data structures support Reactive Streams, not vert.x as such. This makes integrating with their connectors harder than I hoped it to be. 
What we'd need is to connect to the vert.x event bus to a bi-directional flow. I'll give it some more time in a while.

huntc:
...or just avoid vertx for stomp?

huntc:
Any further thoughts on this PR @ennru ?

ennru:
Sorry, @nachinius if we didn't communicate well enough before you put the effort in implementing a STOMP connector based on Vertx. As it seems they have a good implementation of the protocol itself and there are few other implementations available so it would be great to re-use that part. 
As said earlier, I don't like the idea to pull in a lot of Vertx and run the Vertx event loop threads within an Akka Streams connector.

Closing this now.

Issue #853, opened 2018-03-20
S3 Connector - ContentType header not present in ObjectMetadata


chetanmeh:
S3 returns the `Content-Type` header as part of object download response and its expected that this should be accessible from the `ObjectMetadata#contentType`. However as akka-http [drops this header][1] it is not included in the headers passed to `ObjectMetadata`.

Probably `S3Stream#download` should extract them from the `ResponseEntity` and add them to the headers passed to `ObjectMetadata`.

[1]: https://doc.akka.io/docs/akka/2.4.11/scala/http/common/http-model.html#HTTP_Headers

chetanmeh:
Would try create a testcase and fix PR for this issue

ennru:
Fixed with #858

PR #858, opened 2018-03-21
AWS S3: Expose Content-Type as part of ObjectMetaData

chetanmeh:
Fix for #853

To expose the `Content-Type` I had to implement a `CustomContentTypeHeader`  as akka http [by design](https://github.com/akka/akka/issues/16523) does not provide a way to construct this header. 

Further `akka.http.scaladsl.model.ContentType` is a sealed trait so its not possible to have an implementation which extends both `ContentType` and `HttpHeader`. So the logic in `ObjectMetadata` also had to be changed to explicitly match header by name. Not sure if there is a better way possible to do this

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
@@ -13,7 +13,7 @@ import akka.{Done, NotUsed}
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.model.StatusCodes.{NoContent, NotFound, OK}
-import akka.http.scaladsl.model.headers.{ByteRange, CustomHeader, `Content-Length`}
+import akka.http.scaladsl.model.headers.{`Content-Length`, ByteRange, CustomHeader}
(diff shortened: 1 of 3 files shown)


lightbend-cla-validator:
Hi @chetanmeh,

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -221,6 +220,24 @@ private[alpakka] final class S3Stream(settings: S3Settings)(implicit system: Act
...
+    ObjectMetadata(
+      headers ++
+      Seq(
+        `Content-Length`(entity.contentLengthOption.getOrElse(0)),
+        CustomContentTypeHeader(entity.contentType)
+      )
+    )
+
+  //`Content-Type` header is by design not accessible as header. So need to have a custom
+  //header implementation to expose that
```
raboof: Jup. I can see how it is reasonable to construct your own Content-Type header in this case... we might want to introduce a nicer escape hatch for that at some point - though still making it evident that for 'typical' situations you should not construct it yourself.

ennru:
Thank you for this valuable improvement!

PR #862, opened 2018-03-23
JMS: Add support for durable subscriptions

liff:
Support durable topic subscriptions by adding `withTopic(name, subscriptionName)` to `JmsConsumerSettings`.

Documentation is missing but I'll certainly add it if the change looks good otherwise.

### DIFF
--- a/docs/src/main/paradox/jms.md
+++ b/docs/src/main/paradox/jms.md
@@ -472,6 +472,27 @@ Such sink and source can be started the same way as in the previous example.
 * Explicit acknowledgement sources and transactional sources work with topics the same way they work with queues.
 * **DO NOT** set the `sessionCount` greater than 1 for topics. Doing so will result in duplicate messages being delivered. Each topic message is delivered to each JMS session and all the messages feed to the same `Source`. JMS 2.0 created shared consumers to solve this problem and multiple sessions without duplication may be supported in the future.
 
+### Using a durable Topic subscription with a JMS provider
+
+Using a durable topic subscription works mostly like a normal topic source.
+
+You need to specify a client ID for the connection factory you use to create the consumer with, however:
+ 
+Scala
+: @@snip ($alpakka$/jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsConnectorsSpec.scala) { #create-connection-factory-with-client-id }
+
+Java
+: @@snip ($alpakka$/jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsConnectorsTest.java) { #create-connection-factory-with-client-id }
+
+In addition, use `withDurableTopic` and specify a name for the durable subscriber:
+
+Scala
+: @@snip ($alpakka$/jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsConnectorsSpec.scala) { #create-durable-topic-source }
+
+Java
+: @@snip ($alpakka$/jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsConnectorsTest.java) { #create-durable-topic-source }
+
+
 ### Running the example code
 
 The code in this guide is part of runnable tests of this project. You are welcome to edit the code and run it in sbt.
--- a/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
@@ -33,7 +33,7 @@ sealed trait JmsSettings {
 }
 
 sealed trait Destination
-final case class Topic(name: String) extends Destination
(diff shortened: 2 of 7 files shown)


ennru:
Looks very useful.
Would be good to split `Topic` into two types instead of the optional subscriber name.
The name `destination` seems a bit overused - can you come up with something else?

liff:
Added new `DurableTopic` type.

Open for suggestions about naming of `destination` :)

ennru:
What about calling it `settingsDestination`,  in `JmsSession` and sub-classes?

ennru:
... and add something to the documentation, please.

liff:
I don't think the test failure is related to this change. The “publish and consume JMS text messages through a queue with client ack” test fails pretty consistently when i run it locally with `testOnly`.

Curiosly the failure disappears if I add a `.log` stage before `.take` [here](https://github.com/akka/alpakka/blob/master/jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsConnectorsSpec.scala#L375).

I usually get this failure, though:

> javax.jms.IllegalStateException, with message: The Session is closed. (JmsConnectorsSpec.scala:385)

It seems like the session sometimes gets closed before all messages have been delivered and/or acked.

ennru:
Check out #867 which proposes a generic way to create JMS destinations. It might be possible to implement your need via it.

liff:
Looks nice but supporting durable subscriptions need a mechanism to also allow custom `MessageConsumer` creation since we need to do `session.createDurableSubscriber(..)` with the provided `Destination`.

russellyou:
Hi, I am creator of #867, I have updated my PR. 

Destination is expanded like
sealed trait Destination {
  val name: String
  def create(name: String): (jms.Session) => jms.Destination
}

This means you can override create method with your session.createDurableSubscriber(..) . 

Let me know your thought please.

liff:
But [`createDurableSubscriber`](https://docs.oracle.com/javaee/7/api/javax/jms/Session.html#createDurableSubscriber-javax.jms.Topic-java.lang.String-) returns a [`MessageConsumer`](https://docs.oracle.com/javaee/7/api/javax/jms/MessageConsumer.html), not a [`Destination`](https://docs.oracle.com/javaee/7/api/javax/jms/Destination.html), right?

So for a durable subscription, the [creation of a consumer](https://github.com/akka/alpakka/blob/master/jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala#L115) also needs to be customized somehow.

russellyou:
Sorry for missing that. You are right. 

The issue is we try to do different thing according to Destination type. If we agree with the proposed approach in #867, we can also put createConsumer function in Destination to avoid complex pattern matching in createConsumer of JmsConnector.scala

liff:
Sounds good to me. To be continued on #867; I'll close this PR.

ennru:
I assume you'd like to propose a new solution to this now that #867 is merged to master.

liff:
Sure, I'll get it going.

liff:
Well, looking at it now, it seems like it'd be actually easier if `Destination` had:

```scala
val consumer: jms.Session => jms.MessageConsumer
val producer: jms.Session => jms.MessageProducer
```

instead of `=> jms.Destination`, which is a rather intermediate thing after all.

If we keep `create: jms.Session => jms.Destination` in `Destination` then the above functions also need `jms.Session` as an additional parameter.

@russellyou what do you think?

russellyou:
If we replace the session => destination with consumer/producer, I guess we have to change lots of existing code. 

This PR looks like only need consumer, not producer , right?  I would rather add val consumer: jms.Session => jms.MessageConsumer for now. 

Just my suggestion:


sealed trait Destination {
 val name: String
 val create: (jms.Session) => jms.Destination
 val consumer: (jms.Session) => jms.MessageConsumer 
}
final case class Topic(override val name: String, create = session =>session.createTopic(name), consumer = session => session.createConsumer(session.createTopic(name)) ) extends Destination

russellyou:
If we replace the session => destination with consumer/producer, I guess we have to change lots of existing code. 

This PR looks like only need consumer, not producer , right?  I would rather add val consumer: jms.Session => jms.MessageConsumer for now. 

Just my suggestion:




`
sealed trait Destination {
 val name: String
 val create: (jms.Session) => jms.Destination
 val consumer: (jms.Session) => jms.MessageConsumer 
}
final case class Topic(override val name: String, create = session =>session.createTopic(name), consumer = session => session.createConsumer(session.createTopic(name)) ) extends Destination 

`

ennru:
Please move this discussion to a new issue.

jugalps:
I am sort of using similar implementation for durable topics, it would be good to have this PR merged. 
@liff - Is there a new issue created for this feature?

Issue #860, opened 2018-03-22
S3Connector: support for multipart copy object


sfali:
Hi There,

I am looking for support for multipart copy object , I can help implementing this functionality, but need some guidelines how to go about it, looks to me this is slightly different then  multi part upload.

Regards,

Syed

Wizzy-wooz:
+1 - that would be nice to have such functionality

PR #868, opened 2018-03-28
Support for multi part copy in S3, #860

sfali:
#860


s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3Client.scala
```diff
@@ -664,6 +664,213 @@ final class S3Client(s3Settings: S3Settings, system: ActorSystem, mat: Materiali
...
+   * @param sourceBucket the source s3 bucket name
+   * @param sourceKey the source s3 key
+   * @param targetBucket the target s3 bucket name
+   * @param targetKey the target s3 key
+   * @param contentType an optional [[akka.http.javadsl.model.ContentType ContentType]]
+   * @param s3Headers any headers you want to add
+   * @param sse the server side encryption to use
+   * @return a [[java.util.concurrent.CompletionStage CompletionStage]] containing the [[akka.stream.alpakka.s3.javadsl.MultipartUploadResult MultipartUploadResult]] of the uploaded S3 Object
+   */
+  def multipartCopy(sourceBucket: String,
```
ennru: Is it useful to have this many overloads? Users may have a hard time to pick the best overload for their purpose.
I see `putObject` does the same.

docs/src/main/paradox/s3.md
```diff
@@ -85,6 +85,14 @@ Scala
 Java
 : @@snip ($alpakka$/s3/src/test/java/akka/stream/alpakka/s3/javadsl/S3ClientTest.java) { #list-bucket }
 
+### Copy upload (multi part)
+
```
ennru: Please describe what the operation does.

s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3Client.scala
sfali: I wasn't sure whether to add that many overloaded methods or, I saw `PutObject` is having overloaded methods so decided to add overloaded, I figured since  `Java ` doesn't have default parameters that might be reason overloaded methods were added in `PutObject`

ennru: Maybe add something to the method names, so they can more easily be distinguished.
Or make sure good default values are available in the Java API and show those in the example code.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Request.scala
```diff
@@ -15,3 +15,5 @@ case object PutObject extends S3Request
 case object InitiateMultipartUpload extends S3Request
 
 case object UploadPart extends S3Request
+
+case object CopyPart extends S3Request
```
ennru: These classes should become `private` as they are internal only.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -231,6 +237,36 @@ private[alpakka] final class S3Stream(settings: S3Settings)(implicit system: Act
...
+      chunkingParallelism: Int = 4
+  ): RunnableGraph[Future[CompleteMultipartUploadResult]] = {
+    import mat.executionContext
+
+    // Pre step get source meta to get content length (size of the object)
+    val eventualMaybeObjectSize: Future[Option[Long]] =
+      getObjectMetadata(sourceLocation.bucket, sourceLocation.key, sse).map(_.map(_.contentLength))
+    val eventualPartitions =
+      eventualMaybeObjectSize.map(_.map(createPartitions(chunkSize, sourceLocation)).getOrElse(Nil))
+    val partitions = Source.fromFuture(eventualPartitions)
```
ennru: Why do you wrap it into a stream? Isn't the future good enough?

sfali: First step in multi part copy (or upload) is to initiate multi part upload, which is done in `initiateUpload` (which is an existing function, used in `multipartUpload`, the return type of this function is `Source[(MultipartUpload, Int), NotUsed]`) , in order to copy an object first step is to get the content length of the existing object which is done here, the reason for converting into Source is to zip with source of `initiateUpload`.

docs/src/main/paradox/s3.md
```diff
@@ -100,19 +100,72 @@ Java
...
+import akka.stream.alpakka.s3.javadsl.MultipartUploadResult;
+import akka.stream.alpakka.s3.javadsl.S3Client;
+import akka.testkit.javadsl.TestKit;
+import com.amazonaws.SdkClientException;
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.auth.AWSStaticCredentialsProvider;
+import com.amazonaws.auth.BasicAWSCredentials;
+import com.amazonaws.regions.AwsRegionProvider;
+
+public class Test extends TestKit {
```
ennru: Please put the code for the documentation in the tests and import the snippets via Parardox' @@snip support.
You may use the same snippet marker at several places in one file and Paradox will put it together in the order they appear, that way you can pull in correct Java imports from the source.

sfali:
Thank you

ennru:
This PR is now replaced with #943 
@sfali Please use `git rebase master` instead of merging from master for your future PRs.

Issue #870, opened 2018-03-29
CSV: wrong parsing of escaped text within columns


ennru:
Just discovered that the CSV parsing doesn't handle escaping correctly.
A,B\\\\C,D should result in "A", "B\C", "D", but becomes "A", "B\\\\C", "D"

(this is even hard to write so it shows correctly in Markdown, btw)

ennru:
Fixed with #872

PR #872, opened 2018-03-30
CSV: fix wrong parsing of escaped text within columns

tuleism:
Ref #870. Other than the mentioned example, this also fixes for cases where escapeChar appear in the first field (see the tests).


ennru:
Good one.
The bug I identified is in case WithinField the escaped char should be added with `fieldBuffer.init` (right now it's `add`).
The normal character cases should not use `init`.

Another bug occurs when an escape char (or quote char) happens to be split over two ByteStrings.

tuleism:
Thanks!

>Another bug occurs when an escape char (or quote char) happens to be split over two ByteStrings.

Can you give an example of this?

tuleism:
>The bug I identified is in case WithinField the escaped char should be added with fieldBuffer.init (right now it's add).

Ping @ennru: I've fixed this. Can you give example for my comment above?

Issue #874, opened 2018-04-04
FTP: SecureRandom hangs on docker


fmsbeekmans:
In environments with low entropy such as docker the creation of an `SSHClient` in the `ftp` module can take very long. In situations where security is not important such as testing on a build server this can be problematic.

johanandren:
I think it may be the case of /dev/random being used, and that blocking until there is enough entropy (For example https://stackoverflow.com/questions/26021181/not-enough-entropy-to-support-dev-random-in-docker-containers-running-in-boot2d) 

Did you try explicitly pointing it to /dev/urandom instead?

johanandren:
Ups, pasted the wrong link: https://stackoverflow.com/questions/137212/how-to-solve-slow-java-securerandom

fmsbeekmans:
It is indeed but it's an environment where we can't mount `/dev/urandom`. Our proposed solution was to use a different ssh config that uses a different random

ennru:
With #875 in place, it is now possible to inject a custom SSH config.

PR #875, opened 2018-04-04
FTP: Support custom SSH configuration

fmsbeekmans:
Made it possible to use a custom SSH configuration.

An alternative would be the possibility to inject an `SSHClient` instance.

This should help with #874.
Worked together with @HarrisonBaxter

### DIFF
--- a/ftp/src/main/scala/akka/stream/alpakka/ftp/javadsl/FtpApi.scala
+++ b/ftp/src/main/scala/akka/stream/alpakka/ftp/javadsl/FtpApi.scala
@@ -235,19 +235,19 @@ sealed trait FtpApi[FtpClient] { _: FtpSourceFactory[FtpClient] =>
 
   protected[this] implicit def ftpLike: FtpLike[FtpClient, S]
 }
-
+class SftpApi extends FtpApi[SSHClient] with SftpSourceParams
 object Ftp extends FtpApi[FTPClient] with FtpSourceParams
 object Ftps extends FtpApi[FTPClient] with FtpsSourceParams
-object Sftp extends FtpApi[SSHClient] with SftpSourceParams {
+object Sftp extends SftpApi {
 
   /**
-   * Java API: creates a [[akka.stream.alpakka.ftp.javadsl.FtpApi]] of [[net.schmizz.sshj.SSHClient]] with [[akka.stream.alpakka.ftp.impl.SftpSourceParams]]
+   * Java API: creates a [[akka.stream.alpakka.ftp.javadsl.SftpApi]]
    *
    * @param customSshClient custom ssh client
-   * @return A [[akka.stream.alpakka.ftp.javadsl.FtpApi]] of [[net.schmizz.sshj.SSHClient]] with [[akka.stream.alpakka.ftp.impl.SftpSourceParams]]
+   * @return A [[akka.stream.alpakka.ftp.javadsl.SftpApi]]
    */
-  def create(customSshClient: SSHClient): FtpApi[SSHClient] with SftpSourceParams =
-    new FtpApi[SSHClient] with SftpSourceParams {
+  def create(customSshClient: SSHClient): SftpApi =
+    new SftpApi {
       override val sshClient = customSshClient
     }
 }
--- a/ftp/src/main/scala/akka/stream/alpakka/ftp/javadsl/FtpApi.scala
+++ b/ftp/src/main/scala/akka/stream/alpakka/ftp/javadsl/FtpApi.scala
@@ -238,9 +238,18 @@ sealed trait FtpApi[FtpClient] { _: FtpSourceFactory[FtpClient] =>
 
 object Ftp extends FtpApi[FTPClient] with FtpSourceParams
 object Ftps extends FtpApi[FTPClient] with FtpsSourceParams
-object Sftp extends FtpApi[SSHClient] with SftpSourceParams {
-  def withSshClient(customSshClient: SSHClient): FtpApi[SSHClient] with SftpSourceParams =
-    new FtpApi[SSHClient] with SftpSourceParams {
+class Sftp extends FtpApi[SSHClient] with SftpSourceParams
+
+object Sftp extends Sftp {
+
+  /**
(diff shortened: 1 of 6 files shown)


ennru:
Looks good. Could you add it to the docs in `ftp.md` as well, please?

fmsbeekmans:
Changed it such that you can use an `SSHClient` instead. This also allows reuse of an existing `SSHClient` instance. The method name in the scalaDsl is `apply` but in the javaDsl is `withSshClient`, is that considered idiomatic or should they be the same?

ftp/src/main/scala/akka/stream/alpakka/ftp/javadsl/FtpApi.scala
```diff
@@ -238,4 +238,18 @@ sealed trait FtpApi[FtpClient] { _: FtpSourceFactory[FtpClient] =>
...
+
+object Sftp extends Sftp {
+
+  /**
+    * Java API: creates a [[akka.stream.alpakka.ftp.javadsl.Sftp]]
+    *
+    * @param customSshClient custom ssh client
+    * @return A [[akka.stream.alpakka.ftp.javadsl.Sftp]]
+    */
+  def withSshClient(customSshClient: SSHClient): Sftp =
```
ennru: The `withXyz` methods are used have something to replace Scala's copy methods on case classes.
As constructor I'd recommend `create` where `apply` is used in Scala.

ftp/src/main/scala/akka/stream/alpakka/ftp/javadsl/FtpApi.scala
```diff
@@ -238,4 +238,16 @@ sealed trait FtpApi[FtpClient] { _: FtpSourceFactory[FtpClient] =>
...
-object Sftp extends FtpApi[SSHClient] with SftpSourceParams
+object Sftp extends FtpApi[SSHClient] with SftpSourceParams {
+
+  /**
+   * Java API: creates a [[akka.stream.alpakka.ftp.javadsl.FtpApi]] of [[net.schmizz.sshj.SSHClient]] with [[akka.stream.alpakka.ftp.impl.SftpSourceParams]]
+   *
+   * @param customSshClient custom ssh client
+   * @return A [[akka.stream.alpakka.ftp.javadsl.FtpApi]] of [[net.schmizz.sshj.SSHClient]] with [[akka.stream.alpakka.ftp.impl.SftpSourceParams]]
+   */
+  def withSshClient(customSshClient: SSHClient): FtpApi[SSHClient] with SftpSourceParams =
```
ennru: Would it help the Java API to create a class `Sftp` and base the object on it?
I'd still like to call it `create` or `createWithSshClient` as a dual to Scala's `apply`.

HarrisonBaxter:
@ennru is there more work needed for this, or does it look ok to merge?

ftp/src/main/scala/akka/stream/alpakka/ftp/javadsl/FtpApi.scala
ennru: Could you make it return a `Sftp` instance instead of `FtpApi[SSHClient]`?

HarrisonBaxter:
Ferdy and I have done it like this to try and avoid breaking backward
compatibility

ftp/src/main/scala/akka/stream/alpakka/ftp/javadsl/FtpApi.scala
fmsbeekmans: In 013f9c9baae32bafbe2f8dbcbc52adb58e81fff3 we tried to create an `SFTP` class that the current `SFTP` object inherited from. This way all existing code using the `SFTP` object would still work.

Sadly Java doesn't play well with this, it is not able to resolve the methods inherited from the traits on companion object as static methods.

We could make a separate type for instead but it would look a bit awkward.

fmsbeekmans:
Hi @ennru , is there more work to be done before this can be merged?

ftp/src/main/scala/akka/stream/alpakka/ftp/javadsl/FtpApi.scala
ennru: What about introducing 
`class SftpApi extends FtpApi[SSHClient] with SftpSourceParams`
and extending and returning that?

HarrisonBaxter:
We were hoping to achieve the same return type when using a custom ssh
client as the default one. The proposed solution does not allow this.

ftp/src/test/java/akka/stream/alpakka/ftp/examples/ConfigureCustomSSHClient.java
```diff
@@ -0,0 +1,22 @@
...
+import akka.stream.alpakka.ftp.javadsl.FtpApi;
+import akka.stream.alpakka.ftp.javadsl.Sftp;
+import net.schmizz.sshj.DefaultConfig;
+import net.schmizz.sshj.SSHClient;
+
+public class ConfigureCustomSSHClient {
+
+    public ConfigureCustomSSHClient() {
+        SSHClient sshClient = new SSHClient(new DefaultConfig());
+        FtpApi<SSHClient> sftp = Sftp.create(sshClient);
```
ennru: You should change this to `SftpApi` as well.

ftp/src/main/scala/akka/stream/alpakka/ftp/scaladsl/FtpApi.scala
```diff
@@ -169,4 +169,18 @@ sealed trait FtpApi[FtpClient] { _: FtpSourceFactory[FtpClient] =>
 
 object Ftp extends FtpApi[FTPClient] with FtpSourceParams
 object Ftps extends FtpApi[FTPClient] with FtpsSourceParams
-object Sftp extends FtpApi[SSHClient] with SftpSourceParams
+class Sftp extends FtpApi[SSHClient] with SftpSourceParams
```
ennru: Would make sense to give it the same structure in the Scala API, WDYT?

ennru:
Thank you for getting this in place!

PR #877, opened 2018-04-06
DynamoDB: Fix parallelism #876

### LINKED ISSUE #876: AWS DynamoDB doesn't obey parallelism setting
The DynamoDB connector has a parallelism setting as follows:

```properties
  # Max number of in flight requests from the AwsClient
  parallelism = 10
```

However, it doesn't work. The way it's used in the flow is:

```scala
      .via(connection)
      .mapAsync(settings.parallelism) {
        case (Success(response), i) => toAwsResult(response, i)
        case (Failure(ex), i) => Future.failed(ex)
      }
```

Which means that only the `mapAsync` call is affected by the parallelism and not the connection itself. The connection instead gets its parallelism from the `akka.http.host-connection-pool.max-connections` setting which defaults to `4`.

If the `parallelism` setting is higher than the `max-connections` setting then the adapter never reaches the parallelism limit, and if it's lower then it doesn't have much effect as the `mapAsync` is typically faster than the request so it doesn't backpressure much.

gregbeech:
This PR fixes #876 by changing the parallelism in DynamoDB so that it is applied to the HTTP connection pool.

I upped the default parallelism to 32 as this is the nearest power of 2 below 50, which is the default that [AWS used in their own SDK before deprecating the methods](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/dynamodbv2/AmazonDynamoDBAsyncClient.html). I've run up to 128 connections in production (by setting the akka-http config properties) without any problems so 32 seems perfectly safe.

### DIFF
--- a/dynamodb/src/main/resources/reference.conf
+++ b/dynamodb/src/main/resources/reference.conf
@@ -8,6 +8,6 @@ akka.stream.alpakka.dynamodb {
   # The AWS port, https is used as default scheme - 443 will result in not using port in url
   port: -1
 
-  # Max number of in flight requests from the AwsClient
-  parallelism = 10
+  # Max number of in flight requests from the AwsClient - must be a power of 2
+  parallelism = 32
 }
--- a/dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/impl/DynamoClientImpl.scala
+++ b/dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/impl/DynamoClientImpl.scala
@@ -8,6 +8,7 @@ import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.model.MediaType.NotCompressible
 import akka.http.scaladsl.model.{ContentType, MediaType}
+import akka.http.scaladsl.settings.ConnectionPoolSettings
 import akka.stream.Materializer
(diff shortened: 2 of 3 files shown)


gregbeech:
The power of two is from the akka-http requirement for [max open requests](https://doc.akka.io/docs/akka-http/current/configuration.html):

```properties
    # Must be a power of 2 and > 0!
    max-open-requests = 32
```

It's an intriguing requirement but I haven't dug into the code enough to see why it's implemented that way.

gregbeech:
One thing that occurs to me is that with this change [`parallelism` is still used for `mapAsync`](https://github.com/akka/alpakka/blob/master/dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/impl/AwsClient.scala#L81) and it's likely way higher than needed.

For example, we use 128 connections from our production servers but don't need anything like that parallelism for `mapAsync` as it's so much faster (we still use the default of 10 for that). Having another setting for that value seems messy though, as that's really an implementation detail that the user of the library shouldn't have to know/care about.

Any ideas? Perhaps leaving them the same is fine as it'll never get that high, or we could compute a value for it that seems more reasonable such as parallelism/8.

ennru:
Thank you for this improvement. Good to see you funnelling quite some data through it!

PR #873, opened 2018-04-04
Unix Domain Socket: fix some bugs due to races

longshorej:
This fixes a couple small bugs with the unix domain socket connector, both of which would cause its thread to crash depending upon timing.

1) `channel.read` can throw an `IOException` if the socket has subsequently been closed since the `SelectionKey` was selected.

```
Exception in thread "unix-domain-socket-io" java.io.IOException: Bad file descriptor
	at jnr.enxio.channels.Common.read(Common.java:67)
	at jnr.enxio.channels.AbstractNativeSocketChannel.read(AbstractNativeSocketChannel.java:62)
	at jnr.unixsocket.UnixSocketChannel.read(UnixSocketChannel.java:250)
	at akka.stream.alpakka.unixdomainsocket.scaladsl.UnixDomainSocket$.akka$stream$alpakka$unixdomainsocket$scaladsl$UnixDomainSocket$$nioEventLoop(UnixDomainSocket.scala:142)
	at akka.stream.alpakka.unixdomainsocket.scaladsl.UnixDomainSocket$$anon$4.run(UnixDomainSocket.scala:323)
	at java.lang.Thread.run(Thread.java:748)
```

2) `PendingReceiveAck` case where `pendingResult` hasn't completed yet, resulted in `MatchError`

```
Exception in thread "unix-domain-socket-io" scala.MatchError: PendingReceiveAck(akka.stream.impl.QueueSource$$anon$1@2fb54c65,java.nio.HeapByteBuffer[pos=9 lim=9 cap=65536],Future(Success(Enqueued))) (of class akka.stream.alpakka.unixdomainsocket.scaladsl.UnixDomainSocket$PendingReceiveAck)
	at akka.stream.alpakka.unixdomainsocket.scaladsl.UnixDomainSocket$.akka$stream$alpakka$unixdomainsocket$scaladsl$UnixDomainSocket$$nioEventLoop(UnixDomainSocket.scala:139)
	at akka.stream.alpakka.unixdomainsocket.scaladsl.UnixDomainSocket$$anon$4.run(UnixDomainSocket.scala:323)
	at java.lang.Thread.run(Thread.java:748)
```

PR #882, opened 2018-04-09
 Unix Domain Socket - more fixes

longshorej:
Another round of a few small tweaks. Hopefully after this it is quite stable :)

* `channel.write()` can throw an exception if writing to a closed socket, similar to the fix for read, see #873
* Following up on #878, `sendSink` shouldn't be coupled to `receiveSource`. This allows the server to continue to write data when the client has closed its write side.

/cc @huntc


ennru:
Thank you! Great that you fix these details to get a stable solution.

Issue #883, opened 2018-04-09
rabbitmq (amqp) cached connection shutdown listeners, memory leak


drobert:
While investigating some issues with heap size in an application I work on, I opened an hprof file in a memory analyzer tool which suggested a 'leak suspect' to be `akka.stream.alpakka.amqp.AmqpConnectorLogic$$anon$1` (screenshot attached). At the time of the analysis, there were 57,000 instances of this. Looking at the common paths, this appears to be distinct instances of the rabbitmq underlying `shutdownListeners` array.

I am working on distilling this down to a minimally-reproducible block of code, however I think the important aspect is that I'm using the `AmqpCachedConnectionProvider` (backed by `AmqpUriConnectionProvider` with `automaticRelease=true`).

(_Update_: it would be reproducible by re-establishing a new akka stream with a new sink, delivering 1+ messages, then closing the stream. Repeat this process with a cached connection provider and ultimately the heap should exhaust.)

As it stands, each graph started with a single connection will augment the shutdown listeners array by one, and since the connection remains open, shutdown listeners are accumulated on the heap for each created graph.

Note that there is currently no issue with *channel* shutdown listeners as channels are used within a single graph stage and never shared or reused.

![alpakka_amqp_heap](https://user-images.githubusercontent.com/455162/38508296-d553c356-3bec-11e8-9017-5c5b89421c38.png)

drobert:
Despite the two PRs up right now, I think the bigger issue is the lack of test coverage for this use case. I'm going to try to take a pass at this later this evening (I'm in TZ `America/New_York`) but would gladly accept some help as this is my first time contributing to alpakka.

drobert:
Took a pass at adding tests for this case in https://github.com/akka/alpakka/pull/887, but I'm having some issues. What I currently have is passing for me locally, but I left some questions in the PR. I'd appreciate some feedback/help on this as I'm still coming up to speed.

ennru:
Fixed by #887

PR #886, opened 2018-04-09
Fix AMQP cached connection memory leak

juanjoDiaz:
Might close (totally or partially) #883

The problem is that we were setting too many `shutdownListener` (in the connector logic and in the stages) and what is worse we were never removing them. So if the app is running within a retry with backoff kind of block, we are adding more are more handlers to the connection for no reason.

### DIFF
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
@@ -21,17 +21,15 @@ private[amqp] trait AmqpConnectorLogic { this: GraphStageLogic =>
   def whenConnected(): Unit
   def onFailure(ex: Throwable): Unit = failStage(ex)
 
+  protected val shutdownCallback = getAsyncCallback(onFailure)
+  private val shutdownListener = new ShutdownListener {
+    override def shutdownCompleted(cause: ShutdownSignalException): Unit = shutdownCallback.invoke(cause)
+  }
+
   final override def preStart(): Unit =
     try {
       connection = settings.connectionProvider.get
       channel = connection.createChannel()
-
-      val connShutdownCallback = getAsyncCallback[ShutdownSignalException] { ex =>
-        if (!ex.isInitiatedByApplication) failStage(ex)
-      }
-      val shutdownListener = new ShutdownListener {
-        override def shutdownCompleted(cause: ShutdownSignalException): Unit = connShutdownCallback.invoke(cause)
-      }
       connection.addShutdownListener(shutdownListener)
       channel.addShutdownListener(shutdownListener)
 
@@ -73,6 +71,8 @@ private[amqp] trait AmqpConnectorLogic { this: GraphStageLogic =>
 
   /** remember to call if overriding! */
   override def postStop(): Unit = {
+    connection.removeShutdownListener(shutdownListener)
+    channel.removeShutdownListener(shutdownListener)
     if ((channel ne null) && channel.isOpen) channel.close()
     channel = null
     settings.connectionProvider.release(connection)
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
@@ -53,7 +53,7 @@ final class AmqpRpcFlowStage(settings: AmqpSinkSettings, bufferSize: Int, respon
 
       override def whenConnected(): Unit = {
         import scala.collection.JavaConverters._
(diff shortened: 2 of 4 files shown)


amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
```diff
@@ -73,6 +71,8 @@ private[amqp] trait AmqpConnectorLogic { this: GraphStageLogic =>
 
   /** remember to call if overriding! */
   override def postStop(): Unit = {
+    connection.removeShutdownListener(shutdownListener)
+    channel.removeShutdownListener(shutdownListener)
```
drobert: Heh, apparently I put up a PR a few minutes after this one, but in either case: the channel shouldn't require special shutdown listener removal as it is always going to be distinct to this instance. My understanding is that it's the connection that persists, but `channel.close()` should always *actually* close and therefore unlink it's registered shutdown listeners. (Anecdotally in my heap analysis when opening this issue, channel shutdown listeners were not an issue.)

juanjoDiaz: Yeah. Channels are always unique to a stage.

My understanding is that the suhtdownListener is needed just in case the channel is closed by the server or for whatever other reason, so the stage gets informed and fails itself.

drobert: Agreed that the shutdown listener is necessary for channels. I'm suggesting that the explicit removal of the channel shutdown listener may be unnecessary.

juanjoDiaz: Oh yeah! you are right on that one.
It's not really necessary. Just being extra-careful on that one!

ennru:
@juanjoDiaz and @drobert would be great if you two together could get #886 and #886 into one PR to fix #883

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
drobert: I'm not entirely convinced 'extra careful' is a good thing here. Explicitly removing a listener could have some runtime impact, especially if the listener would be naturally removed on channel shutdown anyway and if the listener collection is a thread-safe or locking collection (which, presumably, it is).

I think in either case, regardless of approach taken, I'm nervous about the absence of test coverage here. Perhaps we should brainstorm a way to test this? I think it could be done with a proxying connection that verifies only a single channel shutdown listener is added despite re-using a cached connection?

juanjoDiaz: What side runtime effect do you expect?

Removing a channel shutdown is literally:
```
    @Override
    public void removeShutdownListener(ShutdownListener listener)
    {
        synchronized(this.monitor) {
            this.shutdownListeners.remove(listener);
        }
    }
```
so thread-safe and no exceptions thrown.

Having this explicitly it's more future proof imho since we don't know how this connector will evolve in the future.

I like the changes that you have done to #887. However, it's now an incomplete subset of this fix since it doesn't tackle the duplicated listeners. So I'd still go with this fix.

But I agree 100% on the issue around testing disconnect. That's a known issue in this connector and the MQTT one.
Using a TCP proxy was done on the MQTT connector (see https://github.com/akka/alpakka/blob/master/mqtt/src/test/scala/akka/stream/alpakka/mqtt/scaladsl/MqttSourceSpec.scala#L256) but AFAIK it's not a 100% reliable test.

I'm in EU so not much overlapping TZ with you, but happy to help with the testing if I can :)

drobert: > What side runtime effect do you expect?

`synchronized` will result in locking and slower memory access. So I'm advocating that we avoid unnecessary calls to mutate this collection, especially behind a lock.

> However, it's now an incomplete subset of this fix since it doesn't tackle the duplicated listeners

I feel like we're just not communicating on this, or perhaps just not agreeing on the problem space. There is no current issue with duplicated channel listeners, only connection listeners. The potential for error with shared channels is also very low and represents a misuse of ampq channels, or at least a misuse of alpakka amqp graph stages at the least. 

Perhaps it's more of a personal ethos, but I generally dislike making changes to code that isn't exhibiting a known problem and has no supported use case that would exhibit it. 

This PR represents more change to as-of-yet untested code. The reason I've left my PR up so far is I think it will both solve the immediate problem and the narrow footprint of the change should allow it to be more easily tested with confidence in all changes.

I will ping again when I have a testing proposal in place. As of now, I feel like the two of us are basically at an impasse as to which is the best approach to take to solving #883

ennru:
Replaced by the solution in #887

PR #887, opened 2018-04-09
AMQP: remove shutdown listeners for cached connections

drobert:
Pertaining to https://github.com/akka/alpakka/issues/883 :

On each graph creation an amqp `Connection` is obtained from an `AmqpConnectionProvider` and a shutdown listener is registered on it. However when the `AmqpCachedConnectionProvider` was introduced, the removal of the shutdown listener was never introduced, and the result is effectively a memory leak where shutdown listeners are added to but never removed from the same connection. 

In applications which establish new streams frequently, this can result in myriad shutdown listeners lingering after they are no longer applicable.

### DIFF
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
@@ -22,9 +22,11 @@ sealed trait AmqpConnectionProvider {
 }
 
 /**
-  * Wrapper/proxy implementation of a provider, useful for testing with a spy or mock
-  */
-private [amqp] case class DelegatingConnectionProvider(delegate: AmqpConnectionProvider, f: Connection => Connection = identity) extends AmqpConnectionProvider {
+ * Wrapper/proxy implementation of a provider, useful for testing with a spy or mock
+ */
+private[amqp] case class DelegatingConnectionProvider(delegate: AmqpConnectionProvider,
+                                                      f: Connection => Connection = identity)
+    extends AmqpConnectionProvider {
   override def get: Connection = f(delegate.get)
   override def release(connection: Connection): Unit = delegate.release(connection)
 }
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
@@ -22,6 +22,14 @@ sealed trait AmqpConnectionProvider {
 }
 
 /**
+  * Wrapper/proxy implementation of a provider, useful for testing with a spy or mock
+  */
+private [amqp] case class DelegatingConnectionProvider(delegate: AmqpConnectionProvider, f: Connection => Connection = identity) extends AmqpConnectionProvider {
+  override def get: Connection = f(delegate.get)
+  override def release(connection: Connection): Unit = delegate.release(connection)
+}
+
+/**
  * Connects to a local AMQP broker at the default port with no password.
  */
 case object AmqpLocalConnectionProvider extends AmqpConnectionProvider {
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
@@ -22,16 +22,6 @@ sealed trait AmqpConnectionProvider {
 }
 
 /**
- * Wrapper/proxy implementation of a provider, useful for testing with a spy or mock
- */
-private[amqp] case class DelegatingConnectionProvider(delegate: AmqpConnectionProvider,
-                                                      f: Connection => Connection = identity)
-    extends AmqpConnectionProvider {
-  override def get: Connection = f(delegate.get)
-  override def release(connection: Connection): Unit = delegate.release(connection)
-}
-
-/**
  * Connects to a local AMQP broker at the default port with no password.
  */
 case object AmqpLocalConnectionProvider extends AmqpConnectionProvider {
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
@@ -9,30 +9,18 @@ import com.rabbitmq.client._
 
 import scala.util.control.NonFatal
 
-private[amqp] case class ConnectionWithShutdown(conn: Connection, onShutdown: ShutdownListener) {
-
-  /**
-   * @return a Channel with the shutdown listener pre-registered
-   */
-  def createChannel: Channel = {
-    val ch = conn.createChannel()
-    ch.addShutdownListener(onShutdown)
-    ch
-  }
-
-  /**
-   * Removes the shutdown listener from the connection
-   */
-  def shutdown(): Unit = conn.removeShutdownListener(onShutdown)
-}
-
 /**
  * Internal API
  */
 private[amqp] trait AmqpConnectorLogic { this: GraphStageLogic =>
 
-  private var connection: ConnectionWithShutdown = _
+  private var connection: Connection = _
   protected var channel: Channel = _
+
+  protected lazy val shutdownCallback = getAsyncCallback(onFailure)
+  private lazy val shutdownListener = new ShutdownListener {
+    override def shutdownCompleted(cause: ShutdownSignalException): Unit = shutdownCallback.invoke(cause)
+  }
 
   def settings: AmqpConnectorSettings
   def whenConnected(): Unit
@@ -40,20 +28,11 @@ private[amqp] trait AmqpConnectorLogic { this: GraphStageLogic =>
 
   final override def preStart(): Unit =
     try {
-      val connShutdownCallback = getAsyncCallback[ShutdownSignalException] { ex =>
-        if (!ex.isInitiatedByApplication) failStage(ex)
-      }
-      val shutdownListener = new ShutdownListener {
-        override def shutdownCompleted(cause: ShutdownSignalException): Unit = connShutdownCallback.invoke(cause)
-      }
-
-      connection = {
-        val c = settings.connectionProvider.get
-        c.addShutdownListener(shutdownListener)
-        ConnectionWithShutdown(c, shutdownListener)
-      }
-
+      connection = settings.connectionProvider.get
       channel = connection.createChannel
+
+      connection.addShutdownListener(shutdownListener)
+      channel.addShutdownListener(shutdownListener)
 
       import scala.collection.JavaConverters._
 
@@ -96,8 +75,8 @@ private[amqp] trait AmqpConnectorLogic { this: GraphStageLogic =>
     if ((channel ne null) && channel.isOpen) channel.close()
     channel = null
 
-    connection.shutdown
-    settings.connectionProvider.release(connection.conn)
+    connection.removeShutdownListener(shutdownListener)
+    settings.connectionProvider.release(connection)
     connection = null
   }
 }
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
@@ -9,13 +9,11 @@ import com.rabbitmq.client._
 
 import scala.util.control.NonFatal
 
-private[amqp] case class ConnectionWithShutdown(
-  conn: Connection,
-  onShutdown: ShutdownListener) {
+private[amqp] case class ConnectionWithShutdown(conn: Connection, onShutdown: ShutdownListener) {
 
   /**
-    * @return a Channel with the shutdown listener pre-registered
-    */
+   * @return a Channel with the shutdown listener pre-registered
+   */
   def createChannel: Channel = {
     val ch = conn.createChannel()
     ch.addShutdownListener(onShutdown)
@@ -23,8 +21,8 @@ private[amqp] case class ConnectionWithShutdown(
   }
 
   /**
-    * Removes the shutdown listener from the connection
-    */
+   * Removes the shutdown listener from the connection
+   */
   def shutdown(): Unit = conn.removeShutdownListener(onShutdown)
 }
 
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
@@ -9,12 +9,31 @@ import com.rabbitmq.client._
 
 import scala.util.control.NonFatal
 
+private[amqp] case class ConnectionWithShutdown(
+  conn: Connection,
+  onShutdown: ShutdownListener) {
+
+  /**
+    * @return a Channel with the shutdown listener pre-registered
+    */
+  def createChannel: Channel = {
+    val ch = conn.createChannel()
+    ch.addShutdownListener(onShutdown)
+    ch
+  }
+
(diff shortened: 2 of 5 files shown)


juanjoDiaz:
Duplicates #886

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
```diff
@@ -9,12 +9,29 @@ import com.rabbitmq.client._
 
 import scala.util.control.NonFatal
 
+private[amqp] case class ConnectionWithShutdown(conn: Connection, onShutdown: ShutdownListener) {
```
juanjoDiaz: I'm not convinced that the shutdown listeners belong here since there are stage-specific.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
```diff
@@ -23,17 +40,20 @@ private[amqp] trait AmqpConnectorLogic { this: GraphStageLogic =>
...
       val shutdownListener = new ShutdownListener {
         override def shutdownCompleted(cause: ShutdownSignalException): Unit = connShutdownCallback.invoke(cause)
       }
-      connection.addShutdownListener(shutdownListener)
-      channel.addShutdownListener(shutdownListener)
+
+      connection = {
+        val c = settings.connectionProvider.get
+        c.addShutdownListener(shutdownListener)
+        ConnectionWithShutdown(c, shutdownListener)
```
juanjoDiaz: I find weird adding the connection shutdown and then creating the `ConnectionWithShutdown` instead of doing everything inside the `ConnectionWithShutdown`

drobert:
> And I don't think that the shutdown callbacks belong to the provider instead of the stage.

I'm a little confused here as well. While my initial thought was to add the shutdown listener to the provider (as I mentioned in the reported issue), that is not the implementation I've proposed here. 

A connection is obtained within a stage only once during `preStart` and it's here I decorate the returned connection with some utilities to register and unregister the connection shutdown listener. The connection providers are left unchanged.

The goal here was to minimize change while solving the immediate problem.

> However, it doesn't solve the duplicate shutdown listeners the sinks.

I'm not sure I follow. The conversation feels a bit muddled as there is not a general shutdown listener' problem; there is specifically a *connection* shutdown listener problem. In this PR, channel shutdown listeners are left untouched as they clean themselves up properly, and only the connection shutdown listener removal is enforced per stage.

The Sink-specific code modification made in #886 seem to only impact channels, which I don't believe are necessary changes for the reported issue.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
drobert: Perhaps this syntax is misleading. This is effectively just a tuple to keep the fetched connection and its associated shutdown listener together so that they can be closed together. The listener can't be unregistered without the connection is was registered on so this construct felt necessary.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnector.scala
```diff
@@ -17,21 +17,20 @@ private[amqp] trait AmqpConnectorLogic { this: GraphStageLogic =>
   private var connection: Connection = _
   protected var channel: Channel = _
 
+  protected lazy val shutdownCallback = getAsyncCallback(onFailure)
```
drobert: @juanjoDiaz I've followed your lead in your PR (#886) and have replaced my case class/tuple with a class-level shutdown listener and shutdown callback. I think this keeps the scope of change as small as possible while still accomplishing what the ticket reports.

drobert:
I'm having a difficult time testing this, and part of that is my relative inexperience with akka streams. 

The approach I've taken here:

- Provide a `DelegatingConnectionProvider` that allows a delegate and a modifying function that is called on `.get`. This allows callers to supply spys, fakes, and mocks.
- Use the above to track calls to `addShutdownListener` and `removeShutdownListener`
- Create several graphs and execute them with one value each
- Verify that each graph was executed with one element and called both the add and remove shutdown listener methods once for each graph

Note: I created `AmqpProxyConnection` as a last resort for proxying. Previously I was using Mockito `spy`s, but they are nominally not thread safe and I was getting inconsistent test results. This implementation, while messy, should be thread safe.

Things I do not understand:

- When running this new test (`registers and unregisters a single connection shutdown hook per graph`), I am seeing consistent passes locally when `autoRelease = true` but I'm not seeing the same behavior when `autoRelease = false`. I don't understand why; I would expect `preStart` and `postStop` to be called exactly once for each completed stream.
- I mentioned Mockito wasn't thread-safe, but I don't know why that should be an issue here, given the implicit execution context is `sameThreadExecutionContext`

Last question: I'm unclear on where exactly I should have put this test. Perhaps it belongs in `AmqpConnectorsSpec` instead?

amqp/src/test/scala/akka/stream/alpakka/amqp/AmqpProxyConnection.scala
```diff
@@ -0,0 +1,68 @@
+package akka.stream.alpakka.amqp
+
+import java.net.InetAddress
+import java.util
+
+import com.rabbitmq.client._
+
+class AmqpProxyConnection(delegate: Connection) extends Connection {
```
drobert: A dumb, easily extensible proxy. I'd rather use `Mockito.spy` though I had issues with it as mentioned [here](https://github.com/akka/alpakka/pull/887#issuecomment-380319279)

amqp/src/test/scala/akka/stream/alpakka/amqp/scaladsl/AmqpConnectionProvidersSpec.scala
```diff
@@ -135,9 +147,57 @@ class AmqpConnectionProvidersSpec extends AmqpSpec {
...
+
+      val amqpSink = AmqpSink.simple(
+        AmqpSinkSettings(reusableConnectionProvider)
+          .withRoutingKey(queueName)
+          .withDeclarations(queueDeclaration)
+      )
+
+      val input = Vector("one", "two", "three", "four")
+
+      Future.traverse(input)(in => Source.single(ByteString(in)).runWith(amqpSink))
```
drobert: Given my [earlier issues](https://github.com/akka/alpakka/pull/887#issuecomment-380319279) I'm wondering if this isn't a valid test since I get inconsistent results with `autoRelease = false`

ennru:
You can not rely on `postStop` being called in a timely manner. It may be signalled asynchronously. I believe you can force it to be called by calling `ActorMaterializer.shutdown()`, but then you can't re-use it.

drobert:
> You can not rely on postStop being called in a timely manner.

Well, that would explain it :-)

I'm very open to suggestion on how to test this. I can isolate this test or move it to a new file so that it uses its own actor system and handles these assertions after shutdown.

ennru:
Yes, try creating a new spec and create a new `ActorMaterializer` in every test.

drobert:
Things look pretty good with the test/actor system in its own file. What do we think?

ennru:
Yes, the purpose is much more obvious. So your tests aren't flaky anymore?

@juanjoDiaz Can you review and see, if everything you find important is covered now?

drobert:
> So your tests aren't flaky anymore?

Things seem pretty stable. I'm getting consistent passes locally and travis' builds were successful.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -21,6 +21,16 @@ sealed trait AmqpConnectionProvider {
   def release(connection: Connection): Unit = if (connection.isOpen) connection.close()
 }
 
+/**
```
juanjoDiaz: This should be moved to the test project with the `AmqpProxyConnection`since it's just a test utility.

juanjoDiaz: Actually, I don't even think that you need this provider at all.
You could use the `AmqpConnectionFactoryConnectionProvider` and pass to it a connection factory that returns your spied connection.

Something like:
```
    val factory = new ConnectionFactory {
      override def newConnection(): Connection = new AmqpProxyConnection(conn)
    }
```

amqp/src/test/scala/akka/stream/alpakka/amqp/scaladsl/AmqpGraphStageLogicConnectionShutdownSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+    // actor system is within this test as it has to be shut down in order
+    // to verify graph stage termination
+    implicit val system = ActorSystem(this.getClass.getSimpleName + System.currentTimeMillis())
+    implicit val materializer = ActorMaterializer()
+
+    val spyConnectionPromise = Promise[Connection]
+    val shutdownsAdded = new AtomicInteger()
+    val shutdownsRemoved = new AtomicInteger()
+
+    def spy(conn: Connection): Connection = new AmqpProxyConnection(conn) {
```
juanjoDiaz: I don't think that you need the `AmqpProxyConnection` class.
Following up on my previous comment I think that you could just do:
```
    val factory = new ConnectionFactory {
      override def newConnection(): Connection = new Connection {
        override def addShutdownListener(shutdownListener: ShutdownListener): Unit = {
          shutdownsAdded.incrementAndGet()
          super.addShutdownListener(shutdownListener)
        }

        override def removeShutdownListener(shutdownListener: ShutdownListener): Unit = {
          shutdownsRemoved.incrementAndGet()
          super.removeShutdownListener(shutdownListener)
        }
      }
    }
```
Right?

amqp/src/test/scala/akka/stream/alpakka/amqp/scaladsl/AmqpGraphStageLogicConnectionShutdownSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+      AmqpSinkSettings(reusableConnectionProvider)
+        .withRoutingKey(queueName)
+        .withDeclarations(queueDeclaration)
+    )
+
+    val input = Vector("one", "two", "three", "four")
+
+    Future
+      .traverse(input)(in => Source.single(ByteString(in)).runWith(amqpSink))
+      .flatMap(_ => spyConnectionPromise.future)
```
juanjoDiaz: What's this promise used for?

amqp/src/test/scala/akka/stream/alpakka/amqp/scaladsl/AmqpGraphStageLogicConnectionShutdownSpec.scala
drobert: Hmm, I'm not sure this will work. I think in my current case I need a valid amqp connection, and it's not necessarily the case that extending a default, no-args construction of `Connection` is the safe action. Looking at the default implementation of `ConnectionFactory`, various properties are set, including the connection name, executor pool, etc. What I really want is a wrapper around some Connection created by the underlying factory, decorated to track shutdown listeners. (Note how much actually goes into creating a connection: https://github.com/rabbitmq/rabbitmq-java-client/blob/master/src/main/java/com/rabbitmq/client/ConnectionFactory.java#L978)

I like the idea of using a spy or a mixin, but:
- Mockito's `spy` does not claim to be thread-safe
- I can't dynamically add a mixin to a class I'm not instantiating

So, I'm not entirely sure I can get rid of the `AmqpProxyConnection`

amqp/src/test/scala/akka/stream/alpakka/amqp/scaladsl/AmqpGraphStageLogicConnectionShutdownSpec.scala
drobert: I was using it to ensure that I was really using the same connection each time. Moot point after my most recent rewrite per your suggestions.

amqp/src/test/scala/akka/stream/alpakka/amqp/scaladsl/AmqpGraphStageLogicConnectionShutdownSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  val shutdownsAdded = new AtomicInteger()
+  val shutdownsRemoved = new AtomicInteger()
+
+  override def beforeEach() = {
+    shutdownsAdded.set(0)
+    shutdownsRemoved.set(0)
+  }
+
+  trait ShutdownListenerTracking { self: AmqpProxyConnection =>
```
drobert: A few notes:

I wouldn't have to require `AmqpProxyConnection` as the self-type (`Connection` should be enough), except I couldn't figure out any scala gymnastics to both override `addShutdownListener` and delegate to the super class' default behavior without a stack overflow (infinite loop of this function calling itself). So I require the proxy and get direct access to the delegate.

I could make this trait a bit more general purpose by giving it its own delegate, but it doesn't seem worth the effort to me. Something like this:

```scala
trait ShutdownListenerTracking { self: Connection => 
  protected def delegate: Connection

  override def addShutdownListener(shutdownListener: ShutdownListner): Unit = {
    shutdownsAdded.incrementAndGet
    delegate.addShutdownListener(shutdownListener)
  }
}
```

amqp/src/test/scala/akka/stream/alpakka/amqp/scaladsl/AmqpGraphStageLogicConnectionShutdownSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  "registers and unregisters a single connection shutdown hook per graph" in {
+    // actor system is within this test as it has to be shut down in order
+    // to verify graph stage termination
+    implicit val system = ActorSystem(this.getClass.getSimpleName + System.currentTimeMillis())
+    implicit val materializer = ActorMaterializer()
+
+    val connectionFactory = new ConnectionFactory() {
+      override def newConnection(es: ExecutorService, ar: AddressResolver, name: String): Connection =
```
drobert: @juanjoDiaz I mostly took your advice here but had to override this particular variant of the various, overloaded `newConnection` implementations as they all delegate to this one.

amqp/src/test/scala/akka/stream/alpakka/amqp/scaladsl/AmqpGraphStageLogicConnectionShutdownSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+      AmqpSinkSettings(reusableConnectionProvider)
+        .withRoutingKey(queueName)
+        .withDeclarations(queueDeclaration)
+    )
+
+    val input = Vector("one", "two", "three", "four")
+
+    Future
+      .traverse(input)(in => Source.single(ByteString(in)).runWith(amqpSink))
+      .transformWith(_ => system.terminate())
```
drobert: Cleaned up the system shutdown behavior too so that it gets called on failures as well as successes.

drobert:
>  I think that it could be simplified a bit

I've taken most of your suggested. I left some notes on how I don't think I can quite get rid of the `AmqpProxyConnection` implementation but if you have more thoughts on this, let me know.

drobert:
Can I bump this? I haven't heard any feedback in a couple weeks now.

ennru:
Could you change your review to Approved, please?

ennru:
OK to close #886?

juanjoDiaz:
Yeah. You can close #886  :)

Good work @drobert !

ennru:
The community at its best! Thank you @drobert and @juanjoDiaz to get this fixed together.

Issue #889, opened 2018-04-10
PostgreSQL: Lightweight Change Data Capture 


sebastianharko:
On a branch in my fork,  I have implemented a working, lightweight CDC (change data capture) for PostgreSQL. It is basically a source that streams changes from the database (and emits elements such as RowDeleted, RowUpdated, RowInserted (...) ) . It's basically conceptually similar to things such as Debezium(http://debezium.io) or Hibernate Envers(http://hibernate.org/orm/envers/) , but way easier to get started with -> since it's just an Akka Stream Source :-) :-). 

Again, it's very lightweight (zero dependencies) and uses the "logical_decoding" feature of PostgreSQL (PostgreSQL already comes with this feature, the user just has to enable it and it's very simple and has almost no performance overhead). PG v > 9.4 required though. 

I'm curious if this is something that has a place in the Alpakka project (I'd think it does!!). If so,  please let me know so I can PR my branch (already have proper docs btw).
 
Cheers!

ennru:
Yes, that fits well in Alpakka.
Would be interesting to see if this can be implemented for other "table-oriented" datastores, as well.

sebastianharko:
@ennru Thanks, I'll make a PR. And to answer you questuons yes it can be implemented for other "table-oriented" data stores as well (as a matter of fact my original I also have a semi-functional MS SQL implementation of this lying around).

PR #891, opened 2018-04-10
POC/WIP : PostgreSQL: Lightweight Change Data Capture / Event Log 

sebastianharko:
#889 This is a working, lightweight CDC (change data capture) for PostgreSQL. It is basically a source that streams changes from the database (and emits elements such as RowDeleted, RowUpdated, RowInserted (...) ) . It's basically conceptually similar to things such as Debezium(http://debezium.io),  but way easier to get started with, since it's just an Akka Stream Source :-) :-)..whereas Debezium requires Kafka + a custom plugin you have to compile & install on your PG server (IIRC).

It's very lightweight (zero dependencies) and uses the "logical_decoding" feature of PostgreSQL (PostgreSQL already comes with this feature, the user just has to enable it and it's very simple and has almost no performance overhead). PG v > 9.4 required though.

This PR is WIP but looking for early review/feedback:
Working at the moment on:
- More unit tests
- Java API
- Some hardening around error handling 
- Version with offset management

Cheers!
P.S. I used the "Azure Storage Queue" as a template for this source / example source to start off with.


postgresql-cdc/src/main/scala/akka/stream/alpakka/postgresqlcdc/PgSQLChangeDataCaptureSourceStage.scala
```diff
@@ -0,0 +1,214 @@
...
+          case SlotChange(_, _, ChangeStatement(schemaName, tableName, "INSERT", properties)) =>
+            RowInserted(schemaName, tableName, parseProperties(properties))
+        }
+
+        ChangeSet(transactionId, changes)
+
+    }
+  }.filter(_.changes.nonEmpty).toSet
+
+  object Grammar {
```
sebastianharko: some of the regular expressions below need to be hardened, to catch some edge cases ;-)

sebastianharko:
@ennru apologies for the late reply, haven't had time to work on this in a while (but getting back to it this weekend!!) but here's the answers to your questions:

**Is PostgreSQL CDC considered reliable?**
Yes, CDC via "logical decoding" is reliable (it has corner/edge cases but just like everything else). Debezium (enterprise grade CDC from RedHat) also uses this feature to get the events.  

**What happens if the listening application "misses" a bit for a restart ?**
Currently (present version of the PR), on the PostgreSql side : the change events accumulate in a "slot", as soon as you fetch them from the slot, they get cleared from the slot. This means that if the listening application "misses" a bit for a restart, then it can just resume from where it left off and fetch the events that have accumulated in the slot since the crash/restart. 

(Depends on the meaning of "misses a bit" as well.  Depending on configuration / settings, IIRC PostgreSql will automatically clear the "slot" after awhile (i.e. but we are talking days not hours)).

With this approach, there's a risk of events getting "lost" as they travel from the Alpakka Source to Sink, if the app crashes during one of the intermediary  stages (i.e.```map``` in the example below):    

```
PostgreSQLCapturer(....)
 .map(.....)
 .map(....)
 .map(...)
.to(Sink(...))
```

However, this can be fixed too (will discuss later). 

**Would it be possible to send the table's current state when the stream starts and than get CDC events?**
Yes, sounds like a good option to give users.

ennru:
@sebastianharko Looks like you merged from master, please use rebase instead.

project/Dependencies.scala
```diff
@@ -212,6 +212,13 @@ object Dependencies {
     )
   )
 
+  val PostgreSqlCdc = Seq(
+    libraryDependencies ++= Seq(
+      "org.scala-lang.modules" %% "scala-java8-compat" % "0.9.0", // BSD-style,
```
2m: `scala-java8-compat` comes as a transitive dependency from `akka-actor`. Therefore we should not set it as a dependency of this connector, because of a possible version mismatch.

sebastianharko: thanks. i'll take it out.

sebastianharko:
Hi @2m @ennru -  this is ready for an early review if you have time. 
I tried to follow the **Reference** connector as much as possible and wondering if I did it well :-)  

I still have quite a bit to do: mostly testing the Ack Sink (and also rebasing). The Ack Sink is the tricky part in this (but the source works without the ack sink - if you choose at most once delivery).

Please note that most of the tests are in PostgreSQLCapturerSpec.scala which is a long file and github does not render it by default ( you have to click on load diff), so don't miss that ;-)

Cheers!

sebastianharko:
@ennru @2m Any chance to get some PR review here ? This is almost complete, I just have to add a few more unit tests.

I wrote a sample / demo app that shows how this can work in "real life":
[https://github.com/sebastianharko/samuraicop](https://github.com/sebastianharko/samuraicop)

ennru:
Looks very good, I'll give it a full review on Monday.

docs/src/main/paradox/postgresql-cdc.md
```diff
@@ -0,0 +1,152 @@
...
+
+* It cannot capture events regarding changes to tables' structure (i.e. column removed, table dropped, table
+index added, change of column type etc.).
+* When a row update is captured, the previous version of the row is lost / not available.
+    * Unless the table has a non-default `REPLICA IDENTITY` setting. To get the full previous version of the row: `ALTER TABLE "TABLE_NAME" REPLICA IDENTITY FULL`.
+* It doesn't work with older version of PostgreSQL (i.e < 9.4).
+
+## Important Note
+
+* Remember to drop the slot after you are done using it.
```
ennru: Could you recommend a way to accomplish it?
On stream completion, or Actor system termination?
Maybe a Paradox [warning](https://developer.lightbend.com/docs/paradox/current//directives/callouts.html#warning-callout) is what you need here.

ennru:
Hi @sebastianharko 
Would you be able to push this over the corner?

ennru:
@sebastianharko Ping.

sebastianharko:
@ennru @2m Just finishing up some client work, will get back to this thing this week, the good news it that there's not much to do: just one unit test to test the AckSink and rebasing.

michalbogacz:
Is there any possibility that this will be completed in the near future?

sebastianharko:
@alexanderzafirov @michalbogacz Thanks for the interest in this and finding this useful. Now that I see some real interest, I'll create some time in my calendar for wrapping this up :-) Let me know if you have any feedback and if you'd be willing to help contribute/maintain. Thanks. Cheers!

eambrosio:
Hi @sebastianharko , are you still working on this project? It would be quite useful.

ennru:
With the work on Debezium in #2275, I think this PR might not be as attractive anymore.

sebastianharko:
@ennru I agree, I'm going to close this PR...

Issue #841, opened 2018-03-16
Streaming JSON unmarshalling of any nested structure


markusthoemmes:
When consuming public APIs and trying to embrace streaming I constantly face issues when it comes to unmarshalling JSON. Many many APIs (CouchDB for instance) do not provide me with a "clean" stream of data in form of a JSON array or a JSON object. They rather nest the data that I would like to get the most in some structure of metadata, which could look like this:

```json
{
  "count": 10,
  "rows": [...]
}
```

(This is of course a very simple case, other APIs are much worse, but this isn't about API bashing).

In the above case, https://github.com/knutwalker/akka-stream-json won't help me to parse `rows` in a streaming fashion because jawn only supports a stream of json-objects, a single array or a single value when it comes to the async parser.

I stumbled upon https://github.com/jsurfer/JsonSurfer which implements **exactly** the functionality I'd need to extract (streamingly) exactly the data that I need. I already built a PoC of a `GraphStage` which will support the following interface:

```scala
Flow[ByteString, JValue, NotUsed].via(JsonStream.flow[JValue]("$.rows[*]")
```

The above would now stream every element of that array as it occurs, and drop anything else immediately. JsonSurfer supports even more sophisticated models of filtering the stream even narrower, which is icing on the cake.

If this is deemed useful (it has some overlap with #320, but that doesn't seem to hit the ground anytime soon sadly) I'll straight away open a pull-request and work on productizing this piece of work. I myself cannot count how many times I wished this existed OOTB 😋

ennru:
Would it make sense to select the data array and than stream its content as `ByteString` ignoring the rest of the Json structure?
That way JsonFraming and any Json parser can be applied to that data.

markusthoemmes:
@ennru In theory, sure. In practice though, `JsonSurfer` already parses the JSON, that'd result in duplicate work.

I've written a generic version of this though, you can plugin whatever AST you want to. In reality the syntax is:

```scala
Flow[ByteString, J, NotUsed].via(JsonStream.flow[J]("$.rows[*]")
```

Where there is an implementation of `JsonSupport` for `J` like so:

```scala
trait JsonSupport[T] {
  implicit val support: JsonSupport[T]
  val provider: JsonProvider[_, _, _]
}
```

`JsonProvider[_, _, _]` is a native Jackson interface, so that'll be used under the hood.

[...]

PR #899, opened 2018-04-18
Support streaming of json-values.

markusthoemmes:
This enables parsing a json-structure in a streaming way. It allows to pick the values wanted from the original structure via `JsonPath`, a notation to refer to a specific part of a json-structure.

It never reads the original value in memory entirely (unless that is required by the given json-path), but instead parses the incoming values in chunks. This is especially helpful when consuming APIs, which put the results you're interested in in some nested structure, which isn't simply streamable.

A common example would be:

```
{
  "rows": [...]
}
```

Where the application is only interested in the values contained in `rows`, preferably as they arrive (over a TCP stream for example).

Issue of the initial discussion: #841.


ennru:
Great, thank you. I'll look at it soon.

build.sbt
```diff
@@ -116,6 +117,8 @@ lazy val ironmq = alpakkaProject("ironmq", "ironmq", Dependencies.IronMq)
 
 lazy val jms = alpakkaProject("jms", "jms", Dependencies.Jms, parallelExecution in Test := false)
 
+lazy val json = alpakkaProject("json", "json", Dependencies.Json)
```
ennru: I wonder if the module should be called `jsonStreaming` to leave room for other nifty things with Json.

docs/src/main/paradox/data-transformations/json.md
```diff
@@ -10,20 +12,67 @@ See @scaladoc[ScalaDSL JsonFraming](akka.stream.scaladsl.JsonFraming$) or @scala
...
+  "size": 100,
+  "rows": [
+    {"id": 1, "doc": {}}
+    {"id": 2, "doc": {}}
+    ...
+  ]
+}
+```
+
+The JSON parsing module offers a flow, which allows to stream specific parts of that JSON structure.
```
ennru: JSON streaming module

json/src/main/scala/akka/stream/alpakka/json/JsonParsing.scala
```diff
@@ -0,0 +1,98 @@
...
+
+object JsonParsing {
+
+  /**
+   * A Flow that consumes incoming json in chunks and produces a stream of parseable json values
+   * according to the JsonPath given.
+   *
+   * Supported JsonPath syntax: https://github.com/jsurfer/JsonSurfer#what-is-jsonpath
+   */
+  def flow(path: JsonPath): Flow[ByteString, ByteString, NotUsed] = Flow.fromGraph(new JsonStreamParser(path))
```
ennru: Or call these methods `select`? Would make the DSL quite nice to read.

json/src/main/scala/akka/stream/alpakka/json/JsonParsing.scala
```diff
@@ -0,0 +1,98 @@
...
+import akka.stream.scaladsl.Flow
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
+import akka.stream._
+import akka.util.ByteString
+import org.jsfr.json.compiler.JsonPathCompiler
+import org.jsfr.json.exception.JsonSurfingException
+import org.jsfr.json.path.JsonPath
+import org.jsfr.json.{JsonPathListener, JsonSurferJackson, ParsingContext}
+
+object JsonParsing {
```
ennru: I think it should be `JsonSelect`, `JsonPathSelect` or something like that. The contained Json is still to be parsed - a design decision I like.

json/src/main/scala/akka/stream/alpakka/json/JsonParsing.scala
```diff
@@ -0,0 +1,98 @@
+/*
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.json
```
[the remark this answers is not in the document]
markusthoemmes: I was wondering if this needs a seperate `javadsl`/`scaladsl`, since the parameters will be the same. Or is this to keep binaryCompat in the future? If so, then yeah, that'd make sense.

docs/src/main/paradox/data-transformations/json.md
```diff
@@ -10,20 +12,67 @@ See @scaladoc[ScalaDSL JsonFraming](akka.stream.scaladsl.JsonFraming$) or @scala
...
+
+@@dependency [sbt,Maven,Gradle] {
+  group=com.lightbend.akka
+  artifact=akka-stream-alpakka-json_$scalaBinaryVersion$
+  version=$version$
+
+### Example
+
+To define which parts of the structure you want to stream the module supports
+[JsonPath notation](https://github.com/jsurfer/JsonSurfer#what-is-jsonpath).
```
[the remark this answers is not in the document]
markusthoemmes: I chose the one to github on purpose, because JsonSurfer might not implement the whole of JsonPath's specification. WDYT?

json/src/main/scala/akka/stream/alpakka/json/JsonParsing.scala
markusthoemmes: Hmm, `JsonSelect.select` sounds a bit weird? What about `JsonReader`?

docs/src/main/paradox/data-transformations/json.md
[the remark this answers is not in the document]
ennru: Ok. Please add a few examples directly in the docs.

json/src/main/scala/akka/stream/alpakka/json/JsonParsing.scala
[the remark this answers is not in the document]
ennru: Yes, the API will evolve and then Java and Scala might not use the same.

markusthoemmes:
@ennru Addressed all the inline concerns hopefully. If you want to wait for a self-contained example it could take me another few days, let me know if you think its needed strictly.

build.sbt
```diff
@@ -116,6 +117,8 @@ lazy val ironmq = alpakkaProject("ironmq", "ironmq", Dependencies.IronMq)
 
 lazy val jms = alpakkaProject("jms", "jms", Dependencies.Jms, parallelExecution in Test := false)
 
+lazy val jsonStreaming = alpakkaProject("json-streaming", "json-streaming", Dependencies.JsonStreaming)
```
ennru: The module name may not use dashes (`-`), use `.` instead.

docs/src/main/paradox/data-transformations/json.md
```diff
@@ -10,20 +12,70 @@ See @scaladoc[ScalaDSL JsonFraming](akka.stream.scaladsl.JsonFraming$) or @scala
...
+
+The JSON reading module offers a flow, which allows to stream specific parts of that JSON structure.
+In this particular example, only the `rows` array is interesting for the application, more specifically
+even: Only the `doc` inside each element of the array.
+
+### Artifacts
+
+@@dependency [sbt,Maven,Gradle] {
+  group=com.lightbend.akka
+  artifact=akka-stream-alpakka-json_$scalaBinaryVersion$
```
ennru: The artifact id will be `akka-stream-alpakka-json-streaming`.

json-streaming/src/main/scala/akka/stream/alpakka/json/JsonStreamReader.scala
```diff
@@ -0,0 +1,79 @@
...
+      private val out = shape.out
+      setHandlers(in, out, this)
+
+      private var buffer = Queue.empty[ByteString]
+
+      private val surfer = JsonSurferJackson.INSTANCE
+      private val config = surfer.configBuilder
+        .bind(path, new JsonPathListener {
+          override def onValue(value: Any, context: ParsingContext): Unit =
+            buffer = buffer.enqueue(ByteString(value.toString))
```
2m: `buffer` is a var and is changed from the `onPush` handler. Therefore changing it here as well is not thread-safe. You should create an async callback that enqueues an element to the buffer and invoke that callback from the `JsonPathListener.onValue`.

json-streaming/src/main/scala/akka/stream/alpakka/json/javadsl/JsonReader.scala
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.json.javadsl
+
+import akka.NotUsed
+import akka.stream.alpakka.json.JsonStreamReader
+import akka.stream.scaladsl.Flow
```
2m: s/scaladsl/javadsl

json-streaming/src/main/scala/akka/stream/alpakka/json/JsonStreamReader.scala
markusthoemmes: Will `onPush` ever be called concurrently? I know it's awkward, but the interface itself seems to be synchronous, as in: The handler was already called after `feed` on the parser returns. I'd have to doublecheck this via source-code inspection but my experiments have shown it to work this way until now.

If that's the case, we'd be fine wrt. thread-safety, right?

docs/src/main/paradox/data-transformations/json.md
```diff
@@ -10,20 +12,70 @@ See @scaladoc[ScalaDSL JsonFraming](akka.stream.scaladsl.JsonFraming$) or @scala
...
+    {"id": 1, "doc": {}}
+    {"id": 2, "doc": {}}
+    ...
+  ]
+}
+```
+
+The JSON reading module offers a flow, which allows to stream specific parts of that JSON structure.
+In this particular example, only the `rows` array is interesting for the application, more specifically
+even: Only the `doc` inside each element of the array.
```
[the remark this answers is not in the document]
markusthoemmes: I think `doc` should be fine since the key is `"doc"` (which is what I'm referring to here.)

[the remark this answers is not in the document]
2m: Ahh sorry, I meant to say s/Only/only (a word after a colon starts with an upper case).

json-streaming/src/main/scala/akka/stream/alpakka/json/JsonStreamReader.scala
2m: `onPush` is not called concurrently. But the `buffer` var is modified from `onPush` and then from `onValue` which is a callback from jsonsurfer.

If the jsonsurfer does not have its own threads and the `onValue` callback is called by the same thread that called `parser.feed` in `onPush` then the current implementation is fine.

2m: I just checked and the callback is called from the same thread as the `parser.feed` is called. Therefore this is good and does not need any modifications.

2m:
I tried pushing a commit to your branch with the minor fixups, but it looks that edits from other maintainers are disabled for this PR. Here is the commit which you can cherry pick to the branch and then we are good to merge this one: https://github.com/2m/alpakka/commit/fcf6df5cda195d75bae656b3b1d2aec3365e887e

2m:
Continued in #942

PR #901, opened 2018-04-19
Feature/add configurable credentials to dynamo client

### LINKED ISSUE #900: Enable different credentials to be used in DynamoClient
Currently the `DynamoClient` is fixed to use the default AWS credentials provider chain and this cannot be modified. In some cases, such as needing to access different dynamo instances, with different credentials this is not possible.

Therefore adding the ability to provide credentials directly to the `DynamoSettings` would be useful. In addition, optionally setting static credentials in configuration would also be good.

randomcoder:
* Added static credentials to config and added `AWSCredentialsProvider` parameter to `DynamoSettings`.
* Added some java api `.create` methods and ScalaDoc to `DynamoSettings`
* Updated docs to include new config and credentials setting

Fixes #900

### DIFF
--- a/docs/src/main/paradox/dynamodb.md
+++ b/docs/src/main/paradox/dynamodb.md
@@ -23,11 +23,13 @@ provided by the [DynamoDB Java SDK](http://docs.aws.amazon.com/sdk-for-java/v1/d
 
 If you wish to use static credentials they can be defined in the config
 
+Define Static Credentials in Config
 : @@snip ($alpakka$/dynamodb/src/test/scala/akka/stream/alpakka/dynamodb/DynamoSettingsSpec.scala) { #static-creds }
 
-Alternatively any type of `AWSCredentialsProvider` can be provided programmatically via the default 
-@scaladoc[DynamoSettings](akka.stream.alpakka.dynamodb.impl.DynamoSettings) constructor
+Alternatively any type of `AWSCredentialsProvider` can be provided programmatically via the default @scaladoc[DynamoSettings](akka.stream.alpakka.dynamodb.impl.DynamoSettings) 
+constructor
 
+Supply Credentials Provider to Constructor
 : @@snip ($alpakka$/dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/impl/DynamoSettings.scala) { #init-settings } 
 
 Before you can construct the client, you need an @scaladoc[ActorSystem](akka.actor.ActorSystem), 
--- a/docs/src/main/paradox/dynamodb.md
+++ b/docs/src/main/paradox/dynamodb.md
@@ -18,9 +18,20 @@ The AWS DynamoDB connector provides a flow for streaming DynamoDB requests. For 
 
 ## Usage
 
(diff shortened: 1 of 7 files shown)


dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/impl/DynamoSettings.scala
```diff
@@ -5,20 +5,59 @@
...
+  /**
+   * Scala API: Creates [[DynamoSettings]] from the [[com.typesafe.config.Config Config]] attached to an [[akka.actor.ActorSystem]].
+   */
+  def apply(system: ActorSystem): DynamoSettings = apply(system.settings.config)
+
+  /**
+   * Scala API: Creates [[DynamoSettings]] from a [[com.typesafe.config.Config Config]]
+   */
+  def apply(baseConfig: Config): DynamoSettings = {
+    val config = baseConfig.getConfig("akka.stream.alpakka.dynamodb")
```
ennru: Please pass in the resolved config instance instead so that the user may choose the location in the settings to allow for several configurations.

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/impl/DynamoSettings.scala
```diff
@@ -5,20 +5,59 @@
...
+    val awsCredentialsProvider = Try(config.getConfig("credentials"))
```
ennru: Please test existence with `hasPath` instead.

randomcoder: Is there a specific reason for this? `Try().getOrElse()` does the same thing and `hasPath` forces an `if`. I'm quite happy to, just curious.

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/impl/DynamoSettings.scala
randomcoder: Passed in config resolved from `akka.stream.alpakka.dynamodb`

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/impl/DynamoSettings.scala
randomcoder: Switch to using `hasPath`

randomcoder: Doing this with `hasPath` also breaks the fallback to default credentials is the config is not fully defined, i.e. missing the `credentials.secret-key-id` which is handled tidily with the `Try` form.
If you want me to check both paths explicitly then I can do so, but it seems much more long winded to do so.

docs/src/main/paradox/dynamodb.md
```diff
@@ -18,9 +18,22 @@ The AWS DynamoDB connector provides a flow for streaming DynamoDB requests. For
 
 ## Usage
 
-This connector uses the [default credential provider chain](http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html) provided by the [DynamoDB Java SDK](http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/basics.html) to retrieve credentials.
+This connector will uses the [default credential provider chain](http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html)
```
2m: s/uses/use

ennru:
Thank you for this contribution.

PR #914, opened 2018-04-26
Run git to describe the current git commit

2m:
Using "--all" makes it possible to determine if the current commit is the tip of the master branch even in Travis which checkouts a particular sha of a commit.

@raboof I think this might work in Travis environment and does not depend on sbt-git. WDYT? :)


2m:
CI failure was https://github.com/akka/alpakka/issues/915

PR #925, opened 2018-05-06
Improve AQMP channels shutdown listeners

juanjoDiaz:
I noticed some stuff that was wrong with the channel shutdown listeners

### DIFF
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
@@ -53,14 +53,6 @@ final class AmqpRpcFlowStage(settings: AmqpSinkSettings, bufferSize: Int, respon
 
       override def whenConnected(): Unit = {
         import scala.collection.JavaConverters._
-        val shutdownCallback = getAsyncCallback[(String, Option[ShutdownSignalException])] {
-          case (consumerTag, Some(e)) =>
-            onFailure(
-              new RuntimeException(s"Consumer $queueName with consumerTag $consumerTag shut down unexpectedly", e)
-            )
-          case (consumerTag, None) =>
-            onFailure(new RuntimeException(s"Consumer $queueName with consumerTag $consumerTag shut down unexpectedly"))
-        }
 
         pull(in)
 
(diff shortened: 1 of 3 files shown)


amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpRpcFlowStage.scala
```diff
@@ -53,14 +53,6 @@ final class AmqpRpcFlowStage(settings: AmqpSinkSettings, bufferSize: Int, respon
 
       override def whenConnected(): Unit = {
         import scala.collection.JavaConverters._
-        val shutdownCallback = getAsyncCallback[(String, Option[ShutdownSignalException])] {
```
juanjoDiaz: This was unnecessary since we already have the `shutdownCallback` in the `ConnectorLogic`. Updated the calls to invoke the right method.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSourceStage.scala
```diff
@@ -58,10 +58,6 @@ final class AmqpSourceStage(settings: AmqpSourceSettings, bufferSize: Int)
         // we have only one consumer per connection so global is ok
         channel.basicQos(bufferSize, true)
         val consumerCallback = getAsyncCallback(handleDelivery)
-        val shutdownCallback = getAsyncCallback[Option[ShutdownSignalException]] {
```
juanjoDiaz: Same as in the RPC stage.

I modify the exception messages to be consistent in both stages.

Also, `if (unackedMessages == 0) completeStage()` is wrong logic since once the channel has been shut down, no more messages can be acked so all that's left is failing.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala
```diff
@@ -60,18 +59,7 @@ final class AmqpSinkStage(settings: AmqpSinkSettings)
       private val exchange = settings.exchange.getOrElse("")
       private val routingKey = settings.routingKey.getOrElse("")
 
-      override def whenConnected(): Unit = {
```
juanjoDiaz: This is 100% unnecessary since we are already creating a shutdown connector on the `ConnectorLogic` doing exactly the same.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSinkStage.scala
```diff
@@ -145,18 +133,7 @@ final class AmqpReplyToSinkStage(settings: AmqpReplyToSinkSettings)
     (new GraphStageLogic(shape) with AmqpConnectorLogic {
       override val settings = stage.settings
 
-      override def whenConnected(): Unit = {
```
juanjoDiaz: Same as above.

2m:
Validation failure was: https://github.com/akka/alpakka/issues/927

ennru:
Thank you @juanjoDiaz for keeping an eye on Alpakka!

PR #932, opened 2018-05-07
MQTT: Improve test stability

ennru:
Separate topics between tests better so they do not interfere.
As it turned out the for local tests the docker container had to be cleaned up between two runs, otherwise it would fail for old (unacknowledged) data popping up.
Made sure Java tests and Scala tests don't use the same topic names.

### DIFF
--- a/mqtt/src/test/java/akka/stream/alpakka/mqtt/javadsl/MqttSourceTest.java
+++ b/mqtt/src/test/java/akka/stream/alpakka/mqtt/javadsl/MqttSourceTest.java
@@ -213,8 +213,8 @@ public class MqttSourceTest {
 
   @Test
   public void supportWillMessage() throws Exception {
-    String topic1 = "source-spec/topic1";
-    String willTopic = "source-spec/will";
+    String topic1 = "source-test/topic1";
+    String willTopic = "source-test/will";
     final MqttConnectionSettings baseConnectionSettings = MqttConnectionSettings.create(
             "tcp://localhost:1883",
             "test-java-client",
--- a/mqtt/src/test/java/akka/stream/alpakka/mqtt/javadsl/MqttSourceTest.java
+++ b/mqtt/src/test/java/akka/stream/alpakka/mqtt/javadsl/MqttSourceTest.java
@@ -234,8 +234,9 @@ public class MqttSourceTest {
     //#will-message
 
     // Create a proxy to RabbitMQ so it can be shutdown
+    int proxyPort = 1347; // make sure to keep it separate from ports used by other tests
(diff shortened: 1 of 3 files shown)


mqtt/src/test/java/akka/stream/alpakka/mqtt/javadsl/MqttSourceTest.java
```diff
@@ -213,8 +213,8 @@ public void receiveFromMultipleTopics() throws Exception {
 
   @Test
   public void supportWillMessage() throws Exception {
-    String topic1 = "source-spec/topic1";
-    String willTopic = "source-spec/will";
+    String topic1 = "source-test/topic1";
+    String willTopic = "source-test/will";
```
2m: Good! I remember doing this to other Java tests some time ago. We need to keep an eye on this when adding new MQTT tests.

2m:
Nice find with the overlaping ports! That will surely make it more stable.

2m:
You mentiond that tests error out if the docker container is not restarted between the tests. I just realized that is exactly what happens when running tests on Travis. First all of the tests run using scala 2.11 and then the same tests run with 2.12.

It was not like that before we moved to Travis stages. Before that there were two Travis jobs, one for 2.11 and another for 2.12.

If cleaning up properly is impossible/too much work, we could add something to sbt that starts/stops docker containers during tests. Could solve this issue, and clean up the current `.travis.yml`

ennru:
No problem anymore - with the changes I did the tests do not interfere in the bad way they did before.

mqtt/src/test/java/akka/stream/alpakka/mqtt/javadsl/MqttSourceTest.java
```diff
@@ -252,7 +253,7 @@ public void supportWillMessage() throws Exception {
     MqttSourceSettings settings1 = MqttSourceSettings.create(
       sourceSettings
         .withClientId("source-spec/testator")
-        .withBroker("tcp://localhost:1337")
+        .withBroker("tcp://localhost:" + String.valueOf(proxyPort))
```
juanjoDiaz: This would be a bit cleaner as `s"tcp://localhost:$proxyPort"`

ennru: Yes, but Java does not do that yet, AFAIK ;-)

juanjoDiaz: Oh, my bad! Didn't realize it was the java test.

LGTM, anyway :)

Issue #933, opened 2018-05-07
AMQP: Support for 3.1


giftig:
Hi, after a bit of playing around with using alpakka to connect to ActiveMQ, I discovered that the AMQP protocol used by `alpakka-stream-alpakka-amqp` doesn't match the protocol version used by ActiveMQ:

```
Caused by: com.rabbitmq.client.MalformedFrameException: AMQP protocol version mismatch; we are version 0-9-1, server sent signature 3,1,0,0
```
It seems AMQ doesn't support AMQP version 0.9.1, and I don't see anything in the alpakka documentation about different versions. Is this a known issue? Am I using the wrong connector?

giftig:
I just spotted the clue in the error message, `com.rabbitmq.client.MalformedFrameException`. I take it the AMQP support is more specifically aimed at rabbitmq. I don't see anything about using a different underlying client, though, so is ActiveMQ simply not supported?

juanjoDiaz:
The RabbitMQ client that we use works for any AMQP 0.9.1-compliant server, not only RabbitMQ.

However, it seems that ActiveMQ is using AMQP 1.0 and that's not supported at the moment.

giftig:
Indeed, I also checked it from the other direction but it appears AMQ doesn't support 0.9.1 and has no intention of back-porting older protocol versions: http://activemq.2283324.n4.nabble.com/jira-Closed-AMQ-5332-Support-AMQP-0-9-1-td4686862.html

Since I'm playing around with alpakka in a new project and can choose whichever message queue I want, I simply used RabbitMQ instead of ActiveMQ. It'd be nice to have the more recent AMQP version supported, though.

juanjoDiaz:
It would be nice indeed.

I think that the problem is that AMQP 1.0 is not "just the next version". It's pretty much a new protocol. So some brokers like RabbitMQ decided to stay on 0.9.

Maybe you can change the title to `Add Support for AMQP 1.0` or something like that to make the issue more clear.

giftig:
I'm not sure why AMQ announces its version as 3,1,0,0 if it's AMPQ 1.0, but done. I guess the numbering totally changed with specification v1, and it's not "just a new version", like you said.

PR #947, opened 2018-05-10
Document AQMP versions support

juanjoDiaz:
Documents the issue mentioned in #933

### DIFF
--- a/docs/src/main/paradox/amqp.md
+++ b/docs/src/main/paradox/amqp.md
@@ -1,6 +1,8 @@
 # AMQP Connector
 
-The AMQP connector provides Akka Stream sources and sinks to connect to AMQP servers.
+The AMQP connector provides Akka Stream sources and sinks to connect to AMQP 0.9.1 servers (RabbitMQ, OpenAMQ, etc.).
+
+AMQP 1.0 is currently not supported (Qpid, ActiveMQ, Solace, etc.).
 
 ## Reported issues
 


juanjoDiaz:
I wanted to add a list of AMQP 0.9.1 brokers and 1.0 brokers (since there are more than RabbitMQ and ActiveMQ) but I haven't been able to find a comprehensive list...

I've added at least some names in there :)

giftig:
Might also be worth adding a hint about alternatives to using AMQP with these queues - for example AMQ and Solace both support half a dozen different protocols, including REST interfaces, though whether alpakka can help with any of those protocols I don't know as I'm just trying it out for the first time now. (of course, akka-http is perfectly viable to hit a REST interface though)

Thanks for the quick turnaround on the issue, by the way.

juanjoDiaz:
I'm not sure if it's in the scope of Alpakka's documentation to go deeper on the different MQ brokers and the protocols they support.

We provide a connector to AMQP 0.9.1 and we make a particular note that 1.0 is not supported to avoid unclarity. But I wouldn't go any deeper than that.

Btw, as a side note, Alpakka provides an MQTT connector among others so it's definitely worth looking into them.

giftig:
Fair enough; I thought it might be useful to offer a hint of where else to look if the right AMQP version isn't supported, but given there may be a growing list of AMQP 1.0 applications there, each with their own alternative protocols, that may indeed be too much info.

Issue #944, opened 2018-05-09
S3Connector: Support for download objects using versioning


sfali:
In AWS S3 if a bucket has versioning enable, object can be downloaded using specific version. S3 connector should be supporting downloading of object by specific version, if provided. More over version id should be returned from any put operation.

PR will follow.

ennru:
Fixed with #951

Issue #938, opened 2018-05-08
AWS S3: prepare docker image to run Minio tests


ennru:
The AWS S3 connector code even contains integration tests that can run against Minio.
We could prepare a docker image and run those tests in the CI cycle, as well.
Instructions to run it are found in the `MinioS3IntegrationSpec` comment.

ennru:
@sfali Would be great if you could PR the Docker config you used with #951 and propose a way to add and remove buckets for testing.

sfali:
Sure, I will.

2m:
Fixed in #1395

PR #951, opened 2018-05-13
AWS S3: versioning support

sfali:
Support for download, upload, copy, and delete for a specific version, when the bucket is enabled to support versioning.

#944


docs/src/main/paradox/s3.md
```diff
@@ -93,9 +101,15 @@ Copy an S3 object from source bucket to target bucket using multi part copy uplo
 Scala
 : @@snip ($alpakka$/s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3SinkSpec.scala) { #multipart-copy }
 
+Scala
+: @@snip ($alpakka$/s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3SinkSpec.scala) { #multipart-copy-with-source-version }
+
```
ennru: You should make this two snippet sections, and add some text explaining what it does.

s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3Client.scala
```diff
@@ -195,10 +223,14 @@ final class S3Client(s3Settings: S3Settings, system: ActorSystem, mat: Materiali
    */
   def request(bucket: String,
               key: String,
+              versionId: String,
```
ennru: Wouldn't it be better to go with Optional instead of a nullable String?
We might deprecate the `request` without the parameter.

s3/docker-compose.yaml
```diff
@@ -0,0 +1,54 @@
+version: '3.1'
+
+services:
+  minio1:
```
ennru: Did you implement integration tests against these minio servers?
I so, please send a separate PR with those and remove this section for now.

sfali: Yes I ran against integration tests these minio servers. I noticed that you have an issue, #938, open to run integration tests in CI, in that case there should be a way to create buckets on startup and delete once done.

I'll remove and will send new PR for this.

sfali:
As for creating a PR for docker compose file, as I mentioned in my comments if you are planning to run in CI/CD pipeline then functions need to be added to create and delete buckets at the start and end respectively, let me know about your thoughts.

ennru:
Thank you for growing the S3 connector even more.

Issue #557, opened 2017-11-01
Connector for HDFS


ennru:
Especially writing to HDFS with some file split policy would be interesting to support in Alpakka.
There is a Camel component for more ideas on requirements: `https://camel.apache.org/hdfs2.html`

ennru:
The file split could e.g. be done in a similar way as in #210 logrotation file sink.

burakkose:
If this is still not assigned to anybody, I would like to scratch something. I am planning to implement 'file size rotation,' 'timed rotation' and 'no rotation (basically full dump).'

burakkose:
Just a quick update. I am still working on it. I completed the structure and proof of concept. I need to find time to test and reorganize the project. I will try to find time soon.

Here is my draft repository for it.
https://github.com/burakkose/hdfs-connector

Eventually, I will create a PR.

ennru:
Hi @burakkose 
How far did you get with your HDFS connector? Would be good to get some basic version into Alpakka soonish.

burakkose:
Hey @ennru 
Sorry for the late response. I was incredibly busy. I am working on it. The first PR will only contain writing functionality. Thanks for understanding.

2m:
Implemented in https://github.com/akka/alpakka/pull/965

PR #965, opened 2018-05-21
HDFS: Add sources and flows

burakkose:
Ref: #557

In this pull request, you will find the initial work for Hdfs. The PR is work in progress. There are still some TODOs.

- Documentation (**done**)
- Java Tests (**done**)
- Support HDFS version 2.x (**done**)
- pass-through (**done**)

While I am working on these, please review the code, suggest new functionalities, and help for testing.


hdfs/src/main/scala/akka/stream/alpakka/hdfs/HdfsFlowStage.scala
```diff
@@ -0,0 +1,197 @@
+/*
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.hdfs
```
ennru: It would be great to move internal stuff into an `impl` package. That would improve Java Module readiness.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/javadsl/HdfsSink.scala
```diff
@@ -0,0 +1,123 @@
...
+  /*
+   * Java API: creates a Sink with [[HdfsFlowStage]] for [[CompressionOutputStream]]
+   *
+   * @param fs HDFS FileSystem
+   * @param syncStrategy Sync Strategy
+   * @param rotationStrategy Rotation Strategy
+   * @param compressionCodec a class encapsulates a streaming compression/decompression pair.
+   * @param settings Hdfs writing settings
+   */
+  def compressed(
```
ennru: To reduce the API, you could just have `HdfsSink.data` and let the users connect to `Sink.ignore` for the other cases.

project/Dependencies.scala
```diff
@@ -156,6 +156,21 @@ object Dependencies {
     )
   }
 
+  val Hdfs = {
+    val hadoopVersion = "3.1.0"
+    val catsVersion = "1.0.1"
+    Seq(
+      libraryDependencies ++= Seq(
+        "org.apache.hadoop" % "hadoop-client" % hadoopVersion, // ApacheV2
+        "org.typelevel" %% "cats-core" % catsVersion, // MIT,
```
ennru: You might upgrade to Cats 1.1.0.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/javadsl/HdfsSink.scala
burakkose: Sorry I could not understand. `HdfsSink.data` , `HdfsSink.compressed`, `HdfsSink.sequence` are completely different. What exactly do you mean by letting users connect to `Sink.ignore`

burakkose:
I am not sure about the HDFS version. How can we provide support for 2.x also? Is there any example for it in Alpakka connectors?

hdfs/src/main/scala/akka/stream/alpakka/hdfs/javadsl/HdfsSink.scala
ennru: I think users needing these cases might as well use the `HdfsFlow` and connect it to `Sink.ignore` themselves.

ennru:
No there is no example right now. One way of doing it would be to make the HDFS dependency optional and let the users add their version explicitly.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/javadsl/HdfsSink.scala
burakkose: oh, then can we remove Sink implementations?

burakkose:
I have rearranged the project structure with `impl` package. Shall we move `FilePathGenerator`, `RotationStrategy`, `SyncStrategy` to `model.scala`.

Can you also check the commit https://github.com/akka/alpakka/pull/965/commits/b687ad53576c597eef154a914101fe683f55bd3a , I have realized that we need to use `japi.Pair` for Java

hdfs/src/main/scala/akka/stream/alpakka/hdfs/javadsl/HdfsSink.scala
ennru: Yes, just keep the most basic ones. Most users discover they'll need to do something after sending/writing to the destination.
Speaking of that, what we most often need is some kind of pass-through, a value untouched by the flow, but available afterwards (eg. Kafka offsets for committing). Have you thought about that? It would require a wrapper for `Writable`, AFAICS.

ennru:
The build error is an extra comma which is not supported in Scala 2.11 in `HdfsWriterSpec.scala:463`.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/javadsl/HdfsSink.scala
burakkose: Actually, I was thinking about `pass-through` like Solr connector. I was just not sure to implement. It will be added soon.

ennru:
We build on Scala 2.11 and 2.12. Since 2.12 you're allowed to have a superfluous comma as in
```
Seq(
  1,
  2,
)
```
and that happens to be in there in `HdfsWriterSpec.scala:463`.

burakkose:
Yes, my late time commit fault :)

hdfs/src/main/scala/akka/stream/alpakka/hdfs/javadsl/HdfsSink.scala
burakkose: I would like to consult about it. I was thinking the best strategy for it. Currently, the flow expects `ByteString` as an input. Whenever it rotates the output, it pushes `WriteLog` to the downstream.

If we want to implement `pass-through,` how shall we design it? The dummies way is
```
final case class WriteLog[T](path: String, rotation: Int, passThroughs: Seq[T])
```
However, if we have millions input, keeping this sequence in memory until the flow rotates is super inefficient. 

The second idea is
```
sealed trait OutgoingMessage
final case class RotationMessage(path: String, rotation: Int) extends OutgoingMessage
final case class PassThrough[T](pass: T)  extends OutgoingMessage
```
So flow can push `RotationMessage` when it rotates, and for the rest, it will push `PassThrough`. However, this has a drawback. Let's talk about Kafka example, we write the input to output and send `PassThrough` message with an offset. If something goes wrong in the flow, and if it does not synchronize the output, we will basically fail, but downstream can already commit this offset.

I actually did not like this idea also. Do you have any idea for it?

ennru:
Emitting a message for every incoming is the only reasonable way. 
The user may accumulate pass-throughs if needed. With Kafka the offsets would be committed on a Rotation/Write message.
An alternative could be to have a type
```scala
case class HdfsWritten[T](passThrough: T, status: Option[RotatationMessage])
```

burakkose:
I have added Sources and made `RotationStrategy` and `SyncStrategy` extendable.

docs/src/main/paradox/hdfs.md
```diff
@@ -0,0 +1,234 @@
...
+Scala
+: @@snip ($alpakka$/hdfs/src/test/scala/akka/stream/alpakka/hdfs/HdfsWriterSpec.scala) { #define-data }
+
+Java
+: @@snip ($alpakka$/hdfs/src/test/java/akka/stream/alpakka/hdfs/HdfsWriterTest.java) { #define-data }
+
+
+### Compressed Data Writer
+
+First, create `CompressionCodec`.
```
ennru: By adding 
```
"javadoc.org.apache.hadoop.base_url" -> s"https://hadoop.apache.org/docs/r${hadoopVersion}/api/",
```
in `build.sbt` you can create links to Hadoop's API via `@javadoc`.

docs/src/main/paradox/hdfs.md
```diff
@@ -0,0 +1,234 @@
...
+
+Java
+: @@snip ($alpakka$/hdfs/src/test/java/akka/stream/alpakka/hdfs/HdfsWriterTest.java) { #define-settings }
+
+
+### File path generator
+
+`FilePathGenerator` provides a functionality to generate rotation path in HDFS. 
+@scala[@scaladoc[FilePathGenerator](akka.stream.alpakka.hdfs.FilePathGenerator$).]
+@java[@scaladoc[FilePathGenerator](akka.stream.alpakka.hdfs.FilePathGenerator$).]
```
ennru: No need to use `@scala/@java` when linking to the same class.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/impl/HdfsFlowStage.scala
```diff
@@ -0,0 +1,217 @@
...
+import akka.stream.alpakka.hdfs.impl.strategy.DefaultRotationStrategy.TimeRotationStrategy
+import akka.stream.alpakka.hdfs.impl.writer.HdfsWriter
+import akka.stream.stage._
+import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
+import cats.data.State
+
+/**
+ * Internal API
+ */
+private[hdfs] final class HdfsFlowStage[W, I, C](
```
ennru: Please add even `@akka.annotation.InternalApi` as `private[hdfs]` doesn't protect from Java users using it.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/model.scala
```diff
@@ -0,0 +1,166 @@
...
+   * Java API
+   */
+  def create(): HdfsWritingSettings =
+    HdfsWritingSettings()
+}
+
+final case class IncomingMessage[T, C](source: T, passThrough: C)
+
+object IncomingMessage {
+  // Apply method to use when not using passThrough
```
ennru: Make these doc comments, please.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/model.scala
```diff
@@ -0,0 +1,166 @@
...
+    FilePathGenerator((rc: Long, _: Long) => s"/tmp/alpakka/$rc")
+
+  /**
+   * Java API
+   */
+  def create(): HdfsWritingSettings =
+    HdfsWritingSettings()
+}
+
+final case class IncomingMessage[T, C](source: T, passThrough: C)
```
ennru: I'm not super happy with the `IncomingMessage` name used in several connectors. It can easily become messy when using several connectors in the same codebase. And "incoming" is tied to the point of view of the stage, for the user the data leaves...
Maybe `HdfsWriteMessage` or `HdfsData`?

hdfs/src/test/scala/akka/stream/alpakka/hdfs/HdfsWriterSpec.scala
```diff
@@ -0,0 +1,501 @@
...
+      val logs = Await.result(resF, Duration.Inf)
+      logs.size shouldEqual 1
+
+      val files = getFiles(fs)
+      files.size shouldEqual logs.size
+      files.head.getLen shouldEqual books.map(_.toArray.length).sum
+    }
+
+    "kafka-example - store data with passThrough" in {
+      //#define-kafka-classes
```
ennru: Move the documentation examples to a package like `docs.scaladsl` and `docs.javadsl` to make sure everything you show is accessible.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/model.scala
```diff
@@ -0,0 +1,166 @@
...
+   * Creates [[SizeRotationStrategy]]
+   *
+   * @param count a count of [[FileUnit]]
+   * @param unit a file unit
+   */
+  def size(count: Double, unit: FileUnit): RotationStrategy =
+    SizeRotationStrategy(0, count * unit.byteCount)
+
+  /**
+   * Creates [[CountRotationStrategy]]
```
ennru: The docs would try to link to private classes, which are not included in the scaladocs.
Describe the strategy instead.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/scaladsl/HdfsFlow.scala
```diff
@@ -0,0 +1,231 @@
...
+import org.apache.hadoop.io.{SequenceFile, Writable}
+
+object HdfsFlow {
+
+  private[hdfs] val OnlyRotationMessage: PartialFunction[OutgoingMessage[_], RotationMessage] = {
+    case m: RotationMessage => m
+  }
+
+  /**
+   * Scala API: creates a Flow with [[HdfsFlowStage]] for [[FSDataOutputStream]]
```
ennru: Don't link to the stage. Users should not need to care.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/scaladsl/HdfsFlow.scala
```diff
@@ -0,0 +1,231 @@
...
+
+  /**
+   * Scala API: creates a Flow with [[HdfsFlowStage]] for [[FSDataOutputStream]] with `passThrough` of type `C`
+   *
+   * @param fs file system
+   * @param syncStrategy sync strategy
+   * @param rotationStrategy rotation strategy
+   * @param settings hdfs writing settings
+   */
+  def dataWithPassThrough[C](
```
ennru: Would `P` make an easier type parameter name?

hdfs/src/main/scala/akka/stream/alpakka/hdfs/model.scala
```diff
@@ -0,0 +1,166 @@
...
+
+  // Java-api - with passThrough
+  def create[T, C](source: T, passThrough: C): IncomingMessage[T, C] =
+    IncomingMessage(source, passThrough)
+
+}
+
+sealed abstract class OutgoingMessage[+T]
+final case class RotationMessage(path: String, rotation: Int) extends OutgoingMessage[Nothing]
+final case class WrittenMessage[T](passThrough: T, inRotation: Int) extends OutgoingMessage[T]
```
ennru: Document the `inRotation` value.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/impl/writer/HdfsWriter.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  protected def pathGenerator: FilePathGenerator
+
+  protected def create(fs: FileSystem, file: Path): W
+
+}
+
+private[writer] object HdfsWriter {
+
+  val NewLineByteArray: Array[Byte] = ByteString(System.getProperty("line.separator")).toArray
```
ennru: Is it useful to use the system's separator? It could be provided via the settings instead.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/impl/HdfsFlowStage.scala
```diff
@@ -0,0 +1,217 @@
...
+    state = onPushProgram(grab(inlet))
+      .runS(state)
+      .value
+
+  def onPull(): Unit =
+    tryPull()
+
+  override def preStart(): Unit = {
+    // Schedule timer to rotate output file
+    initialRotationStrategy match {
```
ennru: This should be expressed by something in `RotationStrategy` instead of the type so it becomes extendible.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/javadsl/HdfsFlow.scala
```diff
@@ -0,0 +1,240 @@
...
+import org.apache.hadoop.io.Writable
+import org.apache.hadoop.io.compress.CompressionCodec
+
+object HdfsFlow {
+
+  /**
+   * Java API: creates a Flow with [[akka.stream.alpakka.hdfs.impl.HdfsFlowStage]]
+   * for [[org.apache.hadoop.fs.FSDataOutputStream]]
+   *
+   * @param fs file system
```
ennru: **Hadoop** file system

hdfs/src/main/scala/akka/stream/alpakka/hdfs/impl/HdfsFlowStage.scala
burakkose: Do you have any hint about this? How can i have a strategy in RotationStrategy for preStart. I would consider to pass stateLogic as a parameter and call schedule there but these methods are not visible. How can i trigger schedule in RotationStrategy.

ennru: You could add an `interval: Option[FiniteDuration]` to it and use `foreach` in pre-start to schedule the poll it.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/impl/HdfsFlowStage.scala
```diff
@@ -0,0 +1,219 @@
...
+    rs: RotationStrategy,
+    settings: HdfsWritingSettings,
+    hdfsWriter: HdfsWriter[W, I]
+) extends GraphStage[FlowShape[HdfsWriteMessage[I, C], OutgoingMessage[C]]] {
+
+  private val in = Inlet[HdfsWriteMessage[I, C]](Logging.simpleName(this) + ".in")
+  private val out = Outlet[OutgoingMessage[C]](Logging.simpleName(this) + ".out")
+
+  override val shape: FlowShape[HdfsWriteMessage[I, C], OutgoingMessage[C]] = FlowShape(in, out)
+
```
2m: As the underlying writing abstraction is `java.io.OutputStream` which is blocking, we have to signal to the Akka Stream materializer that it should materialize HdfsFlowStage to a separate dispatcher. This allows other parts of the stream to continue uninpacted when this stage is going to block a thread during writing operations.

Therefore add the following here:
```
override def initialAttributes: Attributes =
  super.initialAttributes and ActorAttributes.IODispatcher
```

hdfs/src/main/scala/akka/stream/alpakka/hdfs/scaladsl/HdfsSource.scala
```diff
@@ -0,0 +1,76 @@
...
+    val it = Iterator
+      .continually {
+        val key = classK.newInstance()
+        val value = classV.newInstance()
+        val hasCurrent = reader.next(key, value)
+        (hasCurrent, (key, value))
+      }
+      .takeWhile(_._1)
+      .map(_._2)
+    Source.fromIterator(() => it)
```
2m: I assume that `SequenceFile.Reader`s `next` operation is blocking as well. We will have to put this source on a separate dispatcher then as well: `.addAttributes(Attributes(ActorAttributes.IODispatcher))`.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/scaladsl/HdfsSource.scala
```diff
@@ -0,0 +1,76 @@
...
+   * @param codec a streaming compression/decompression pair
+   * @param chunkSize the size of each read operation, defaults to 8192
+   */
+  def compressed(
+      fs: FileSystem,
+      path: Path,
+      codec: CompressionCodec,
+      chunkSize: Int = 8192
+  ): Source[ByteString, Future[IOResult]] =
+    StreamConverters.fromInputStream(() => codec.createInputStream(fs.open(path)), chunkSize)
```
2m: A source created by `fromInputStream` already runs on the IO dispatcher, therefore it is fine here.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/impl/HdfsFlowStage.scala
burakkose: Thank you for this comment. While I was running it, i was also profiling the app. I guess this is a nice improvement.

hdfs/src/test/scala/akka/stream/alpakka/hdfs/HdfsWriterSpec.scala
burakkose: Actually, this is not a real example, and it contains test assertions for pass-through logic. I would like to have it as part of tests. Shall we keep it as it is?

burakkose:
@2m, @ennru did you have a chance to review last changes?

hdfs/src/main/scala/akka/stream/alpakka/hdfs/scaladsl/HdfsSource.scala
```diff
@@ -70,7 +71,9 @@ object HdfsSource {
       }
       .takeWhile(_._1)
       .map(_._2)
-    Source.fromIterator(() => it)
+    Source
+      .fromIterator(() => it)
+      .addAttributes(Attributes(IODispatcher))
```
2m: Use `ActorAttributes.IODispatcher` here as well since the `DefaultAttributes.IODispatcher` is in the `impl` package.

2m:
Thank you for the ping. Looking very good. Just one last nitpick an we are good to merge.

burakkose:
Done

PR #971, opened 2018-05-23
A couple of fixes in MQTT tests

2m:
* fixed clientId reuse between java and scala tests
* fixed `pull` call from paho library thread context


2m:
Issues fixed in this PR might not be a source of MQTT test instabilities, but should be fixed nevertheless.

One of the more serious sources of MQTT test instabilities has been fixed in https://github.com/akka/akka/pull/25135

Issue #976, opened 2018-05-25
It should be possible to enable explicit ssl mode for file transfers with FTPS


anzecesar:
Listing files on a FTPS server that requires encryption for data files fails without setting `ftpClient.execPROT("P")`.

The server returns
```
LIST /
431 The data connection must be SSL-secured. Please configure your client to use explicit or implicit SSL for data connections.
```

of secondary concern is also, that this error stays silent and the stream terminates with success, with 0 files listed.

I would like to add it to [FtpOperations trait](https://github.com/akka/alpakka/blob/a24bbb40f246ab61cd5be34527429c8991de88f1/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/FtpOperations.scala#L19-L32), but that requires possibly breaking changes to the method signature. 

FtpSettings are then not really common between FTP and FTPS, if `FtpsSettings` gain e.g. `explicitSSL`.

I'm happy to do it, if someone can point me in the right direction.

anzecesar:
I've implemented this here: https://github.com/akka/alpakka/pull/977

ennru:
Fixed with #977

PR #975, opened 2018-05-25
FTP Use FTPSClient for ftps

anzecesar:
I ran into issues connecting to a ftps server and traced it down to this.

I still have an issue with needing to call `execPROT("P")`, to transfer files over ssl too,
but I'll open an issue for discussion first.

PR #977, opened 2018-05-27
FTP  Added `explicitSSL` parameter to `FtpsSettings` (#976).

anzecesar:
I've tested it agains the server, which requires encrypted data transfer and it works.

I'm not super sure how to add this case to the specs yet.

### DIFF
--- a/ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
+++ b/ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
@@ -94,13 +94,15 @@ object FtpSettings {
  * @param credentials credentials (username and password)
  * @param binary specifies the file transfer mode, BINARY or ASCII. Default is ASCII (false)
  * @param passiveMode specifies whether to use passive mode connections. Default is active mode (false)
+ * @param explicitSSL specifies whether to use ssl encryption for data connections. Default is no (false).
  */
 final case class FtpsSettings(
     host: InetAddress,
     port: Int = FtpsSettings.DefaultFtpsPort,
     credentials: FtpCredentials = AnonFtpCredentials,
     binary: Boolean = false,
-    passiveMode: Boolean = false
+    passiveMode: Boolean = false,
+    explicitSSL: Boolean = false
 ) extends FtpFileSettings {
   def withPort(port: Int): FtpsSettings =
     copy(port = port)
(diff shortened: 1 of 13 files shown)


anzecesar:
This still needs to be rebased, once #975 is merged.

ennru:
I wonder if we could find a good way to hook into the creation of the ftp client so that any special configuration could be added.
For JMS we have a `CustomDestination` which allows for such a hook.

anzecesar:
@ennru Good point. 

I'll take a look at the JMS code, when I get a chance.

ennru:
Yes, a general hook to allow for custom configuration of the FTP client is what I'd like to see. By default, it would just be empty.
What about `configureConnection`?

ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
```diff
@@ -74,6 +77,9 @@ final case class FtpSettings(
 
   def withPassiveMode(passiveMode: Boolean): FtpSettings =
     copy(passiveMode = passiveMode)
+
+  def withConfigureConnection(configureConnection: FTPClient => Unit): FtpSettings =
```
anzecesar: @ennru I'm assuming this is intended mostly for the Java api. Should I make this take a `Function[FTPClient, Unit]`? Or do I need to use `Void`? I never made Java api in Scala before :).

I want to add an example to the docs, but I'm struggling with how to use this in Java 😅

2m: It would be [`java.util.function.Consumer`](https://docs.oracle.com/javase/8/docs/api/java/util/function/Consumer.html)

anzecesar: Tnx. I was unable to convert that to Scala. Luckily, Java's lambda fit in elegantly.

anzecesar:
@ennru lmk if I should make any more changes. I'm using this in my project already, and am happy with it :).

ennru:
But it has a problem with the Java API on Scala 2.11.

ennru:
@anzecesar I added a Java-specific method taking a `j.u.f.Consumer` explicitly to support Java with the Scala 2.11, compiled code.

ennru:
Thank you for your efforts, great to see this improving!

anzecesar:
@ennru Tnx. Happy to contribute :).

PR #978, opened 2018-05-27
Fix exception thrown when AMQP can't connect to the server.

juanjoDiaz:
I noticed that when the connection fails (wrong credentials or whatever) the connection is still null so there is a NullPointerException when trying to remove listeners.

However, the exception is not caught in the tests. Probably because it's thrown in a different thread?
Any advice on how to test or improve this is more than welcome :)


2m:
Nice catch!

There is a `try` in the preStart that catches all of the exceptions. This is why you probably do not see that exception in the tests.

You could create a custom `AmqpConnectionProvider` that returns a null or throws an exception when new connection is requested. And then you could assert that `connectionProvider.release(connection)` got called.

juanjoDiaz:
I'll still try to look into the test at some point. Just been too busy lately :)

Issue #981, opened 2018-05-28
Aplakka Google PubSub with Emulator: unsupported Content-Type exception


damianbl:
I run `akka-stream-alpakka-google-cloud-pub-sub` version 0.19 with Google PubSub Emulator and get this exception:

> akka.http.scaladsl.unmarshalling.Unmarshaller$UnsupportedContentTypeException: Unsupported Content-Type, supported: application/json
> 	at akka.http.scaladsl.unmarshalling.Unmarshaller$UnsupportedContentTypeException$.apply(Unmarshaller.scala:158)
> 	at akka.http.scaladsl.unmarshalling.Unmarshaller$EnhancedFromEntityUnmarshaller$$anonfun$forContentTypes$extension$1$$anonfun$apply$24$$anonfun$apply$25.apply(Unmarshaller.scala:114)
> 	at akka.http.scaladsl.unmarshalling.Unmarshaller$EnhancedFromEntityUnmarshaller$$anonfun$forContentTypes$extension$1$$anonfun$apply$24$$anonfun$apply$25.apply(Unmarshaller.scala:111)
> 	at akka.http.scaladsl.unmarshalling.Unmarshaller$$anon$1.apply(Unmarshaller.scala:58)
> 	at akka.http.scaladsl.unmarshalling.Unmarshaller$$anonfun$transform$1$$anonfun$apply$2$$anonfun$apply$3.apply(Unmarshaller.scala:23)

damianbl:
The actual content type I get is `application/octet-stream`

damianbl:
It fails in GooglePubSubSource:34:

`session.getToken()`

damianbl:
It should be fixed with this PR:

https://github.com/akka/alpakka/pull/982

PR #982, opened 2018-05-28
[Google Cloud PubSub] Fix getting token when running in emulated mode

damianbl:
- it should not try to fetch the token from the session when running in emulated mode
- fixes [981](https://github.com/akka/alpakka/issues/981)

### DIFF
--- a/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSource.scala
+++ b/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSource.scala
@@ -31,13 +31,27 @@ private final class GooglePubSubSource(projectId: String,
       def fetch(implicit mat: Materializer): Unit = {
         import mat.executionContext
 
-        val req = session.getToken().flatMap { token =>
-          httpApi.pull(project = projectId, subscription = subscription, accessToken = token, apiKey = apiKey)
+        if (httpApi.isEmulated) {
+          httpApi
+            .pull(project = projectId, subscription = subscription, maybeAccessToken = None, apiKey = apiKey)
+            .onComplete { tr =>
+              callback.invoke(tr -> mat)
+            }
+          state = Fetching
+        } else {
+          session
+            .getToken()
+            .flatMap { token =>
+              httpApi.pull(project = projectId,
+                           subscription = subscription,
+                           maybeAccessToken = Some(token),
+                           apiKey = apiKey)
+            }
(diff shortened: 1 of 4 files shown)


google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSource.scala
```diff
@@ -31,13 +31,27 @@ private final class GooglePubSubSource(projectId: String,
...
 
-        val req = session.getToken().flatMap { token =>
-          httpApi.pull(project = projectId, subscription = subscription, accessToken = token, apiKey = apiKey)
-        }
-        req.onComplete { tr =>
-          callback.invoke(tr -> mat)
+        if (httpApi.isEmulated) {
+          httpApi
+            .pull(project = projectId, subscription = subscription, maybeAccessToken = None, apiKey = apiKey)
+            .onComplete { tr =>
```
ennru: To make it obvious that the `onComplete` is shared between both branches you might go back to the `val req` by assigning the if clause to it.

damianbl: Sure, changed. Is this what you meant?

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSource.scala
```diff
@@ -31,13 +31,27 @@ private final class GooglePubSubSource(projectId: String,
       def fetch(implicit mat: Materializer): Unit = {
         import mat.executionContext
 
-        val req = session.getToken().flatMap { token =>
-          httpApi.pull(project = projectId, subscription = subscription, accessToken = token, apiKey = apiKey)
+        val req = if (httpApi.isEmulated) {
+          val resp = httpApi
+            .pull(project = projectId, subscription = subscription, maybeAccessToken = None, apiKey = apiKey)
+          state = Fetching
```
ennru: If you move the state change to its original spot it get even better...

damianbl: Fixed.

ennru:
Thank you for fixing this!

damianbl:
Thanks for accepting this. 
When we have it released, it will resolve the last issue I have when integrating with Google PubSub using Alpakka.

Thanks!

PR #989, opened 2018-05-31
fix potential memory leak because retry never ends

CatTail:


CatTail:
@2m can you help to merge and publish a new version of alpakka kinesis library

2m:
I rebased this PR ontop of the current master and will merge this when Travis build finishes.

After merge, there will be a snapshot version published automatically. Will you be able to use the snapshot version?

PR #995, opened 2018-06-03
Udp connector

2m:
Based on UDP from akka.io

Continuation of @radekg previous attempt.

Fixes https://github.com/akka/akka/issues/17185

Missing docs.


radekg:
@2m done, thanks for a ping! Not sure who / how to trigger the checks.

huntc:
Should this not be modeled on the Akka stream Tcp interface? (https://doc.akka.io/japi/akka/current/akka/stream/scaladsl/Tcp.html)

For consistency, I modelled the Unix Domain Socket interfaces of Tcp. 

I accept that there will be some differences with UDP though.

2m:
@huntc yes, it is different because the protocol semantics are very different. For example the `bind` flow in TCP materializes to a source of incoming connections. There is no corresponding concept in UDP.

When it comes to the outgoing connection flow in TCP, a similar shape factory exists in the UDP connector, but again, there is no connection concept.

I say the biggest difference is that lack of the connection concept. Which also means that every incoming or outgoing message in the UDP connector carries the socket address as well.

huntc:
Sure, just raising for where there are similarities, may be we should be consistent. I’ve not looked at this in any detail to be more specific. Thanks.

udp/src/main/scala/akka/stream/alpakka/udp/impl/UdpBind.scala
```diff
@@ -0,0 +1,98 @@
...
+
+import akka.actor.{ActorRef, ActorSystem}
+import akka.io.{IO, Udp}
+import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
+import akka.stream.alpakka.udp.UdpMessage
+import akka.stream.stage._
+
+import scala.concurrent.{Future, Promise}
+
+class UdpBindLogic(localAddress: InetSocketAddress, boundPromise: Promise[InetSocketAddress])(
```
huntc: A top level comment for type would be great. 

Should be private/marked internal?

udp/src/main/scala/akka/stream/alpakka/udp/scaladsl/Udp.scala
```diff
@@ -0,0 +1,43 @@
...
+  /**
+   * Creates a flow that upon materialization binds to the given `localAddress`. All incoming
+   * messages to the `localAddress` are emitted from the flow. All incoming messages to the flow
+   * are sent to the remote address contained in the message.
+   */
+  def bindFlow(
+      localAddress: InetSocketAddress
+  )(implicit sys: ActorSystem): Flow[UdpMessage, UdpMessage, Future[InetSocketAddress]] =
+    Flow.fromGraph(new UdpBindFlow(localAddress))
+}
```
huntc: Why not just send/bind as names?

radekg: @huntc this story hangs on github for so long that I've completely forgot why I used such names. It probably is a good idea but I'm wondering what's the point in investing more time into this while there is zero pick up from the maintainers to get this in.

That's not to say there's no will to improve, the reason why I've abandoned work on this was because there was no indication from the maintainers on the desired direction. If there is the will to get this in, one will find the time to resolve review requests.

2m: Hi @radekg. This PR is an indication of wanting to get this in. :)

@huntc having "flow"/"sink" in the factory methods is a pattern that has been emerging in various Alpakka connectors. It is especially useful when there is a need to differentiate a factory based on the return type but which takes exactly the same arguments.

However I aggree that "fireAndForget" might be a bit too verbose. Having "sendFlow", "sendSink" and "bindFlow" sounds good to me.

radekg: I have no problems with such modifications. Apologies if my communication from yesterday sounded harsh.

2m:
Adressed feedback regarding scaladoc, naming and documentation.

docs/src/main/paradox/udp.md
```diff
@@ -0,0 +1,67 @@
+# UDP
+
+The UDP connector provides Akka Stream flows that allow to send and receive UDP messages.
```
ennru: Wouldn't "datagram" be the technically correct term instead of "messages"?

docs/src/main/paradox/udp.md
```diff
@@ -0,0 +1,67 @@
...
+: @@snip ($alpakka$/udp/src/test/scala/akka/stream/alpakka/udp/UdpSpec.scala) { #bind-address }
+
+Java
+: @@snip ($alpakka$/udp/src/test/java/akka/stream/alpakka/udp/UdpTest.java) { #bind-address }
+
+A Flow created from `Udp.bindFlow` will bind to the given address. All messages coming from the network
+to the bound address will be sent downstream. Messages received from the upstream will be sent to their
+corresponding destination addresses.
+
+The flow materializes to the @scala[`Future[InetSocketAddress]`]@java[`CompletionStage[InetSocketAddress]`] which
```
ennru: Java uses "crocodile brackets" for Generics...

udp/src/main/scala/akka/stream/alpakka/udp/Model.scala
```diff
@@ -0,0 +1,18 @@
...
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.udp
+
+import java.net.InetSocketAddress
+
+import akka.util.ByteString
+
+final case class UdpMessage(data: ByteString, remote: InetSocketAddress)
```
ennru: `Datagram`? Make it MiMa friendly?

udp/src/main/scala/akka/stream/alpakka/udp/impl/UdpBind.scala
```diff
@@ -0,0 +1,91 @@
...
+    extends GraphStageLogic(shape) {
+
+  private def in = shape.in
+  private def out = shape.out
+
+  private var listener: ActorRef = _
+
+  override def preStart(): Unit = {
+    implicit val sender = getStageActor(processIncoming).ref
+    IO(Udp) ! Udp.Bind(sender, localAddress)
```
ennru: Would this stage need some binding timeout support?

udp/src/main/scala/akka/stream/alpakka/udp/javadsl/Udp.scala
```diff
@@ -0,0 +1,43 @@
...
+import scala.compat.java8.FutureConverters._
+
+object Udp {
+
+  /**
+   * Creates a flow that will send all incoming [UdpMessage] messages to the remote address
+   * contained in the message. All incoming messages are also emitted from the flow for
+   * subsequent processing.
+   */
+  def sendFlow(sys: ActorSystem): Flow[UdpMessage, UdpMessage, NotUsed] =
```
ennru: Spell out `system`, please.

udp/src/test/java/akka/stream/alpakka/udp/UdpTest.java
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.udp;
```
ennru: Move documentation code to `docs.javadsl`

udp/src/test/scala/akka/stream/alpakka/udp/UdpSpec.scala
```diff
@@ -0,0 +1,126 @@
+/*
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.udp
```
ennru: ... and `docs.scaladsl`.

udp/src/main/scala/akka/stream/alpakka/udp/impl/UdpBind.scala
2m: I am not sure. If, for some reason, binding fails, the UDP manager actor will send back `Udp.CommandFailed` message, which this stage is going to handle.

2m:
Take additional care when squashing this PR via GitHub UI. Since commits in this PR contain multiple authors, only the last author is considered when squashing via GitHub UI.

PR #1001, opened 2018-06-07
AMQP - allow setting of autoAck on underlying channel.basicConsume call

### LINKED ISSUE #1000: AMQP Fire-and-forget (autoAck) message consumption
It would be good to be able to configure zero-acknowledgement-required message consumption.

Where there are vast quantities of incoming messages and it is not essential to process them all, then it significantly reduces load on the exchange server to use fire-and-forget mode.

https://github.com/akka/alpakka/blob/master/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpSourceStage.scala#L123 could bemade configurable via settings.

Synesso:
Fixes #1000 

The introduced field is `NamedQueueSourceSettings.ackRequired` rather than `autoAck` (being the name of the affected parameter in underlying amqp-client lib) to avoid confusion with the existing autoAck functionality (cc #97).

New integration test only tests that it isn't broken. Given the flag is merely passed to underlying library, I think we can assume that it works as intended.

### DIFF
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
@@ -25,6 +25,7 @@ final case class NamedQueueSourceSettings(
     declarations: immutable.Seq[Declaration] = immutable.Seq.empty,
     noLocal: Boolean = false,
     exclusive: Boolean = false,
+    ackRequired: Boolean = true,
     consumerTag: String = "default",
     arguments: Map[String, AnyRef] = Map.empty
 ) extends AmqpSourceSettings {
@@ -34,6 +35,8 @@ final case class NamedQueueSourceSettings(
   def withNoLocal(noLocal: Boolean): NamedQueueSourceSettings = copy(noLocal = noLocal)
 
   def withExclusive(exclusive: Boolean): NamedQueueSourceSettings = copy(exclusive = exclusive)
+
+  def withAckRequired(ackRequired: Boolean): NamedQueueSourceSettings = copy(ackRequired = ackRequired)
 
   def withConsumerTag(consumerTag: String): NamedQueueSourceSettings = copy(consumerTag = consumerTag)
 
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpSourceStage.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpSourceStage.scala
@@ -114,7 +114,7 @@ private[amqp] final class AmqpSourceStage(settings: AmqpSourceSettings, bufferSi
         def setupNamedQueue(settings: NamedQueueSourceSettings): Unit =
           channel.basicConsume(
             settings.queue,
-            false, // never auto-ack
+            !settings.ackRequired,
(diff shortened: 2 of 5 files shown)


Synesso:
Using this the queue connections are made with Ack Required = false, as confirmed in the cloudamqp console. But the actors receiving messages will only process the first message and not continue to consume further. Something additional is required to make this work, and I'm not sure what. I'm going to investigate further tomorrow. Does anybody have any thoughts on what might be needed?

Synesso:
Added a failing spec to highlight the failure to consume messages after the first batch.

Synesso:
Modified `AmqpSourceStage` to ensure that channel n/ack is not called when source settings indicate ack is not required. I still have the following concerns:

- Do I need to modify the flow stage code, which seems similar?
- Is the implementation, pattern matching on queue setting type, calling for a refactor?
- In practical use, I am still not able to make this work. My queues are terminated after the first batch of messages arrive. There are no errors logged. I do not yet know the cause for this.

ennru:
Hi @Synesso 

Thank you for engaging with the Amqp connector and improving it.

Which side is terminated when it doesn't work for you?

Synesso:
Hi @ennru.

The client remains active and free from errors. After the first batch of messages are read, then the queue is purged from the server. I see it immediately disappear from the rabbit console. 

Although I have since implemented what I need with the raw rabbit driver (thanks to deadlines), I am keen to re-introduce Alpakka. The following client code gives the correct effect. Specifically, a named, temporary queue connected to a permanent fan-out exchange with fire-and-forget messaging.

```
    val handler: ActorRef = ???
    val f = new ConnectionFactory()
    f.setHandshakeTimeout(20000)
    f.setNetworkRecoveryInterval(10000)
    f.setUri(URI)
    val connection = f.newConnection()
    val channel = connection.createChannel()
    val queue = channel.queueDeclare(queueName, false, true, true, null)
    channel.queueBind(queue.getQueue, exchange, "")
    channel.basicConsume(
      name,
      true,
      new DefaultConsumer(channel) {
        override def handleDelivery(consumerTag: String,
                                    envelope: Envelope,
                                    properties: BasicProperties,
                                    body: Array[Byte]): Unit = {
          handler ! new String(body, Codec.UTF8.charSet)
        }
      }
    )
```

ennru:
Sorry, I haven't had the time to look into your specific case. When I looked into the AMQP connector I decided to reorganise it a bit into the pattern we want to follow in all connectors very soon.
Some code you touched moved around a bit, I hope you will be able to get back to it and adopt this PR to these changes by rebasing.

2m:
Hi @Synesso, would be able to rebase this PR on top of the current master?

Synesso:
@2m It's now rebased. The AmqpDocsTest failures are not the same each time, are intermittent & are also present in current master. Even though the test queue names are suffixed with `currentTimeInMillis` I suspect we might be getting name collisions...? I've made each test's queue names distinct.

2m:
I see that even with these topic name changes there are intermittent test failures. Having different topic names is good nevertheless.

So I am going to merge this PR and open a new ticket to track intermittent failures.

2m:
Ahh. I might have jumped the gun here. The new test is also flaky, which you have described above. Sorry for not noticing that. Created ticket: https://github.com/akka/alpakka/issues/1110

PR #1006, opened 2018-06-07
Unix Domain Socket - fix bug causing write side closure of socket to be swallowed

longshorej:
Unix Domain Socket - If writing fails, we need to fail the `sent` promise with the exception so that it can bubble up appropriately. This PR does that.

Additionally, the use of `BroadcastHub` was preventing the sink cancelling the attached `Source`, so now using `Sink.fromGraph(Flow[ByteString]...)`.

### DIFF
--- a/unix-domain-socket/src/main/scala/akka/stream/alpakka/unixdomainsocket/scaladsl/UnixDomainSocket.scala
+++ b/unix-domain-socket/src/main/scala/akka/stream/alpakka/unixdomainsocket/scaladsl/UnixDomainSocket.scala
@@ -19,7 +19,7 @@ import akka.actor.{
   ExtensionIdProvider
 }
 import akka.stream._
-import akka.stream.scaladsl.{BroadcastHub, Flow, Keep, Sink, Source, SourceQueueWithComplete}
+import akka.stream.scaladsl.{Flow, Keep, Sink, Source, SourceQueueWithComplete}
 import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
 import akka.util.ByteString
 import jnr.enxio.channels.NativeSelectorProvider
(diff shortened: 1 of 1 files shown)


longshorej:
cc @huntc

unix-domain-socket/src/main/scala/akka/stream/alpakka/unixdomainsocket/scaladsl/UnixDomainSocket.scala
```diff
@@ -251,47 +252,46 @@ object UnixDomainSocket extends ExtensionId[UnixDomainSocket] with ExtensionIdPr
...
-          sendSource
-            .expand { bytes =>
-              if (bytes.size <= sendBufferSize) {
-                List(bytes).toIterator
+
+    val sendSink = Sink.fromGraph(
+      Flow[ByteString]
+        .expand { bytes =>
+          if (bytes.size <= sendBufferSize) {
+            List(bytes).toIterator
```
viktorklang: There's also: `Iterator.single(bytes)`

longshorej: Thanks! - fixed that. Originally I tried to keep changes not related to the bug to a minimum but might as well. Also changed the collection type there to a `Vector` as we're appending.

I presume `Iterator.single` is special cased and performs a bit better?

viktorklang: @longshorej Fear not! :) In Akka "Apply the Boy Scout Rule whenever you have the chance to." - https://github.com/akka/akka/blob/master/CONTRIBUTING.md#additional-guidelines

It does indeed have a performance difference: List(bytes).toIterator allocates both a new List, then allocates a new Iterator. Iterator.single() only allocates the Iterator.

2m:
Validation failures are due to the Bintray repo problems. Unix domain socket tests completed successfully.

PR #1011, opened 2018-06-12
New module: Text en-/decoding

ennru:
This module offers flows for character set encoding, decoding and transcoding.


build.sbt
```diff
@@ -162,6 +163,8 @@ lazy val sqs = alpakkaProject("sqs",
 
 lazy val sse = alpakkaProject("sse", "sse", Dependencies.Sse)
 
+lazy val text = alpakkaProject("text", "text")
```
2m: Move down a bit.

text/src/main/scala/akka/stream/alpakka/text/impl/CharsetDecodingFlow.scala
```diff
@@ -0,0 +1,23 @@
...
+import java.nio.charset.Charset
+
+import akka.stream.stage.{GraphStage, GraphStageLogic}
+import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
+import akka.util.ByteString
+
+/**
+ * Decodes a stream of bytes into a stream of characters, using a supplied {@link Charset}.
+ */
+private[text] class CharsetDecodingFlow(incoming: Charset) extends GraphStage[FlowShape[ByteString, String]] {
```
2m: InternalApi

text/src/main/scala/akka/stream/alpakka/text/impl/CharsetLogic.scala
```diff
@@ -0,0 +1,194 @@
...
+import java.nio.charset.{Charset, CharsetDecoder, CharsetEncoder}
+
+import akka.stream.{FlowShape, Inlet, Outlet}
+import akka.stream.stage.{GraphStageLogic, InHandler, OutHandler}
+import akka.util.ByteString
+
+/**
+ * ByteBuffer to CharBuffer decoding logic.
+ */
+private[impl] trait Decoding {
```
2m: InternalApi here and below in the file.

build.sbt
ennru: What do you mean?

Voltir:
@ennru - I guess its been awhile since I last checked on the state of alpakka, but better late then never? This library does solve a good chunk of the problem around text encodings and is quite useful in its own right, so thank you for that. 

I still think its problematic that alpakka-csv uses ByteString and Bytes though. - I think there is room in the ecosystem for "CharStrings" and that the various text parsing transforms should really rely on that as a primitive. For example, even today alpakka-csv can't handle certain delimiters, even if the stream is already UTF-8 encoded. At the end of the day, UTF-8 is a variable length encoding where a single character can be encoded with more than one byte - and alpakka-csv and friends are operating on too low level of a primitive to deal with that.

ennru:
I agree, the byte-based parsing is not optimal and doing it char-based is much more correct.
On the other hand, this solves most use cases well without introducing a new wrapper type like `CharString`. Even going from a `ByteString` to a `CharString` will require translating the whole thing properly into chars and you might get some left-over bytes.

Voltir:
@ennru One thing we did to help in that regard was write a utility flow that attempts to canonicalize ByteStrings (assumed to be text in an unknown encoding) into UTF-8. After testing a number of libraries, we settled on using this com.ibm.icu.text.CharsetDetector as the basis for doing the detection. Would a flow like that be something alpakka is interested in?

Also, if we did want to go further and explore something like a "CharString" - where do you think something like that would belong? It seems like its not quite at the level of akka, with its akka.util.ByteString, but its also not quite at the alpakka csv library level either.

ennru:
The `com.ibm.icu.text.CharsetDetector`-base flow would be a great addition to this Alpakka Text module.

There is no obvious home for the `CharString`, you're right. Making a jar just for that seems not great either, but that way both Alpakka Text and CSV could depend on it.

PR #1018, opened 2018-06-13
Describe includeDocumentVersion property of ElasticsearchSourceSettings

takezoe:
Add description about `includeDocumentVersion` property of `ElasticsearchSourceSettings` to the document.

This property is defined here:
https://github.com/akka/alpakka/blob/3902367aa8a9097d6f676adecb8fb475b76505af/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/scaladsl/ElasticsearchSourceSettings.scala#L17

and used in `ElasticsearchSourceStage`:
https://github.com/akka/alpakka/blob/3902367aa8a9097d6f676adecb8fb475b76505af/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSourceStage.scala#L80-L88


ennru:
Great, thank you.

PR #1022, opened 2018-06-14
JMS examples in Java

ennru:


doc-examples/src/main/java/jms/javasamples/JmsToHttpGet.java
```diff
@@ -0,0 +1,100 @@
...
+
+
+    Pair<KillSwitch, CompletionStage<Done>> pair =
+        jmsSource                                                  //: String
+            .map(ByteString::fromString)                           //: ByteString   (2)
+            .map(bs ->
+                HttpRequest.create("http://localhost:8080/hello")
+                    .withEntity(bs)
+            )                                                      //: HttpRequest  (3)
+            .mapAsyncUnordered(4, http::singleRequest)             //: HttpResponse (4)
```
2m: Here and below move the argument to the variable so it is more clear what this number does.

PR #1028, opened 2018-06-16
Jackson 2.9.6

sullis:


2m:
Failure was https://github.com/akka/alpakka/issues/1031

PR #1030, opened 2018-06-17
Cassandra: avoiding isExhausted calls + prefetching results when buffer is half full

### LINKED ISSUE #1021: Cassandra: avoid isExhausted calls
example code:
-------------------
https://github.com/hochgi/cassandra-migrate-keyspace-from-cluster

Problem:
------------
When running with locally (on my laptop), fetching data from a remote data-center, and ingesting to a local data center I encountered the following error which broke the stream:

```
[error] com.datastax.driver.core.exceptions.OperationTimedOutException: [remote.domain.com/12.34.56.78:9042] Timed out waiting for server response
[error] 	at com.datastax.driver.core.exceptions.OperationTimedOutException.copy(OperationTimedOutException.java:44)
[error] 	at com.datastax.driver.core.exceptions.OperationTimedOutException.copy(OperationTimedOutException.java:26)
[error] 	at com.datastax.driver.core.DriverThrowables.propagateCause(DriverThrowables.java:37)
[error] 	at com.datastax.driver.core.ArrayBackedResultSet$MultiPage.prepareNextRow(ArrayBackedResultSet.java:314)
[error] 	at com.datastax.driver.core.ArrayBackedResultSet$MultiPage.isExhausted(ArrayBackedResultSet.java:270)
[error] 	at akka.stream.alpakka.cassandra.CassandraSourceStage$$anon$1$$anon$2.onPull(CassandraSourceStage.scala:42)
[error] 	at akka.stream.impl.fusing.GraphInterpreter.processPull(GraphInterpreter.scala:526)
[error] 	at akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:422)
[error] 	at akka.stream.impl.fusing.GraphInterpreterShell.runBatch(ActorGraphInterpreter.scala:588)
[error] 	at akka.stream.impl.fusing.GraphInterpreterShell$AsyncInput.execute(ActorGraphInterpreter.scala:472)
[error] 	at akka.stream.impl.fusing.GraphInterpreterShell.processEvent(ActorGraphInterpreter.scala:563)
[error] 	at akka.stream.impl.fusing.ActorGraphInterpreter.akka$stream$impl$fusing$ActorGraphInterpreter$$processEvent(ActorGraphInterpreter.scala:745)
[error] 	at akka.stream.impl.fusing.ActorGraphInterpreter.akka$stream$impl$fusing$ActorGraphInterpreter$$shortCircuitBatch(ActorGraphInterpreter.scala:735)
[error] 	at akka.stream.impl.fusing.ActorGraphInterpreter$$anonfun$receive$1.applyOrElse(ActorGraphInterpreter.scala:761)
[error] 	at akka.actor.Actor.aroundReceive(Actor.scala:517)
[error] 	at akka.actor.Actor.aroundReceive$(Actor.scala:515)
[error] 	at akka.stream.impl.fusing.ActorGraphInterpreter.aroundReceive(ActorGraphInterpreter.scala:670)
[error] 	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:588)
[error] 	at akka.actor.ActorCell.invoke(ActorCell.scala:557)
[error] 	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:258)
[error] 	at akka.dispatch.Mailbox.run(Mailbox.scala:225)
[error] 	at akka.dispatch.Mailbox.exec(Mailbox.scala:235)
[error] 	at akka.dispatch.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
[error] 	at akka.dispatch.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
[error] 	at akka.dispatch.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
[error] 	at akka.dispatch.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
[error] Caused by: com.datastax.driver.core.exceptions.OperationTimedOutException: [remote.domain.com/12.34.56.78:9042] Timed out waiting for server response
[error] 	at com.datastax.driver.core.RequestHandler$SpeculativeExecution.onTimeout(RequestHandler.java:790)
[error] 	at com.datastax.driver.core.Connection$ResponseHandler$1.run(Connection.java:1382)
[error] 	at io.netty.util.HashedWheelTimer$HashedWheelTimeout.expire(HashedWheelTimer.java:663)
[error] 	at io.netty.util.HashedWheelTimer$HashedWheelBucket.expireTimeouts(HashedWheelTimer.java:738)
[error] 	at io.netty.util.HashedWheelTimer$Worker.run(HashedWheelTimer.java:466)
[error] 	at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:138)
[error] 	at java.lang.Thread.run(Thread.java:748)
[error] [FAILURE!]
```
Cause:
---------
Problem lies with this line:  
https://github.com/akka/alpakka/blob/master/cassandra/src/main/scala/akka/stream/alpakka/cassandra/CassandraSourceStage.scala#L42

`rs.isExhausted` may throw an exception, which might break the stream.  

Possible fix:
---------------
A way around it, is not to call it at all.
Perhaps something like:

```scala
var minimumPrefetchedRows: Int = 0
var askingForMore = false
val setAskingForMore: () => Unit = _

  setAskingForMore = getAsyncCallback[Unit]{ _ =>  
    askingForMore = false 
  }.invoke
  …
  val futRs = futStmt.flatMap { stmt =>
    // OK to assign outside of AsyncCallback since this is the only write
    // and var reads uses the value as an optimization for eager fetching
    // which is not required.
    minimumPrefetchedRows = math.max(1, stmt.getFetchSize() / 2)
    session.executeAsync(stmt).asScala()
  }
  …
}

setHandler(
  out,
  new OutHandler {
    override def onPull(): Unit = maybeRs.foreach { rs: ResultSet =>
      if (!askingForMore && rs.getAvailableWithoutFetching < minimumPrefetchedRows) {
        askingForMore = true
        rs.fetchMoreResults.asScala.onComplete(setAskingForMore)
      }
      Option(rs.one()).fold(completeStage())(push(out, _))
    }
  }
)
```

I understand that the code was designed such that we only call `rs.one()` if we are sure we have an available row in the result set, so that we don't block the dispatcher thread when more data needs to be fetched. But thread is blocked anyway if we call `rs.isExhausted`.

My suggestion is to pre-fetch rows when we reach half the size of buffer.
Since we don't handle timeouts at all, I think it's preferable to ignore exceptions thrown on the datastax' driver thread, and thus we don't deal with timeouts.

Since we start asking when the buffer is full, we will attempt again when downstream polls, since `rs.getAvailableWithoutFetching` is still less than `minimumPrefetchedRows`.

In the unlikely case we have a real problem, downstream will exhaust all pre-fetched rows, and then call `rs.one()`, which will be blocked, and if we still have a problem, it'll still throw, but at least it will happen after we tried several times without failing on the first timeout.

This code also have the benefit of rarely blocking the dispatcher thread, compared to original implementation which blocks every `stmt.getFetchSize()` rows.

hochgi:
Fixes #1021 

* also using `Future.foreach` instead of `map`/`flatMap` for side effect
* not using redundant scala Future implicit class conversion (we are only interested in the side effect of `AsyncCallback` anyway…)
* not replacing boxed `ResultSet` on every fetch (`ResultSet` is reusable, no need to generate garbage for gc)
* general boyscouting

### DIFF
--- a/cassandra/src/main/scala/akka/stream/alpakka/cassandra/GuavaFutures.scala
+++ b/cassandra/src/main/scala/akka/stream/alpakka/cassandra/GuavaFutures.scala
@@ -6,9 +6,24 @@ package akka.stream.alpakka.cassandra
 
 import com.google.common.util.concurrent.{FutureCallback, Futures, ListenableFuture}
 
-import scala.concurrent.{Future, Promise}
+import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.util.{Failure, Success, Try}
 
 private[cassandra] object GuavaFutures {
+
+  def invokeTryCallback[T](
+      listenableFuture: ListenableFuture[T],
+      executor: java.util.concurrent.Executor = ExecutionContext.global
+  )(callback: Try[T] => Unit): Unit =
+    Futures.addCallback(
+      listenableFuture,
+      new FutureCallback[T] {
+        override def onSuccess(result: T): Unit = callback(Success(result))
+        override def onFailure(t: Throwable): Unit = callback(Failure(t))
+      },
+      executor
+    )
+
   implicit final class GuavaFutureOpts[A](val guavaFut: ListenableFuture[A]) extends AnyVal {
     def asScala(): Future[A] = {
       val p = Promise[A]()
--- a/cassandra/src/main/scala/akka/stream/alpakka/cassandra/GuavaFutures.scala
+++ b/cassandra/src/main/scala/akka/stream/alpakka/cassandra/GuavaFutures.scala
@@ -11,10 +11,8 @@ import scala.util.{Failure, Success, Try}
 
 private[cassandra] object GuavaFutures {
 
-  def invokeTryCallback[T](
-      listenableFuture: ListenableFuture[T],
-      executor: java.util.concurrent.Executor = ExecutionContext.global
-  )(callback: Try[T] => Unit): Unit =
+  def invokeTryCallback[T](listenableFuture: ListenableFuture[T],
+                           executor: java.util.concurrent.Executor)(callback: Try[T] => Unit): Unit =
     Futures.addCallback(
       listenableFuture,
       new FutureCallback[T] {
--- a/cassandra/src/main/scala/akka/stream/alpakka/cassandra/CassandraSourceStage.scala
+++ b/cassandra/src/main/scala/akka/stream/alpakka/cassandra/CassandraSourceStage.scala
@@ -17,51 +17,60 @@ final class CassandraSourceStage(futStmt: Future[Statement], session: Session) e
   val out: Outlet[Row] = Outlet("CassandraSource.out")
   override val shape: SourceShape[Row] = SourceShape(out)
 
-  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
-    new GraphStageLogic(shape) {
-      var maybeRs = Option.empty[ResultSet]
-      var futFetchedCallback: AsyncCallback[Try[ResultSet]] = _
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+    var maybeRs = Option.empty[ResultSet]
+    var futFetchedCallback: Try[ResultSet] => Unit = _
+    var isFetching = true // set to true until prestart's callback will set to false
+    var minimumPreFetchSize: Int = _
 
-      override def preStart(): Unit = {
-        implicit val ec = materializer.executionContext
-
-        futFetchedCallback = getAsyncCallback[Try[ResultSet]](tryPushAfterFetch)
-
-        val futRs = futStmt.flatMap(stmt => session.executeAsync(stmt).asScala())
-        futRs.onComplete(futFetchedCallback.invoke)
-      }
-
-      setHandler(
-        out,
-        new OutHandler {
-          override def onPull(): Unit = {
-            implicit val ec = materializer.executionContext
-
-            maybeRs match {
-              case Some(rs) if rs.getAvailableWithoutFetching > 0 => push(out, rs.one())
-              case Some(rs) if rs.isExhausted => completeStage()
-              case Some(rs) =>
-                // fetch next page
-                val futRs = rs.fetchMoreResults().asScala()
-                futRs.onComplete(futFetchedCallback.invoke)
-              case None => () // doing nothing, waiting for futRs in preStart() to be completed
-            }
-          }
-        }
-      )
-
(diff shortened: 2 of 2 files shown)


cassandra/src/main/scala/akka/stream/alpakka/cassandra/GuavaFutures.scala
```diff
@@ -6,9 +6,24 @@ package akka.stream.alpakka.cassandra
...
 
-import scala.concurrent.{Future, Promise}
+import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.util.{Failure, Success, Try}
 
 private[cassandra] object GuavaFutures {
+
+  def invokeTryCallback[T](
+      listenableFuture: ListenableFuture[T],
+      executor: java.util.concurrent.Executor = ExecutionContext.global
```
ennru: Please remove the default parameter

cassandra/src/main/scala/akka/stream/alpakka/cassandra/GuavaFutures.scala
```diff
@@ -6,9 +6,24 @@ package akka.stream.alpakka.cassandra
...
+    Futures.addCallback(
+      listenableFuture,
+      new FutureCallback[T] {
+        override def onSuccess(result: T): Unit = callback(Success(result))
+        override def onFailure(t: Throwable): Unit = callback(Failure(t))
+      },
+      executor
+    )
+
   implicit final class GuavaFutureOpts[A](val guavaFut: ListenableFuture[A]) extends AnyVal {
```
ennru: Is this needed anymore?

hochgi: yes. it is used in Sink:
https://github.com/akka/alpakka/blob/2dac1e082b2afcb419610e8221844ae22f8338cb/cassandra/src/main/scala/akka/stream/alpakka/cassandra/scaladsl/CassandraSink.scala#L24

ennru:
Would you be able to add a few more tests, as well?
What I see the Cassandra tests cover a bare minimum right now.

ennru:
For inspiration how to force disconnects you may look at `MqttSourceSpec` "mqtt source" should "automatically reconnect", which uses a TCP proxy.

ennru:
Thank you for improving this. Would be great to add some more tests as discussed, if you get to it.

PR #1033, opened 2018-06-18
AMQP: Reorganise packages

ennru:
* Moved internal classes to `impl` package and annotated with `@InternalApi`
* Moved code for documentation snippets into `docs.scaladsl|javadsl` packages
* Smaller code improvements


amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -4,39 +4,45 @@
...
+  def declarations: immutable.Seq[Declaration]
 }
 
 sealed trait AmqpSourceSettings extends AmqpConnectorSettings
 
 final case class NamedQueueSourceSettings(
     connectionProvider: AmqpConnectionProvider,
     queue: String,
-    declarations: Seq[Declaration] = Seq.empty,
+    declarations: immutable.Seq[Declaration] = Nil,
```
2m: This is not ideal. I think instead of importing `s.c.immutable` we should add type and companion object alias in the connector base package object: 

```scala
package object amqp {
  type Seq[A] = immutable.Seq[A]
  val Seq = immutable.Seq
}
```

in `... /alpakka/ampq/package.scala` file.

amqp/src/test/scala/docs/scaladsl/AmqpDocsSpec.scala
```diff
@@ -0,0 +1,223 @@
+/*
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package docs.scaladsl
```
2m: Nice. Shorter packages for code that will be included in the docs is great!

PR #1032, opened 2018-06-18
Add MiMa

ennru:
Added 2m's repo, required for `sbt-paradox-local`, as well.

PR #1038, opened 2018-06-19
sbt-dynver 3.0.0

sullis:


2m:
Thanks for the PR. Unfortunately there is another PR (#1032) which changes to even later sbt-dynver version so we get `previousStableVersion` sbt task.

PR #1039, opened 2018-06-19
Reference connector 

2m:
~This is work in progress.~ However early feedback is very welcome.

### DIFF
--- a/contributor-advice.md
+++ b/contributor-advice.md
@@ -1,5 +1,10 @@
 # Some advice for Alpakka contributors
 
+## Reference connector
+
+To inspect how all of the below listed guidelines are to be implemented in practice, take a look at
+[the reference](reference/) connector. Feel free to use the reference connector as a starting point for new connectors.
+
 ## API design
 
 ### Public factory methods
@@ -10,10 +15,14 @@ simple and easy to use.
 
 ### Flows
 
+> Reference connector [model classes](reference/src/main/scala/akka/stream/alpakka/reference/model.scala)
+
 When designing Flows, consider adding an extra field to the in- and out-messages which is passed through. A common
 use case we see, is committing a Kafka offset after passing data to another system.
 
 ### Java APIs
+
+> Reference connector [Java API factory methods](reference/src/main/scala/akka/stream/alpakka/javadsl/Reference.scala)
 
 Alpakka, same as Akka, aims to keep 100% feature parity between the various language DSLs.
 
@@ -26,17 +35,18 @@ Alpakka, same as Akka, aims to keep 100% feature parity between the various lang
 
 ### Settings
 
+> Reference connector [settings classes](reference/src/main/scala/akka/stream/alpakka/reference/settings.scala)
+
 Most technologies will have a couple of configuration settings that will be needed for several Sinks, Flows, or Sinks. 
 Create case classes collecting these settings instead of passing them in every method.
 
-Create a default instance in the companion object with good defaults which can be updated via `copy` or `withXxxx` methods.
+Create a default instance in the companion object with good defaults which can be updated via `withXxxx` methods.
 
-As Java can't use the `copy` methods, add `withXxxx` methods to specify certain fields in the settings instance.
+Add `withXxxx` methods to specify certain fields in the settings instance.
 
 In case you see the need to support reading the settings from `Config`, offer a method taking the `Config` instance so
 that the user can apply a proper namespace.
 Refrain from using `akka.stream.alpakka` as Config prefix, prefer `alpakka` as root namespace.
-
 
 ## Implementation details
 
@@ -50,7 +60,6 @@ Which licenses are compatible with Apache 2 are defined in [this doc](http://www
 
 Dependency licenses will be checked automatically by the sbt Whitesource plug-in. 
 
-
 ### Packages & Scoping
 
 Use `private`, `private[connector]` and `final` extensively to limit the API surface.
@@ -59,10 +68,12 @@ Use `private`, `private[connector]` and `final` extensively to limit the API sur
 | -----------------------------------------|------------------------
 | `akka.stream.alpakka.connector.javadsl`  | Java-only part of the API, normally factories for Sources, Flows and Sinks
 | `akka.stream.alpakka.connector.scaladsl` | Scala-only part of the API, normally factories for Sources, Flows and Sinks
-| `akka.stream.alpakka.connector`          | Shared API, eg. settings classes, implementation if thorowly made private
-| `akka.stream.alpakka.connector.impl`     | Optional, implementation in separate package
+| `akka.stream.alpakka.connector`          | Shared API, eg. settings classes
+| `akka.stream.alpakka.connector.impl`     | Implementation in separate package
 
 ### Graph stage checklist
+
+> Reference connector [operator implementations](reference/src/main/scala/akka/stream/alpakka/reference/impl/)
 
 * Keep mutable state within the `GraphStageLogic` only 
 * Open connections in `preStart`
@@ -84,8 +95,10 @@ as a Flow. This can be easily done by reusing the Flow implementation:
 
 You do not need to expose every configuration a Flow offers this way -- Focus on the most important ones.
 
-
 ## Test
+
+> Reference connector [Scala](reference/src/test/scala/docs/scaladsl/ReferenceSpec.scala) and
+> [Java](reference/src/test/java/docs/javadsl/ReferenceTest.java) tests
 
 Write tests for all public methods and possible settings.
 
@@ -95,6 +108,8 @@ Please ensure that you limit the amount of resources used by the containers.
 
 ## Documentation
 
+> Reference connector [paradox documentation](docs/src/main/paradox/refernece.md)
+
 Using [Paradox](https://github.com/lightbend/paradox) syntax (which is very close to markdown), create or complement
 the documentation in the `docs` module.
 Prepare code snippets to be integrated by Paradox in the tests. Such example should be part of real tests and not in
@@ -102,5 +117,5 @@ unused methods.
 
 Use ScalaDoc if you see the need to describe the API usage better than the naming does.
 
-Run `sbt docs/local:paradox` to generate reference docs while developing. Generated documentation can be 
+Run `sbt docs/Local/paradox` to generate reference docs while developing. Generated documentation can be 
 found in the `./docs/target/paradox/site/local` directory.
--- a/docs/src/main/paradox/reference.md
+++ b/docs/src/main/paradox/reference.md
@@ -31,6 +31,7 @@ Show an example code snippet of how a source of this connector can be created.
 
 Scala
 : @@snip [snip]($alpakka$/reference/src/test/scala/docs/scaladsl/ReferenceSpec.scala) { #source }
+Language specific text can follow immediately after a snippet so it is switched with the snippet tab.
 
 Java
 : @@snip [snip]($alpakka$/reference/src/test/java/docs/javadsl/ReferenceTest.java) { #source }
--- a/docs/src/main/paradox/reference.md
+++ b/docs/src/main/paradox/reference.md
@@ -1,0 +1,64 @@
+# Reference
+
+This is the reference documentation for an Alpakka connector. This section should contain
+a general overview of the connector and mention the libraries and APIs that the connector
+is using. Also it should link to external resources that might help to learn
+about the technology connector is using.
+
+### Reported issues
+
+[Tagged issues at Github](https://github.com/akka/alpakka/labels/p%3Areference)
+
+## Artifacts
+
+@@dependency [sbt,Maven,Gradle] {
+  group=com.lightbend.akka
+  artifact=akka-stream-alpakka-reference_$scalaBinaryVersion$
+  version=$version$
+}
+
+## Usage
+
+Give a brief description of the usage of this connector. If you want to mention a
+class name, make sure to link to the API docs:
+@scaladoc[ReferenceReadMessage](akka.stream.alpakka.reference.ReferenceReadMessage).
+
+### Reading messages
+
+If any of the API classes are different between Scala and Java, link to both API docs:
+@scala[@scaladoc[Reference](akka.stream.alpakka.reference.scaladsl.Reference$)]
+@java[@scaladoc[Reference](akka.stream.alpakka.reference.javadsl.Reference$)].
+
+Show an example code snippet of how a source of this connector can be created.
+
+Scala
+: @@snip [snip]($alpakka$/reference/src/test/scala/docs/scaladsl/ReferenceSpec.scala) { #source }
+
+Java
+: @@snip [snip]($alpakka$/reference/src/test/java/docs/javadsl/ReferenceTest.java) { #source }
+
+### Writing messages
+
+Show an example code snippet of how a flow of this connector can be created.
+
+Scala
+: @@snip [snip]($alpakka$/reference/src/test/scala/docs/scaladsl/ReferenceSpec.scala) { #flow }
+
+Java
+: @@snip [snip]($alpakka$/reference/src/test/java/docs/javadsl/ReferenceTest.java) { #flow }
+
+### Running the example code
+
+The code in this guide is part of runnable tests of this project. You are welcome to edit the code and run it in sbt.
+
+Scala
+:   ```
+    sbt
+    > reference/testOnly *.ReferenceSpec
+    ```
+    
+Java
+:   ```
+    sbt
+    > reference/testOnly *.ReferenceTest
+    ```
--- a/docs/src/main/paradox/reference.md
+++ b/docs/src/main/paradox/reference.md
@@ -35,6 +35,9 @@ Scala
 Java
 : @@snip [snip]($alpakka$/reference/src/test/java/docs/javadsl/ReferenceTest.java) { #source }
 
+Wrap language specific text with language specific directives,
+like @scala[`@scala` for Scala specific text]@java[`@java` for Java specific text].
+
 ## Writing messages
 
 Show an example code snippet of how a flow of this connector can be created.
--- a/docs/src/main/paradox/reference.md
+++ b/docs/src/main/paradox/reference.md
@@ -3,7 +3,7 @@
 This is the reference documentation for an Alpakka connector. This section should contain
 a general overview of the connector and mention the libraries and APIs that the connector
 is using. Also it should link to external resources that might help to learn
-about the technology connector is using.
+about the technology the connector is using.
 
 ### Reported issues
 
--- a/docs/src/main/paradox/reference.md
+++ b/docs/src/main/paradox/reference.md
@@ -17,13 +17,11 @@ about the technology connector is using.
   version=$version$
 }
 
-## Usage
+## Reading messages
 
 Give a brief description of the usage of this connector. If you want to mention a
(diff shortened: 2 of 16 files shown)


2m:
Missing pieces:
* [ ] API that uses `ExecutionContext` / `Executor`
* [ ] paradox docs
* [ ] private implementation in `impl` package

docs/src/main/paradox/reference.md
```diff
@@ -0,0 +1,64 @@
...
+  version=$version$
+}
+
+## Usage
+
+Give a brief description of the usage of this connector. If you want to mention a
+class name, make sure to link to the API docs:
+@scaladoc[ReferenceReadMessage](akka.stream.alpakka.reference.ReferenceReadMessage).
+
+### Reading messages
```
ennru: Make the interesting sections on level 2 to let them show in the page menu.
Possibly removing the Usage section.

docs/src/main/paradox/reference.md
```diff
@@ -0,0 +1,64 @@
...
+
+If any of the API classes are different between Scala and Java, link to both API docs:
+@scala[@scaladoc[Reference](akka.stream.alpakka.reference.scaladsl.Reference$)]
+@java[@scaladoc[Reference](akka.stream.alpakka.reference.javadsl.Reference$)].
+
+Show an example code snippet of how a source of this connector can be created.
+
+Scala
+: @@snip [snip]($alpakka$/reference/src/test/scala/docs/scaladsl/ReferenceSpec.scala) { #source }
```
ennru: Add text underneath to show it is flipped with the selected language.

2m: Done.

2m:
Addressed PR feedback and implemented all of the aspects that were initially planned. Ready to be merged or for more feedback.

docs/src/main/paradox/reference.md
ennru: Ah, I start to believe you are not aware of this feature. Adding text after a snippet without an empty line will flip in the same tab.

reference/src/main/scala/akka/stream/alpakka/reference/impl/ReferenceFlow.scala
```diff
@@ -0,0 +1,62 @@
...
+    override def onPull(): Unit = pull(in)
+  })
+
+  /**
+   * Cleanup logic
+   */
+  override def postStop(): Unit = {}
+}
+
+@InternalApi final class ReferenceFlow() extends GraphStage[FlowShape[ReferenceWriteMessage, ReferenceWriteMessage]] {
```
ennru: Add `private[reference]`

reference/src/main/scala/akka/stream/alpakka/reference/impl/ReferenceSource.scala
```diff
@@ -0,0 +1,69 @@
...
+  })
+
+  /**
+   * Cleanup logic
+   */
+  override def postStop(): Unit = {}
+}
+
+@InternalApi final class ReferenceSource(settings: SourceSettings)
+    extends GraphStageWithMaterializedValue[SourceShape[ReferenceReadMessage], Future[NotUsed]] {
```
ennru: Wouldn't `Done` be a better type/value for the startup future?

reference/src/main/scala/akka/stream/alpakka/reference/model.scala
```diff
@@ -0,0 +1,139 @@
...
+    case Success(_) => null
+    case Failure(ex) => ex
+  }
+
+  private def copy(data: immutable.Seq[ByteString] = data, bytesRead: Try[Int] = bytesRead) =
+    new ReferenceReadMessage(data, bytesRead)
+}
+
+object ReferenceReadMessage {
+  def apply(): ReferenceReadMessage = new ReferenceReadMessage()
```
ennru: `create` for Java? 
Even if this is constructed internally only, for testing it should be made available.

reference/src/main/scala/akka/stream/alpakka/reference/model.scala
```diff
@@ -0,0 +1,139 @@
...
+   */
+  @varargs def withData(data: ByteString*): ReferenceWriteMessage =
+    copy(data = data.to[immutable.Seq])
+
+  /**
+   * Java API
+   *
+   * Java setter needs to take Java Long class and convert to Scala Long.
+   */
+  @varargs def withMetrics(metrics: Pair[String, JavaLong]*): ReferenceWriteMessage =
```
ennru: Even add `def withMetrics(key: String, value: JavaLong)`?

reference/src/main/scala/akka/stream/alpakka/reference/settings.scala
```diff
@@ -0,0 +1,149 @@
...
+       |  pollInterval   = $pollInterval
+       |)""".stripMargin
+}
+
+object SourceSettings {
+
+  /**
+   * Factory method for Scala.
+   */
+  def apply(): SourceSettings = new SourceSettings()
```
ennru: If there are settings without good defaults (eg. connection details), these should appear in `apply/create`.

docs/src/main/paradox/reference.md
2m: Ahh, I see. Never used it like this.

reference/src/main/scala/akka/stream/alpakka/reference/impl/ReferenceSource.scala
2m: Yes, think it is.

reference/src/main/scala/akka/stream/alpakka/reference/model.scala
2m: Good point.

reference/src/main/scala/akka/stream/alpakka/reference/model.scala
2m: Do you think it is worth it? The equivalent of this would be `withMetrics(Pair.create(key, value))`

reference/src/main/scala/akka/stream/alpakka/reference/settings.scala
```diff
@@ -0,0 +1,149 @@
...
+   * Make singleton objects extend an abstract class with the same name.
+   * This makes it possible to refer to the object type without `.type`.
+   */
+  sealed abstract class None extends Authentication
+  case object None extends None
+
+  /**
+   * Java API
+   */
+  def createNone: None = None
```
sebastianharko: quick question: why not just def none: None = None ?

project/Dependencies.scala
```diff
@@ -216,6 +216,13 @@ object Dependencies {
     )
   )
 
+  val Reference = Seq(
+    // connector specific library dependencies and resolver settings
+    libraryDependencies ++= Seq(
+      "org.scala-lang.modules" %% "scala-java8-compat" % "0.9.0" // BSD-style
```
sebastianharko: another quick question: so every connector should include this ?

ennru: If it needs help converting between Java and Scala types (collections, futures, durations) - so most often, yes.

reference/src/main/scala/akka/stream/alpakka/reference/settings.scala
ennru: Java APIs most often use a verb as prefix.

reference/src/main/scala/akka/stream/alpakka/reference/settings.scala
```diff
@@ -0,0 +1,149 @@
...
+ * Settings class constructor is private and not exposed as API.
+ * Adding or removing arguments to methods with default values is not binary
+ * compatible. However, since the constructor is private, it will be possible
+ * to add or remove attributes without introducing binary incompatibilities.
+ */
+final class SourceSettings private (
+    val clientId: Option[String] = None,
+    val authentication: Authentication = Authentication.None,
+    val pollInterval: Duration = 5.seconds
+) {
```
sebastianharko: another question: where would you actually validate these SourceSettings ? What if , for example, clientId is not optional and absolutely required for the Source to work but the user provides SourceSettings.withAuthentication().withPollInterval()  (and accidentally leaving clientId to None) ???

reference/src/main/scala/akka/stream/alpakka/reference/model.scala
ennru: It would let Java users get away without learning about Akka's `Pair`.

2m: Okay. Then lets change this setter and instead of `Pair` let it take `Map.Entry`.

ennru: The ugly thing is the warning "Unchecked generics array creation" which needs to be oppressed by `@SuppressWarnings("unchecked")`.

reference/src/main/scala/akka/stream/alpakka/reference/settings.scala
2m: Good point. For such mandatory attributes, we should not have default values and take them as parameters in the `apply` and `create` factory methods. Added a commit that illustrates that.

docs/src/main/paradox/reference.md
```diff
@@ -0,0 +1,66 @@
+# Reference
+
+This is the reference documentation for an Alpakka connector. This section should contain
+a general overview of the connector and mention the libraries and APIs that the connector
+is using. Also it should link to external resources that might help to learn
+about the technology connector is using.
```
ennru: "technology THE connector"

reference/src/main/scala/akka/stream/alpakka/reference/model.scala
```diff
@@ -0,0 +1,141 @@
...
+    copy(bytesRead = bytesRead)
+
+  /**
+   * Java API
+   *
+   * If the model class is meant to be also consumed from the user API,
+   * but the attribute class is Scala specific, create getter for Java API.
+   */
+  def getData(): JavaList[ByteString] = {
+    import scala.collection.JavaConverters._
```
ennru: import on top level instead?

reference/src/test/java/docs/javadsl/ReferenceTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+/**
+ * Append "Test" to every Java test suite.
+ */
+public class ReferenceTest {
+
+  static ActorSystem sys;
+  static Materializer mat;
+
+  final static String ClientId = "test-client-id";
```
ennru: `clientId` - capitalised values/variables are very uncommon in Java.

project/Dependencies.scala
2m: But the java8 compat library comes as a transitive dependency to akka-actor. Therefore we should not add it explicitly here and instead use the transitively resolved one.

That will also make sure that the right version of java8-compat is used.

I'll remove it from here.

2m:
Added pointers from contributor advice and refactored to setters that take Java collection types.

reference/src/main/scala/akka/stream/alpakka/reference/model.scala
```diff
@@ -0,0 +1,137 @@
...
+import scala.collection.immutable
+import scala.collection.JavaConverters._
+import akka.japi.Pair
+
+import scala.util.{Failure, Success, Try}
+
+/**
+ * Use "Read" in message data types to signify that the message was read from outside.
+ */
+final class ReferenceReadMessage private (
```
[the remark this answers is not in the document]
2m: Most of the time `unapply` for a such a message is not very useful. Such messages usually contain a bytestring and possibly something for passthrough. Therefore I think it is fine to leave unapplies out.

Authors are of course free to add unapply methods, if, for a specific connector, unapply for Read or Write message does make sense.

ennru:
Should this connector be set to `NoPublish`?

contributor-advice.md
```diff
@@ -26,18 +35,19 @@ Alpakka, same as Akka, aims to keep 100% feature parity between the various lang
 
 ### Settings
 
+> Reference connector [settings classes](reference/src/main/scala/akka/stream/alpakka/reference/settings.scala)
```
patriknw: lowercase `settings.scala`?

contributor-advice.md
```diff
@@ -59,11 +68,13 @@ Use `private`, `private[connector]` and `final` extensively to limit the API sur
 | -----------------------------------------|------------------------
 | `akka.stream.alpakka.connector.javadsl`  | Java-only part of the API, normally factories for Sources, Flows and Sinks
 | `akka.stream.alpakka.connector.scaladsl` | Scala-only part of the API, normally factories for Sources, Flows and Sinks
-| `akka.stream.alpakka.connector`          | Shared API, eg. settings classes, implementation if thorowly made private
-| `akka.stream.alpakka.connector.impl`     | Optional, implementation in separate package
+| `akka.stream.alpakka.connector`          | Shared API, eg. settings classes
+| `akka.stream.alpakka.connector.impl`     | Implementation in separate package
```
patriknw: perhaps even "Internal implementation in separate package"

contributor-advice.md
```diff
@@ -95,12 +108,14 @@ Please ensure that you limit the amount of resources used by the containers.
 
 ## Documentation
 
+> Reference connector [paradox documentation](docs/src/main/paradox/refernece.md)
```
patriknw: typo refernece

docs/src/main/paradox/reference.md
```diff
@@ -0,0 +1,66 @@
...
+
+If any of the API classes are different between Scala and Java, link to both API docs:
+@scala[@scaladoc[Reference](akka.stream.alpakka.reference.scaladsl.Reference$)]
+@java[@scaladoc[Reference](akka.stream.alpakka.reference.javadsl.Reference$)].
+
+Show an example code snippet of how a source of this connector can be created.
+
+Scala
+: @@snip [snip]($alpakka$/reference/src/test/scala/docs/scaladsl/ReferenceSpec.scala) { #source }
+Language specific text can follow immediately after a snippet so it is switched with the snippet tab.
```
patriknw: how does that work after the last snippet? Ah, no newline, sneaky.

patriknw: I think I prefer text in `@scala[]` `@java[]`

reference/src/main/scala/akka/stream/alpakka/reference/impl/ReferenceFlow.scala
```diff
@@ -0,0 +1,63 @@
...
+import akka.annotation.InternalApi
+import akka.event.Logging
+import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
+import akka.stream.alpakka.reference.ReferenceWriteMessage
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
+
+/**
+ * private package hides the class from the API in Scala. However it is still
+ * visible in Java. Use "InternalApi" annotation to communicate to Java developers
+ * that this is private API.
```
patriknw: No `INTERNAL API` doc comment? Is the annotation visible in JavaDoc?

reference/src/main/scala/akka/stream/alpakka/reference/model.scala
```diff
@@ -0,0 +1,137 @@
...
+    metrics.mapValues(JavaLong.valueOf).asJava
+
+  private def copy(data: immutable.Seq[ByteString] = data, metrics: Map[String, Long] = metrics) =
+    new ReferenceWriteMessage(data, metrics)
+
+  override def toString: String =
+    s"""ReferenceWriteMessage(
+       |  data       = $data
+       |  metrics    = $metrics
+       |)""".stripMargin
```
patriknw: I'd avoid line breaks in toString, since it might be used for logging, and line breaks in log messages are just a pain (e.g. grep)

reference/src/main/scala/akka/stream/alpakka/reference/settings.scala
```diff
@@ -0,0 +1,151 @@
...
+                   pollInterval: Duration = pollInterval) =
+    new SourceSettings(clientId, traceId, authentication, pollInterval)
+
+  override def toString: String =
+    s"""SourceSettings(
+       |  clientId       = $clientId
+       |  traceId        = $traceId
+       |  authentication = $authentication
+       |  pollInterval   = $pollInterval
+       |)""".stripMargin
```
patriknw: linebreaks

reference/src/main/scala/akka/stream/alpakka/reference/settings.scala
```diff
@@ -0,0 +1,151 @@
...
+ * Settings class constructor is private and not exposed as API.
+ * Adding or removing arguments to methods with default values is not binary
+ * compatible. However, since the constructor is private, it will be possible
+ * to add or remove attributes without introducing binary incompatibilities.
+ */
+final class SourceSettings private (
+    val clientId: String,
+    val traceId: Option[String] = None,
+    val authentication: Authentication = Authentication.None,
+    val pollInterval: Duration = 5.seconds
```
patriknw: probably more common to use `FiniteDuration`

reference/src/main/scala/akka/stream/alpakka/reference/settings.scala
```diff
@@ -0,0 +1,151 @@
...
+    def withVerifierPredicate(verifier: Predicate[String]): Provided =
+      copy(verifier = verifier.asScala)
+
+    private def copy(verifier: String => Boolean = verifier) =
+      new Provided(verifier)
+
+    override def toString: String =
+      s"""Authentication.Provided(
+         |  verifier       = $verifier
+         |)""".stripMargin
```
patriknw: \n

reference/src/main/scala/akka/stream/alpakka/reference/settings.scala
```diff
@@ -0,0 +1,151 @@
...
+     */
+    def apply(): Provided = new Provided()
+  }
+
+  /**
+   * Java API
+   *
+   * Factory method needed to access nested object.
+   */
+  def createProvided() = Provided()
```
patriknw: explicit return type

contributor-advice.md
2m: Yes, which would mean that there are multiple logical units of compilation - settings classes. This is from https://docs.scala-lang.org/style/files.html

Usually a connector will have `settings.scala` and `model.scala`. The latter is for read/write messages.

docs/src/main/paradox/reference.md
2m: I agree. I can see how easily this functionality can break by unintentional adding a linebreak after the snip directive.

reference/src/main/scala/akka/stream/alpakka/reference/impl/ReferenceFlow.scala
2m: Ahh, forgot that. We do not generate javadoc so I forgot to check.

2m:
Thank you @patriknw, feedback addressed in https://github.com/akka/alpakka/pull/1108!

contributor-advice.md
patriknw: ok, that's a style I don't use, because I don't like files with many unrelated classes. Often there is one main class, and a few related that can be grouped together and then the file should have the name of the main class.

but I guess it's a matter of taste

PR #1044, opened 2018-06-25
Upgrade to the latest paradox

2m:
Also fixes a problem with placeholders in links in the Slick page where it did not work (https://github.com/lightbend/paradox/issues/178). Changed to use `extref` directive instead.


docs/src/main/paradox/file.md
```diff
@@ -141,20 +119,20 @@ Scala
 :   ```
     sbt
     // tail source
-    > akka-stream-alpakka-file/test:runMain akka.stream.alpakka.file.scaladsl.FileTailSourceSpec /some/path/toa/file
+    > file/Test/runMain akka.stream.alpakka.file.scaladsl.FileTailSourceSpec /some/path/toa/file
```
ennru: Good find, do people find the "Running the code" useful?

docs/src/main/paradox/slick.md
```diff
@@ -1,6 +1,6 @@
 # Slick (JDBC)
 
-The Slick connector provides Scala and Java DSLs to create a `Source` to stream the results of a SQL database query and a `Flow`/`Sink` to perform SQL actions (like inserts, updates, and deletes) for each element in a stream. It is built on the [Slick](http://slick.lightbend.com/) library to interact with a long list of [supported relational databases](http://slick.lightbend.com/doc/$slickVersion$/supported-databases.html).
+The Slick connector provides Scala and Java DSLs to create a `Source` to stream the results of a SQL database query and a `Flow`/`Sink` to perform SQL actions (like inserts, updates, and deletes) for each element in a stream. It is built on the [Slick](http://slick.lightbend.com/) library to interact with a long list of @extref[supported relational databases](slick:supported-databases.html).
```
ennru: But the links did work, this is shorter though.

docs/src/main/paradox/file.md
2m: The fact that no one discovered this before seems to indicate that not so much.

docs/src/main/paradox/slick.md
2m: It was broken after https://github.com/akka/alpakka/pull/1015 and can be seen here: https://developer.lightbend.com/docs/alpakka/snapshot/slick.html

PR #1048, opened 2018-06-26
AWS Kinesis: fix exponential backoff retry for Firehose

### LINKED ISSUE #1045: Kinesis Firehose exponential backoff uses initial delay for backoff factor base
The exponential backoff in the Kinesis Firehose support is implemented incorrectly (or at least surprisingly):
https://github.com/akka/alpakka/blob/f3c1e894c8f62ba7c22b6926974592dbde4cb7ca/kinesis/src/main/scala/akka/stream/alpakka/kinesisfirehose/KinesisFirehoseFlowStage.scala#L87
Note that the base of the power is the initial retry delay rather than a constant (typically 2). Contrast with the same code in the Kinesis Streams code with the expected base of 2:
https://github.com/akka/alpakka/blob/f3c1e894c8f62ba7c22b6926974592dbde4cb7ca/kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisFlowStage.scala#L87

dana-harrington:
Fixes #1045

### DIFF
--- a/kinesis/src/main/scala/akka/stream/alpakka/kinesisfirehose/KinesisFirehoseFlowStage.scala
+++ b/kinesis/src/main/scala/akka/stream/alpakka/kinesisfirehose/KinesisFirehoseFlowStage.scala
@@ -84,7 +84,7 @@ private[kinesisfirehose] final class KinesisFirehoseFlowStage(
           inFlight -= 1
           waitingRetries.put(retryToken, Job(attempt + 1, errors.map(_._2)))
           scheduleOnce(retryToken, backoffStrategy match {
-            case Exponential => scala.math.pow(retryBaseInMillis, attempt).toInt millis
+            case Exponential => retryBaseInMillis * scala.math.pow(2, attempt - 1).toInt millis
             case Linear => retryInitialTimeout * attempt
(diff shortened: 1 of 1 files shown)


kinesis/src/main/scala/akka/stream/alpakka/kinesisfirehose/KinesisFirehoseFlowStage.scala
```diff
@@ -84,7 +84,7 @@ private[kinesisfirehose] final class KinesisFirehoseFlowStage(
           inFlight -= 1
           waitingRetries.put(retryToken, Job(attempt + 1, errors.map(_._2)))
           scheduleOnce(retryToken, backoffStrategy match {
-            case Exponential => scala.math.pow(retryBaseInMillis, attempt).toInt millis
+            case Exponential => retryBaseInMillis * scala.math.pow(2, attempt - 1).toInt millis
```
2m: Lets make the logic identical to the one in the `KinesisFlowStage` https://github.com/akka/alpakka/pull/989/files

Which is to remove `retryBaseInMillis` completely.

PR #1050, opened 2018-06-27
HBase: Allow multiple mutations

davideicardi:
I need to perform different kind of mutations on HBase given an input message.
Until now  only `put` is supported. 

My idea is to return a list of `Mutation` from the `converter` function instead of a single `Put` instance.
Then I have changed:

```
table.put(settings.converter(msg))
```

with:

```
val mutations = settings.converter(msg)
for (mutation <- mutations) {
	mutation match {
		case x: Put => table.put(x)
		case x: Delete => table.delete(x)
		case x: Append => table.append(x)
		case x: Increment => table.increment(x)
    }
}
```
This should allow to perform different kind of mutations (delete, increment, append) and also perform no mutations (useful if you have some business logic to skip some records).

It can probably solve #328 and #326

What do you think?

### DIFF
--- a/hbase/src/test/java/akka/stream/alpakka/hbase/javadsl/HBaseStageTest.java
+++ b/hbase/src/test/java/akka/stream/alpakka/hbase/javadsl/HBaseStageTest.java
@@ -56,7 +56,7 @@ public class HBaseStageTest {
   }
 
   // #create-converter
-  Function<Person, Put> hBaseConverter =
+  Function<Person, List<Mutation>> hBaseConverter =
       person -> {
         Put put = null;
         try {
@@ -66,7 +66,7 @@ public class HBaseStageTest {
         } catch (UnsupportedEncodingException e) {
           e.printStackTrace();
         }
-        return put;
+        return Arrays.asList(put);
       };
   // #create-converter
 
--- a/hbase/src/test/java/akka/stream/alpakka/hbase/javadsl/HBaseStageTest.java
+++ b/hbase/src/test/java/akka/stream/alpakka/hbase/javadsl/HBaseStageTest.java
@@ -18,6 +18,7 @@ import akka.stream.javadsl.Source;
 import akka.testkit.javadsl.TestKit;
 import org.apache.hadoop.hbase.HBaseConfiguration;
 import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.Mutation;
 import org.apache.hadoop.hbase.client.Put;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
--- a/hbase/src/test/java/akka/stream/alpakka/hbase/javadsl/HBaseStageTest.java
+++ b/hbase/src/test/java/akka/stream/alpakka/hbase/javadsl/HBaseStageTest.java
@@ -18,8 +18,7 @@ import akka.stream.javadsl.Source;
 import akka.testkit.javadsl.TestKit;
 import org.apache.hadoop.hbase.HBaseConfiguration;
 import org.apache.hadoop.hbase.TableName;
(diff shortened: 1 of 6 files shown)


ennru:
Yes, this change looks very useful, yet simple. 
It will change the API a little bit, but that's OK AFAICS.

We installed a Java code formatter, please rebase.

davideicardi:
Thanks @ennru just rebased! 

There is the possibility to take an "unstable" package from some continuous build? Or what do you suggest to try this change in my project? (sorry for the stupid question but I'm new on this ecosystem...)

ennru:
Great, thank you.
Could you please add test cases exercising the other mutations?

Once it is merged to master it will be part of a snapshot build.

davideicardi:
Sorry I need your help ... I have some problems running HBase tests.
If I leave the code as is and run `sbt hbase/test` no tests are discovered. Also looking inside the Travis build I don't see any tests related to HBase: https://travis-ci.org/akka/alpakka/jobs/398182454

I'm missing something?

To let sbt find some tests I need to remove:

`hbaseIT.foreach { hbase =>`

Removing the above for each tests are executed but with errors:

`java.util.concurrent.TimeoutException: Futures timed out after [10 seconds]`

In docker compose I don't see any hbase instance. It is correct? I suspect that there isn't a valid hbase configuration...

davideicardi:
For now I just added some tests but only for paradox doc (they are not executed, see prev message). I also added some documentation notes.

hbase/src/test/java/akka/stream/alpakka/hbase/javadsl/HBaseStageTest.java
```diff
@@ -65,10 +65,35 @@ public static void teardown() {
         } catch (UnsupportedEncodingException e) {
           e.printStackTrace();
         }
-        return put;
+        return Arrays.asList(put);
```
ennru: Use `Collections.singletonList(put)` instead.

hbase/src/test/java/akka/stream/alpakka/hbase/javadsl/HBaseStageTest.java
```diff
@@ -65,10 +65,35 @@ public static void teardown() {
...
+          Append append = new Append(String.format("id_%d", person.id).getBytes("UTF-8"));
+          append.add(
+              "info".getBytes("UTF-8"), "aliases".getBytes("UTF-8"), person.name.getBytes("UTF-8"));
+
+          Delete delete = new Delete(String.format("id_%d", person.id).getBytes("UTF-8"));
+
+          Increment increment = new Increment(String.format("id_%d", person.id).getBytes("UTF-8"));
+          increment.addColumn("info".getBytes("UTF-8"), "age".getBytes("UTF-8"), 1);
+
+          return Arrays.asList(append, put, delete, increment);
```
ennru: This example does not make much sense, could you come up with something more realistic?

davideicardi: In my real use case I just need to `Put` or `Delete` depending of some message data.

For the pull request, now I have tried to split every mutations in a separate example. Probably a more realistic scenario require a more complex data model. 

About the `Append` I'm not sure how to use it. If I understand correctly it will append bytes to a single cell, I can't see a real usage for now (maybe for a log appender?)...

In any case I think that the point of this pull requests is to permit sending different kind of mutations. Explaining how to use HBase mutations is probably out of scope in my opinion. Eventually we can link some official HBase doc?

thanks

ennru:
You are right, I wasn't aware that there are not runnable tests in the HBase module.  I created #1058.

ennru:
Thank you for making the HBase connector much more useful.

Issue #1020, opened 2018-06-13
DynamoDB: Java API uses scala futures


ennru:
The `akka.stream.alpakka.dynamodb.javadsl.DynamoClient` methods should be changed to return `CompletionStage`.

PR #1055, opened 2018-06-29
Migrate DynamoDB Java API to return CompletionStage instead of Future…

adamtrousdale:
… (#1020)

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -83,6 +83,7 @@ object Dependencies {
 
   val DynamoDB = Seq(
     libraryDependencies ++= Seq(
+      "org.scala-lang.modules" %% "scala-java8-compat" % "0.8.0", // BSD-style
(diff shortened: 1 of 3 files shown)


project/Dependencies.scala
```diff
@@ -83,6 +83,7 @@ object Dependencies {
 
   val DynamoDB = Seq(
     libraryDependencies ++= Seq(
+      "org.scala-lang.modules" %% "scala-java8-compat" % "0.8.0", // BSD-style
```
2m: Dependency to java8-compat comes transitively from akka-actor. So it is not needed to include it here as well.

adamtrousdale:
Ah my bad, I missed that. Removed the explicit dependency.

ennru:
Fixes #1020

ennru:
Thank you for getting this done! Keep going.

PR #1056, opened 2018-06-29
Unix Domain Socket - ensure that interest in READ is removed upon EOF

longshorej:
Fixes an issue observed with unix domain socket connector where when the read side was closed, the NIO event loop would go into an infinite loop with `channel.read` returning `-1`.

We should have be deregistering interest in `READ` when completing the queue.

Related PR: https://github.com/akka/alpakka/pull/878

/cc @huntc

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -312,7 +312,7 @@ object Dependencies {
 
   val UnixDomainSocket = Seq(
     libraryDependencies ++= Seq(
-      "com.github.jnr" % "jnr-unixsocket" % "0.18" // BSD/ApacheV2/CPL/MIT as per https://github.com/akka/alpakka/issues/620#issuecomment-348727265
+      "com.github.jnr" % "jnr-unixsocket" % "0.19" // BSD/ApacheV2/CPL/MIT as per https://github.com/akka/alpakka/issues/620#issuecomment-348727265
     )
   )
 
--- a/unix-domain-socket/src/main/scala/akka/stream/alpakka/unixdomainsocket/scaladsl/UnixDomainSocket.scala
+++ b/unix-domain-socket/src/main/scala/akka/stream/alpakka/unixdomainsocket/scaladsl/UnixDomainSocket.scala
@@ -168,10 +168,12 @@ object UnixDomainSocket extends ExtensionId[UnixDomainSocket] with ExtensionIdPr
                       val pendingResult = queue.offer(ByteString(buffer))
                       pendingResult.onComplete(_ => sel.wakeup())
                       sendReceiveContext.receive = PendingReceiveAck(queue, buffer, pendingResult)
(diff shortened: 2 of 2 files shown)


project/Dependencies.scala
```diff
@@ -312,7 +312,7 @@ object Dependencies {
 
   val UnixDomainSocket = Seq(
     libraryDependencies ++= Seq(
-      "com.github.jnr" % "jnr-unixsocket" % "0.18" // BSD/ApacheV2/CPL/MIT as per https://github.com/akka/alpakka/issues/620#issuecomment-348727265
+      "com.github.jnr" % "jnr-unixsocket" % "0.19" // BSD/ApacheV2/CPL/MIT as per https://github.com/akka/alpakka/issues/620#issuecomment-348727265
```
huntc: What’s the fix they made? I’m wondering if they fixed a native buffer issue, in which case perhaps an alternate PR for this with consideration of using native buffers. There’s a FixMe somewhere in this code with a ref to the bug...

unix-domain-socket/src/main/scala/akka/stream/alpakka/unixdomainsocket/scaladsl/UnixDomainSocket.scala
```diff
@@ -170,6 +170,7 @@ object UnixDomainSocket extends ExtensionId[UnixDomainSocket] with ExtensionIdPr
                       sendReceiveContext.receive = PendingReceiveAck(queue, buffer, pendingResult)
                       key.interestOps(key.interestOps() & ~SelectionKey.OP_READ)
                     } else {
+                      key.interestOps(key.interestOps() & ~SelectionKey.OP_READ)
```
huntc: This can now move outside the conditional block given that both conditions remove the subscription.

project/Dependencies.scala
longshorej: Socket options weren't set correctly -- i.e. big endian was always used. This would cause buffer sizes to be quite wrong on intel arch: https://github.com/jnr/jnr-unixsocket/commit/aa2c1e43493edc9f6ca2af45a789a7bacbf196a1

https://github.com/jnr/jnr-unixsocket/issues/48 is still open though.

huntc:
Got it. Thanks.

unix-domain-socket/src/main/scala/akka/stream/alpakka/unixdomainsocket/scaladsl/UnixDomainSocket.scala
2m: Yup, moving it outside the conditional would be good.

longshorej: Moved!

ennru:
Thank you, good to see the nifty details fixed.

PR #1057, opened 2018-06-30
DynamoDB client interface rather than class

gregbeech:
Implementing `DynamoClient` as a class, and in particular a final class, means that it's not possible to mock for testing repositories that use it.

This change converts it to an interface and returns an anonymous class implementing it from the apply method, which will be source-compatible with `DynamoClient(settings)`, although not with `new DynamoClient(settings)`.

Removing the `final` modified from the class would be another option, which would keep source compatibility in either of the above cases, but this seems like a cleaner approach to me.


ennru:
Thank you, we want to improve testability of Alpakka flows, so this is great.

PR #1059, opened 2018-07-03
Add method to create KinesisFlowSettings from stream name

takezoe:
`KinesisFlowSettings` has a method to create by number of shards. This pull request adds a method to create it by a shard name. This method fetches number of shards from the actual stream instead of specifying number of shards directly.

### DIFF
--- a/kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisFlowSettings.scala
+++ b/kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisFlowSettings.scala
@@ -4,7 +4,12 @@
 
 package akka.stream.alpakka.kinesis
 
+import java.util
+
 import KinesisFlowSettings._
+import com.amazonaws.services.kinesis.AmazonKinesisAsync
+import com.amazonaws.services.kinesis.model.Shard
+
 import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.duration._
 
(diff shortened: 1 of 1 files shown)


ennru:
Great addition! Though I wonder if it should live somewhere else than the settings class. I would be very surprised when I find out configuring a setting leads to remote calls. 
Possibly, you could introduce some `KinesisAdmin` or `KinesisInfo` helper that uses the client library and contains some `shardsByStreamName` method?

takezoe:
It's makes sense to implement remote API calls elsewhere. As another consideration, I think that creating `KinesisFlowSettings` from an actual number of shards is better as the default behavior.

How about creating `KinesisFlowSettings` (with API call) inside `KinesisFlow` factory if it's not given by users?

ennru:
In that case, you need to build something that defers the API call until stream materalization. 
I think the best solution is some kind of settings factory using the Kinesis client.

takezoe:
Isn't it a method that I added?

ennru:
Yes, it is, but I would like settings classes to be non-active, not doing any calls. So basically the same code in a class indicating with its name that it will do a remote call to gather information. 
You should return a `Future[KinesisFlowSettings]` and a `CompletionStage[KinesisFlowSettings]` instead of the blocking code, as well.

takezoe:
@ennru I see... I understood that this pull request isn't good anyway.

If I'll create a helper to fetch number of shards, I can put it to my project or library. There is no need for it's in Alpakka. Or If I'll create a factory to create `KinesisFlowSettings` from the stream information, I have to create both Java and Scala API unlike this simple implementation.

Thanks for your advice.

takezoe:
I close this pull request once.

PR #1099, opened 2018-07-08
XML subtree flow

takezoe:
Add `XmlPasring.subtree` which extracts elements match a certain path. This flow is similar to `XmlPasring.subslice` but pushes matched elements and their child nodes to the downstream as `org.w3c.dom.Element` instead of pure `ParseEvent`. It needs more memory, but it makes easier to handle iteration of sub tree of XML document.

If this pull request is acceptable, I will update the documentation as well.


2m:
Thanks for the PR. I think the implementation looks good! It would be great if you could amend the docs. Then we can merge this!

takezoe:
I updated the document as well.

PR #1104, opened 2018-07-11
Elasticsearch: Support update and delete by incoming message types

takezoe:
Currently, `ElasticsearchFlow` supports only **insert** and **upsert**. We can configure a flow as either one of them.

This pull request adds support for **update** and **delete** by message types by introducing following incoming message types:

- `IncomingIndexMessage`
- `IncomingUpdateMessage`
- `IncomingUpsertMessage`
- `IncomingDeleteMessage`

This also means it makes possible to handle all request types in single flow.

I'm not sure whether I should modify existing `ElasticsearchFlow` interfaces to support this functionality. I think it doesn't need so much effort to update existing code which uses `ElasticsearchFlow` to this version, but breaking source code compatibility is a fact.

If keeping source code compatibility is critical in Alpakka connectors, I would like to add new flow which has this functionality.

### DIFF
--- a/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
+++ b/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
@@ -306,7 +306,6 @@ class ElasticsearchFlowStage[T, C](
         val successMsgs = messageResults.filter(_.error.isEmpty)
         if (successMsgs.nonEmpty) {
           // push the messages that DID succeed
-          //val resultForSucceededMsgs = successMsgs.map(_.r)
           emit(out, Future.successful(successMsgs))
         }
       }
--- a/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
+++ b/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
@@ -415,9 +415,9 @@ class ElasticsearchFlowStage[T, C](
             ).toString
           case x: IncomingUpdateMessage[T, C] =>
             "\n" + JsObject(
-              "doc" -> writer.convert(x.source).parseJson,
+              "doc" -> writer.convert(x.source).parseJson
             ).toString
-          case x: IncomingDeleteMessage[T, C] =>
+          case _: IncomingDeleteMessage[T, C] =>
             ""
         }
 
--- a/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
+++ b/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
@@ -22,64 +22,191 @@ import akka.NotUsed
 import org.apache.http.message.BasicHeader
 import org.apache.http.util.EntityUtils
 
-object IncomingMessage {
+object IncomingIndexMessage {
   // Apply method to use when not using passThrough
   def apply[T](id: Option[String], source: T): IncomingMessage[T, NotUsed] =
-    IncomingMessage(id, source, NotUsed)
+    IncomingIndexMessage(id, source, NotUsed)
 
   // Apply method to use when not using passThrough
   def apply[T](id: Option[String], source: T, version: Long): IncomingMessage[T, NotUsed] =
-    IncomingMessage(id, source, NotUsed, Option(version))
+    IncomingIndexMessage(id, source, NotUsed, Option(version))
 
   // Java-api - without passThrough
   def create[T](id: String, source: T): IncomingMessage[T, NotUsed] =
-    IncomingMessage(Option(id), source)
+    IncomingIndexMessage(Option(id), source)
 
   // Java-api - without passThrough
   def create[T](id: String, source: T, version: Long): IncomingMessage[T, NotUsed] =
-    IncomingMessage(Option(id), source, version)
+    IncomingIndexMessage(Option(id), source, version)
 
   // Java-api - without passThrough
   def create[T](source: T): IncomingMessage[T, NotUsed] =
-    IncomingMessage(None, source)
+    IncomingIndexMessage(None, source)
 
   // Java-api - with passThrough
   def create[T, C](id: String, source: T, passThrough: C): IncomingMessage[T, C] =
-    IncomingMessage(Option(id), source, passThrough)
+    IncomingIndexMessage(Option(id), source, passThrough)
 
   // Java-api - with passThrough
   def create[T, C](id: String, source: T, passThrough: C, version: Long): IncomingMessage[T, C] =
-    IncomingMessage(Option(id), source, passThrough, Option(version))
+    IncomingIndexMessage(Option(id), source, passThrough, Option(version))
 
   // Java-api - with passThrough
   def create[T, C](id: String, source: T, passThrough: C, version: Long, indexName: String): IncomingMessage[T, C] =
-    IncomingMessage(Option(id), source, passThrough, Option(version), Option(indexName))
+    IncomingIndexMessage(Option(id), source, passThrough, Option(version), Option(indexName))
 
   // Java-api - with passThrough
   def create[T, C](source: T, passThrough: C): IncomingMessage[T, C] =
-    IncomingMessage(None, source, passThrough)
+    IncomingIndexMessage(None, source, passThrough)
 }
 
-final case class IncomingMessage[T, C](id: Option[String],
-                                       source: T,
-                                       passThrough: C,
-                                       version: Option[Long] = None,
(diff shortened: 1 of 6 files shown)


elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
```diff
@@ -22,64 +22,191 @@ import akka.NotUsed
...
 }
 
-final case class IncomingMessage[T, C](id: Option[String],
-                                       source: T,
-                                       passThrough: C,
-                                       version: Option[Long] = None,
-                                       indexName: Option[String] = None) {
+object IncomingUpdateMessage {
+  // Apply method to use when not using passThrough
+  def apply[T](id: String, source: T): IncomingMessage[T, NotUsed] =
```
2m: I think we should have factory `apply` and `create` methods only with the arguments that are mandatory. Then optional arguments can be set with `with***` builder methods. You can take a look at the [reference connector](https://github.com/akka/alpakka/blob/master/reference/src/main/scala/akka/stream/alpakka/reference/model.scala) which follows that convention.

2m: This would also bring elasticsearch connector one step closer to https://github.com/akka/alpakka/issues/1070

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
```diff
@@ -167,30 +293,30 @@ class ElasticsearchFlowStage[T, C](
...
+        failedMessages = failedMsgs.map(_.message) // These are the messages we're going to retry
         scheduleOnce(NotUsed, settings.retryInterval.millis)
 
-        val successMsgs = messageResults.filter(_.r.success)
+        val successMsgs = messageResults.filter(_.error.isEmpty)
         if (successMsgs.nonEmpty) {
           // push the messages that DID succeed
-          val resultForSucceededMsgs = successMsgs.map(_.r)
-          emit(out, Future.successful(resultForSucceededMsgs))
+          //val resultForSucceededMsgs = successMsgs.map(_.r)
```
2m: Feel free to remove commented out code.

takezoe:
> Have you considered alternative approach to keep IncomingMessage and introduce an operation argument instead?

I tried this approach at first, but there were some differences in these operations such as:

- `IncomingDeleteMessage` doesn't have `source` which means a payload of the message
- `id` is required in `IncomingUpdateMessage` and `IncomingDeleteMessage`, but it's optional in `IncomingIndexMessage` and `IncomingUpsertMessage`

So It think it's inpossible to represent these operations as a single message type.

takezoe:
> So It think representing these operations as a single message type is not good idea.

I changed my mind because I realized that representing these parameter as `Option` is not so bad in fact. Also I applied the rule that provides factory methods with only mandatory parameters and `withXxx` methods for other parameters.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSinkSettings.scala
```diff
@@ -20,7 +20,7 @@ final case class ElasticsearchSinkSettings(bufferSize: Int = 10,
                                            retryInterval: Int = 5000,
                                            maxRetry: Int = 100,
                                            retryPartialFailure: Boolean = false,
-                                           docAsUpsert: Boolean = false,
+                                           //docAsUpsert: Boolean = false,
```
2m: Feel free to remove.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchSinkSettings.scala
```diff
@@ -34,8 +34,8 @@ final case class ElasticsearchSinkSettings(bufferSize: Int = 10,
   def withRetryPartialFailure(retryPartialFailure: Boolean): ElasticsearchSinkSettings =
     copy(retryPartialFailure = retryPartialFailure)
 
-  def withDocAsUpsert(docAsUpsert: Boolean): ElasticsearchSinkSettings =
-    copy(docAsUpsert = docAsUpsert)
+//  def withDocAsUpsert(docAsUpsert: Boolean): ElasticsearchSinkSettings =
+//    copy(docAsUpsert = docAsUpsert)
```
2m: Feel free to remove.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
```diff
@@ -22,64 +22,118 @@ import akka.NotUsed
...
-                                       version: Option[Long] = None,
-                                       indexName: Option[String] = None) {
+object IncomingDeleteMessage {
+  // Apply method to use when not using passThrough
+  def apply[T](id: String): IncomingMessage[T, NotUsed] =
+    IncomingMessage(Some(id), None, NotUsed, Delete, None, None)
+
+  // Apply method to use when using passThrough
+  def apply[T, C](id: String, passThrough: C): IncomingMessage[T, C] =
+    IncomingMessage(Some(id), None, passThrough, Delete, None, None)
```
2m: I think we should only have a constructor with `id`. Passthrough could be added by calling `withPassthrough` builder method.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
```diff
@@ -22,64 +22,118 @@ import akka.NotUsed
...
-  def apply[T](id: Option[String], source: T, version: Long): IncomingMessage[T, NotUsed] =
-    IncomingMessage(id, source, NotUsed, Option(version))
+  def apply[T](id: String, source: T): IncomingMessage[T, NotUsed] =
+    IncomingMessage(id = Some(id), source = Some(source), passThrough = NotUsed, operation = Index)
 
-  // Java-api - without passThrough
-  def create[T](id: String, source: T): IncomingMessage[T, NotUsed] =
-    IncomingMessage(Option(id), source)
+  // Apply method to use when using passThrough
+  def apply[T, C](source: T, passThrough: C): IncomingMessage[T, C] =
```
2m: Since we are already breaking source compatibility (`IncomingMessage` -> `IncomingIndexMessage`) we can remote this combinatorial explosion of factory methods. Lets keep only one and then have other combinations supported by calling `withXXX` builder methods.

takezoe: Specifying `source` and `passThrough` at the instance creation time is needed to fix type of `T` and `C`. If we don't specify them here, we need to call this method with type parameters like:
```scala
IncomingIndexMessage[Book, Offset](id).withSource(book).withPassTrough(offset)
```
instead of
```scala
IncomingIndexMessage(id, book, offset)
``` 
However Java API needs specify type parameters, anyway...

How do you think about it?

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
takezoe: Same as below.

takezoe:
@2m Thanks for your review. I'm worrying about factory methods of `IncomingMessage` as I commented above. I would like to hear your opinion.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala
2m: The `withSource` and `withPassThrough` methods can change the type parameters of the `IncommingMessage`:

```scala
def withSource[S](source: S): IncomingMessage[S, C] =
  copy(source = Some(source))

def withPassThrough[P](passThrough: P): IncomingMessage[T, P] =
  copy(passThrough = passThrough)
```

With these it is possible to use builder methods without specifying the type parameters upfront:

```scala
IncomingIndexMessage("id")
  .withSource(Book())
  .withPassThrough("pass-through")
```

takezoe: That's right! I removed combinatorial factory methods except methods which take required parameters.

takezoe:
@2m Thanks for your comment! I removed combinatorial factory methods except methods which take required parameters. Only `IncomingIndexMessage` has 2 patterns of factory methods, but they are both patterns of required parameters.

takezoe:
@2m Thanks! But we must update docs as well. I will create an another pull request soon.

PR #1105, opened 2018-07-12
Update library dependencies of xml project

takezoe:


project/Dependencies.scala
```diff
@@ -324,8 +324,8 @@ object Dependencies {
 
   val Xml = Seq(
     libraryDependencies ++= Seq(
-      "com.fasterxml" % "aalto-xml" % "1.0.0", // ApacheV2,
-      "org.scala-lang.modules" %% "scala-java8-compat" % "0.8.0" // BSD-style
+      "com.fasterxml" % "aalto-xml" % "1.1.0", // ApacheV2,
+      "org.scala-lang.modules" %% "scala-java8-compat" % "0.9.0" // BSD-style
```
2m: `scala-java8-compat` comes as a transitive dependency with `akka-actor`. And it is different version depending on what Scala version is used. Therefore we should remove explicit dependency to `scala-java8-compat` so we always use the one that comes with `akka-actor`.

takezoe:
@2m Thanks! I remove scala-java8-compat dependency from xml project.

Upgrading and reconsidering library dependencies in all projects might be needed. I would like to work on it in another pull requests.

2m:
Thanks.

You can use [`sbt-updates`](https://github.com/rtimush/sbt-updates) plugin that makes it easy to get information about libraries that need to be updated.

PR #1121, opened 2018-07-21
Remove "//#upload" comments to avoid duplicate codes in the document

nwtgck:
Thank you very much for your wonderful project! 

I found duplicate codes in the following document.

screen shot 2018-07-22 at 0 05 27  

(from: <https://developer.lightbend.com/docs/alpakka/current/s3.html>)

I'm not sure a mechanism to how the document created. But I guess that "//#upload" comments created a part of documents, so I removed one of them.


nwtgck:
Thank you very much for your response, @2m . Is there anything to do?

PR #1125, opened 2018-07-31
Refactor AMQP case classes to private classes 

### LINKED ISSUE #1064: AMQP: become 1.0 ready
- [X] Move internal code to `impl` package
- [X] Mark internal code `@InternalApi` and `private[connector]`
- [X] Move documentation example code to packages `docs.javadsl` and `docs.scaladsl`

2m:
Fixes #1064

### DIFF
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
@@ -6,7 +6,10 @@ package akka.stream.alpakka.amqp
 
 import akka.annotation.InternalApi
 
+import java.util.{List => JavaList, Map => JavaMap}
+
 import scala.collection.immutable
+import scala.collection.JavaConverters._
 
 /**
  * Internal API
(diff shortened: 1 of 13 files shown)


amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -19,37 +22,71 @@ sealed trait AmqpConnectorSettings {
...
+    val exclusive: Boolean = false,
+    val ackRequired: Boolean = true,
+    val consumerTag: String = "default",
+    val arguments: Map[String, AnyRef] = Map.empty
 ) extends AmqpSourceSettings {
-  @annotation.varargs
-  def withDeclarations(declarations: Declaration*): NamedQueueSourceSettings = copy(declarations = declarations.toList)
 
-  def withNoLocal(noLocal: Boolean): NamedQueueSourceSettings = copy(noLocal = noLocal)
+  def withDeclarations(declaration: immutable.Seq[Declaration]): NamedQueueSourceSettings =
```
ennru: Scala API

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -19,37 +22,71 @@ sealed trait AmqpConnectorSettings {
...
+  def withAckRequired(ackRequired: Boolean): NamedQueueSourceSettings =
+    copy(ackRequired = ackRequired)
 
-  def withConsumerTag(consumerTag: String): NamedQueueSourceSettings = copy(consumerTag = consumerTag)
+  def withConsumerTag(consumerTag: String): NamedQueueSourceSettings =
+    copy(consumerTag = consumerTag)
 
-  def withArguments(argument: (String, AnyRef), arguments: (String, AnyRef)*): NamedQueueSourceSettings =
-    copy(arguments = (argument +: arguments).toMap)
+  def withArguments(arguments: Map[String, AnyRef]): NamedQueueSourceSettings =
```
ennru: Scala API

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -58,20 +95,34 @@ object NamedQueueSourceSettings {
...
-    declarations: immutable.Seq[Declaration] = Nil,
-    routingKey: Option[String] = None
+final class TemporaryQueueSourceSettings private (
+    val connectionProvider: AmqpConnectionProvider,
+    val exchange: String,
+    val declarations: immutable.Seq[Declaration] = Nil,
+    val routingKey: Option[String] = None
 ) extends AmqpSourceSettings {
+
+  def withDeclarations(declaration: immutable.Seq[Declaration]): TemporaryQueueSourceSettings =
```
ennru: Scala API

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -80,44 +131,69 @@ object TemporaryQueueSourceSettings {
...
+    val declarations: immutable.Seq[Declaration] = Nil
+) extends AmqpConnectorSettings {
+
+  def withExchange(exchange: String): AmqpSinkSettings =
+    copy(exchange = Some(exchange))
+
+  def withRoutingKey(routingKey: String): AmqpSinkSettings =
+    copy(routingKey = Some(routingKey))
+
+  def withDeclarations(declaration: immutable.Seq[Declaration]): AmqpSinkSettings =
```
ennru: Scala API

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -128,85 +204,125 @@ object AmqpSinkSettings {
...
-  def withArguments(argument: (String, AnyRef), arguments: (String, AnyRef)*): QueueDeclaration =
-    copy(arguments = (argument +: arguments).toMap)
+  def withAutoDelete(autoDelete: Boolean): QueueDeclaration =
+    copy(autoDelete = autoDelete)
 
-  @annotation.varargs
-  def withArguments(argument: akka.japi.Pair[String, AnyRef],
-                    arguments: akka.japi.Pair[String, AnyRef]*): QueueDeclaration =
-    copy(arguments = (argument +: arguments).map(_.toScala).toMap)
+  def withArguments(arguments: Map[String, AnyRef]): QueueDeclaration =
```
ennru: Scala API

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -128,85 +204,125 @@ object AmqpSinkSettings {
...
+    val exchange: String,
+    val routingKey: Option[String] = None,
+    val arguments: Map[String, AnyRef] = Map.empty
 ) extends Declaration {
+
   def withRoutingKey(routingKey: String): BindingDeclaration = copy(routingKey = Some(routingKey))
 
-  def withArguments(argument: (String, AnyRef), arguments: (String, AnyRef)*): BindingDeclaration =
-    copy(arguments = (argument +: arguments).toMap)
+  def withArguments(arguments: Map[String, AnyRef]): BindingDeclaration =
```
ennru: Scala API

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -128,85 +204,125 @@ object AmqpSinkSettings {
...
+
   def withDurable(durable: Boolean): ExchangeDeclaration = copy(durable = durable)
 
   def withAutoDelete(autoDelete: Boolean): ExchangeDeclaration = copy(autoDelete = autoDelete)
 
   def withInternal(internal: Boolean): ExchangeDeclaration = copy(internal = internal)
 
-  def withArguments(argument: (String, AnyRef), arguments: (String, AnyRef)*): ExchangeDeclaration =
-    copy(arguments = (argument +: arguments).toMap)
+  def withArguments(arguments: Map[String, AnyRef]): ExchangeDeclaration =
```
ennru: Scala API

amqp/src/test/java/akka/stream/alpakka/amqp/javadsl/AmqpConnectorsTest.java
```diff
@@ -75,7 +73,7 @@ public void throwIfCanNotConnect() throws Throwable {
         AmqpSink.createSimple(
             AmqpSinkSettings.create(connectionProvider)
                 .withRoutingKey(queueName)
-                .withDeclarations(queueDeclaration));
+                .withDeclarations(Collections.singletonList(queueDeclaration)));
```
ennru: Would it make sense to have a `withDeclaration` (or `withDeclarations(String)`) to support the singleton case, which seems common?

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -19,37 +22,71 @@ sealed trait AmqpConnectorSettings {
...
-final case class NamedQueueSourceSettings(
-    connectionProvider: AmqpConnectionProvider,
-    queue: String,
-    declarations: immutable.Seq[Declaration] = immutable.Seq.empty,
-    noLocal: Boolean = false,
-    exclusive: Boolean = false,
-    ackRequired: Boolean = true,
-    consumerTag: String = "default",
-    arguments: Map[String, AnyRef] = Map.empty
+final class NamedQueueSourceSettings private (
```
juanjoDiaz: Is adding `private` at the end of the class declaration a convention that you are following accros the entire alpakka project?

I personally find more readable `private final class NamedQueueSourceSettings (` or even `final private class NamedQueueSourceSettings (` :)

ennru: Adding `private` before the argument list makes the constructor private, not the class.

2m:
I do not think we should add "Scala API" to the methods that have "similar" Java API. It would leave methods that work with both languages but do not have such comments in a bit of a strange situation.

amqp/src/test/java/akka/stream/alpakka/amqp/javadsl/AmqpConnectorsTest.java
2m: I think that seems to be worth it, yes.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -80,44 +137,72 @@ object TemporaryQueueSourceSettings {
...
+) extends AmqpConnectorSettings {
+
+  def withExchange(exchange: String): AmqpSinkSettings =
+    copy(exchange = Some(exchange))
+
+  def withRoutingKey(routingKey: String): AmqpSinkSettings =
+    copy(routingKey = Some(routingKey))
+
+  def withDeclaration(declaration: Declaration): AmqpSinkSettings =
+    copy(declarations = immutable.Seq(declaration))
```
ennru: I think it should add to the current list of values and not drop things set before.

2m: Interesting. This would be different from all the other setters, but it would still have `with` prefix. I think that can catch users surprised.

We could have setters (adders) that do not drop current values, but then we should call them differently.

ennru: Yes, could be surprising.
`addDeclaration` sounds OK, but let's skip that for now.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -6,7 +6,10 @@ package akka.stream.alpakka.amqp
 
 import akka.annotation.InternalApi
 
+import java.util.{List => JavaList, Map => JavaMap}
```
ennru: I think it is better to spell out the whole types where needed. In case a Java user jumps to the source, it would be easier to understand.

PR #1129, opened 2018-08-03
Use absolute links to snippets

2m:
Which is now possible since paradox 0.3.6.


2m:
Merging since this is a trivial change and will conflict with further package renames in all connectors.

Issue #277, opened 2017-04-26
Go through stages and look for resource leaks


johanandren:
When the Materializer or ActorSystem is killed there will be no invocations of `onUpstreamFailure` or any of the other handler callbacks, stages must make sure to clean up resources in `postStop`. Go through all stages and look for resource leaks and fix

(The same work as done in akka/akka#22781)

ennru:
Most tests use `assertAllStagesStopped` in one way or another.

PR #1131, opened 2018-08-03
Sink and Source for Apache Parquet

dannylesnik:
This commit contains Scala and Java DSL for Sink and Source. I don't think there is a need to create Flow stage as well since I can't find and use case for it.

It contains Unit tests for both DSL.


juanjoDiaz:
I haven't gone through this in depth but this PR seems very similar to #720 and I can see some of the same issues that I highlighted there:

> I don't think that we should create connectors for specific file formats.
> This should be a flow to format the content and then use the File sink to write to disk.

and on the same line, I don't think that we should have a dependency to hadoop at all. We should only have a dependency to Avro-Parquet

dannylesnik:
@juanjoDiaz 
Thank you for the feedback, Some points. 

- In my commit I provided also Sink and Source. So I believe I extended #720 with Source implementation as well.

- You mention that there should be flow format and File as Sink. In my opinion this is good reason, but don't forget that ParquetReader and Writer can write and read to HDFS directly (we are using it this way on our production systems. In m my Sink and Source can come directly from files located in HDFS.
This is how we are using it on production.
               1) Source is query to Elasticsearch with elastic4s. Sink is parquet file located directly on HDFS.
               2) Run Spark job on this data, which generates parquet output files.
               3) Use parquet output files from HDFS as the Source, and Elastic4s as sink to stream job results to HDFS.
Using parquet Source and Sink, we can do this functionality by several line of codes.

- and the last, I don't depend on hadoop.  In my commit the scope for hadoop files is `TEST`. I use Avro-Parquet library only as the dependency.

ennru:
I see you have trouble with Paradox, we recently changed to absolute paths. You can remove the `$alpakka$` prefix.

dannylesnik:
@ennru Thank you for your help. This issue fixed. Now have some weird ftp unit test failure.

ennru:
Looks very promising!

Without looking into the details, I'd like to ask you to change a few things in the structure
- move internal classes into an `impl` package
- move the example code to `docs.scaladsl` and `docs.javadsl` (to surface any visibility problems)
- add the module to `.travis.yml`
- akka-stream, junit and akka-stream-testkit dependencies come in from `Common`
- please annotate other dependencies with their license

dannylesnik:
Hi @ennru 

Just made the changes as you asked me.

avroparquet/src/main/scala/akka/stream/alpakka/avroparquet/impl/AvroParquetFlow.scala
```diff
@@ -0,0 +1,45 @@
...
+
+          override def onUpstreamFinish(): Unit = {
+            //super.onUpstreamFinish()
+            writer.close()
+            completeStage()
+          }
+
+          override def onUpstreamFailure(ex: Throwable): Unit = {
+            super.onUpstreamFailure(ex)
+            writer.close()
```
ennru: Please free resources in `postStop`. See #277

avroparquet/src/main/scala/akka/stream/alpakka/avroparquet/impl/AvroParquetFlow.scala
```diff
@@ -0,0 +1,45 @@
+package akka.stream.alpakka.avroparquet.impl
+import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
+import org.apache.avro.generic.GenericRecord
+import org.apache.parquet.hadoop.ParquetWriter
+
+private[avroparquet] class AvroParquetFlow(writer: ParquetWriter[GenericRecord])
```
ennru: Add `@InternalApi` and a ScalaDoc comment stating `INTERNAL API` to all internal classes.

avroparquet/src/main/scala/akka/stream/alpakka/avroparquet/impl/AvroParquetFlow.scala
```diff
@@ -0,0 +1,45 @@
...
+            writer.close()
+            completeStage()
+          }
+
+          override def onUpstreamFailure(ex: Throwable): Unit = {
+            super.onUpstreamFailure(ex)
+            writer.close()
+          }
+
+          @scala.throws[Exception](classOf[Exception])
```
ennru: Does the annotation add any value?

avroparquet/src/main/scala/akka/stream/alpakka/avroparquet/impl/AvroParquetSource.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+
+    setHandler(
+      out,
+      new OutHandler {
+
+        override def onDownstreamFinish(): Unit = {
+          super.onDownstreamFinish()
+          reader.close()
```
ennru: Use `postStop` here, as well.

avroparquet/src/main/scala/akka/stream/alpakka/avroparquet/impl/AvroParquetFlow.scala
dannylesnik: Done.

avroparquet/src/main/scala/akka/stream/alpakka/avroparquet/impl/AvroParquetFlow.scala
dannylesnik: Done.

avroparquet/src/main/scala/akka/stream/alpakka/avroparquet/impl/AvroParquetFlow.scala
dannylesnik: Done

avroparquet/src/main/scala/akka/stream/alpakka/avroparquet/impl/AvroParquetSource.scala
dannylesnik: Done

dannylesnik:
Hi @ennru.

Committed all the changes you requested to do after your code review. 

Regarding what you suggest. I don't think that ByteString is the correct abstraction, since this is Avro Column record and I need schema to manipulate it.  
Working heavily with Akka Streams and Hadoop Echo System. I believe that 99% percent of this code usage would be storing events (which obviously are  case classes or raw Jsons, fetched from DB, received from HTTP Layer or even from Actor Publisher) on HDFS (or any other distributed storage), to run Map-Reduce operation, which stores result in Parquet and move output back to the system. I don't believe that we need any other format on the way from schema based parquet record to case class. 

What I suggest might be useful is to make `AvroParquetFlow` public API and use it as `.via(FlowStage[GenericRecord,GenericRecord]) ` in case where Store to Parquet is not a last stage of the stream.

ennru:
Ok, makes sense. Having access to a flow is the most important bit. 

My advice for the documentation snippets was not clear enough. Please move the snippets to `/avroparquet/src/test/java/docs/javadsl` and `/avroparquet/src/test/java/docs/scaladsl`.

You need to add Parquet in `connectors.md` to get it listed.

dannylesnik:
- Created public Java and Scala DSL for API for AvroParquetFlow
- Added Spec for AvroParquetFlow
- Added documentation section for AvroParquetFlow in Paradox
- Added Parquet for connectors.md
- `Examples.java` moved to `/avroparquet/src/test/java/docs/javadsl`

avroparquet/src/main/scala/akka/stream/alpakka/avroparquet/impl/AvroParquetFlow.scala
```diff
@@ -0,0 +1,53 @@
...
+            completeStage()
+
+          override def onUpstreamFailure(ex: Throwable): Unit = {
+            super.onUpstreamFailure(ex)
+            writer.close()
+          }
+
+          override def onPush(): Unit = {
+            val obtainedValue = grab(in)
+            writer.write(obtainedValue)
```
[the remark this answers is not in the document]
danelkotev: @ennru Hi, can you give a brief explanation what  IODispatcher is and why it's necessary here?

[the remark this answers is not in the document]
ennru: Yes, the `IODispatcher` can be configured centrally in Akka. Its size may need to be adapted if there are many blocking things happening in your Actor system. And yes, it should be selected for all stages executing blocking IO.

dannylesnik:
- Logging configured

- All API unit tests moved to `docs.javadoc` and `docs.scaladoc`

- Added more explanation and imports in documentation

- Added IO Dispatcher to Flow and Sink.

- generated avroparquet.html using code/paradox. Looks OK.

avroparquet/src/test/java/docs/javadsl/AvroParquetSinkTest.java
```diff
@@ -0,0 +1,118 @@
...
+  @After
+  public void tearDown() {
+    system = null;
+    materializer = null;
+    File index = new File(folder);
+    String[] entries = index.list();
+    if (entries != null) {
+      for (String s : entries) {
+        File currentFile = new File(index.getPath(), s);
+        currentFile.delete();
```
ennru: Did you consider to use `File.createTempFile`  and `File.deleteOnExit`?

avroparquet/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,13 @@
+<configuration>
+    <appender name="FILE" class="ch.qos.logback.core.FileAppender">
+        <file>target/files.log</file>
```
ennru: The log should be called after the module, `avroparquet.log`.

docs/src/main/paradox/connectors.md
```diff
@@ -8,6 +8,7 @@
 * [Apache Geode](geode.md)
 * [Apache Kafka](kafka.md)
 * [Apache Kudu](kudu.md)
+* [Apache Parquet](avroparquet.md)
```
ennru: Was it intentional to use "Apache" instead of "Avro"? Both are Apache projects, but it would better to follow the module name.

avroparquet/src/test/java/docs/javadsl/AvroParquetSinkTest.java
dannylesnik: `File.createTempFile` can't work here, because I'm creating Hadoop Path file directly from path String, however `File.deleteOnExit` plays perfectly here. Thanks.

dannylesnik:
Changes committed.

dannylesnik:
@ennru .
Changes committed.

ennru:
Thank you for your contribution! Keep them coming.

Issue #1139, opened 2018-08-06
Unix Domain Socket half close behaviour broken


huntc:
There are various issues in the UnixDomainSocket support of half closed connections. The following tests will fail:

```scala
    "allow the client to close the connection" in {
      val file = Files.createTempFile("UnixDomainSocketSpec2", ".sock").toFile
      file.delete()
      file.deleteOnExit()

      val sendBytes = ByteString("Hello")

      val binding =
        UnixDomainSocket().bindAndHandle(Flow[ByteString]
                                           .delay(5.seconds)
                                           .map(identity),
                                         file)

      binding.flatMap { connection =>
        Source
          .single(sendBytes)
          .via(UnixDomainSocket().outgoingConnection(new UnixSocketAddress(file), halfClose = false))
          .runWith(Sink.headOption)
          .flatMap {
            case e if e.isEmpty => connection.unbind().map(_ => succeed)
          }
      }
    }

    "close the server once the client is also closed" in {
      val file = Files.createTempFile("UnixDomainSocketSpec3", ".sock").toFile
      file.delete()
      file.deleteOnExit()

      val sendBytes = ByteString("Hello")

      val binding =
        UnixDomainSocket().bindAndHandle(
          Flow
            .fromSinkAndSource(Sink.ignore, Source.single(sendBytes).delay(1.second)),
          file,
          halfClose = true
        )

      binding.flatMap { connection =>
        Source
          .empty[ByteString]
          .via(UnixDomainSocket().outgoingConnection(file))
          .runWith(Sink.headOption)
          .flatMap {
            case e if e.nonEmpty => connection.unbind().map(_ => succeed)
          }
      }
    }
```

@longshorej and I have a WIP that we will submit shortly.

2m:
Thanks for the heads-up. Looking forward to the PR!

PR #1142, opened 2018-08-07
Unix Domain Socket - Improve halfClose handling

longshorej:
This PR includes a number of improvements to the `halfClose` handling of the Unix Domain Socket. `halfClose` is particularly relevant for clients (and enabled by default), as they will close their write side to signal end of input, but will still wish to read data from the server.

The main changes here are to call `shutdownOutput` and `shutdownInput` where appropriate, instead of closing the channel. Additionally, a bug in JNR is worked around which was causing some NIO events to be lost.

Fixes https://github.com/akka/alpakka/issues/1139

### DIFF
--- a/unix-domain-socket/src/main/java/jnr/enxio/channels/KQSelector.java
+++ b/unix-domain-socket/src/main/java/jnr/enxio/channels/KQSelector.java
@@ -1,0 +1,350 @@
+/*
+ * Copyright (C) 2008 Wayne Meissner
+ *
+ * This file is part of the JNR project.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// FIXME: This file can be removed once https://github.com/jnr/jnr-enxio/pull/28 is in.
+
+package jnr.enxio.channels;
+
+import jnr.constants.platform.Errno;
+import jnr.ffi.Memory;
+import jnr.ffi.Pointer;
+import jnr.ffi.StructLayout;
+import jnr.ffi.TypeAlias;
+import jnr.ffi.provider.jffi.NativeRuntime;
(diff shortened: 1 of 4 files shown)


huntc:
Just to expand on the macOS issue that I observed - I noticed a periodic but common failure on this test:

https://github.com/akka/alpakka/pull/1142/files#diff-8be0724cc2d84e3eb4ab27041b91c914R86

What I noticed was that the server side would indicate that it is interested in accepting connections (`OP_ACCEPT`), but it would then never receive an `OP_ACCEPT` event in the NIO loop i.e. `isAcceptable` on a selected key would never return `true`. I've no idea why this or even how this could be so.

Calling all Java NIO experts! (@viktorklang)

huntc:
@ennru @2m Any thoughts on our woes re. this PR? https://github.com/akka/alpakka/pull/1142#issuecomment-411226368 I'm thinking that we should push forward with it and flag my discovery as a known issue. WDYT?

huntc:
I had another look at this today. I didn't get much further. However, I'm baffled re. ScalaTest why my particular test isn't timing out at least. Shouldn't Scala Tests's `AsyncWordSpecLike` eventually timeout if a `Future[Assertion]` doesn't complete?

huntc:
I've fixed the issue and observed that all now appears well on macOS (FreeBSD). There was a bug in JNR which took me a very long time to discover (!). Here's the JNR PR I made: https://github.com/jnr/jnr-enxio/pull/28

Meanwhile, I've included the JNR patch into the PR. JNR doesn't appear to receive a lot of attention so I'd personally prefer not to have to wait on it.

unix-domain-socket/src/main/java/jnr/enxio/channels/KQSelector.java
```diff
@@ -0,0 +1,350 @@
...
+
+  @Override
+  protected SelectionKey register(AbstractSelectableChannel ch, int ops, Object att) {
+    KQSelectionKey k = new KQSelectionKey(this, (NativeSelectableChannel) ch, ops);
+    synchronized (regLock) {
+      Descriptor d = new Descriptor(k.getFD());
+      descriptors.put(k.getFD(), d);
+      d.keys.add(k);
+      handleChangedKey(d);
+    }
```
huntc: For reference, the synchronized block is what changed in the JNR PR.

huntc:
Failing CI due to some AMQP thing - nothing to do with this PR, I don't think.

huntc:
Hi @ennru and team - in case it isn't clear (there are a lot of comments!), this PR is complete from Jason and my perspective. :-)

huntc:
Our pleasure! Don’t forget to squash on merging! :-)

PR #1146, opened 2018-08-13
Add AMQP support for all SSL configuration variants

### LINKED ISSUE #1145: No convenient AmqpConnectionProvider with TLS certificate verification
By default, the Java RabbitMQ client will use the `TrustEverythingTrustManager`, which trusts every certificate.

In https://github.com/akka/alpakka/blob/f6ecfa1dac41514a51340c031db5d2970ff46b62/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala#L124 we use `ConnectionFactory#useSslProtocol` but only the one-argument version which lets you specify the protocol. This still uses the default trust manager. Instead, we should allow use of one of the other overloads of `useSslProtocol`, either specifying an `SSLContext` or a protocol string and `TrustManager`.

The workaround is to use the `AmqpConnectionFactoryConnectionProvider` and configure the `ConnectionFactory` manually.

juanjoDiaz:
Closes #1145

This change breaks the current API but it's the only way I could think of adding support for all the possible variants of this configuration.

Wdyt??

On an unrelated note, I noticed that there are some other RabbitMQ options missing from Alpakka. Once this PR gets in, I'll also add those missing options.


ennru:
Thank you for jumping on this one. Please be aware of https://github.com/akka/alpakka/issues/1064 as we're trying to increase chances for staying binary compatible, most settings classes will be changed to non-case classes.

juanjoDiaz:
Thanks for the warning. It should be trivial to convert this into a  non-case class.
Maybe it makes more sense to get this in and then modify all the classes to non-case classes than the other way around?

Just for my own understanding. Could you briefly elaborate on why non-case classes are better here?

juanjoDiaz:
I just saw #1125 which does the non-casing. Should I wait for that to be merged? or how should we proceed?

2m:
#1125 has been merged. Please rebase and convert `AmqpSSLConfiguration` to a regular class.

juanjoDiaz:
Two questions:
- Should `AmqpCredentials` also be moved to a non-case class?
- In line 126, I pattern match to see what kind of SSL config the user is using. Any tips on how to do that with regular classes without having to if-else every possible scenario?

ennru:
Now I've merged https://github.com/akka/alpakka/pull/1163, would be great if you'd like to rebase and continue this.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -199,6 +210,46 @@ object AmqpCredentials {
     AmqpCredentials(username, password)
 }
 
+final class AmqpSSLConfiguration(val protocol: Option[String] = None,
```
ennru: Make the constructor private and add `apply` and `create` methods taking either `protocol` or `context` to the companion instead, please.

PR #1147, opened 2018-08-13
Reference flow that needs a resource

2m:
Introduces an extension that is used to hold a resource per ActorSystem. One of the usecases for such pattern is the DynamoDB connector.


ennru:
I think this example of an Akka extension to control the resource lifecycle deserves a bit more textual description.

2m:
Added scaladoc motivating resource extension pattern.

Issue #1031, opened 2018-06-18
Failed: JmsConnectorsSpec sink disconnect exceptional completion


2m:
```
[info] - should sink disconnect exceptional completion *** FAILED *** (95 milliseconds)
[info]   ActiveMQConnection {id=ID:travis-job-988bd44f-80d4-4010-a51f-9a1858f06dd2-40599-1529123777756-220:1,clientId=ID:travis-job-988bd44f-80d4-4010-a51f-9a1858f06dd2-40599-1529123777756-219:1,started=false} was not null (JmsConnectorsSpec.scala:584)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.MatchersHelper$.indicateFailure(MatchersHelper.scala:346)
[info]   at org.scalatest.Matchers$AnyShouldWrapper.shouldBe(Matchers.scala:7149)
[info]   at akka.stream.alpakka.jms.scaladsl.JmsConnectorsSpec$$anonfun$1$$anonfun$apply$mcV$sp$17$$anonfun$apply$mcV$sp$38.apply(JmsConnectorsSpec.scala:584)
[info]   at akka.stream.alpakka.jms.scaladsl.JmsConnectorsSpec$$anonfun$1$$anonfun$apply$mcV$sp$17$$anonfun$apply$mcV$sp$38.apply(JmsConnectorsSpec.scala:560)
[info]   at akka.stream.alpakka.jms.JmsSpec.withServer(JmsSpec.scala:48)
[info]   at akka.stream.alpakka.jms.scaladsl.JmsConnectorsSpec$$anonfun$1$$anonfun$apply$mcV$sp$17.apply$mcV$sp(JmsConnectorsSpec.scala:560)
[info]   at akka.stream.alpakka.jms.scaladsl.JmsConnectorsSpec$$anonfun$1$$anonfun$apply$mcV$sp$17.apply(JmsConnectorsSpec.scala:560)
[info]   at akka.stream.alpakka.jms.scaladsl.JmsConnectorsSpec$$anonfun$1$$anonfun$apply$mcV$sp$17.apply(JmsConnectorsSpec.scala:560)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
...
```

[...]

PR #1148, opened 2018-08-14
JMS producer multi session

### LINKED ISSUE #1144: JmsProducerStage is single-session only (fix available)
Currently, the JmsProducerStage offers limited throughput only, as it uses a single Jms session only. Thankfully, https://github.com/akka/alpakka/blob/master/jms/src/main/scala/akka/stream/alpakka/jms/JmsProducerStage.scala#L40 indicates very honestly that there is still work to be done.

I've prepared a branch in my fork that I could contribute back to fix this issue: https://github.com/akka/alpakka/compare/master...andreas-schroeder:jms-producer-multi-session
This solution is heavily inspired by Akka Stream's wonderful MapAsyc stage.

Would this be a worthwhile contribution? If so, I would take care of going through the process of contributing, and make this ready for PR.

andreas-schroeder:
- JmsProducerStage now supports multiple sessions by behaving
  similarly to MapAsync
- Producers are pooled and parallelism is controlled by how producers
  are returned to the pool.
- Created a separate JmsMessageProducer class to keep the stage logic
  more focussed on coordinating elements within the stage.
- Grouped the logic for creating the execution context to use in the
  producer stage as well as in the consumer stage.

Fixes #1144

### DIFF
--- a/docs/src/main/paradox/jms.md
+++ b/docs/src/main/paradox/jms.md
@@ -192,6 +192,24 @@ Scala
 Java
 : @@snip [snip](/jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsConnectorsTest.java) { #run-flow-producer }
 
+### Configuring the Producer
+
+Scala
+: @@snip [snip](/jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsSettingsSpec.scala) { #producer-settings }
+
+Java
+: @@snip [snip](/jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsSettingsTest.java) { #producer-settings }
+
+The producer can be configured with the following settings.
+
+* `connectionFactory` (mandatory) the factory to use for creating Jms connections.
+* `destination` (mandatory) the destination (queue or topic) to send Jms messgages to.
+* `credentials` (optional) username and password to use for authentication to the Jms broker .
+* `sessionCount` (defaults to 1) the number of parallel sessions to use for sending Jms messages. Increasing the 
+  number of parallel sessions increases throughput at the cost of message ordering. While the messages may arrive
+  out of order at the Jms broker, the producer flow outputs messages in the order they are received.
+* `timeToLive`  (optional) the time messages should be kept on the Jms broker. This setting can be overridden on 
+  individual messages. If not set, messages will never expire.
 
 ## Receiving messages from a JMS provider
 
--- a/jms/src/main/scala/akka/stream/alpakka/jms/Exceptions.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/Exceptions.scala
@@ -1,0 +1,35 @@
+/*
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.jms
+
+trait RecoverableException extends Exception
+
+trait NonRecoverableException extends Exception
+
+case class UnsupportedMessagePropertyType(propertyName: String, propertyValue: Any, message: JmsMessage)
+    extends Exception(
+      s"Jms property '$propertyName' has unknown type '${propertyValue.getClass.getName}'. " +
+      "Only primitive types and String are supported as property values."
+    )
+    with NonRecoverableException
+
+case class NullMessageProperty(propertyName: String, message: JmsMessage)
+    extends Exception(
+      s"null value was given for Jms property '$propertyName'."
+    )
+    with NonRecoverableException
+
+case class UnsupportedMapMessageEntryType(entryName: String, entryValue: Any, message: JmsMapMessage)
+    extends Exception(
+      s"Jms MapMessage entry '$entryName' has unknown type '${entryValue.getClass.getName}'. " +
+      "Only primitive types, String, and Byte array are supported as entry values."
+    )
+    with NonRecoverableException
+
+case class NullMapMessageEntry(entryName: String, message: JmsMapMessage)
+    extends Exception(
+      s"null value was given for Jms MapMessage entry '$entryName'."
+    )
+    with NonRecoverableException
--- a/jms/src/main/scala/akka/stream/alpakka/jms/Exceptions.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/Exceptions.scala
@@ -4,32 +4,38 @@
 
 package akka.stream.alpakka.jms
 
-trait RecoverableException extends Exception
+/**
+ * Marker trait indicating that the exception thrown is intermittent. The failed operation might succeed if tried again.
+ */
+trait RetriableException extends Exception
 
-trait NonRecoverableException extends Exception
+/**
+ * Marker trait indicating that the exception thrown is persistent. The operation will always fail when retried.
+ */
+trait NonRetriableException extends Exception
 
 case class UnsupportedMessagePropertyType(propertyName: String, propertyValue: Any, message: JmsMessage)
     extends Exception(
       s"Jms property '$propertyName' has unknown type '${propertyValue.getClass.getName}'. " +
       "Only primitive types and String are supported as property values."
     )
-    with NonRecoverableException
+    with NonRetriableException
 
 case class NullMessageProperty(propertyName: String, message: JmsMessage)
     extends Exception(
       s"null value was given for Jms property '$propertyName'."
     )
-    with NonRecoverableException
+    with NonRetriableException
 
 case class UnsupportedMapMessageEntryType(entryName: String, entryValue: Any, message: JmsMapMessage)
     extends Exception(
       s"Jms MapMessage entry '$entryName' has unknown type '${entryValue.getClass.getName}'. " +
       "Only primitive types, String, and Byte array are supported as entry values."
     )
-    with NonRecoverableException
+    with NonRetriableException
 
 case class NullMapMessageEntry(entryName: String, message: JmsMapMessage)
     extends Exception(
       s"null value was given for Jms MapMessage entry '$entryName'."
     )
-    with NonRecoverableException
+    with NonRetriableException
--- a/jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
@@ -40,7 +40,7 @@ private[jms] trait JmsConnector { this: GraphStageLogic =>
       onSessionOpened(session)
     }
 
-  private[jms] def executionContext(attributes: Attributes): ExecutionContext = {
+  protected def executionContext(attributes: Attributes): ExecutionContext = {
     val dispatcher = attributes.get[ActorAttributes.Dispatcher](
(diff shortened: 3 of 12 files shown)


jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
```diff
@@ -90,6 +102,96 @@ private[jms] trait JmsConnector { this: GraphStageLogic =>
...
+    properties.foreach {
+      case (key, v) =>
+        v match {
+          case v: String => message.setStringProperty(key, v)
+          case v: Int => message.setIntProperty(key, v)
+          case v: Boolean => message.setBooleanProperty(key, v)
+          case v: Byte => message.setByteProperty(key, v)
+          case v: Short => message.setShortProperty(key, v)
+          case v: Long => message.setLongProperty(key, v)
+          case v: Double => message.setDoubleProperty(key, v)
```
WellingR: If the properties contain a value which is not of any of these types, would it make sense to explicitly throw an exception instead of the default MatchError?

andreas-schroeder: In general, I would say yes.

However, the Jms producer crashes on any exception, and does not distinguish between retriable errors and non-retriable ones - it doesn't perform any retries. So in the end, the behavior would still be very similar if we introduced a custom exception here. It would make much more sense to introduce a custom exception if the Jms producer stage did retries, and would not try to do those retries if the property value type is of an unexpected type.

(please note:  this is part of the code I moved out of the stage into a separate class - I didn't modify it)

ennru:
Test compilation fails for Scala 2.11.

andreas-schroeder:
@ennru thanks for notifying and sorry for not seeing this myself. I will take care of fixing the issues.

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
andreas-schroeder: @WellingR after giving it some thought, those exceptions can be handy in supervision logic to, so here is a proposal - what do you think? https://github.com/akka/alpakka/pull/1148/commits/a2805765cdd09868f878f2d514788d1914d984b1

andreas-schroeder:
@ennru compilation is now fixed, I just happen to add exception classes as suggested by @WellingR .

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
```diff
@@ -40,11 +40,23 @@ private[jms] trait JmsConnector { this: GraphStageLogic =>
       onSessionOpened(session)
     }
 
-  private[jms] def initSessionAsync(dispatcher: Dispatcher): Future[Unit] = {
-    ec = materializer match {
+  private[jms] def executionContext(attributes: Attributes): ExecutionContext = {
```
ennru: This and a few other things in this class should be just regular `private` or `protected`.

jms/src/main/scala/akka/stream/alpakka/jms/JmsProducerStage.scala
```diff
@@ -21,11 +28,14 @@ private[jms] final class JmsProducerStage[A <: JmsMessage](settings: JmsProducer
...
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
     new GraphStageLogic(shape) with JmsConnector {
 
-      private var jmsProducer: MessageProducer = _
-      private var jmsSession: JmsSession = _
+      private lazy val decider = inheritedAttributes.mandatoryAttribute[SupervisionStrategy].decider
+
+      private val jmsProducers: Buffer[JmsMessageProducer] = Buffer(settings.sessionCount, settings.sessionCount)
+
+      private val buffer: Buffer[Holder[A]] = Buffer(settings.sessionCount, settings.sessionCount)
```
ennru: Would `producerWithMessage` describe better what it holds?

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsConnectorsSpec.scala
```diff
@@ -782,5 +785,92 @@ class JmsConnectorsSpec extends JmsSpec {
 
       result.futureValue should ===(input)
     }
+
+    "publish and consume strings through a queue with multiple sessions" in withServer() { ctx =>
+      //#connection-factory
```
ennru: You need to remove the copied `//#` as that marks what Paradox cuts out for the documentation.
It would be very useful if you added a section on the producer settings in the documentation.

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
andreas-schroeder: agreed; I wanted to follow the style of the existing implementation, I'll change that to simpler visibility restrictions.

jms/src/main/scala/akka/stream/alpakka/jms/JmsProducerStage.scala
andreas-schroeder: how about `inFlightMessagesWithProducer`?

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsConnectorsSpec.scala
andreas-schroeder: I see, I wasn't sure what `//#` does. I'll add producer settings documentation as well.

jms/src/main/scala/akka/stream/alpakka/jms/Exceptions.scala
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.jms
+
+/**
+ * Marker trait indicating that the exception thrown is intermittent. The failed operation might succeed if tried again.
+ */
+trait RetriableException extends Exception
```
WellingR: It makes sense to mark some exception cases as `NonRetriableException`. However the other way around `RetriableException` does not make as much sense.

The essence is that you want to point out for some specific exceptions that a retry makes no sense. However for all other exceptions, a retry might make sense (so there is no real need for the marker trait `RetriableException` especially since the code does not even use this).

jms/src/main/scala/akka/stream/alpakka/jms/Exceptions.scala
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Marker trait indicating that the exception thrown is persistent. The operation will always fail when retried.
+ */
+trait NonRetriableException extends Exception
```
WellingR: Since this is a jms library, maybe we should call this a `NonRetriableJmsException`.

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
WellingR: Looks better!
I had some comments about the exception traits though.

andreas-schroeder: 👍 I've followed your suggestions :)

andreas-schroeder:
Travis failed because an AMQP test failed: https://travis-ci.org/akka/alpakka/jobs/420470777#L680

docs/src/main/paradox/jms.md
```diff
@@ -192,6 +192,24 @@ Scala
...
+The producer can be configured with the following settings.
+
+* `connectionFactory` (mandatory) the factory to use for creating Jms connections.
+* `destination` (mandatory) the destination (queue or topic) to send Jms messgages to.
+* `credentials` (optional) username and password to use for authentication to the Jms broker .
+* `sessionCount` (defaults to 1) the number of parallel sessions to use for sending Jms messages. Increasing the 
+  number of parallel sessions increases throughput at the cost of message ordering. While the messages may arrive
+  out of order at the Jms broker, the producer flow outputs messages in the order they are received.
+* `timeToLive`  (optional) the time messages should be kept on the Jms broker. This setting can be overridden on 
+  individual messages. If not set, messages will never expire.
```
andreas-schroeder: Note that I didn't document `acknowledgeMode`. This is because I read various threads on Jms  stating that this setting is not considered when producing records.

andreas-schroeder:
Looks like it is always the same AMQP test that is consistently failing:
https://travis-ci.org/akka/alpakka/jobs/420947066#L631
https://travis-ci.org/akka/alpakka/jobs/420780309#L810
Should I rebase this PR on top of master?

ennru:
Don't worry for the AMQP test, it is not your fault. See #1167.

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
```diff
@@ -79,8 +91,8 @@ private[jms] trait JmsConnector { this: GraphStageLogic =>
     }
 
     val sessionCount = jmsSettings match {
-      case settings: JmsConsumerSettings =>
-        settings.sessionCount
+      case settings: JmsConsumerSettings => settings.sessionCount
+      case settings: JmsProducerSettings => settings.sessionCount
```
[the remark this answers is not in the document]
andreas-schroeder: Will do :) I try to find some time today to finish things up.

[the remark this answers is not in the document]
andreas-schroeder: Actually, the `JmsBrowseSettings` would then need to implement `sessionCount` as well, which would be hard-coded to 1. Should we go for that?

andreas-schroeder:
Test failure is https://travis-ci.org/akka/alpakka/jobs/422226062#L715 
I'm not 100% sure, but that test looks flaky to me, as the broker is stopped after the stream is started. If stopping the broker is delayed long enough for the stream to be actually started, a cached connection will be created. I can reproduce the failure with a `Thread.sleep(1000)` before `ctx.broker.stop()` here: https://github.com/akka/alpakka/blob/master/jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsConnectorsSpec.scala#L580

andreas-schroeder:
well, now I am at a loss at how to stabilize the failing test - any ideas welcome :)

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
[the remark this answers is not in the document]
ennru: Yes, makes sense to me.

ennru:
We've seen failure on that from time to time: https://github.com/akka/alpakka/issues/1031
If you find a way to make it stable, please do (eg. in another PR).

andreas-schroeder:
Now, it's an ftp test that fails: https://travis-ci.org/akka/alpakka/jobs/422635814#L644

ennru:
Thank you for this great addition to the JMS connector!

Issue #829, opened 2018-03-11
SQS refactor


aserrallerios:
Honestly, the SQS module is a bit overloaded, could use better naming conventions and have a more consistent design.

Also, it'd be interesting to study how to use `AmazonSQSBufferedAsyncClient` properly in alpakka.

ennru:
Yes, I'm looking into the SQS connector (see #922) as we have test instabilities and I agree we can improve the API. I try to improve the Java API experience, but I will try to change it in a backwards compatible way for now.

ennru:
Now that the Alpakka 1.0 changes are merged, do you think there are more changes we should consider?

aserrallerios:
Still not sure about the `AmazonSQSBufferedAsyncClient`. It could make part of the alpakka features redundant. But, it's kinda out of scope, so this issue could be closed.

ennru:
I created issue #1238 to discuss AmazonSQSBufferedAsyncClient further.
Closing this.

PR #1149, opened 2018-08-15
AWS SQS towards 1.0 - major API changes

### LINKED ISSUE #1091: AWS SQS: become Alpakka 1.0 ready
- [x] Move internal code to `impl` package

ennru:
Fixes #1091.

The diff is quite huge as a lot of code has moved around.

I changed the API a bit to remove the use of tuples as flow data type, mainly for the Java side of things.
So it might be good if @aserrallerios can have a look regarding is ideas in #829.


ennru:
Thank you @aserrallerios, very good input.
I've taken your ideas into account and made even larger changes to the API to get it much more consistent.

ennru:
I even wonder if `SqsAckFlow` is the best name, wouldn't `SqsActionFlow` describe better what it offers?

aserrallerios:
My 2 cents:

But `Publish` can also be an action :)

And after giving it a thought, I think that we should keep them separated, as the input of both streams is different:

* Ack (IN: a consumed message from SQS)
   * ignore
   * delete
   * republish
* Publish (IN: a message body)

Which term can include the three `Ack` types but not `Publish`? For lack of a better term I'd stick to `Ack`.

ennru:
Yes, separated they need to be. I just contemplated if `Ack` is misleading as more sounds data receiving thing where you can acknowledge received messages. But it is about updating the queue instead.

2m:
Regarding `SqsAckFlow` I also think that having ignore delete and republish all under `ack` can be a bit confusing. Normally `ack` is meant as a positive acknowledgement of a message.

What about `SqsConfirmFlow`? Or `SqsAfterPublishFlow`?

aserrallerios:
And what about dropping the `Ignore` action? If someone wants to ignore messages he doesn't even need a custom Flow/Sink.

I'm not sold on `Confirm`, usually the confirm term is reserved as security double-check step before executing an action.

2m:
Good point about dropping `Ignore`. If we have only delete and republish left, then calling it `SqsAckFlow` is not that bad.

We could also use `Ack` and `Nak` (negative acknowledgement) in the connector for delete and republish actions.

2m:
Needs a rebase.

PR #1155, opened 2018-08-17
Google BigQuery integration - WIP

tg44:
Hy!

I had a small time finally to port [this](https://github.com/emartech/rdb-connector-bigquery) to an alpakka connector!

It's a huge code base, but at least its battle-tested. If you have any thoughts pls share it. 

(The current state is mostly a ported/refactored/reorganized version of the emartech repo, so I added my team as coauthor. I hope this is not a problem.)


ennru:
Sorry, for not getting to it yet... will try to now.

tg44:
Thanks for the review, I will correct the code based on your suggestions.

Exposing from the impl package can be solved by moving some code/model/helper from the impl to outside. The ordering and the misused implicit param is trivial to solve too. I think I will find some time for these on the next week.

The Try.toOption is an interesting question. We expected that the user who provides these functions will test the input function and log if bad things happens. As I browsed the code to make a proper answer, I found some interesting things which are there as leftovers, so maybe I can clean something up from there. I think I can erase the Option, and just let the code explode, if the user gives us a bad parser...

Yapp, I can write a tutorial how to test this on google, maybe export some code from our original codebase. (Will do with the refactors.)

I don't see how this code could be an akka extension. If we don't count the first two functions (which I think just some helper/example for devs whose don't really want to read the whole documentation), the others are generating Sources. I don't think we need to wrap this more, but I'm open for suggestions.

Spray vs. Java: I can offer an api which works with a `String => Option[T]` (or if I do the refactor I mentioned earlier a `String => T`), but that `bytestring => json => {java-api} string => {usercode} json => {usercode}T` could hurt performance if you streaming terrabytes of data...  I can live with that, but I think its less practical... If the user don't care; he/she can call a .prettyPrint and then handle it as a string, or can use the given JsObject as a map so he/she will save some conversions. Idk maybe with some documentation this can be more clear, but I'm pretty sure that the JsObject as an interface gives more place to play with (even if java ppl not familiar with spray-json). (But ofc. if you have any concrete suggestion I'm open to them :D )

tg44:
I cleaned up a lot of things, but I need to test this implementation against the real service. I will migrate some integration tests.

tg44:
Added end to end tests! It can be configured from env-vars.

For reproduction on CI you need to get a new BigQuery up and running, and get the json from the 4th point:
https://cloud.google.com/bigquery/docs/quickstarts/quickstart-client-libraries

After that you need to copy the things from the json to env-vars defind in the https://github.com/akka/alpakka/pull/1155/files#diff-2e3c5dd1e41f0d06a9455df3380aff92 file.
The cert is a bit tricky, in some cases you need to replace lineendings to \n-s so the envvar can be written in a single line. (for ex. if you use the .env plugin you have to make it as a single line)

tg44:
Ahh I found out today that I cleaned more code then needed so I will need to reinsert them... I forgot that if the query returns with an empty result (bcs its not finished yet) the parsers will fail (thats why we parsed everything as an option) instead of just wait a bit more to get the result set. :( So I will rewrite that part of the stream.

ennru:
@tg44 We might be ready for a new milestone release of Alpakka soon, do you think you will pick this up again soonish?

tg44:
Yeah I try to make it work again in the next days.

tg44:
Soo I added back the previously refactored functionality + added the newly implemented "try to cancel the query on timeout" too.

I have no idea if its alpakka-1.0 ready or not, some bincomp and case-class pro should check that out.

ennru:
Great, I'll have a look.

ennru:
Ping @tg44 - any chance to push this further?

gkatzioura:
Hi @ennru I am interested in using it due to a project. Is it ok to make another pull request? Will take some time but I went through the process on adding an integration before.

tg44:
I think this integration is not in very bad shape, probably it only needs a rebase and some minor refactors with the mind of bincomp. I even think if you just copy-paste my code to a new project it will work as expected. I would really glad if you could fix this PR instead of reimplementing the whole thing :)

gkatzioura:
> I think this integration is not in very bad shape, probably it only needs a rebase and some minor refactors with the mind of bincomp. I even think if you just copy-paste my code to a new project it will work as expected. I would really glad if you could fix this PR instead of reimplementing the whole thing :)

Hi @tg44! Thanks for your work . I am not intending on copy pasting your code since I don't thing it is appropriate from many perspectives (licensing, plagiarism, using it without reaching the official Alpakka codebase  etc). My goal is to have a BigQuery integration. Your work would help in terms of using it as an example, and giving a big head start but I definitely don't want to get into the trouble of cp.

tg44:
I meant, that I think understanding and fixing this PR is much less work than starting it from scratch :) (Especially that this is tested, and battle-tested :D ) If you need any help to understand the code ping me!

ennru:
Hi @gkatzioura,
It would be great if you can fit the loose ends together for a Google BigQuery connector. As @tg44 says, please don't start over, take this branch and push it to your repository to work on, than open a new PR from that branch.
Collaboration is the beauty of open source.

gkatzioura:
Hi @ennru and @tg44 
This is a catch up message and the progress so far.
I have been able to proceed with the changes requested by @ennru  and currently I am working on the testing and evaluating the options (mocking, Hoverfly etc). Will get back soon with a pull request based on this branch.

ennru:
Superseded by #2106

PR #1162, opened 2018-08-23
Google Pub Sub connector built on Akka Grpc

2m:
This is an alternative connector to [Google Pub Sub](https://developer.lightbend.com/docs/alpakka/current/google-cloud-pub-sub.html) that uses gRPC protocol.

This is a continuation of https://github.com/akka/alpakka/pull/759 with the difference that it uses Akka Grpc instead of Google Java gRPC libraries.

This is still WIP, as it needs the following:

* [x] separate Java and Scala extensions, so Java sub/pub client is not created when only Scala one is used or vice versa.


2m:
Code is ready to be reviewed.

Docs still need an update, since those have been copied from the current Google Pub Sub connector.

google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/impl/AkkaGrpcSettings.scala
```diff
@@ -0,0 +1,55 @@
...
+      |port = ${config.port}
+      |
+      |$sslConfig
+      |""".stripMargin
+
+    val settings =
+      GrpcClientSettings.fromConfig(
+        ConfigFactory
+          .parseString(akkaGrpcConfig)
+          .withFallback(sys.settings.config.getConfig("akka.grpc.client.\"*\""))
```
2m: Since we do not want to expose akka-grpc config settings, and the ssl-context loading part in   `GrpcClientSettings` [is private](https://github.com/akka/akka-grpc/blob/607599b0408c5e8bbdeb5bf59c07e9863c7ae566/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala#L123), we build `GrpcClientSettings` from Config and then set the remaining attributes (`useTls` and `callCredentials`) using setters.

//cc @raboof

build.sbt
```diff
@@ -114,6 +115,18 @@ lazy val googleCloudPubSub = alpakkaProject(
...
 )
 
+lazy val googleCloudPubSubGrpc = alpakkaProject(
+  "google-cloud-pub-sub-grpc",
+  "google.cloud.pubsub.grpc",
+  Dependencies.GooglePubSubGrpc,
+  akkaGrpcCodeGeneratorSettings ~= { _.filterNot(_ == "flat_package") },
+  akkaGrpcGeneratedSources := Seq(AkkaGrpc.Client),
+  akkaGrpcGeneratedLanguages := Seq(AkkaGrpc.Scala, AkkaGrpc.Java),
+  javaAgents += Dependencies.GooglePubSubGrpcAlpnAgent % "test",
```
2m: @raboof will the users of alpakka-google-pub-sub-grpc connector need to add `javaAgent` to their build manually?

raboof: I'm not completely sure if the Agent is needed only for the server side or also for the client - did you try it and see it fail without?

2m: Without it client gets stuck after:
```
...
2018-08-24 13:37:54,819 DEBUG [ExampleApp-akka.actor.default-dispatcher-3] [i.g.n.s.io.netty.handler.ssl.OpenSsl]  Supported protocols (OpenSSL): [SSLv2Hello, SSLv3, TLSv1, TLSv1.1, TLSv1.2]
2018-08-24 13:37:54,819 DEBUG [ExampleApp-akka.actor.default-dispatcher-3] [i.g.n.s.io.netty.handler.ssl.OpenSsl]  Default cipher suites (OpenSSL): [TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384, TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256, TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256, TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA, TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA, TLS_RSA_WITH_AES_128_GCM_SHA256, TLS_RSA_WITH_AES_128_CBC_SHA, TLS_RSA_WITH_AES_256_CBC_SHA]
```

2m:
Updated the docs to reflect the configuration of this connector.

google-cloud-pub-sub-grpc/src/test/scala/docs/scaladsl/ExampleUsage.scala
```diff
@@ -0,0 +1,101 @@
...
+import com.google.pubsub.v1.pubsub._
+
+import scala.collection.immutable.Seq
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+class ExampleUsage {
+
+  //#init-mat
+  implicit val system = ActorSystem()
```
ennru: Annotate with types, please.

google-cloud-pub-sub-grpc/src/test/scala/docs/scaladsl/ExampleUsage.scala
```diff
@@ -0,0 +1,101 @@
...
+  //#init-client
+  val projectId = "test-XXXXX"
+  val topic = "topic1"
+  val subscription = "subscription1"
+
+  val config = PubSubConfig(system.settings.config)
+  //#init-client
+
+  //#publish-single
+  val publishMessage: PubsubMessage =
```
ennru: Make the relevant imports show in the snippet.

google-cloud-pub-sub-grpc/src/test/java/docs/javadsl/ExampleUsageJava.java
```diff
@@ -0,0 +1,111 @@
...
+    Flow<PublishRequest, PublishResponse, NotUsed> publishFlow = GooglePubSub.publish(1, system);
+
+    CompletionStage<List<PublishResponse>> publishedMessageIds =
+        source.via(publishFlow).runWith(Sink.seq(), materializer);
+    // #publish-single
+
+    // #publish-fast
+    Source<PubsubMessage, NotUsed> messageSource = Source.single(publishMessage);
+    messageSource
+        .groupedWithin(1000, FiniteDuration.apply(1, "min"))
```
ennru: Use `java.time.Duration` instead.

google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/impl/GrpcCredentials.scala
```diff
@@ -0,0 +1,24 @@
...
+import scala.collection.JavaConverters._
+
+/**
+ * Internal API
+ */
+@InternalApi private[grpc] object GrpcCredentials {
+
+  def applicationDefault(): CallCredentials =
+    MoreCallCredentials.from(
+      GoogleCredentials.getApplicationDefault.createScoped(List("https://www.googleapis.com/auth/pubsub").asJava)
```
ennru: What about using `Collections.singletonList` directly?

google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/javadsl/GooglePubSubExternal.scala
```diff
@@ -0,0 +1,52 @@
...
+      .setStreamAckDeadlineSeconds(0)
+      .build()
+
+    subscriber.client
+      .streamingPull(
+        Source
+          .single(request)
+          .concat(
+            Source
+              .tick(Duration.ZERO, pollInterval, ())
```
ennru: Shouldn't this be `pollInterval` as initial delay, as well?

google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/javadsl/GooglePubSubExternal.scala
```diff
@@ -0,0 +1,52 @@
...
+              .map(_ => subsequentRequest)
```
ennru: Why not use `subsequentRequest` in `tick`?

docs/src/main/paradox/google-cloud-pub-sub-grpc.md
```diff
@@ -0,0 +1,107 @@
...
+
+Java
+: @@snip (/google-cloud-pub-sub-grpc/src/test/java/docs/javadsl/ExampleUsageJava.java) { #subscribe-auto-ack }
+
+## Subscribing
+
+To receive message from a subscription, first we create a `StreamingPullRequest` with a FQRS of a subscription and
+a deadline for acknowledgemts in seconds. Google requires that only the first `StreamingPullRequest` has the subscription
+and the deadline set. This connector takes care of that and clears up the subscription FQRS and the deadline for subsequent
+`StreamingPullRequest` messages.
```
ennru: Add something about the `acknowledge` sink here.

docs/src/main/paradox/google-cloud-pub-sub-grpc.md
```diff
@@ -0,0 +1,107 @@
...
+
+    // receive messages from a subsciptions
+    > google-cloud-pub-sub-grpc/Test/run subscribe <project-id> <subscription-name>
+
+    // publish a single message to a topic
+    > google-cloud-pub-sub-grpc/Test/run publish-single <project-id> <topic-name>
+
+    // continually publish a message stream to a topic
+    > google-cloud-pub-sub-grpc/Test/run publish-stream <project-id> <topic-name>
+    ```
```
ennru: Possibly add something about the tools used in the integration tests, should be useful for users of Alpakka, as well.

docs/src/main/paradox/google-cloud-pub-sub-grpc.md
```diff
@@ -0,0 +1,107 @@
+# Google Cloud Pub/Sub gRPC
+
+The google cloud pub/sub connector provides a way to connect to google clouds managed pub/sub https://cloud.google.com/pubsub/.
+
+This connector communicates to PubSub via gRPC protocol. The integration between Akka Stream and gRPC is handled by the
```
ennru: Possibly add a callout with Google's explanation of what Pub/Sub offers. Good for googleability.

ennru: And a note on the difference to the other Alpakka connector...
And vice versa.

google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/javadsl/GooglePubSubExternal.scala
2m: Yup. Nice catch.

2m:
Pushed commits that address feedback comments.

I had no problem generating api docs when running `google-cloud-pub-sub-grpc/doc` task. Which task did you use?

I see that scalapb [has support](https://scalapb.github.io/customizations.html#scalapb-file-level-options) for generating scala files under a custom package. However those scalapb specific options need to be defined in the `proto` files. I did not find a way how to control the package name from the outside when proto files come from somewhere else. 

Added `ApiMayChange` on the clients that come from akka-grpc. I do not think `ApiMayChange` is needed on the akka stream operator factories, as those do not expose anything from akka-grpc.

raboof:
Looked a bit at generating Scala files into a custom package, but didn't see a way to achieve this. Submitted https://github.com/scalapb/ScalaPB/issues/495 . Of course to have scaladsl/javadsl, we'd need the same for the protoc Java code generation as well.

docs/src/main/paradox/google-cloud-pub-sub-grpc.md
```diff
@@ -0,0 +1,139 @@
...
+    > google-cloud-pub-sub-grpc/test
+    ```
+
+There is also an @github[ExampleApp](/google-cloud-pub-sub-grpc/src/test/scala/docs/scaladsl/ExampleApp.scala) that can be used
+to test publishing to topics and receiving messages from subscriptions.
+
+To run the example app you will need to configure a project and Pub/Sub in Google Cloud and provide your own credentials.
+
+sbt
+:   &#9;
```
ennru: Is that an HTML tab char?

2m: Yes it is. :) Workaround for https://github.com/lightbend/paradox/issues/161

docs/src/main/paradox/google-cloud-pub-sub-grpc.md
```diff
@@ -0,0 +1,139 @@
...
+
+    ```bash
+    env GOOGLE_APPLICATION_CREDENTIALS=/path/to/application/credentials.json sbt
```
ennru: What is this `sbt` at the end? Should be on a new line I guess.

google-cloud-pub-sub-grpc/src/main/resources/reference.conf
```diff
@@ -0,0 +1,12 @@
+alpakka.google.cloud.pubsub.grpc {
+  host = "pubsub.googleapis.com"
+  port = 443
+
+  # Set to "none" to disable TLS
+  rootCa = "GoogleInternetAuthorityG3.crt"
```
ennru: Wouldn't it be better to put the certificate in a sub-directory, so that it works well even with über-jars?

docs/src/main/paradox/google-cloud-pub-sub-grpc.md
2m: This command runs sbt with modified environment variables (as opposed to modifying environment variables globally and then running sbt).

google-cloud-pub-sub-grpc/src/main/resources/reference.conf
2m: Why would having the certificate in the top directory not work with uber-jars?

ennru: It works. Just thought it might be good to show context to where it is used. Maybe irrelevant.

2m:
I just realized, that there is no akka-grpc for 2.11 released currently (https://github.com/akka/akka-grpc/issues/345). But we need to discuss if we really need it. Maybe Alpakka 1.0 will drop 2.11 and be released for Scala 2.12 and 2.13.

2m:
Refactored compiled-only tests that are used in the documentation snippets to runnable ones.

2m:
Everything is green. This is ready to go in!

2m:
Thank you @CremboC for the initial implementation of this. :) I hope you can try it out and see how it works in your use-case.

PR #1164, opened 2018-08-24
Solr: support update/delete/atomic update operations and batch those

giena:
At this moment, there's a little problem of performance with SolR component because there's no asynchronous part in the code. So the batch size is always 1. 
I propose an alternative by keeping the inspiration from the Elastic Search component.
The solr update is now encapsulated in a future, and the async callbacks are used to make everything coherent (in failure and in success).
Still using the elastic search code, I will provide a notion of operation, to allow the removal of documents.
Please, have a look.

### DIFF
--- a/solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
+++ b/solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
@@ -7,16 +7,21 @@ package akka.stream.alpakka.solr
 import java.net.{ConnectException, SocketException}
 
 import akka.NotUsed
+
+import scala.concurrent.blocking
 import akka.stream.alpakka.solr.SolrFlowStage.{Finished, Idle, Sending}
-import akka.stream.stage.{GraphStage, InHandler, OutHandler, TimerGraphStageLogic}
+import akka.stream.stage._
 import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
 import org.apache.http.NoHttpResponseException
 import org.apache.solr.client.solrj.SolrClient
+import org.apache.solr.client.solrj.response.UpdateResponse
 import org.apache.solr.common.{SolrException, SolrInputDocument}
 
 import scala.collection.JavaConverters._
 import scala.collection.mutable
 import scala.concurrent.Future
+import scala.util.{Failure, Success}
+import scala.concurrent.ExecutionContext.Implicits.global
 import scala.util.control.NonFatal
 
 object IncomingMessage {
@@ -71,12 +76,15 @@ private[solr] final class SolrFlowLogic[T, C](
     messageBinder: T => SolrInputDocument
 ) extends TimerGraphStageLogic(shape)
     with OutHandler
-    with InHandler {
+    with InHandler
+    with StageLogging {
 
   private var state: SolrFlowState = Idle
   private val queue = new mutable.Queue[IncomingMessage[T, C]]()
   private var failedMessages: Seq[IncomingMessage[T, C]] = Nil
   private var retryCount: Int = 0
+  private val failureHandler = getAsyncCallback[(Seq[IncomingMessage[T, C]], Throwable)](handleFailure)
+  private val responseHandler = getAsyncCallback[(Seq[IncomingMessage[T, C]], Int)](handleResponse)
 
   setHandlers(in, out, this)
 
@@ -85,6 +93,7 @@ private[solr] final class SolrFlowLogic[T, C](
 
   override def onPush(): Unit = {
     queue.enqueue(grab(in))
+
     state match {
       case Idle => {
         state = Sending
@@ -121,16 +130,24 @@ private[solr] final class SolrFlowLogic[T, C](
       pull(in)
     }
 
-  private def handleFailure(messages: Seq[IncomingMessage[T, C]], exc: Throwable): Unit =
+  private def handleFailure(args: (Seq[IncomingMessage[T, C]], Throwable)): Unit = {
+    val (messages, exc) = args
     if (retryCount >= settings.maxRetry) {
+      log.warning(s"Received error from solr. Giving up after $retryCount tries. Error: ${exc.toString}")
       failStage(exc)
     } else {
       retryCount = retryCount + 1
+      log.warning(
+        s"Received error from solr. (re)tryCount: $retryCount maxTries: ${settings.maxRetry}. Error: ${exc.toString}"
+      )
       failedMessages = messages
       scheduleOnce(NotUsed, settings.retryInterval)
     }
+  }
 
-  private def handleResponse(messages: Seq[IncomingMessage[T, C]], status: Int): Unit = {
+  private def handleResponse(args: (Seq[IncomingMessage[T, C]], Int)): Unit = {
+    val (messages, status) = args
+    log.debug(s"Handle the response with $status")
     retryCount = 0
     val result = messages.map(m => IncomingMessageResult(m.source, m.passThrough, status))
 
@@ -146,6 +163,9 @@ private[solr] final class SolrFlowLogic[T, C](
         case _ => state = Idle
       }
     } else {
+      if (log.isDebugEnabled) {
+        log.debug(s"Remains ${nextMessages.size} to send")
+      }
       sendBulkToSolr(nextMessages)
     }
   }
@@ -153,22 +173,31 @@ private[solr] final class SolrFlowLogic[T, C](
   private def handleSuccess(): Unit =
     completeStage()
 
+  private def fSendBulkToSolr(messages: Seq[IncomingMessage[T, C]]): Future[UpdateResponse] = {
+    val docs = messages.map(message => messageBinder(message.source))
+    Future {
+      blocking {
+        client.add(collection, docs.asJava, settings.commitWithin)
+      }
+    }
+  }
+
   private def sendBulkToSolr(messages: Seq[IncomingMessage[T, C]]): Unit =
-    try {
-      val docs = messages.map(message => messageBinder(message.source))
-      val response = client.add(collection, docs.asJava, settings.commitWithin)
-      handleResponse(messages, response.getStatus)
-    } catch {
-      case NonFatal(exc) =>
-        val rootCause = SolrException.getRootCause(exc)
-        if (shouldRetry(rootCause)) {
-          handleFailure(messages, exc)
-        } else {
-          val status = exc match {
-            case e: SolrException => e.code()
-            case _ => -1
-          }
-          handleResponse(messages, status)
+    fSendBulkToSolr(messages).onComplete {
+      case Success(response) => responseHandler.invoke((messages, response.getStatus))
+      case Failure(exception) =>
+        exception match {
+          case NonFatal(exc) =>
+            val rootCause = SolrException.getRootCause(exc)
+            if (shouldRetry(rootCause)) {
+              failureHandler.invoke((messages, exception))
+            } else {
+              val status = exc match {
+                case e: SolrException => e.code()
+                case _ => -1
+              }
+              responseHandler.invoke((messages, status))
+            }
         }
     }
 
--- a/solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
+++ b/solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
@@ -17,45 +17,131 @@ import org.apache.solr.client.solrj.SolrClient
 import org.apache.solr.client.solrj.response.UpdateResponse
 import org.apache.solr.common.{SolrException, SolrInputDocument}
 
-import scala.collection.JavaConverters._
 import scala.collection.mutable
 import scala.util.{Failure, Success}
 import scala.util.control.NonFatal
+import scala.collection.JavaConverters._
 
-object IncomingMessage {
-  // Apply method to use when not using passThrough
+private object IncomingMessage {
+  // Apply methods to use when not using passThrough
   def apply[T](source: T): IncomingMessage[T, NotUsed] =
-    IncomingMessage(Update, None, Some(source), NotUsed)
-
-  def apply[T](operation: Operation, source: T): IncomingMessage[T, NotUsed] =
-    IncomingMessage(operation, None, Some(source), NotUsed)
+    IncomingMessage(Update, None, None, Option(source), Map.empty, NotUsed)
 
   def apply(id: String): IncomingMessage[NotUsed, NotUsed] =
-    IncomingMessage(Delete, Some(id), None, NotUsed)
+    IncomingMessage(Delete, None, Option(id), None, Map.empty, NotUsed)
+
+  def apply(idField: String, id: String, field: String, updates: Map[String, Any]): IncomingMessage[NotUsed, NotUsed] =
+    IncomingMessage(AtomicUpdate, Option(idField), Option(id), None, Map(field -> updates), NotUsed)
+
+  // Apply methods to use with passThrough
+  def apply[T, C](source: T, passThrough: C): IncomingMessage[T, C] =
+    IncomingMessage(Update, None, None, Option(source), Map.empty, passThrough)
+
+  def apply[C](id: String, passThrough: C): IncomingMessage[NotUsed, C] =
+    IncomingMessage(Delete, None, Option(id), None, Map.empty, passThrough)
+
+  def apply[C](idField: String,
+               id: String,
+               field: String,
+               updates: Map[String, Any],
+               passThrough: C): IncomingMessage[NotUsed, C] =
+    IncomingMessage(AtomicUpdate, Option(idField), Option(id), None, Map(field -> updates), passThrough)
 
   // Java-api - without passThrough
   def create[T](source: T): IncomingMessage[T, NotUsed] =
     IncomingMessage(source)
 
-  // Java-api - with passThrough
-  def create[T, C](source: T, passThrough: C): IncomingMessage[T, C] =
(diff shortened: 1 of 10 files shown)


ennru:
Thank you for making the Solr connector non-blocking!

Now that the blocking call is executed asynchronously, you should take care of selecting a proper execution context for it - the global one is not a good choice.
A better default is using Akka's `akka.stream.default-blocking-io-dispatcher` and possibly making it configurable. For inspiration how to use it, look into `JmsConsumerStage`.

ennru:
I've now understood that you can make the async stuff much simpler.
1. Select the `IODispatcher` for the stage
```
override protected def initialAttributes: Attributes = Attributes(ActorAttributes.IODispatcher)
```
2. Remove the `ExecutionContext`, `Future` and `blocking` from the client calls (the stage will execute on the IODispatcher)

This should make the code much simpler.

giena:
@ennru : Not sure to see what you mean. Here a code snippet of the onPush method:
    queue.enqueue(grab(in))

    state match {
      case Idle => {
        state = Sending
        val messages = (1 to settings.bufferSize).flatMap { _ =>
          queue.dequeueFirst(_ => true)
        }
        sendBulkToSolr(messages)
      }
      case _ => ()
    }

    tryPull()
  }
Even if i use IODispatcher, i will send message 1 by 1, if sendBulkToSolr doesnot use Futures...

ennru:
Ah ok, I finally see how you are trying to add asynchronous updates to Solr. I'm afraid your solution will not be able to guarantee the order of updates to Solr, as Futures not necessarily are executed in the order they are created. Are you confident the client library is even thread-safe?

For Elasticsearch client API is prepared for this use-case, it is not added by Alpakka.

If you can ignore the order of updates to Solr in your use-case, you might construct a flow using the Graph DSL, that runs several Solr stages in parallel.

giena:
It is why i use getAsyncCallback on the completion (success or failure) of my future. Since my batch is ordered and the buffer is managed by onPush, all my updates from a source will be ordered in my sink. There is no doubt. For me the client is thread safe (used by us in multithreaded application since some years), ans thus, if we cannot guarantee the order as you said, the elastic search implementation can't do it too. This is the same code.

ennru:
I'm not sure what you're referring to in the Elasticsearch way of doing this. The Alpakka Elasticsearch connector sends all operations as one JSON using the ES client's `performRequestAsync` (see https://github.com/akka/alpakka/blob/master/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchFlowStage.scala#L299).

giena:
The solr client sends documents as a batch too...
https://github.com/worldline-messaging/alpakka/blob/master/solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala#L180
I don't understand why you make some difference...

ennru:
Yes, the Solr client sends batches. But your proposal first runs a `groupBy(operation)` after that calls the client with batches per operation wrapped in futures. The order Solr will see the messages will be something different from what the incoming stream contained.
It would be more correct to implement some `takeWhile(Operation)` and update Solr with those, before taking the next batch.
You should not wrap the call to a client in Futures, the stage's asynchronicity must be enough.
(Side note: `blocking` should not be used in Futures.)

giena:
Ok, see what you mean now. Thank you! I will try to make it.

giena:
To get the best performance with SolR, you need to batch. It's what i do. To batch documents, i need to enqueue them before calling SolR. If i do not use futures, and if i use IODispatcher as you described, i scale horizontally perhaps, but the batch size is always 1 element. And the performance are the same as the initial version. I tell that because i test it. So i keep going with futures, but now the order is respected (takeWhile).

ennru:
Ok, if the Solr API does much better with batches of data, I'd suggest this stage should accept `Seq[IncomingMessage]` and it is the user's responsibility to find the best level of batching with eg. `groupedWithin`. The stage should not apply batching itself.

giena:
Ok. I think we are not so far now.

solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
```diff
@@ -8,51 +8,164 @@ import java.net.{ConnectException, SocketException}
...
-import scala.collection.JavaConverters._
-import scala.collection.mutable
-import scala.concurrent.Future
+import scala.annotation.tailrec
 import scala.util.control.NonFatal
+import scala.collection.JavaConverters._
 
-object IncomingMessage {
-  // Apply method to use when not using passThrough
+private object IncomingMessage {
```
ennru: Why did this become `private`?

I see you're creating different operations depending on the parameter lists, that is very confusing.
Please, have a look at my proposed changes to the Elasticsearch connector, as well: https://github.com/akka/alpakka/pull/1179 (I'm not saying you should do the whole Alpakka 1.0 thing in this PR.)

solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
```diff
@@ -117,60 +228,129 @@ private[solr] final class SolrFlowLogic[T, C](
...
 
-    if (nextMessages.isEmpty) {
-      state match {
-        case Finished => handleSuccess()
-        case _ => state = Idle
+  private def handleSuccess(): Unit =
+    completeStage()
+
+  private def updateBulkToSolr(messages: Seq[IncomingMessage[T, C]]): UpdateResponse = {
+    val binder = messageBinder.get
```
ennru: This might fail with a very unexpected error. Fail with a message users might understand, please.

solr/src/main/scala/akka/stream/alpakka/solr/javadsl/SolrFlow.scala
```diff
@@ -22,29 +23,41 @@ object SolrFlow {
    * Java API: creates a [[akka.stream.alpakka.solr.SolrFlowStage]] for [[SolrInputDocument]]
    * from [[IncomingMessage]] to sequences of [[IncomingMessageResult]].
    */
-  def document(
+  def documents(
```
ennru: Please keep the original signatures and complement with the `Seq`-versions (by putting the single element in a Seq).

solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
```diff
@@ -8,51 +8,164 @@ import java.net.{ConnectException, SocketException}
...
+
+final case class IncomingMessageResult[T, C](id: Option[String], sourceOpt: Option[T], passThrough: C, status: Int)
 
 private[solr] final class SolrFlowStage[T, C](
     collection: String,
     client: SolrClient,
     settings: SolrUpdateSettings,
-    messageBinder: T => SolrInputDocument
-) extends GraphStage[FlowShape[IncomingMessage[T, C], Future[Seq[IncomingMessageResult[T, C]]]]] {
+    messageBinder: Option[T => SolrInputDocument]
```
ennru: Did you make use of this being an `Option`?

solr/src/test/java/akka/stream/alpakka/solr/SolrTest.java
```diff
@@ -98,9 +103,12 @@ public void solrInputDocumentStream() throws Exception {
                 tuple -> {
                   Book book = tupleToBook.apply(tuple);
                   SolrInputDocument doc = bookToDoc.apply(book);
-                  return IncomingMessage.create(doc);
+                  List<IncomingMessage<SolrInputDocument, NotUsed>> list = new ArrayList<>();
+                  list.add(IncomingUpdateMessage.create(doc));
+                  return list;
```
ennru: Instead of building the lists in user code, the examples should show `groupedWithin`.

solr/src/test/java/akka/stream/alpakka/solr/SolrTest.java
```diff
@@ -322,13 +351,136 @@ public void testKafkaExample() throws Exception {
...
+    assertEquals(expect, result);
+  }
+
+  @Test
+  public void atomicUpdateDocuments() throws Exception {
+    // Copy collection1 to collection2 through document stream
+    createCollection("collection8"); // create a new collection
+    TupleStream stream = getTupleStream("collection1");
+
+    // #run-document
```
ennru: Did you check the generated documentation? It looks like the `#run-document` tag is used for different things.

solr/src/test/scala/akka/stream/alpakka/solr/SolrSpec.scala
```diff
@@ -67,18 +68,20 @@ class SolrSpec extends WordSpecLike with Matchers with BeforeAndAfterAll {
         .map { tuple: Tuple =>
           val book: Book = tupleToBook(tuple)
           val doc: SolrInputDocument = bookToDoc(book)
-          IncomingMessage(doc)
+          Seq(IncomingUpdateMessage(doc))
```
ennru: Same about `groupedWithin`.

solr/src/test/scala/akka/stream/alpakka/solr/SolrSpec.scala
```diff
@@ -67,18 +68,20 @@ class SolrSpec extends WordSpecLike with Matchers with BeforeAndAfterAll {
...
         }
         .runWith(
-          SolrSink.document(
+          SolrSink.documents(
             collection = "collection2",
             settings = SolrUpdateSettings(commitWithin = 5)
-          )
+          )(cluster.getSolrClient)
```
ennru: Why are you passing the Solr client explicitly?

solr/src/test/scala/akka/stream/alpakka/solr/SolrSpec.scala
```diff
@@ -313,12 +322,284 @@ class SolrSpec extends WordSpecLike with Matchers with BeforeAndAfterAll {
...
     }
   }
 
+  "Un-typed Solr connector" should {
+    "consume and delete documents" in {
+      // Copy collection1 to collection2 through document stream
+      createCollection("collection7") //create a new collection
+      val stream = getTupleStream("collection1")
+
+      //#run-document
```
ennru: Duplicated `#run-document` even here.

solr/src/test/scala/akka/stream/alpakka/solr/SolrSpec.scala
giena: Because the implicit was another solr client, and we do not have to instantiate multiple clients. We could instantiate an implicit lazily with this instance.

solr/src/test/java/akka/stream/alpakka/solr/SolrTest.java
giena: Yes it should

solr/src/test/java/akka/stream/alpakka/solr/SolrTest.java
giena: No

giena:
@ennru I've done my best to complete your main requirements in this last code review. But i won't work on this project for the next weeks. However I will follow this pull request to help people if necessary. And i hope it will be successfully merged soon. Thank you for your ideas and support. See you.

solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
```diff
@@ -84,14 +257,12 @@ private[solr] final class SolrFlowLogic[T, C](
...
     state match {
       case Idle => {
         state = Sending
-        val messages = (1 to settings.bufferSize).flatMap { _ =>
-          queue.dequeueFirst(_ => true)
-        }
-        sendBulkToSolr(messages)
+        sendBulkToSolr(messagesIn)
       }
       case _ => ()
```
2m: Since the messages are no longer enqueued they will be lost here if the `state` is not `Idle`.

giena: Yep, you're right. Did you think we do not need anymore state management? Or something else?

2m: I have not looked further, maybe that is already the case, but it this stage should not be pushed when it is in state other than `Idle`. That means, that this stage should only pull, when it is in state `Idle`.

giena: I ve chosen the logical option: I do not manage the internal state since there's no more internal batching

giena:
Let's go for the merge? ;-)

solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
```diff
@@ -4,179 +4,375 @@
...
-      case _ => false
+  private def sendBulkToSolr(messages: Seq[IncomingMessage[T, C]]): Unit = {
+
+    @tailrec
+    def send(toSend: Seq[IncomingMessage[T, C]]): UpdateResponse = {
+      val operation = toSend.head.operation
+      //Just take a subset of this operation
+      val current = toSend.takeWhile { m =>
+        m.operation == operation
+      }
```
2m: Instead of calling `takeWhile` and `dropWhile` below, you can call `partition` and get both of the `Seq` back in one go.

solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
```diff
@@ -4,179 +4,375 @@
...
+      val response = operation match {
+        case Upsert => updateBulkToSolr(current)
+        case AtomicUpdate => atomicUpdateBulkToSolr(current)
+        case DeleteByIds => deleteBulkToSolrByIds(current)
+        case DeleteByQuery => deleteEachByQuery(current)
+      }
+      //Now take the remaining
+      val remaining = toSend.dropWhile(m => m.operation == operation)
+      if (remaining.nonEmpty) {
+        send(remaining) //Important: Not really recursive, because the future breaks the recursion
```
2m: I think this comment is outdated, is it not?

solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
```diff
@@ -4,179 +4,375 @@
...
+      if (message.idFieldOpt.isEmpty || message.idFieldValueOpt.isEmpty) {
+        throw new IllegalArgumentException("idfield name and idfield value should be set")
+      }
 
-    if (nextMessages.isEmpty) {
-      state match {
-        case Finished => handleSuccess()
-        case _ => state = Idle
+      doc.addField(message.idFieldOpt.get, message.idFieldValueOpt.get)
+      if (client.isInstanceOf[CloudSolrClient]) {
```
2m: Replace `isInstanceOf` with a pattern match:

```scala
client match {
  case c: CloudSolrClient => ...
  case _ => ...
}```

File: solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala [hunk 14]
```diff
@@ -4,179 +4,375 @@
...
-    }
+  private def updateBulkToSolr(messages: Seq[IncomingMessage[T, C]]): UpdateResponse = {
+    val docs = messages
+      .map(
+        message =>
+          message.sourceOpt.map { source: T =>
+            messageBinder(source)
+        }
+      )
+      .flatten
```
2m: map + flatten = flatMap

Replace to:

```scala
messages.flatMap(_.sourceOpt.map(messageBinder))
```

solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
```diff
@@ -4,179 +4,375 @@
...
-          }
-          handleResponse(messages, status)
-        }
+  private def deleteEachByQuery(messages: Seq[IncomingMessage[T, C]]): UpdateResponse = {
+    val responses = messages.map { message =>
+      val query = message.queryOpt.get
+      if (log.isDebugEnabled) log.debug(s"Delete by the query $query")
+      client.deleteByQuery(collection, query, settings.commitWithin)
     }
+    responses.filter(r => r.getStatus != 0).headOption.getOrElse(responses.head)
```
2m: filter + headOption = find

2m:
Almost there. :) If you could fix the last nitpicks, then this can go in!

solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
giena: Thx

solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
giena: Thx

solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
giena: Thx

solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
giena: Thx

solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
giena: Thx

giena:
Thx @2m for your help. Do you think we could merge? ;-)

PR #1168, opened 2018-08-27
Unix Domain Socket: Become 1.0 ready

ennru:
* Move internal code to impl; API tests to docs

As part of #1093


ennru:
@huntc and @longshorej I tried to press this module into our new recommended mould, I hope it makes sense to you.

PR #1177, opened 2018-08-30
Added passthrough pattern example and documentation

davideicardi:
As per @ennru  request here https://github.com/akka/alpakka-kafka/pull/525#issuecomment-415782601  
I have moved the pass through example inside the pattern documentation.

The general idea is to have a simple way to pass a message though an arbitrary flow maintaining the original message, also if the flow perform a trasformation. This is especially useful when integrating different flows together.

I have also simplified it.

      val source = Source(List(1, 2, 3))

       // Int to Unit
      val printerFlow =
        Flow[Int]
          .map(println)
      // Int to Int
      val plus10Flow =
        Flow[Int]
          .map(_ + 10)

       val ret = source
        .via(PassThroughFlow(printerFlow))
        .via(plus10Flow)
        .runWith(Sink.seq)

I don't know how to explain it in a clear way... but in my opinion it is quite useful.


ennru:
I believe there is a more general use-case, as well. Eg. when you read from Kafka and publish data via a flow that doesn't support pass-through data itself.

Something like this:
```
                    a=>provided flow=>a1
                   /                    \
                  /                      \
(a, b) => unzip --                        -- zip => (a1, b)
                  \                      /
                   \                    /
                    -------------------
```

It is very important to understand that the provided flow must produce exactly one element per consumed stream element.

ennru:
What do you think about my variant of the Pass-Through flow?

davideicardi:
@ennru The first version of the PassThrough flow worked as you just have described, but I noticed that it is not very convinient to work with. I don't always have a Tuple...

The new version is more generic: take any input A, pass it to a provided flow and then returns again A. The provided flow output is discarded because currently I don't have seen usage.
I have used this new flow in many scenarios. Recently for example I have used it in a source like this:

```
MQTT_Source => Passthrough(writeTo_HBASE_Flow) => Passthrough(writeTo_Kafka_Flow) => MQTT_Commit
```
I just want to write the original message, don't need a Tuple in input/output.

On the input side I think that my new version is more generic (doesn't require a Tuple but a Tuple can be used). On the output side I usually don't need the transformed message, I just want to "pass-through" this flow and get back the original message.

But you are the boss :wink:, I let you decide. Eventually both versions can be implemented... we just have to find a good name (the most difficult part  :smiley:)

doc-examples/src/test/scala/akka/stream/alpakka/eip/scaladsl/PassThroughExamples.scala
```diff
@@ -0,0 +1,109 @@
...
+object PassThroughFlow {
+  def apply[A, T](processingFlow: Flow[A, T, NotUsed]): Graph[FlowShape[A, A], NotUsed] =
+    Flow.fromGraph(GraphDSL.create() { implicit builder =>
+      {
+        import GraphDSL.Implicits._
+
+        val tuplerizer = builder.add(Flow[A].map(x => (x, x)))
+        val detuplerizer = builder.add(Flow[(T, A)].map(_._2))
+
+        val unzip = builder.add(Unzip[A, A]())
```
ennru: You can replace `tupelizer` and `Merge` with `Broadcast(2)`.

doc-examples/src/test/scala/akka/stream/alpakka/eip/scaladsl/PassThroughExamples.scala
```diff
@@ -0,0 +1,109 @@
...
+        val zip = builder.add(Zip[T, A]())
+
+        val passThroughFlow = Flow[A].map(identity)
```
ennru: You should remove `passThroughFlow` and connect the `merge.out(1)` directly with the `zip.in1`.
If you want to keep the Ascii-art, you may surround it with `// format:off` and `on`.

davideicardi: Sorry, it is not clear to me how should I use `Merge` and `Broadcast` ... reading the documentation it is not clear how to use it in my use case. Can you provide an example. Why they are preferred instead of `Zip`/`Unzip`?

davideicardi:
I have tried to make the flow more generic using a syntax like:

- `PassThroughFlow(writeFlow, Keep.both)` : to get the Tuple in output
- `PassThroughFlow(writeFlow, Keep.right)` : to just get the original value

Can this be a solution?

I have added `// format: off` around ASCII art, but don't know how exactly to use `Merge` and `Broadcast` as you have suggested, sorry.

davideicardi:
About Java version, I find it difficult to write... I don't see a lot of examples of custom graph written in Java. Maybe because I have used Java a long time ago but I found quite difficult to write the equivalent flow. And unfortunately I don't have a lot of time to investigate further ... I appreciate your help on this.

ennru:
Yes, reusing `Keep` is a great idea - if explained well...

Regarding `Broadcast` (not merge as I wrote), it would look like this
```scala
val broadcast = builder.add(scaladsl.Broadcast[A](2))
val zip = builder.add(Zip[T, A]())
val detuplerizer = builder.add(Flow[(T, A)].map(x => output(x._1, x._2)))

// format: off
broadcast.out(0) ~> processingFlow ~> zip.in0
broadcast.out(1)         ~>           zip.in1
                                      zip.out ~> detuplerizer
// format: on
```

...and it was that Ascii-art that needs the format-switch, not the docs.
Leave the Java-version out, and we'll try to add it.

doc-examples/src/test/scala/akka/stream/alpakka/eip/scaladsl/PassThroughExamples.scala
```diff
@@ -0,0 +1,124 @@
...
+        Merge
+
+        val broadcast = builder.add(Broadcast[A](2))
+        val zip = builder.add(Zip[T, A]())
+        val detuplerizer = builder.add(Flow[(T, A)].map(x => output(x._1, x._2)))
+
+        // format: off
+        broadcast.out(0) ~> processingFlow ~> zip.in0
+        broadcast.out(1)         ~>           zip.in1
+                                              zip.out ~> detuplerizer
```
2m: You can also use `ZipWith` to combine `zip` and `detuplerizer` together.

doc-examples/src/test/scala/akka/stream/alpakka/eip/scaladsl/PassThroughExamples.scala
```diff
@@ -0,0 +1,124 @@
...
+object PassThroughFlow {
+  def apply[A, T](processingFlow: Flow[A, T, NotUsed]): Graph[FlowShape[A, (T, A)], NotUsed] =
+    apply[A, T, (T, A)](processingFlow, Keep.both)
+
+  def apply[A, T, O](processingFlow: Flow[A, T, NotUsed], output: (T, A) => O): Graph[FlowShape[A, O], NotUsed] =
+    Flow.fromGraph(GraphDSL.create() { implicit builder =>
+      {
+        import GraphDSL.Implicits._
+
+        Merge
```
2m: Leftover?

ennru:
Thank you!

PR #1181, opened 2018-09-01
Jms per message destination

andreas-schroeder:
Extends the Jms message model with messages having destinations, and implement a flow (`directedMessageFlow`) that consumes messages with destinations only.

### DIFF
--- a/jms/src/main/scala/akka/stream/alpakka/jms/scaladsl/JmsProducer.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/scaladsl/JmsProducer.scala
@@ -15,8 +15,18 @@ object JmsProducer {
   /**
    * Scala API: Creates an [[JmsProducer]] for [[JmsMessage]]s
    */
-  def flow[T <: JmsMessage](settings: JmsProducerSettings): Flow[T, T, NotUsed] =
+  def flow[T <: JmsMessage](settings: JmsProducerSettings): Flow[T, T, NotUsed] = {
+    require(settings.destination.isDefined, "Producer destination must be defined in regular producer flow")
     Flow.fromGraph(new JmsProducerStage(settings))
+  }
+
+  /**
+   * Scala API: Creates an [[JmsProducer]] for [[JmsDirectedMessage]]s
+   */
+  def directedMessageFlow[T <: JmsDirectedMessage](settings: JmsProducerSettings): Flow[T, T, NotUsed] = {
+    require(settings.destination.isEmpty, "Producer destination must be undefined in directed producer flow")
+    Flow.fromGraph(new JmsProducerStage(settings))
+  }
 
   /**
    * Scala API: Creates an [[JmsProducer]] for [[JmsMessage]]s
--- a/docs/src/main/paradox/jms.md
+++ b/docs/src/main/paradox/jms.md
@@ -192,6 +192,26 @@ Scala
 Java
 : @@snip [snip](/jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsConnectorsTest.java) { #run-flow-producer }
 
+### Sending messages with per-message destinations
+
+The producer supports also a message flow that allows to set per-message destinations. In the flow settings that
+are passed in, it is not allowed to provide a queue or topic:
+
+Scala
+: @@snip [snip](/jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsConnectorsSpec.scala) { #create-directed-flow-producer }
+
+Java
+: @@snip [snip](/jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsConnectorsTest.java) { #create-directed-flow-producer }
+
+The destination can then be defined per message:
+
+Scala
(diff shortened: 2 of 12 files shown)


andreas-schroeder:
There are multiple ways of implementing the interface, however, and I'm not 100% sure the solution I propose here is the best one. 

There are two ways per-message destinations can be offered at the API level:
- force the user to use separate message types (directed messages) and use those in the flow (this is this proposal).
- force the user to provide a default destination and extend regular messages with an optional destination (that overrides the default if set). This would mean that a default destination _must_ always be provided.

While I provide the first here, the latter now seems to be more lightweight, so I'm slightly in favor of reducing the message model options back to what it initially was.

andreas-schroeder:
Ok, I will boil this proposal down to a simpler model.

jms/src/main/scala/akka/stream/alpakka/jms/scaladsl/JmsProducer.scala
```diff
@@ -15,8 +15,10 @@ object JmsProducer {
   /**
    * Scala API: Creates an [[JmsProducer]] for [[JmsMessage]]s
    */
-  def flow[T <: JmsMessage](settings: JmsProducerSettings): Flow[T, T, NotUsed] =
+  def flow[T <: JmsMessage](settings: JmsProducerSettings): Flow[T, T, NotUsed] = {
+    require(settings.destination.isDefined, "Producer destination must be defined in regular producer flow")
```
ennru: Great to check early on.

jms/src/main/scala/akka/stream/alpakka/jms/impl/SoftReferenceCache.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  def lookup(key: K, default: => V): V = {
+    val lfence = memoryFenceGenerator // force a lfence (read memory barrier)
+    cache.get(key) match {
+      case Some(ref) =>
+        ref.get match {
+          case Some(value) => value
+          case None =>
+            purgeCache() // facing a garbage collected soft reference, purge other entries.
+            update(key, default, lfence)
```
jrudolph: Wouldn't it still be possible to access the `mutable.HashMap` concurrently? This is not safe.

The most simple solution would be to use regular synchronization if there is no need for utmost performance. Otherwise, the next best solution would be to use an immutable `HashMap inside of an `AtomicReference` and the CAS pattern at least for writing to the cache.

jms/src/main/scala/akka/stream/alpakka/jms/scaladsl/JmsProducer.scala
andreas-schroeder: I was also thinking of introducing an intermediate `JmsIncompleteProducerSettings` type that gets the destination methods `toQueue`, `toTopic` etc, and that those methods yield a `JmsProducerSettings` instance - similar to the `JmsMessage` and `JmsDirectedMessage` types that were previously in this PR. Would break compatibility, but maybe something worthwhile for the 1.0 release?

jms/src/main/scala/akka/stream/alpakka/jms/impl/SoftReferenceCache.scala
andreas-schroeder: sure, let's keep things simple 👍when re-thinking what's the whole flow, I think it might be possible to actually piggy-back synchronization on the futureCB https://github.com/akka/alpakka/blob/b107b80622c274627c3d74846527a6cfbf235197/jms/src/main/scala/akka/stream/alpakka/jms/JmsProducerStage.scala#L97 , and then all this delicate volatile-based memory fences would be a red herring anyway.

ennru: If you think you can simplify it with that solution, please go ahead in this PR.

andreas-schroeder: @ennru I decided to go for a simple lock here: https://github.com/akka/alpakka/blob/bf30b5d863e9de0f9215732935e60a290b7e8edd/jms/src/main/scala/akka/stream/alpakka/jms/impl/SoftReferenceCache.scala#L17 since there will be no concurrent access to the cache (only sequential by different threads), the cost of `sychronized` should be low. I'm not 100% sure about the synchronization guarantees of async callbacks, is there some docs?

ennru: Yes, I saw that before I LGTMed, but didn't see your comment before today...

The async callbacks are guaranteed to be allowed to access everything in the stage logic. See [Using asynchronous side-channels](https://doc.akka.io/docs/akka/current/stream/stream-customize.html#using-asynchronous-side-channels)

andreas-schroeder: getting that one test right was an interesting challenge, so here you go :) https://github.com/akka/alpakka/pull/1181/commits/050b3af33567444a9accba1df3ef75ca75f279ae

Issue #1183, opened 2018-09-03
SSE retry loop when event lines breach 4096 chars


Synesso:
Edit: 

There are two parts to this.

1. The client silently fails to parse lines greater than the limit imposed by akka-http. It should have some sort of error behavior. (This may be an akka-http bug). The problem probably also occurs for breaches of [the 8k event size limit](https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/sse/EventStreamUnmarshalling.scala#L39-L43)
2. The Alpakka SSE module does not provide a mechanism for changing this limit, whereas the underlying akka-http API does. #1185 fixes this.

----

https://github.com/Synesso/alpakka-sse-possible-bug is an MVCE for this bug. 

The logic:

```scala
    implicit val system = ActorSystem("sse-test")
    implicit val materializer = ActorMaterializer()
    implicit val executionContext = system.dispatcher

    def requestWithDebugging(r: HttpRequest): Future[HttpResponse] = {
      println("\n=== Reconnection ===")
      println("Request headers:")
      val request = r
      request.headers.foreach(h => println(s"* $h"))
      println()
      val f = Http().singleRequest(r)
      f.onComplete {
        case Success(r2) =>
          println("Response headers: ")
          r2.headers.foreach(h => println(s"* $h"))
        case _ =>
      }
      f
    }

    val work: Future[Done] =
      EventSource(Uri("https://horizon.stellar.org/transactions"), requestWithDebugging, Some("85025096736968704"), 1.second)
        .mapConcat{ case ServerSentEvent(data, eventType, id, retry) =>
          print(s"eventType: $eventType, id:$id, retry:$retry, data:$data")
          (if (eventType.contains("open")) None else Some(data)).to[collection.immutable.Iterable]
        }.toMat(Sink.foreach(println))(Keep.right).run()

    work.onComplete(_ => system.terminate())
    Await.ready(system.whenTerminated, Duration.Inf)
```

There are several subdomains and SSE endpoints under stellar.org. Data is parsed by the SSE adaptor until eventually the component gets into a loop where no more events are found. It seems that the http request headers are correct, but no events are parsed.

Streams that start from the most recent event will run for anywhere up to 10 or 100 events before they meet an event that begins the no-more-events loop. That is to say, 1 event in ~10 to 100 will trigger this bug.

[...]

PR #1185, opened 2018-09-03
SSE: Allow provision of custom EventStreamUnmarshalling in EventSource

Synesso:
#1183 describes how events that exceed line length of 4096 bytes silently fail, leading to a retry loop.

This PR allows the user to provide their own `EventStreamUnmarshalling` instance, thereby overriding the default values for `maxEventSize` and `maxLineSize`.

It does not fix the silent failure, but allows users to avoid it.


Synesso:
@hseeberger That sounds perfect. Both `akka.http.sse.max-event-size` and `akka.http.sse.max-line-size`. Shelving this PR.

ktoso:
I agree that it should be fixed in Akka HTTP and will review over there, thanks for the help @Synesso

PR #1186, opened 2018-09-04
MQTT: towards 1.0

### LINKED ISSUE #1084: MQTT: become Alpakka 1.0 ready
- [ ] Move internal code to `impl` package
- [ ] Mark internal code `@InternalApi` and `private[connector]`
- [ ] Move documentation example code to packages `docs.javadsl` and `docs.scaladsl`

ennru:
I tried to tackle the MQTT connector step by step in separate commits.

API changes include
1. `MqttConnectionSettings` is now a regular class, Scala users must use the `withXyz` methods instead of `copy`
1. The materialized values for Sinks were inconsistent. It is now always stream completion (some used to be MQTT subscription established).
1. `MqttSourceSettings` is deprecated now (it was used for both Sources and Flows). Instead both `MqttConnectionSettings` and `MqttSubscriptions` should be passed.
1. `MqttCommittableMessage` became `MqttMessageWithAck` and uses `ack` instead of `messageArrivedComplete`.

Fixes #1084


mqtt/src/main/scala/akka/stream/alpakka/mqtt/impl/MqttFlowStage.scala
```diff
@@ -0,0 +1,309 @@
...
+    }
+    (logic, subscriptionPromise.future)
+  }
+}
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+object MqttFlowStage {
```
2m: `private[mqtt]`?

mqtt/src/main/scala/akka/stream/alpakka/mqtt/javadsl/MqttSource.scala
```diff
@@ -7,31 +7,72 @@ package akka.stream.alpakka.mqtt.javadsl
...
   /**
-   * Java API: create an [[MqttSource]] with a provided bufferSize.
+   * Create a source subscribing to MQTT messages (without a commit handle).
+   *
+   * The materialized value completes on successful connection to the MQTT broker.
+   *
+   * @param bufferSize max number of messages read from MQTT before back-pressure applies
+   * @deprecated use atMostOnce with MqttConnectionSettings and MqttSubscriptions instead
+   */
+  @deprecated("use atMostOnce with MqttConnectionSettings and MqttSubscriptions instead", "0.21")
```
2m: Just a reminder that we will need to grep those version values and replace to the first milestone version before we release the milestone.

mqtt/src/main/scala/akka/stream/alpakka/mqtt/model.scala
```diff
@@ -0,0 +1,77 @@
...
+
+import scala.compat.java8.OptionConverters._
+
+final class MqttMessage private (val topic: String,
+                                 val payload: akka.util.ByteString,
+                                 val qos: Option[MqttQoS],
+                                 val retained: Boolean) {
+
+  /** Java API */
+  def getTopic: String = topic
```
2m: Is this for backwards compatibility? Usually we would not add getters for attributes that have java-friendly types.

ennru: I was a bit indifferent there, I believe we don't need Java getters at all for messages filled by the user. But for messages created by the connector we might support Java getters for all fields, I guess.

Issue #1188, opened 2018-09-05
JMS: Should have commit/rollback timeout for transactional consumer


akara:
The stream gets stuck if message processing fails, whether this is due to application code, recovery by the stream decider, etc. A failure to commit/rollback must not get the stream stuck. The message failing to signal back should be rolled back automatically after timeout.

ennru:
Looking at https://github.com/akka/alpakka/pull/1831 shows this has been available for quite a while now.
Closing.

PR #1189, opened 2018-09-05
JMS: Introducing commit timeouts and use Future instead of BlockingQueue

### LINKED ISSUE #1187: JMS: Transactional consumer to be lenient to underlying threading anomalies
The JMS specification requires the session to be bound to threads and the current implementation using the commit and rollback queue relies on this requirement to work properly. We have seen anomalies causing messages to get out of sequence on underlying reconnects. With the current implementation this can cause the stream to collapse.

akara:
for signaling commits and rollbacks (fixes #1187 and #1188).

Using `Future` also cuts down the complex logic for tombstones that prevented wrong commits at stream shutdown and makes a clear association between the envelope and the waiting thread. The downside is the cost of creating a `Future` for every envelope.


akara:
Note this fix is only pertaining to the transactional consumer.

akara:
@ennru Happy to do so, just not clear what you mean by using timers in the stage. Each envelope needs to be directly associated with the JMS `onMessage` invocation processing it. As long as we can achieve it with better ways than `Future`, I'm totally happy to make this change.  Just need to be clear how you'd want to achieve this with a timer. Thanks!

ennru:
I haven't thought it through, but AFAICS you'd need to call `rollback` after the given timeout iff no commit/rollback happened. So I think you can schedule a timer in the stage with a handle to the emitted envelope and call the rollback if needed.

akara:
The thread that times out is a blocking thread, sadly. The commit/rollback needs to happen on that thread. The `Future` serves a few purposes: 1) the association between the envelope and the `onMessage` (and the thread running it), 2) inter-thread communication between the thread running the stream calling `commit`/`rollback` and the `onMessage` thread, 3) timing out and rollback if `commit`/`rollback` is not being called - i.e. a message being handled by the decider does not starve the stream. I can use `akka.pattern.after` with a `Scheduler` to achieve (3), but I'd need some mechanism for inter-thread communication. For instance, I can use a `Semaphore` to signal between threads. A per-envelope `Semaphore` serves (1) and partially (2). I also need some alternate container to hold the `commit`/`rollback` command in combination with the synchronization structure to serve the rest of (2).

The synchronization structure also serves as the blocking point for the `onMessage` session thread (similar to how I use `Await` in this code). I'm just not sure all this is worth the additional complexity that can be easily solved by using a `Future`.

I can also replace the `scala.concurrent.Future` with a `java.util.concurrent.CompletableFuture` making use of the same semantics without using the `Await` call. Logically there is no difference.

ennru:
Ok, it makes sense to continue with the Futures-approach.

akara:
This PR is also in good shape for review.

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
```diff
@@ -71,6 +76,7 @@ final case class JmsConsumerSettings(connectionFactory: ConnectionFactory,
                                      bufferSize: Int = 100,
                                      selector: Option[String] = None,
                                      acknowledgeMode: Option[AcknowledgeMode] = None,
+                                     ackTimeout: Duration = Duration.Inf,
```
ennru: Wouldn't it be better to have a finite default? The uncommitted futures might pile up.

jms/src/main/scala/akka/stream/alpakka/jms/JmsConsumerStage.scala
```diff
@@ -162,32 +162,29 @@ final class JmsTxSourceStage(settings: JmsConsumerSettings)
...
-                        action(envelope)
-                      } catch {
-                        case _: StopMessageListenerException => listenerStopped = true // Tombstone.
-                        case e: IllegalArgumentException => handleError.invoke(e) // Invalid envelope. Fail the stage.
-                        case e: JMSException => handleError.invoke(e)
-                      }
+                    try {
+                      val envelope = TxEnvelope(message, session)
+                      handleMessage.invoke(envelope)
+                      val action = Await.result(envelope.commitFuture, settings.ackTimeout)
```
ennru: Whose thread is blocked here? How many outstanding commits can it take?

ennru:
@andreas-schroeder Another PR where I'd appreciate a review by you...

jms/src/main/scala/akka/stream/alpakka/jms/JmsConsumerStage.scala
akara: This is the JMS session thread (which is blocking anyway... previously just on the blocking queue). The number of outstanding commits is the number of sessions we open.

In our production we do see anomalies and deviations from the JMS spec in edge cases when using the previous implementation of blocking queue. Switching to futures makes it more lenient to such anomalies and has 0 incidents over the past month. We could use either this or the Java `CompletableFuture`. Both cases we do block in similar fashion, though.

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
akara: You're right. I tried to retain the previous behavior. The question is what timeout makes sense? We're using a 1 second timeout for our use cases and it seems to be stable. Let me know if there is a rationale for a different value.

akara:
@ennru Looks like the compile failures have nothing to do with this PR. Please let me know if I read it wrong.

akara:
Thanks! The status here still shows failing in this screen. But when you look into the test itself it shows passing. Looks like a Travis issue.

andreas-schroeder:
@ennru I'll grab this one for review 👍

akara:
Thanks for your feedback and suggestions. Really appreciate you taking the time.

Just to clarify, the change proposed is not changing much from the current implementation, that is also blocking on the consumer thread. The current implementation is only blocking on the `BlockingQueue` instead of a `Future`. And the fix is also a narrow scope fix to be more lenient towards JMS implementations that do not treat the session per thread rule as strictly as it should, causing confusion with the queueing implementation we had. Given, the chances are very low but at the volume we're dealing with we see such issues every now and then, causing stream collapses on production systems. Enough collapses over a period of time (weeks) would reduce the active streams to the point we'll have a large backlog in the message broker. This proposal is merely just to bring a mechanism we have successfully employed in production back to open source. While I do not want to preclude the suggested redesign, I have to say such work would have a much broader scope than this PR alone.

Let's think about the redesign scenario. The case where we consume and publish in the same transaction is definitely an interesting case to solve. Given the threading architecture of JMS, it would not be an easy task to achieve on Akka Streams. But if there is a need, there is a way. I just don't want to jump the gun and try to design for such scenarios without seeing a concrete use case.

Next let's consider the `receiveNoWait` use case, we are still bound to the commit/rollback command executing on the same thread. This thread is unlikely going to be used for stream processing. And in the single message/single commit scenario (the current scenario we have been supporting) this thread will also be doing one single message at a time, just polling instead of a JMS callback.

If we are considering a scenario where multiple messages can be committed/rolled back on the same session at once, we will have to work off a single session (no session concurrency) or expose the session awareness to the stream, allowing the stream to selectively commit/rollback a session. This can be achieved in ways similar to how akka-stream-kafka handles multiple partitions, by allowing batching up the commit offsets to one larger commit, or really multiple commits, one per partition.

BUT, JMS is semantically very different from Kafka. Kafka does only an offset commit. There is no way known to me to selectively commit/rollback messages so a rollback is not an option. The design pattern around rolling back is to re-publish the messages to a separate retry topic. That can be done through JMS if we do not support rollbacks.

To support JMS in the current semantics, with session parallelism, I'm still a bit stuck at how we can make this happen without increasing the complexity of the user API. But all in all I do agree the Tx source address one particular use case where the user can selectively commit/rollback(retry) particular messages. It just happens to be the production use case we have.

I know I left more questions than answers. I'm happy to get on a call to hash out a design that would work for single message commit/rollback as well as multiple message commit/rollback as you suggested. Not having to block a thread would be a big plus. But polling may also have its own consequences we have to be careful about.

Since all current sources (plain, ack, tx) are essentially blocking on the JMS session thread one way or the other, such a redesign will need to account for all three use cases, too. Thanks!

andreas-schroeder:
hi  @akara, thanks for your detailed answer. Actually, I was confusing the current (0.20) `JmsTxSourceStage` behavior with the `JmsAckSourceStage` behavior (the latter annoyed me actually, while I hadn't worked with the former). You are right: with your changes, the blocking behavior is the same as before, so that there is no degradation here. Thanks for being patient and re-iterating that point ;)

I agree with you that multi-message transaction support (be it only consume or consume/produce) would introduce a much more complex API, and that it shouldn't be done without a solid use case. So a re-design with `receiveNoWait` is out of scope for this PR.

To conclude (cc @ennru), this PR is fine both from the code and from the design, I've got no more open questions.

ennru:
Thank you @akara, for your passion to make Alpakka better for everybody!

akara:
Thanks @andreas-schroeder @ennru ! If we want to get on a journey defining the next step of JMS behavior, please count me in. Feel free to reach out to me anytime. IMO, we're trying to bridge two quite incompatible technologies (JMS is definitely not alone). I hold some hope on correcting these in the work happening in the Eclipse community. But till such technology becomes mature, we'll have to deal with enterprise integration through incompatible paradigms like this one.

Issue #772, opened 2018-02-12
OrientDb: type erasure warnings in OrientDBFlowStage


ennru:
Just noticed we have some serious type erasure warnings in the new OrientDBFlowStage class:
```
./alpakka/orientdb/src/main/scala/akka/stream/alpakka/orientdb/OrientDBFlowStage.scala:173:49: abstract type pattern T is unchecked since it is eliminated by erasure
               case OIncomingMessage(typeRecord: T, passThrough: C) =>
```
The other warnings could be circumvented by not naming type `C` explicitly.

PR #1191, opened 2018-09-06
OrientDb: towards 1.0

ennru:
This PR just moves the code to the designated packages and simplifies the code a tiny bit. There are many more things to clean up. (eg. #772)

Fixes parts of #1085

### DIFF
--- a/orientdb/src/main/scala/akka/stream/alpakka/orientdb/model.scala
+++ b/orientdb/src/main/scala/akka/stream/alpakka/orientdb/model.scala
@@ -24,5 +24,4 @@ final case class OIncomingMessage[T, C](oDocument: T, passThrough: C)
 
 final case class OOutgoingMessage[T](oDocument: T)
 
-case class OSQLResponse[T](error: Option[String], result: Option[OSQLResult[T]])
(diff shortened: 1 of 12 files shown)


orientdb/src/main/scala/akka/stream/alpakka/orientdb/model.scala
```diff
@@ -0,0 +1,27 @@
...
+  // Java-api - with passThrough
+  def create[T, C](oDocument: T, passThrough: C) =
+    OIncomingMessage(oDocument, passThrough)
+}
+
+final case class OIncomingMessage[T, C](oDocument: T, passThrough: C)
+
+final case class OOutgoingMessage[T](oDocument: T)
+
+case class OSQLResponse[T](error: Option[String], result: Seq[OOutgoingMessage[T]])
```
2m: I see that it will be tricky to un-case class those as they are used in pattern matches in the implementation. Do we have manually written unapply methods in any of the connectors already?

ennru: Not that I'm aware of. We need to discuss what is feasible.

Issue #1193, opened 2018-09-06
JMS: Retry create connection on startup and connection failure


akara:
When a connection errors out, disconnects, or broker is down, the consumer should have the option to poll for broker to be up and continue normally.

ennru:
Fixed with #1194

PR #1194, opened 2018-09-06
JMS: Adding connection timeouts and retries (fixes #1192 and #1193)

### LINKED ISSUE #1192: JMS: Connection.start() occasionally hangs depending on broker state
We see some occurrence of issues that look like https://issues.apache.org/jira/browse/AMQ-1375 where `Connection.start()` hang forever, causing the stream to starve. While this is obviously an ActiveMQ or even broker issue, we should give a time budget how long it takes to make a connection and take appropriate actions if the connection cannot be made.

akara:
All good for review now.

### DIFF
--- a/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
@@ -85,19 +85,7 @@ final case class ConnectionRetrySettings(connectTimeout: FiniteDuration = 10.sec
   /** Hypothetical retry time, not accounting for maxBackoff. */
   def waitTime(retryNumber: Int): FiniteDuration =
     (initialRetry * Math.pow(retryNumber, backoffFactor)).asInstanceOf[FiniteDuration]
-
-  /** Max possible time it will take to timeout, including connection timeouts and the wait time between retries.  */
-  def maxWaitTime: Duration =
-    if (maxRetries < 0) Duration.Inf
-    else {
-      val totalWaitTime = (0 to maxRetries).map(connectTimeout + waitTime(_)).reduce(_ + _)
-      val totalTimeout = connectTimeout * (maxRetries + 1)
-      totalTimeout + totalWaitTime
-    }
-
 }
-
-case class ConnectionRetryException(message: String, cause: Throwable) extends Exception(message, cause)
 
 object JmsConsumerSettings {
 
--- a/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
@@ -141,7 +141,7 @@ final case class JmsProducerSettings(connectionFactory: ConnectionFactory,
                                      destination: Option[Destination] = None,
                                      credentials: Option[Credentials] = None,
                                      sessionCount: Int = 1,
-                                     timeToLive: Option[java.time.Duration] = None,
+                                     timeToLive: Option[Duration] = None,
                                      acknowledgeMode: Option[AcknowledgeMode] = None)
     extends JmsSettings {
   def withCredential(credentials: Credentials): JmsProducerSettings = copy(credentials = Some(credentials))
@@ -151,7 +151,10 @@ final case class JmsProducerSettings(connectionFactory: ConnectionFactory,
   def withQueue(name: String): JmsProducerSettings = copy(destination = Some(Queue(name)))
   def withTopic(name: String): JmsProducerSettings = copy(destination = Some(Topic(name)))
   def withDestination(destination: Destination): JmsProducerSettings = copy(destination = Some(destination))
-  def withTimeToLive(ttl: java.time.Duration): JmsProducerSettings = copy(timeToLive = Some(ttl))
+  def withTimeToLive(ttl: java.time.Duration): JmsProducerSettings =
+    copy(timeToLive = Some(Duration.fromNanos(ttl.toNanos)))
+  def withTimeToLive(ttl: Duration): JmsProducerSettings = copy(timeToLive = Some(ttl))
+  def withTimeToLive(ttl: Long, unit: TimeUnit): JmsProducerSettings = copy(timeToLive = Some(Duration(ttl, unit)))
   def withAcknowledgeMode(acknowledgeMode: AcknowledgeMode): JmsProducerSettings =
     copy(acknowledgeMode = Option(acknowledgeMode))
 }
--- a/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
@@ -4,10 +4,12 @@
 
 package akka.stream.alpakka.jms
 
-import java.time.Duration
 import java.util.concurrent.atomic.AtomicBoolean
+
 import javax.jms
 import javax.jms.{ConnectionFactory, Message}
+
+import scala.concurrent.duration._
 
 case class AckEnvelope private[jms] (message: Message, private val jmsSession: JmsAckSession) {
 
@@ -27,6 +29,7 @@ case class TxEnvelope private[jms] (message: Message, private val jmsSession: Jm
 
 sealed trait JmsSettings {
   def connectionFactory: ConnectionFactory
+  def connectionRetrySettings: ConnectionRetrySettings
   def destination: Option[Destination]
   def credentials: Option[Credentials]
   def acknowledgeMode: Option[AcknowledgeMode]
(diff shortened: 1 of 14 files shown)


jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
```diff
@@ -20,26 +23,18 @@ private[jms] trait JmsConnector { this: GraphStageLogic =>
...
-  private def onConnection = getAsyncCallback[Connection] { c =>
+  private def onConnection = getAsyncCallback[jms.Connection] { c =>
     jmsConnection = Some(c)
   }
 
-  private def onSession =
-    getAsyncCallback[JmsSession] { session =>
-      jmsSessions :+= session
-      onSessionOpened(session)
-    }
```
akara: Moved to `SourceStageLogic` in `JmsConsumerStage.scala` for this function only applies to consumers

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
```diff
@@ -49,49 +44,111 @@ private[jms] trait JmsConnector { this: GraphStageLogic =>
       case d => d
     }
 
-    materializer match {
-      case m: ActorMaterializer => m.system.dispatchers.lookup(dispatcher.dispatcher)
-      case x => throw new IllegalArgumentException(s"Stage only works with the ActorMaterializer, was: $x")
-    }
+    ActorMaterializerHelper.downcast(materializer).system.dispatchers.lookup(dispatcher.dispatcher)
```
akara: Replaced with Akka library call providing the same functionality.

akara:
Hmm, the build failure does not give me any indication what I could have messed up.

ennru:
No worries, it's not related to your changes.
Please be aware of https://github.com/akka/alpakka/pull/1181 which will reach master before this.

ennru:
fixes #1192 and #1193

ennru:
fixes #1192 and #1193

akara:
Thanks for the heads up! No worries, I'll rebase when time comes.

akara:
Added tests with mock connection failure and start issues. Note this change only affects consumers for now. A bit more work required to make it work with producers.

ennru:
The other JMS PR is now merged, you've compilation errors with Scala 2.11.

akara:
We have been focusing on the consumer (`Source`) side more than the publishing side as a source fail will indeed lead to stream collapse, something we cannot have in a production environment. No decider can help with that. So the reconnect feature currently only applies to the consumer, while the connect timeout and retries apply to both.

Making the publisher safely deal with reconnects (without message losses) is not quite trivial, especially with the sessions pooling and publishes in-flight. I am contemplating whether publisher reconnects should be a separate task/PR altogether. Please voice your opinion.

ennru:
Sounds good. Please, focus on the consumer side and leave the producer changes out.

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsConnectorsSpec.scala
```diff
@@ -552,12 +558,19 @@ class JmsConnectorsSpec extends JmsSpec with MockitoSugar {
         JmsProducerSettings(connectionFactory).withQueue("numbers")
       )
 
-      val completionFuture: Future[Done] = Source
-        .failed[JmsTextMessage](new RuntimeException("Simulated error"))
```
akara: This test was unreliable. Since setting the `jmsConnection` in `JmsConnector` is done through an async callback, the early arrival of the failure will cause the `jmsConnection` (which is not referenced) to not be closed. Need to cause the failure mid-stream instead of at the beginning for this test to be reliable.

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
```diff
@@ -48,36 +53,115 @@ private[jms] trait JmsConnector[S <: JmsSession] { this: GraphStageLogic =>
...
-    val sessions: Seq[Future[S]] = openSessions()
-    val allSessions = Future.sequence(sessions)
+  protected def createSession(connection: jms.Connection, createDestination: jms.Session => jms.Destination): JmsSession
+
+  sealed trait ConnectionStatus
+  case object Connecting extends ConnectionStatus
+  case object Connected extends ConnectionStatus
+  case object TimedOut extends ConnectionStatus
+
+  protected def initSessionAsync(withReconnect: Boolean = true): Unit = {
```
akara: `withReconnect` is set to `false` for the publisher at this point, until we can be certain reconnects behaving correctly with the publisher. Then even the `if` block below can be taken out.

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
```diff
@@ -30,10 +35,6 @@ private[jms] trait JmsConnector[S <: JmsSession] { this: GraphStageLogic =>
 
   protected val fail: AsyncCallback[Throwable] = getAsyncCallback[Throwable](e => failStage(e))
 
-  private val onConnection: AsyncCallback[jms.Connection] = getAsyncCallback[jms.Connection] { c =>
-    jmsConnection = Some(c)
-  }
```
akara: Removed the connection setting using `AsyncCallback` and set directly instead. Some tests rely on the connection to be immediately available after establishing and fail if not. We minimize the risk of such failure. Also connection is only accessed at startup and shutdown. Making it `@volatile` does not incur runtime overhead.

akara: Hmm, apparently this is still not enough. That test is still failing despite tightening up the timeline.

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
```diff
@@ -48,36 +49,115 @@ private[jms] trait JmsConnector[S <: JmsSession] { this: GraphStageLogic =>
...
+    }
+
+    val timeoutFuture = after(jmsSettings.connectionRetrySettings.connectTimeout, system.scheduler) {
+      // Even if the timer goes off, the connection may already be good. We use the
+      // status field and an atomic compareAndSet to see whether we should indeed time out, or just return
+      // the connection. In this case it does not matter which future returns. Both will have the right answer.
+      if (status.compareAndSet(Connecting, TimedOut)) {
+        connectionRef.get.foreach(_.close())
+        connectionRef.set(None)
+        Future.failed(new TimeoutException("Timed out trying to establish connection"))
```
ennru: You could possibly add the applicable values in the error messages.

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
akara: OK. Need to guarantee the connection is closed in an async system. To do so, we need to hold the connection in a future to ensure the connection will be closed even if it is resolved later than the `postStop` is called. This should settle it. But this test has creeped the scope of change.

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
```diff
@@ -58,13 +61,52 @@ object AcknowledgeMode {
...
+
+  /** Hypothetical retry time, not accounting for maxBackoff. */
+  def waitTime(retryNumber: Int): FiniteDuration =
+    (initialRetry * Math.pow(retryNumber, backoffFactor)).asInstanceOf[FiniteDuration]
+
+  /** Max possible time it will take to timeout, including connection timeouts and the wait time between retries.  */
+  def maxWaitTime: Duration =
+    if (maxRetries < 0) Duration.Inf
+    else {
+      val totalWaitTime = (0 to maxRetries).map(connectTimeout + waitTime(_)).reduce(_ + _)
```
andreas-schroeder: super-cool idea to have that maxWaitTime on the settings to see how long it will wait in total! 
How about 
`val totalWaitTime = (0 to maxRetries).map(connectTimeout + waitTime(_).min(maxBackoff)).reduce(_ + _)`
to account for `maxBackoff`?

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
```diff
@@ -58,13 +61,52 @@ object AcknowledgeMode {
...
+      val totalTimeout = connectTimeout * (maxRetries + 1)
+      totalTimeout + totalWaitTime
+    }
+
+}
+
+case class ConnectionRetryException(message: String, cause: Throwable) extends Exception(message, cause)
```
andreas-schroeder: how about moving that exception to `JmsExceptions.scala`? I think either all exceptions should live there, or all live in `Jms.scala`

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
andreas-schroeder: 👍

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
```diff
@@ -48,36 +50,121 @@ private[jms] trait JmsConnector[S <: JmsSession] { this: GraphStageLogic =>
...
+      case NonFatal(t) =>
+        val retrySettings = jmsSettings.connectionRetrySettings
+        import retrySettings._
+        val nextN = n + 1
+        if (maxRetries >= 0 && nextN > maxRetries) { // Negative maxRetries treated as infinite.
+          if (maxRetries == 0) Future.failed(t)
+          else Future.failed(ConnectionRetryException(s"Could not establish connection after $n retries.", t))
+        } else {
+          val delay = if (maxed) maxBackoff else waitTime(nextN)
+          if (delay >= maxBackoff) {
```
andreas-schroeder: why not just
```scala
          val delay = waitTime(nextN).min(maxBackoff)
          after(delay, system.scheduler) {
            openConnectionWithRetry(startConnection, nextN)
          }
```
? :)

jms/src/main/scala/akka/stream/alpakka/jms/JmsConsumerStage.scala
```diff
@@ -208,7 +208,7 @@ final class JmsTxSourceStage(settings: JmsConsumerSettings)
 abstract class SourceStageLogic[T](shape: SourceShape[T],
                                    out: Outlet[T],
                                    settings: JmsConsumerSettings,
-                                   attributes: Attributes)
+                                   inheritedAttributes: Attributes)
```
andreas-schroeder: great rename, that one was bothering me as well :+1:

jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsConnectorsTest.java
```diff
@@ -732,6 +741,43 @@ public void directedProducerFlow() throws Exception {
...
+
+          Try<List<Message>> tryResult = tryFuture.toCompletableFuture().get();
+          long endTime = System.currentTimeMillis();
+
+          assertTrue("Total retry is too short", endTime - startTime > 100L + 400L + 900L + 1600L);
+          assertTrue("Result must be a failure", tryResult.isFailure());
+          Throwable exception = tryResult.failed().get();
+          assertTrue(
+              "Did not fail with a ConnectionRetryException",
+              ConnectionRetryException.class.isAssignableFrom(exception.getClass()));
```
andreas-schroeder: ain't that just `exception instanceof ConnectionRetryException`?

jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsConnectorsTest.java
```diff
@@ -732,6 +741,43 @@ public void directedProducerFlow() throws Exception {
...
+          assertTrue(
+              "Cause of failure is not a JMSException",
+              JMSException.class.isAssignableFrom(exception.getCause().getClass()));
```
andreas-schroeder: `exception.getCause() instanceof JMSException`, or am I wrong?

jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsConnectorsTest.java
```diff
@@ -732,6 +741,43 @@ public void directedProducerFlow() throws Exception {
...
+        });
+  }
+
```
andreas-schroeder: :+1: for that test. Great idea to check for the total time passed.

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
akara: Since `waitTime` is an exponential function, there is a very small chance a very large backoff may overflow, throwing this logic into haywire. That's why I really want to stop calculating the `waitTime` after max is reached.

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
akara: Sure. Will do.

jms/src/test/java/akka/stream/alpakka/jms/javadsl/JmsConnectorsTest.java
akara: No, you're absolutely right. Was looking for a better way to express as well. Tells me my Java is getting a little rusty :-) Let me make this change, and the next one.

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
akara: After thinking through this, we do not need the `maxWaitTime` any longer. It was used for establishing the blocking time waiting the connection in `JmsProducerStage`. After rebasing from your change where the connection is handled asynchronously there, there is no longer a need for this method.  Let me remove it.

akara:
Thank you very much for the comments @andreas-schroeder! Let me get the changes in by this weekend, before @ennru merges. Yes, we have to look at how to invalidate the in-flight publishes in case of a reconnect. Your expertise is greatly appreciated.

ennru:
Thank you for investing your efforts into the JMS connector!

Issue #36, opened 2016-11-11
CouchBase connector


cheleb:
Hi
  I can provide Couchbase (sink) connector.

  Any interest ?

patriknw:
that is useful, please go ahead

chbatey:
Worth checking out http://reactivecouchbase.org/ first

ennru:
Thanks for the pointer, we should link to that for now.

PR #1198, opened 2018-09-08
Scala DSL for Couchbase Source, Flow and Sink based on official Java SDK.

dannylesnik:
This is Sink Source and Flow for Couchbase based on official SDK.
I can provide Java implementation for Sink and Source, However we need to think how to make this work for Flow. 

I know there is ReactiveCouchbase side project, However this is the alternative based on Java SDK. I'll enrich this API in next few days.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,9 @@ jobs:
       - CMD=+cassandra/testChanged
     - env: CMD=+csv/testChanged
     - env:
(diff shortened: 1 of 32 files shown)


ennru:
I'm afraid you did merge changes in master into your branch. Please rebase your changes on top of master instead.

ennru:
Fixes #36

PR #1204, opened 2018-09-11
FTP: towards 1.0

### LINKED ISSUE #1072: FTP: become Alpakka 1.0 ready
- [X] Move internal code to `impl` package

ennru:
API CHANGE

* Un-case classed settings, use the factories and the `withXyz` methods instead of named arguments.
* Credentials must be created via factory methods in `FtpCredentials`.

Fixes #1072

### DIFF
--- a/ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
+++ b/ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
@@ -132,7 +132,7 @@ object FtpSettings {
 
   /** Java API */
   def create(
-      host: java.net.InetAddress,
+      host: java.net.InetAddress
   ): FtpSettings = apply(
     host
   )
@@ -219,7 +219,7 @@ object FtpsSettings {
 
   /** Java API */
   def create(
-      host: java.net.InetAddress,
+      host: java.net.InetAddress
   ): FtpsSettings = apply(
     host
   )
--- a/ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
+++ b/ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
@@ -4,7 +4,6 @@
 
 package akka.stream.alpakka.ftp
 
-import akka.stream.alpakka.ftp.FtpCredentials.AnonFtpCredentials
 import java.net.InetAddress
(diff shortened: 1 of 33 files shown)


ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
```diff
@@ -60,45 +61,81 @@ sealed abstract class FtpFileSettings extends RemoteFileSettings {
...
-  /** Default FTP port */
+  /** Default FTP port (21) */
   final val DefaultFtpPort = 21
 
+  /** Scala API */
+  def apply(
+      host: java.net.InetAddress
+  ): FtpSettings = new FtpSettings(
+    host,
+    port = DefaultFtpPort,
```
2m: We have default parameters in the private constructor in the reference connector:

https://github.com/akka/alpakka/blob/909cdd29ad39ad6d8f85dd32c2534adcfcfecfa1/reference/src/main/scala/akka/stream/alpakka/reference/model.scala#L81-L84

ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
```diff
@@ -255,18 +404,36 @@ sealed abstract class SftpIdentity {
...
+  ): RawKeySftpIdentity = new RawKeySftpIdentity(
+    privateKey,
+    privateKeyFilePassphrase,
+    publicKey
+  )
+
+  override def toString: String =
+    "RawKeySftpIdentity(" +
+    s"privateKey(length)=${privateKey.length}," +
+    s"privateKeyFilePassphrase=${privateKeyFilePassphrase.isDefined}," +
```
2m: Maybe the label should also include that we are printing the value of `isDefined` here, like: `privateKeyFilePassphrase.isDefined`

ennru: When I looked into that, I felt the urge to improve the credentials creation experience altogether.

ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
ennru: We should change that, as it could easily lead to buggy implementations of the manually-written copy method.

PR #1206, opened 2018-09-11
DynamoDB: towards 1.0

### LINKED ISSUE #1069: DynamoDB: become Alpakka 1.0 ready
- [x] Move internal code to `impl` package
- [x] Move API code (settings) out of `impl` package

ennru:
API CHANGE
* A few classes used in the public API moved out of the `impl` package (eg. `DynamoSettings`)
* `DynamoSettings` now supports configuration via `withXyz` methods (in addition to reading from `Config`)
* To use unencrypted http, `tls` needs explicitly to be set to false. 

To be decided:
* HTTPS is now selected on the settings flag `tls`, is there a better name?
* This module is irregular in that the user creates a client and uses factory methods on it, instead of factory methods taking the client as an argument. 

Fixes #1069


dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/DynamoSettings.scala
```diff
@@ -0,0 +1,122 @@
...
+
+  require(host.nonEmpty, "A host name must be provided.")
+  require(port > -1, "A port number must be provided.")
+
+  def withRegion(value: String): DynamoSettings = copy(region = value)
+  def withHost(value: String): DynamoSettings = copy(host = value)
+  def withPort(value: Int): DynamoSettings =
+    if (value == 443) {
+      copy(port = value, tls = true)
+    } else copy(port = value)
```
2m: I am not sure about these "smart" builders. There could be a case where server expects non-secure connections on port `443` and then it could catch user off-guard. Do we have examples of "smart" builder methods anywhere else?

ennru: That was my try to keep it a bit more backwards-compatible. The old logic was to switch to https for port 443, only.

ennru: I removed the 443 magic, existing users need to switch off TLS explicitly.

PR #1215, opened 2018-09-17
Add react.sphere.it presentation

ennru:


gosubpl:
Many thanks guys, glad you've liked it!

Issue #892, opened 2018-04-15
Is there an appetite for an alternate MQTT connector


huntc:
I've just observed the MQTT component use 8 threads in my live production environment. I believe the number relates to Paho's blocking based approach. 8 threads equate to 2MiB of memory in stack space given a 256k stack size (the min). IMHO, 2MiB is excessive as no threads should have to be dedicated solely to MQTT i.e. MQTT I/O should be non-blocking and a consumer of the regular Akka dispatchers.

Would there be interest in a re-write of the MQTT connector, perhaps without the bells and whistles of Paho in terms of function? My intent would be to provide an MQTT connector based purely on Akka streams with no other dependencies. The new connector would satisfy the major use-cases of subscribing/posting messages securely. My approach would be to provide a minimal starting point on which others in the community could build on.

ennru:
Yes, please give it a go to implement a non-blocking MQTT connector.

schrepfler:
Is there any repo where this development might be happening?

huntc:
Not to my knowledge. Things have been a bit hectic here.

huntc:
@ennru I'm starting to think about this again, particularly in light of a problem I've noticed where [Paho is retaining many, many threads](https://github.com/eclipse/paho.mqtt.java/issues/402). This latter issue is causing us some pain in production.

My thinking is to compose existing stages. So, to subscribe to a topic, receive subscriptions and publish to that same topic, it might go something like this:

```scala
Source.single(Mqtt.Connect("someuser", "somepassword"))
  .via(Tcp().outgoingConnection("localhost", 1883))
  .flatMapConcat {
    case Mqtt.ConnectAck =>
      Source.single(Mqtt.Subscribe("sometopic"))
        .flatMapConcat {
          case Mqtt.SubscribeAck =>
            Source.single(Mqtt.Publish("sometopic", ByteString("hi")))
          case x => 
            x // Let through anything else including PubRec pass through
        }
    case x => 
      x // Let through anything else
  }
  .runForeach {
    case Mqtt.PubRec(topic, message) => println(s"Received: $message from $topic")
    case x                           => println(s"Received: $x")
  }
```

I realize that this operates at a lower level than at present, and perhaps higher level compositions should be declared. My thinking is to start with the lower-level components, composing as much out there as possible.

Thoughts?

2m:
Interesting.

[...]

PR #1225, opened 2018-09-22
MQTT Streaming

huntc:
This is a re-write of the existing [MQTT subproject](https://developer.lightbend.com/docs/alpakka/current/mqtt.html). It is written entirely in Akka Streams i.e. there is no dependency on Paho or any other MQTT library.

My main rationale for the re-write is that the existing subproject uses the Paho library, which is quite liberal in terms of the number of threads it uses. Thread stacks == memory (~256k per thread), and my services operate in memory-constrained environments--the environment that MQTT is actually targetted at. There should be other benefits associated with being purely dependent on Akka including performance. It should also be easier for us to control the fixing of downstream problems.

Another point of departure from the existing subproject is that connection management is considered an external concern. My view is that you should be able to form a connection with an MQTT broker in any way that you choose e.g. use the Akka stream TCP library. By focusing its concerns, this also makes the implementation of the subproject much simpler.

Fixes: https://github.com/akka/alpakka/issues/892


juanjoDiaz:
Hi @huntc,

Is there any reason to have this as a separate connector instead of simply replacing the Paho-based one?

huntc:
> Hi @huntc,
> 
> Is there any reason to have this as a separate connector instead of simply replacing the Paho-based one?

The interfaces and approaches taken are quite different in that connection management is external to this package for one thing. I didn't want to be constrained by the existing design either.

huntc:
There’s now some sample usage at OK, I'm done for the day. I've embarked on the MQTT session flow and have a working test. This should give you something to code to if you get that far: https://github.com/akka/alpakka/pull/1225/files#diff-0eb7c9cf7a5d0199db39e44f73cbb764R33

huntc:
@2m @ennru @longshorej  I've run out of time today, but I was taking an early stab at the Java API to keep on top of things. I'm having a problem with the following line in my test and wondered if you had any immediate ideas?

https://github.com/akka/alpakka/pull/1225/files#diff-7da9c0cb9b8c08348b0ab3f6d71e1486R66

2m:
When it comes to `Either` in javadsl, we usually create a use-case-specific class that has methods to test if it is a success or a failure and with the corresponding accessors for the underlying values.

2m:
Not an `Either` but here is an example for Java API for a `Try`: https://github.com/akka/alpakka/blob/8c4d25d797b84e19910e21139bc565e67a77fccb/reference/src/main/scala/akka/stream/alpakka/reference/model.scala#L43-L56

mqtt-streaming/src/test/scala/docs/scaladsl/MqttSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+
+  val settings = SessionFlowSettings(100)
+
+  "MQTT connector" should {
+
+    "flow through a session" in {
+      import MqttCodec._
+
+      val server = TestProbe()
+      val pipeToServer = Flow[ByteString].mapAsync(1)(msg => server.ref.ask(msg).mapTo[ByteString])
```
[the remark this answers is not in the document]
huntc: Thanks. I'm trying to avoid exposing Akka Typed outside of the internal classes. This code may well end up in snippets.

huntc:
> Not an `Either` but here is an example for Java API for a `Try`:
> 
> [alpakka/reference/src/main/scala/akka/stream/alpakka/reference/model.scala](https://github.com/akka/alpakka/blob/8c4d25d797b84e19910e21139bc565e67a77fccb/reference/src/main/scala/akka/stream/alpakka/reference/model.scala#L43-L56)
> 
> Lines 43 to 56 in [8c4d25d](/akka/alpakka/commit/8c4d25d797b84e19910e21139bc565e67a77fccb)
> 
>    def getBytesRead(): Integer = bytesRead match { 
>      case Success(bytesRead) => bytesRead 
>      case Failure(_) => null 
>    } 
>   
>    /** 
>     * Java API 
>     * 
>     * Return the exception if the Try contains a Failure, otherwise return a null. 
>     */ 
>    def getBytesReadFailure(): Throwable = bytesRead match { 
>      case Success(_) => null 
>      case Failure(ex) => ex 
>    }

I've now got something like:

```java
        CompletionStage<List<MqttCodec.ControlPacketResult>> result =
            Source.from(Stream.<ControlPacket>of(connect, subscribe).collect(Collectors.toList()))
                .via(Mqtt.sessionFlow(settings).join(pipeToServer))
                .runWith(Sink.seq(), mat);
```

i.e. the introduction of a `ControlPacketResult`:

```scala
  /**
   * JAVA API
   *
   * Conveniently wraps [[Either[DecodeError, ControlPacket]] as [[Either]] is horrid in Java.
   */
  final case class ControlPacketResult(v: Either[DecodeError, ControlPacket]) extends AnyVal {
    def getControlPacket: Optional[ControlPacket] =
      v match {
        case Right(cp) => Optional.of(cp)
        case Left(_) => Optional.empty()
      }

    def getControlPacketError: Optional[DecodeError] =
      v match {
        case Right(_) => Optional.empty()
        case Left(de) => Optional.of(de)
      }
  }
```

WDYT?

2m:
Yup. Way better than returning nulls as in the reference connector. Will have to update that to use `Optional` as well.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/MqttFrameStage.scala
```diff
@@ -0,0 +1,94 @@
...
+    val i = bytesReceived.iterator
+    val _ = i.drop(1) // Length starts at offset 1
+    i.decodeRemainingLength() match {
+      case Right(remainingLength) =>
+        val headerSize = bytesReceived.size - i.len
+        val packetSize = remainingLength + headerSize
+        if (packetSize <= maxPacketSize) {
+          if (bytesReceived.size >= packetSize) {
+            val (b0, b1) = bytesReceived.splitAt(packetSize)
+            frames(maxPacketSize, b1, bytesToEmit :+ b0)
```
viktorklang: appending to a list is not a fast operation. May I suggest switching to a Vector?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/MqttFrameStage.scala
```diff
@@ -0,0 +1,94 @@
...
+      bytesToEmit: List[ByteString]
+  ): Either[IllegalStateException, (immutable.Iterable[ByteString], ByteString)] = {
+    import MqttCodec._
+
+    val i = bytesReceived.iterator
+    val _ = i.drop(1) // Length starts at offset 1
+    i.decodeRemainingLength() match {
+      case Right(remainingLength) =>
+        val headerSize = bytesReceived.size - i.len
+        val packetSize = remainingLength + headerSize
```
viktorklang: Are there any overflow/underflow considerations to be made here?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/MqttFrameStage.scala
huntc: You may! Done. https://www.youtube.com/watch?v=fVq4_HhBK8Y

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/MqttFrameStage.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  private val in = Inlet[ByteString]("MqttFrame.in")
+  private val out = Outlet[ByteString]("MqttFrame.out")
+
+  override def shape: FlowShape[ByteString, ByteString] = FlowShape.of(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) {
+
+      private var bytesReceived = ByteString.empty
```
viktorklang: Move this into the InHandler?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/settings.scala
```diff
@@ -0,0 +1,31 @@
...
+  /**
+   * Java API
+   *
+   * Factory method for Java.
+   */
+  def create(maxPacketSize: Int): SessionFlowSettings =
+    SessionFlowSettings(maxPacketSize)
+}
+
+final class SessionFlowSettings private (val maxPacketSize: Int) {
```
viktorklang: No validation of maxPacketSize?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/MqttFrameStage.scala
viktorklang: :D

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/MqttFrameStage.scala
huntc: I don't believe so... have I missed something?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/MqttFrameStage.scala
huntc: Done

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/settings.scala
huntc: I'm unsure what it could be validated to...

huntc: Actually, I guess it cannot exceed 0xffff and it must be greater or equal to zero... lemme add something

huntc: Done

2m:
Very interesting to see how it will shape up using Akka Typed!

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/model.scala
```diff
@@ -0,0 +1,1175 @@
...
+        case Right(_) => Optional.empty()
+        case Left(de) => Optional.of(de)
+      }
+
+    def getControlPacket: Optional[ControlPacket] =
+      v match {
+        case Right(cp) => Optional.of(cp)
+        case Left(_) => Optional.empty()
+      }
+  }
```
huntc: @2m I've arrived at this convention for mapping Scala's either to Java. I use it now in a couple of places. It tends to read well when used.

huntc:
@2m @ennru Here's my first attempt at Akka Typed - it feels great! Any pointers or refinements are appreciated: https://github.com/akka/alpakka/pull/1225/files#diff-56c4012ad771afcefa103eef3416692f

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientConnector.scala
```diff
@@ -0,0 +1,94 @@
...
+      disconnected(Uninitialized(data.settings))
+    // TODO: UnsubscribedReceivedLocally, SubscribeReceivedLocally, PublishReceivedFromRemote, PubRelReceivedFromRemote, PubAckReceivedLocally, PubRecReceivedLocally, PubCompReceivedLocally, PublishReceivedLocally, PubRecReceivedLocally, PubCompReceivedLocally, PubAckReceivedFromRemote, PubRelReceivedFromRemote
+  }
+
+  // Actions
+
+  import MqttCodec._
+
+  def forwardConnectToRemote(connect: Connect, remote: ActorRef[ByteString]): Unit =
+    remote ! connect.encode(ByteString.newBuilder).result()
```
hseeberger: Nice!

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientConnector.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  // State event handling
+
+  def disconnected(data: Uninitialized): Behavior[Event] = Behaviors.receiveMessagePartial {
+    case ConnectReceivedLocally(connect, connectData, remote) =>
+      forwardConnectToRemote(connect, remote)
+      serverConnect(ConnectReceived(connect, connectData, data.settings))
+  }
+
+  def serverConnect(data: ConnectReceived): Behavior[Event] = Behaviors.withTimers { timer =>
```
hseeberger: I don't know the MQTT domain well, so this might be nonsense, but for a state something like "connecting" or "waitingForConnAck" might be a better name.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientConnector.scala
```diff
@@ -0,0 +1,94 @@
...
+  /*
+   * Construct with the starting state
+   */
+  def apply(settings: MqttSessionSettings): Behavior[Event] =
+    disconnected(Uninitialized(settings))
+
+  // Our FSM data and events
+
+  sealed abstract class Data(val settings: MqttSessionSettings)
+  final case class Uninitialized(override val settings: MqttSessionSettings) extends Data(settings)
```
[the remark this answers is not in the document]
huntc: The code linting is requiring that I have override val.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientConnector.scala
```diff
@@ -0,0 +1,94 @@
...
+      extends Data(settings)
+  final case class ConnAckReceived(connect: Connect, connAck: ConnAck, override val settings: MqttSessionSettings)
+      extends Data(settings)
+
+  sealed abstract class Event
+  final case class ConnectReceivedLocally(connect: Connect, connectData: ConnectData, remote: ActorRef[ByteString])
+      extends Event
+  final case class ConnAckReceivedFromRemote(connAck: ConnAck, local: ActorRef[(ConnAck, ConnectData)]) extends Event
+  case object ReceiveConnAckTimeout extends Event
+  case object LostConnection extends Event
```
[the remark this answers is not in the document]
huntc: Done.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientConnector.scala
huntc: I think the current name makes sense. From my state diagram:

image

huntc:
Thanks for the review @hseeberger !

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientConnector.scala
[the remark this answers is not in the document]
hseeberger: Really? We should cachange that. @2m ?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientConnector.scala
```diff
@@ -0,0 +1,252 @@
...
+      subscribe.topicFilters.foreach { topicFilter =>
+        val (topicName, _) = topicFilter
+        context.spawn(Subscriber(subscribeData, remote, data.packetIdAllocator, data.settings), "topic-" + topicName) // FIXME: How about existing children? getOrElse...?
+      }
+      Behaviors.same
+    case (context, SubAckReceivedFromRemote(subAck, local)) =>
+      context.children.foreach {
+        case child: ActorRef[Subscriber.Event] @unchecked =>
+          child ! Subscriber
+            .SubAckReceivedFromRemote(subAck, local) // FIXME: We should probably watch our own subscribers as there will be other types of child actor
```
huntc: @hseeberger Any thoughts on how to improve this?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientConnector.scala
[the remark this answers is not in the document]
2m: It's not me, it's scalac. :)

`override val` is required because that case class extends an abstract class. Changing `Data` to `sealed trait` fixes it:

```
sealed trait Data {
  val packetIdAllocator: ActorRef[PacketIdAllocator.Request]
  val settings: MqttSessionSettings
}
```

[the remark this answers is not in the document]
hseeberger: Nice, did not know that. Thanks!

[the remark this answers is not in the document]
hseeberger: @huntc is using a trait an option?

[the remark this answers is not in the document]
huntc: Perhaps abstract classes vs traits no longer have much meaning in Scala given that Java 8 byte code permits traits to have fields. However, the purist in me still thinks there is a "isa" vs "hasa" argument to be had... and there's precedent for `abstract class` elsewhere in Alpakka. :-)

All of that aside, traits don't always save you code. Look at this before/after scenario in our code:

*before*
```scala
sealed abstract class ControlPacket(val packetType: ControlPacketType, val flags: ControlPacketFlags)

case object Reserved1 extends ControlPacket(ControlPacketType.Reserved1, ControlPacketFlags.ReservedGeneral)

case object Reserved2 extends ControlPacket(ControlPacketType.Reserved2, ControlPacketFlags.ReservedGeneral)
```

*after*
```scala
sealed trait ControlPacket {
  def packetType: ControlPacketType
  def flags: ControlPacketFlags
}

case object Reserved1 extends ControlPacket {
  val packetType: ControlPacketType = ControlPacketType.Reserved1
  val flags: ControlPacketFlags = ControlPacketFlags.ReservedGeneral
}

case object Reserved2 extends ControlPacket {
  val packetType: ControlPacketType = ControlPacketType.Reserved2
  val flags: ControlPacketFlags = ControlPacketFlags.ReservedGeneral
}
```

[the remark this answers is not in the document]
hseeberger: I consider the second more readable, because long lines give me hard times. But anyway, that's just me.

huntc:
Hey @ennru and @2m - Just to let you know that this PR is now fully complete from an MQTT client perspective. I also lifted some of the existing MQTT tests that you had to ensure that connectivity to a real MQTT broker would work - and it did - first time! :-) 

I'm now embarking on the MQTT server code as part of this PR, as we need it ourselves (we're using MQTT outside of a broker scenario).

Feel free to give the PR a heavy review. The code is 100% functional and complete when compared to the existing library and Paho.

ennru:
@huntc Thank you so much for your effort!
This is impressive work and it is fun to see Akka Typed in a real application.

We'll review thoroughly, I can't promise we'll get to it even this month - but we'll try.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ServerState.scala
```diff
@@ -0,0 +1,290 @@
...
+      extends Event(connectionId)
+  final case class ConnectionLost(override val connectionId: ByteString) extends Event(connectionId)
+  final case class ClientConnectionFree(override val connectionId: ByteString) extends Event(connectionId)
+
+  // State event handling
+
+  private val ClientConnectionNamePrefix = "client-connection-"
+
+  private def mkActorName(name: String): String =
+    name.getBytes(StandardCharsets.UTF_8).map(_.toHexString).mkString
```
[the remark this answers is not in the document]
huntc: Thanks for this - fixed.

huntc:
@ennru @2m Both the client and server are now functionally complete. There is one known bug due to my incorrect handling of Akka Typed actor failures. My problem is explained here: https://github.com/akka/akka/issues/25803.

@longshorej and I will now review this 6k LOC PR (!) and test it further against our own usage. It'd be great to see this PR merged by the end of the week if possible. Perhaps I could accelerate things by taking one or both of you through the code.

Meanwhile, for a comparison, here's some sample client code which is reasonably equivalent to the existing MQTT connector doco. The client is connecting to a real broker.

First, create a flow:

```scala
      val settings = MqttSessionSettings()
      val session = ActorMqttClientSession(settings)

      val connection = Tcp().outgoingConnection("localhost", 1883)

      val mqttFlow: Flow[Command[_], Either[MqttCodec.DecodeError, Event[_]], NotUsed] =
        Mqtt
          .clientSessionFlow(session)
          .join(connection)
```

...and then run some commands through it:

```scala
      val (commands, events) =
        Source
          .queue(3, OverflowStrategy.fail)
          .via(mqttFlow)
          .drop(3)
          .toMat(Sink.head)(Keep.both)
          .run()

      commands.offer(Command(Connect(clientId, ConnectFlags.None)))
      commands.offer(Command(Subscribe(topic)))
      commands.offer(Command(Publish(topic, ByteString("ohi"))))
```

The resulting `event` will be the `Publish` echoed back by the MQTT broker given our subscription to the topic.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ServerState.scala
```diff
@@ -0,0 +1,799 @@
...
+
+/*
+ * Handles events in relation to a specific client connection
+ */
+@InternalApi private[streaming] object ClientConnection {
+
+  /*
+   * No ACK received - the subscription failed
+   */
+  case object ClientConnectionFailed extends Exception
```
patriknw: I'd use `with NoStackTrace` here, since the stackTrace is questionable when sharing same exception instance with `object`

same in other places where object Exception is used

huntc: Thanks!

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ActorName.scala
```diff
@@ -0,0 +1,18 @@
...
+import java.nio.charset.StandardCharsets
+
+import akka.annotation.InternalApi
+
+/*
+ * Provides the ability to form valid actor names
+ */
+@InternalApi object ActorName {
+  def mkName(name: String): String =
+    name.getBytes(StandardCharsets.UTF_8).map(b => ("0" + b.toHexString).takeRight(2)).mkString
```
patriknw: what is the reason/advantage over using `java.net.URLEncoder`?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -0,0 +1,526 @@
...
+  // State event handling
+
+  private val ConsumerNamePrefix = "consumer-"
+  private val ProducerNamePrefix = "producer-"
+  private val SubscriberNamePrefix = "subscriber-"
+  private val UnsubscriberNamePrefix = "unsubscriber-"
+
+  def disconnected(data: Uninitialized): Behavior[Event] = Behaviors.receivePartial {
+    case (context, ConnectReceivedLocally(connect, connectData, remote)) =>
+      implicit val mat: Materializer = ActorMaterializer()(context.system)
```
patriknw: Creating a materializer that is bound to the ActorSystem from inside an actor is an anti pattern (possible memory leak). I'd recommend sharing one materializer by passing in constructor parameter or putting it in an extension. For actor bound materializer there is an open PR: https://github.com/akka/akka/pull/25568

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -0,0 +1,526 @@
...
+  }
+
+  def serverConnected(data: ConnAckReceived): Behavior[Event] = Behaviors.withTimers { timer =>
+    if (data.keepAlive.toMillis > 0)
+      timer.startSingleTimer("send-pingreq", SendPingReqTimeout, data.keepAlive)
+    else
+      data.remote.complete() // We'll never be sending pings so free up the command channel for other things
+
+    Behaviors
+      .receivePartial[Event] {
```
patriknw: Is the reason that you use `receivePartial` instead of `receive` that you don't care about several of the messages in certain FSM states? Otherwise the total function has the nice property that the compiler will help with exhaustiveness checks of sealed traits.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ServerState.scala
```diff
@@ -0,0 +1,800 @@
...
+                  context.spawn(Producer(publish, publishData, selfAsRemote, data.producerPacketRouter, data.settings),
+                                producerName),
+                  ProducerFree(publish.topicName)
+                )
+                clientConnected(data)
+              case _ =>
+                clientConnected(
+                  data.copy(pendingLocalPublications = data.pendingLocalPublications :+ (publish.topicName -> prl))
+                )
+            }
```
patriknw: code style: some of these bigger actors might benefit from extracting the code into smaller methods, instead of putting everything in one big pattern match. I think that improves readability and navigation.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
huntc: Thanks for the pick-up. Changed.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
huntc: Yes, this is deliberate and as illustrated by your documentation: https://doc.akka.io/docs/akka/2.5.17/typed/fsm.html. Would you personally list each event, given that there can be many?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ServerState.scala
huntc: I'm on the fence on that one... Let's consult the master of style - @hseeberger ? :-)

huntc:
@patriknw:

> Not a thorough review at all, but I was curious about how Akka Typed was used. Looks pretty clean. Would love to hear your feedback after writing that much actor code, @huntc

Akka Typed was great! Easy to learn and apply given my previous Akka knowledge. I really appreciated the type safety and I'm sure it saved me from trouble many times. Perhaps the only gap in the doco is to explain that `Props` becomes an `apply` function--very straightforward, but it wasn't obvious to me until I watched one of @hseeberger 's presentations.

I still think that `watchWith` could benefit from an additional message to use in the case of failure, or perhaps a `Failable` trait that can be mixed into the `watchWith` message with a `failure` field to indicate optional failure. Being able to discern regular stopping from failure is important.

I'm also wondering if stashing can be made easier. While I didn't find it hard to do myself, I did find myself doing it a bit... Perhaps a `StashedBehaviors` might be useful (or something like that).

That's about it though. I can't see myself going back to Untyped. Ship it!

huntc:
@hseeberger Are you able to re-review? With thanks.

mqtt-streaming/src/test/java/docs/javadsl/MqttFlowTest.java
```diff
@@ -0,0 +1,219 @@
...
+  public void establishServerBidirectionalConnectionAndSubscribeToATopic()
+      throws InterruptedException, ExecutionException, TimeoutException {
+    String clientId = "flow-spec/flow";
+    String topic = "source-spec/topic1";
+    String host = "localhost";
+    int port = 9884;
+
+    // #create-streaming-bind-flow
+    MqttSessionSettings settings = MqttSessionSettings.create();
+    MqttServerSession session = new ActorMqttServerSession(settings, materializer, system);
```
ennru: I wonder if you should have a `create` method in `MqttServerSession` instead? 
Many are puzzled by this kind of distinction (as we have with `Materializer` and `ActorMaterializer`).

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/model.scala
```diff
@@ -0,0 +1,1175 @@
...
+  def apply(topicFilter: String): Subscribe =
+    new Subscribe(topicFilter)
+}
+
+/**
+ * 3.8 SUBSCRIBE - Subscribe to topics
+ * http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html
+ */
+final case class Subscribe @InternalApi private[streaming] (packetId: PacketId,
+                                                            topicFilters: Seq[(String, ControlPacketFlags)])
```
ennru: Offer a `getTopicFilters: java.util.List[akka.japi.Pair[String, ControlPacketFlags]]` as Java API, please.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/model.scala
```diff
@@ -0,0 +1,1175 @@
...
+  /**
+   * Send a command to an MQTT session with data to carry through into
+   * any related event.
+   * @param command The command to send
+   * @param carry The data to carry through
+   * @tparam A The type of data to carry through
+   */
+  def apply[A](command: ControlPacket, carry: A): Command[A] =
+    new Command(command, carry)
+}
```
ennru: Would it make sense to offer `create` methods for the important `ControlPacket` subtypes to get `Command.pubAck(packetId)`?

mqtt-streaming/src/test/scala/docs/scaladsl/MqttFlowSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+      val clientId = "flow-spec/flow"
+      val topic = "source-spec/topic1"
+
+      //#create-streaming-flow
+      val settings = MqttSessionSettings()
+      val session = ActorMqttClientSession(settings)
+
+      val connection = Tcp().outgoingConnection("localhost", 1883)
+
+      val mqttFlow: Flow[Command[_], Either[MqttCodec.DecodeError, Event[_]], NotUsed] =
```
ennru: You could add a type alias for `Either[MqttCodec.DecodeError, Event[_]]` to make this read better.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -0,0 +1,520 @@
...
+    )
+
+  // Our FSM data, FSM events and commands emitted by the FSM
+
+  sealed abstract class Data(val consumerPacketRouter: ActorRef[RemotePacketRouter.Request[Consumer.Event]],
+                             val producerPacketRouter: ActorRef[LocalPacketRouter.Request[Producer.Event]],
+                             val subscriberPacketRouter: ActorRef[LocalPacketRouter.Request[Subscriber.Event]],
+                             val unsubscriberPacketRouter: ActorRef[LocalPacketRouter.Request[Unsubscriber.Event]],
+                             val settings: MqttSessionSettings)
+  final case class Uninitialized(
```
ennru: Empty lines aren't that expensive, please add a few.
(Possibly putting some Scaladoc on the super-class to keep the grouping you tried to achieve.)

docs/src/main/paradox/mqtt-streaming.md
```diff
@@ -0,0 +1,87 @@
...
+
+MQTT stands for MQ Telemetry Transport. It is a publish/subscribe messaging protocol, designed for constrained devices and low-bandwidth, high-latency or unreliable networks. The design principles are to minimize network bandwidth and device resource requirements whilst also attempting to ensure reliability and some degree of assurance of delivery. These principles also turn out to make the protocol ideal of the emerging “machine-to-machine” (M2M) or “Internet of Things” world of connected devices, and for mobile applications where bandwidth and battery power are at a premium.  
+
+Further information on [mqtt.org](https://mqtt.org/).
+
+@@@ 
+
+The Alpakka MQTT connector provides an Akka Stream flow to connect to MQTT brokers. In addition, a flow is provided so that you can implement your own MQTT server in the case where you do not wish to use a broker--MQTT is a fine protocol for directed client/server interactions, as well as having an intermediary broker.
+
+The library has no dependencies other than those of Akka Streams i.e. it is entirely reactive. As such, there should be a significant performance advantage given its pure-Akka foundations.
```
ennru: Please add a reference to the Paho-based module, as well.
And add a reference to this module in `mqtt.md`...

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
patriknw: it's perfectly fine, I was just curious. If it was just about two or three messages I would probably list them to keep the exhaustiveness checks, but I fully understand that it can be too much boilerplate if many

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/model.scala
```diff
@@ -0,0 +1,1175 @@
...
+  val SUBSCRIBE = ControlPacketType(8)
+  val SUBACK = ControlPacketType(9)
+  val UNSUBSCRIBE = ControlPacketType(10)
+  val UNSUBACK = ControlPacketType(11)
+  val PINGREQ = ControlPacketType(12)
+  val PINGRESP = ControlPacketType(13)
+  val DISCONNECT = ControlPacketType(14)
+  val Reserved2 = ControlPacketType(15)
+}
+final case class ControlPacketType(underlying: Int) extends AnyVal
```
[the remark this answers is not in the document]
huntc: The `underlying` field should be public.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/model.scala
```diff
@@ -0,0 +1,1175 @@
...
+                                                          topicName: String,
+                                                          packetId: Option[PacketId],
+                                                          payload: ByteString)
+    extends ControlPacket(ControlPacketType.PUBLISH, flags) {
+
+  /**
+   * 3.3 PUBLISH – Publish message
+   * http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html
+   */
+  def this(flags: ControlPacketFlags, topicName: String, payload: ByteString) =
```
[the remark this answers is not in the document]
huntc: I'm a little conflicted here as I've never really understood why we prefer Java developers to use a `create` method. In Java, it is idiomatic to create an object by using `new`. They'd be a lot of change for me to affect here, so I'd like to better understand why `create` would be preferred.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
huntc: I think that this is a coding style thing, and perhaps a little subjective. I won't mind at all if you'd like to do this once my PR is merged.

docs/src/main/paradox/mqtt-streaming.md
huntc: What would you like me to say in reference to Paho?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/MqttSessionSettings.scala
```diff
@@ -0,0 +1,188 @@
...
+    copy(receiveUnsubAckTimeout = receiveUnsubAckTimeout)
+
+  def withMaxClientConnectionStashSize(maxClientConnectionStashSize: Int): MqttSessionSettings =
+    copy(maxClientConnectionStashSize = maxClientConnectionStashSize)
+
+  def withClientTerminationWatcherBufferSize(clientTerminationWatcherBufferSize: Int): MqttSessionSettings =
+    copy(clientTerminationWatcherBufferSize = clientTerminationWatcherBufferSize)
+
+  def withServerSendBufferSize(serverSendBufferSize: Int): MqttSessionSettings =
+    copy(serverSendBufferSize = serverSendBufferSize)
```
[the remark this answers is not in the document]
huntc: Done (pending push).

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/model.scala
huntc: Done. Pending push.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/model.scala
huntc: Again, I'm conflicted about `create`. The Java developer can use `new` here, quite nicely.

mqtt-streaming/src/test/scala/docs/scaladsl/MqttFlowSpec.scala
huntc: I already have a Java `DecodeErrorOrEvent`, so I'm unsure what I'd call it. I generally prefer not to hide the types also.

huntc:
Hi @ennru - most of your requests are in now.

> A great read!
Thanks!

> I don't know the details of the MQTT protocol, but I'm sure you've done your spec reading properly.
There are no guarantees. ;-) Suffice to state though, this new MQTT machinery is at the heart of what we're doing for our Cisco IoT project. Any problems will be jumped on fast out of necessity.

> My comments are mostly focused on the API experience, especially for potential Java users.
'hopefully addressed.

> Looking forward to getting this in very soon.
Not as much as me! ;-)

Thanks again.

docs/src/main/paradox/mqtt-streaming.md
ennru: "Alpakka contains another MQTT connector which is based on the Eclipse Paho client." (and in the other direction in `mqtt.md`)
Maybe point out some differences.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/model.scala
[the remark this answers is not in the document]
ennru: Making the constructor private doesn't hide the field.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/model.scala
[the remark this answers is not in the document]
ennru: Yes, I guess you're right about `new` and Java. It's mostly that I find factory methods easier to evolve.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/model.scala
ennru: This would work nice for Scala, as well: instead of `Command(Connect(clientId,...))` you'd have `Command.connect(clientId)`
But yeah, just an idea.

project/Dependencies.scala
```diff
@@ -249,6 +249,14 @@ object Dependencies {
     )
   )
 
+  val MqttStreaming = Seq(
+    libraryDependencies ++= Seq(
+      "com.typesafe.akka" %% "akka-actor-typed" % AkkaVersion, // ApacheV2
+      "com.typesafe.akka" %% "akka-actor-testkit-typed" % AkkaVersion, // ApacheV2
```
ennru: This lacks a `% Test`.

mqtt-streaming/src/test/java/docs/javadsl/MqttFlowTest.java
ennru: Any comment to this?

huntc:
> "Alpakka contains another MQTT connector which is based on the Eclipse Paho client." (and in the other direction in mqtt.md)
Maybe point out some differences.

Done.

huntc:
> This would work nice for Scala, as well: instead of Command(Connect(clientId,...)) you'd have Command.connect(clientId)
But yeah, just an idea.

A `Command` is not the same as a `CommandPacket`, and an `Event` also has a `CommandPacket`. `CommandPacket` can optionally declare a `carry` value and `Event` can optionally declare the related `carry`. I don't see merit in conflating these concerns.

project/Dependencies.scala
huntc: Fixed

huntc:
> Making the constructor private doesn't hide the field.

Fixed

project/Dependencies.scala
huntc: Fixed.

huntc:
> This would work nice for Scala, as well: instead of Command(Connect(clientId,...)) you'd have Command.connect(clientId)
But yeah, just an idea.

`Command` and `Event` both contain `ControlPacket`s - they're different kinds of things. I don't see merit in conflating the control packet into these types.

huntc:
@ennru I believe I've addressed your comments.

I've also added JMH benchmarking. With the regular benchmarking caveats, this new library of mine is about 13% faster than its Paho counterpart - that's with zero optimisations. Results:

```
[info] Benchmark                          Mode  Cnt      Score       Error  Units
[info] MqttPerf.serverPublish            thrpt   25  32940.962 ± 10220.687  ops/s
[info] streaming.MqttPerf.serverPublish  thrpt   25  37430.264 ±  5937.542  ops/s
```

I'm using "at least once" to get some ack-ing going. Both clients are using this PR's server implementation to provide as much consistency as possible. I'm also using TCP.

Impress your friends: `mqtt-streaming-bench/jmh:run`

build.sbt
```diff
@@ -162,6 +162,9 @@ lazy val mongodb = alpakkaProject("mongodb", "mongodb", Dependencies.MongoDb)
 lazy val mqtt = alpakkaProject("mqtt", "mqtt", Dependencies.Mqtt)
 
 lazy val mqttStreaming = alpakkaProject("mqtt-streaming", "mqttStreaming", Dependencies.MqttStreaming)
+lazy val mqttStreamingBench = alpakkaProject("mqtt-streaming-bench", "mqttStreamingBench", Seq.empty)
+  .enablePlugins(JmhPlugin)
+  .dependsOn(mqtt, mqttStreaming)
```
2m: I think it would be better to move jmh tests to their corresponding (`mqtt` and `mqttStreaming`) projects.

huntc: I believe that they must be in their own sub project given code generation. Do I have that wrong? I recall reading something to that effect from the sbt-jmh doc.

2m: Ahh, you are right. sbt-jmh does indeed say that. Lets leave it as it is now then. Probably in the future we will want one project for all benchmarks. Otherwise it will end up with a cambrian explosion of benchmark projects.

huntc:
Thanks for approving @ennru - just to let you know that I'm profiling/stressing the code. May I let you know when I'm done, so that I can feel confident this works well?

huntc:
@ennru I discovered a bug while stressing the connector and have now fixed it. The problem was associated with sometimes dropping PUBLISH messages given the same packet id.

Some new benchmarks also as I've slowed them down to be more deliberate about ensuring QoS 1:

```
[info] Benchmark                          Mode  Cnt     Score     Error  Units
[info] MqttPerf.serverPublish            thrpt   25  2631.280 ±  98.332  ops/s
[info] streaming.MqttPerf.serverPublish  thrpt   25  2326.748 ± 270.795  ops/s
```

So, we're actually a bit slower than Paho, which doesn't surprise me given Paho's maturity. However, I've not profiled this code yet and so there should be some opportunities for improvement down the track. I'm confident that its memory usages is better though, and that its API is a little nicer and certainly more flexible.

huntc:
@ennru I'm now pretty happy with this. I've done some memory profiling and don't see anything out of the ordinary.

We will be continuing our tests, but if you're happy then please feel free to merge it.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/RequestState.scala
```diff
@@ -0,0 +1,410 @@
...
+    Behaviors.receiveMessage {
+      case Register(registrant: ActorRef[A], replyTo) if nextPacketId.underlying <= MaxPacketId.underlying =>
+        replyTo ! Registered(nextPacketId)
+        main(registrantsByPacketId + (nextPacketId -> registrant), PacketId(nextPacketId.underlying + 1))
+      case _: Register[A] =>
+        Behaviors.same // We cannot allocate any more. This will eventually cause a timeout to occur on the requestor.
+      case Unregister(packetId) =>
+        val remainingPacketIds = registrantsByPacketId - packetId
+        val revisedNextPacketId = if (remainingPacketIds.nonEmpty) {
+          val maxPacketId = PacketId(remainingPacketIds.keys.map(_.underlying).max)
```
viktorklang:
```suggestion
          val maxPacketId = remainingPacketIds.keys.maxBy(_.underlying)
```

huntc:
@ennru Be sure to squash those commits on merging! ;-)

ennru:
Thank you for this huge effort! I hope it will serve your needs well and get great appreciation from the community.

huntc:
Thanks!

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -397,33 +397,39 @@ import scala.util.{Failure, Success}
...
-    }
+    Behaviors
+      .supervise(Behaviors.receiveMessagePartial[Event] {
+        case AcquiredPacketId(packetId) =>
+          data.remote ! ForwardSubscribe(packetId)
+          serverSubscribe(
+            ServerSubscribe(packetId, data.subscribeData, data.packetRouter, data.settings)
+          )
+        case UnobtainablePacketId =>
+          throw SubscribeFailed
```
dwijnand: Why throw, rather than stopped?

huntc: The actor performing the subscription (the one you're looking at) is being watched by its parent. The parent needs to understand *how* the child terminates i.e. whether it was because of failure or whether it completes normally. Throwing an exception in the subject actor is the only way that an observing actor can discern this.

dwijnand: Ah right!  And why is it preferable in this situation to throw rather than send a message back?

huntc:
I could have sent an error style message back, although I still have to handle for child actor failure anyhow (actors can fail for various reasons). It seemed easier this way. :-)

dwijnand:
OK, thanks.  Just learning, and found this PR via https://github.com/akka/akka/issues/25803.

huntc:
Hey no worries! Keep the questions coming. Akka Typed was new to me also, so still learning. Happy to exchange ideas.

hseeberger:
I have to chim in to second Dale: please don't throw, even if it might be more convenient. But throwing opens a side channel that is hard to detect/understand. Better add a message to the protocol of the parent/supervisor for every expected/well-defined exception (like `UnobtainablePacketId`). Only use specific supervision for "unexpected/hardly-defined" exceptions.

huntc:
@hseeberger I must already handle the case where a child actor has terminated with failure. Adding an additional message that communicates failure would require additional logic. I do not see the advantage in that.

Also, Akka relies on exceptions to detect child actor failure.

dwijnand:
I think it's the same as you require additional logic to suppress that exception's logging (which was temporarily backed out due to that but).

Is perhaps the exception log suppressing logic less invasive or verbose than the message handling one would be?

huntc:
> On 5 Nov 2018, at 08:19, Dale Wijnand <notifications@github.com> wrote:
> 
> Is perhaps the exception log suppressing logic less invasive or verbose than the message handling one would be?
Avoiding throwing my own exception means I would have to have logic handling both an exception *and* the failure message. It would be additional code.

No matter what, I must handle the child actor termination and discern if it was because of failure.

patriknw:
We created issue https://github.com/akka/akka/issues/25803 related to this, and when I thought some more about it afterwards I'm skeptical to this exception passing via watch because sending exceptions in remote messages is difficult (serialization problems). Just having them for watch of local child actors doesn't align well with location transparency principles.

dwijnand:
I still think it's comparable.

Your solution, with regards handling the termination of a child actor is:
1. handle gracefully termination (i.e with no failure)
2. handle any termination failures (including expected `SubscribeFailed`)
3. specifically opt-out of logging if the exception is `SubscribeFailed`

The alternative to point 3 is to:

1. add handling of another expected _message_ in parent, no need to avoid logging

In the child it would probably look along the lines of

```scala
        case UnobtainablePacketId =>
          replyTo ! SubscribeFailed
          Behaviors.stopped
```

at which point the gracefully termination handling would happen, rather than the termination failure + logging opt-out handling.

One upside is it would be a subtype of your message type (potential for using match exhaustion), rather than matching `Throwable`.

PR #1227, opened 2018-09-22
Jms producer retries

### LINKED ISSUE #1226: Jms: Producer should retry sending messages
Thanks to @akara, the Jms connector now has reconnect capabilities that are used during initial connection in both producer and consumers, but only the consumer stage leverages the reconnect logic while the stream is running.

What is missing is to enable reconnection in the producer stage, and implement sending in a way that retries a configurable amount of time on failure.

For this, I have a proposal available that we can discuss (PR will follow).

andreas-schroeder:
This is my proposal on how to achieve producer retries with reconnects.

I also simplified some code along the line (which I hope is okay for everyone).
Also, I added a missing async callback on disconnect that incidentally also changes the amount of retries performed in the reconnect test (I will point out in annotations where).

Documentation is lacking for now, I will add some examples once we are fine with the design and the way it works.

Fixes #1226

### DIFF
--- a/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
@@ -86,18 +86,17 @@ final case class ConnectionRetrySettings(connectTimeout: FiniteDuration = 10.sec
   def withMaxRetries(maxRetries: Int): ConnectionRetrySettings = copy(maxRetries = maxRetries)
   def withInfiniteRetries(): ConnectionRetrySettings = withMaxRetries(-1)
 
-  /** Hypothetical retry time, not accounting for maxBackoff. */
   def waitTime(retryNumber: Int): FiniteDuration =
-    (initialRetry * Math.pow(retryNumber, backoffFactor)).asInstanceOf[FiniteDuration]
+    (initialRetry * Math.pow(retryNumber, backoffFactor)).asInstanceOf[FiniteDuration].min(maxBackoff)
 }
 
 object SendRetrySettings {
   def create(): SendRetrySettings = SendRetrySettings()
 }
 
-final case class SendRetrySettings(initialRetry: FiniteDuration = 50.millis,
-                                   backoffFactor: Double = 2,
-                                   maxBackoff: FiniteDuration = 10.seconds,
+final case class SendRetrySettings(initialRetry: FiniteDuration = 20.millis,
+                                   backoffFactor: Double = 1.5,
+                                   maxBackoff: FiniteDuration = 500.millis,
                                    maxRetries: Int = 10) {
   def withInitialRetry(delay: FiniteDuration): SendRetrySettings = copy(initialRetry = delay)
   def withInitialRetry(delay: Long, unit: TimeUnit): SendRetrySettings =
@@ -109,9 +108,8 @@ final case class SendRetrySettings(initialRetry: FiniteDuration = 50.millis,
   def withMaxRetries(maxRetries: Int): SendRetrySettings = copy(maxRetries = maxRetries)
   def withInfiniteRetries(): SendRetrySettings = withMaxRetries(-1)
 
-  /** Hypothetical retry time, not accounting for maxBackoff. */
   def waitTime(retryNumber: Int): FiniteDuration =
-    (initialRetry * Math.pow(retryNumber, backoffFactor)).asInstanceOf[FiniteDuration]
+    (initialRetry * Math.pow(retryNumber, backoffFactor)).asInstanceOf[FiniteDuration].min(maxBackoff)
 }
 
 object JmsConsumerSettings {
--- a/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
@@ -84,6 +84,30 @@ final case class ConnectionRetrySettings(connectTimeout: FiniteDuration = 10.sec
   def withMaxBackoff(maxBackoff: Long, unit: TimeUnit): ConnectionRetrySettings =
     copy(maxBackoff = Duration(maxBackoff, unit))
   def withMaxRetries(maxRetries: Int): ConnectionRetrySettings = copy(maxRetries = maxRetries)
+  def withInfiniteRetries(): ConnectionRetrySettings = withMaxRetries(-1)
+
+  /** Hypothetical retry time, not accounting for maxBackoff. */
+  def waitTime(retryNumber: Int): FiniteDuration =
+    (initialRetry * Math.pow(retryNumber, backoffFactor)).asInstanceOf[FiniteDuration]
+}
+
+object SendRetrySettings {
+  def create(): SendRetrySettings = SendRetrySettings()
+}
+
+final case class SendRetrySettings(initialRetry: FiniteDuration = 50.millis,
+                                   backoffFactor: Double = 2,
+                                   maxBackoff: FiniteDuration = 10.seconds,
+                                   maxRetries: Int = 10) {
+  def withInitialRetry(delay: FiniteDuration): SendRetrySettings = copy(initialRetry = delay)
+  def withInitialRetry(delay: Long, unit: TimeUnit): SendRetrySettings =
+    copy(initialRetry = Duration(delay, unit))
+  def withBackoffFactor(backoffFactor: Double): SendRetrySettings = copy(backoffFactor = backoffFactor)
+  def withMaxBackoff(maxBackoff: FiniteDuration): SendRetrySettings = copy(maxBackoff = maxBackoff)
+  def withMaxBackoff(maxBackoff: Long, unit: TimeUnit): SendRetrySettings =
+    copy(maxBackoff = Duration(maxBackoff, unit))
+  def withMaxRetries(maxRetries: Int): SendRetrySettings = copy(maxRetries = maxRetries)
+  def withInfiniteRetries(): SendRetrySettings = withMaxRetries(-1)
 
   /** Hypothetical retry time, not accounting for maxBackoff. */
   def waitTime(retryNumber: Int): FiniteDuration =
@@ -133,6 +157,7 @@ object JmsProducerSettings {
 
 final case class JmsProducerSettings(connectionFactory: ConnectionFactory,
                                      connectionRetrySettings: ConnectionRetrySettings = ConnectionRetrySettings(),
+                                     sendRetrySettings: SendRetrySettings = SendRetrySettings(),
                                      destination: Option[Destination] = None,
                                      credentials: Option[Credentials] = None,
                                      sessionCount: Int = 1,
@@ -142,6 +167,8 @@ final case class JmsProducerSettings(connectionFactory: ConnectionFactory,
   def withCredential(credentials: Credentials): JmsProducerSettings = copy(credentials = Some(credentials))
   def withConnectionRetrySettings(settings: ConnectionRetrySettings): JmsProducerSettings =
     copy(connectionRetrySettings = settings)
+  def withSendRetrySettings(settings: SendRetrySettings): JmsProducerSettings =
+    copy(sendRetrySettings = settings)
   def withSessionCount(count: Int): JmsProducerSettings = copy(sessionCount = count)
   def withQueue(name: String): JmsProducerSettings = copy(destination = Some(Queue(name)))
   def withTopic(name: String): JmsProducerSettings = copy(destination = Some(Topic(name)))
--- a/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
@@ -154,7 +154,9 @@ final case class JmsProducerSettings(connectionFactory: ConnectionFactory,
     copy(acknowledgeMode = Option(acknowledgeMode))
 }
 
-final case class Credentials(username: String, password: String)
+final case class Credentials(username: String, password: String) {
+  override def toString = s"Credentials($username,${"*" * password.length})"
+}
 
 object JmsBrowseSettings {
 
--- a/jms/src/main/scala/akka/stream/alpakka/jms/JmsBrowseStage.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/JmsBrowseStage.scala
@@ -10,9 +10,8 @@ import akka.stream.{ActorAttributes, Attributes, Outlet, SourceShape}
 import akka.stream.stage.{GraphStage, GraphStageLogic, OutHandler}
 import java.util.{Enumeration => JEnumeration}
 
-private[jms] final class JmsBrowseStage(settings: JmsBrowseSettings) extends GraphStage[SourceShape[Message]] {
-  private val queue = settings.destination.getOrElse { throw new IllegalArgumentException("Destination is missing") }
-
+private[jms] final class JmsBrowseStage(settings: JmsBrowseSettings, queue: Destination)
+    extends GraphStage[SourceShape[Message]] {
   private val out = Outlet[Message]("JmsBrowseStage.out")
   val shape = SourceShape(out)
 
--- a/jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
@@ -30,6 +30,8 @@ private[jms] trait JmsConnector[S <: JmsSession] {
 
   protected var jmsSessions = Seq.empty[S]
 
+  protected def destination: Destination
+
   protected def jmsSettings: JmsSettings
 
   protected def onSessionOpened(jmsSession: S): Unit = {}
@@ -178,14 +180,9 @@ private[jms] trait JmsConsumerConnector extends JmsConnector[JmsConsumerSession]
 
   override def openSessions(onConnectionFailure: jms.JMSException => Unit): Future[Seq[JmsConsumerSession]] =
     openRecoverableConnection(startConnection = true, onConnectionFailure).flatMap { connection =>
-      val createDestination = jmsSettings.destination match {
-        case Some(destination) => destination.create
-        case _ => throw new IllegalArgumentException("Destination is missing")
-      }
-
       val sessionFutures =
         for (_ <- 0 until jmsSettings.sessionCount)
-          yield Future(createSession(connection, createDestination))
+          yield Future(createSession(connection, destination.create))
       Future.sequence(sessionFutures)
     }(ExecutionContexts.sameThreadExecutionContext)
 }
@@ -200,14 +197,9 @@ private[jms] trait JmsProducerConnector extends JmsConnector[JmsProducerSession]
 
   def openSessions(onConnectionFailure: jms.JMSException => Unit): Future[Seq[JmsProducerSession]] =
     openRecoverableConnection(startConnection = false, onConnectionFailure).flatMap { connection =>
-      val createDestination = jmsSettings.destination match {
-        case Some(destination) => destination.create
-        case _ => throw new IllegalArgumentException("Destination is missing")
-      }
-
       val sessionFutures =
         for (_ <- 0 until jmsSettings.sessionCount)
-          yield Future(createSession(connection, createDestination))
+          yield Future(createSession(connection, destination.create))
       Future.sequence(sessionFutures)
     }(ExecutionContexts.sameThreadExecutionContext)
 }
--- a/jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
@@ -16,6 +16,7 @@ import akka.stream.{ActorAttributes, ActorMaterializerHelper, Attributes}
 import javax.jms
 
 import scala.concurrent.{ExecutionContext, Future, TimeoutException}
+import scala.util.Try
 import scala.util.control.NonFatal
 
 /**
@@ -37,6 +38,8 @@ private[jms] trait JmsConnector[S <: JmsSession] {
   protected def onSessionOpened(jmsSession: S): Unit = {}
 
   protected val fail: AsyncCallback[Throwable] = getAsyncCallback[Throwable](e => failStage(e))
+
+  private val connectionFailedCB: AsyncCallback[Throwable] = getAsyncCallback[Throwable](e => connectionFailed(e))
 
   private val onSession: AsyncCallback[S] = getAsyncCallback[S] { session =>
     jmsSessions :+= session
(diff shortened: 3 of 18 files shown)


jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
```diff
@@ -84,6 +84,30 @@ final case class ConnectionRetrySettings(connectTimeout: FiniteDuration = 10.sec
   def withMaxBackoff(maxBackoff: Long, unit: TimeUnit): ConnectionRetrySettings =
     copy(maxBackoff = Duration(maxBackoff, unit))
   def withMaxRetries(maxRetries: Int): ConnectionRetrySettings = copy(maxRetries = maxRetries)
+  def withInfiniteRetries(): ConnectionRetrySettings = withMaxRetries(-1)
```
andreas-schroeder: Added some helper methods for indefinite retries

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
```diff
@@ -84,6 +84,30 @@ final case class ConnectionRetrySettings(connectTimeout: FiniteDuration = 10.sec
...
+}
+
+object SendRetrySettings {
+  def create(): SendRetrySettings = SendRetrySettings()
+}
+
+final case class SendRetrySettings(initialRetry: FiniteDuration = 50.millis,
+                                   backoffFactor: Double = 2,
+                                   maxBackoff: FiniteDuration = 10.seconds,
+                                   maxRetries: Int = 10) {
```
andreas-schroeder: I chose the default send retries to be much more tight than the reconnect retries. If other default settings make more sense, we should provide those.

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
```diff
@@ -154,7 +181,9 @@ final case class JmsProducerSettings(connectionFactory: ConnectionFactory,
     copy(acknowledgeMode = Option(acknowledgeMode))
 }
 
-final case class Credentials(username: String, password: String)
+final case class Credentials(username: String, password: String) {
+  override def toString = s"Credentials($username,${"*" * password.length})"
```
andreas-schroeder: *-ing out passwords since they would be printed out otherwise in error messages I added

jms/src/main/scala/akka/stream/alpakka/jms/JmsBrowseStage.scala
```diff
@@ -10,9 +10,8 @@ import akka.stream.{ActorAttributes, Attributes, Outlet, SourceShape}
 import akka.stream.stage.{GraphStage, GraphStageLogic, OutHandler}
 import java.util.{Enumeration => JEnumeration}
 
-private[jms] final class JmsBrowseStage(settings: JmsBrowseSettings) extends GraphStage[SourceShape[Message]] {
-  private val queue = settings.destination.getOrElse { throw new IllegalArgumentException("Destination is missing") }
-
+private[jms] final class JmsBrowseStage(settings: JmsBrowseSettings, queue: Destination)
```
andreas-schroeder: Destination is now checked on flow creation, and passed explicitly into the stages

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
```diff
@@ -60,20 +65,18 @@ private[jms] trait JmsConnector[S <: JmsSession] {
...
-
-    def failureHandler(ex: Throwable) =
-      if (withReconnect && ex.isInstanceOf[jms.JMSException]) initSessionAsync()
-      else fail.invoke(ex)
+  protected def initSessionAsync(): Unit = {
 
-    val allSessions = openSessions(failureHandler)
-    allSessions.failed.foreach(failureHandler)
+    val allSessions = openSessions()
+    allSessions.failed.foreach(connectionFailedCB.invoke)
```
andreas-schroeder: here, I think we must switch back to the graph stage threads with the asyncCallback

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
```diff
@@ -136,33 +138,20 @@ private[jms] trait JmsConnector[S <: JmsSession] {
...
-            connection.close() // best effort closing the connection.
-          } catch {
-            case _: Throwable =>
-          }
-          jmsSessions = Seq.empty
-
-          onConnectionFailure(ex)
+        override def onException(ex: jms.JMSException): Unit = {
+          Try(connection.close()) // best effort closing the connection.
+          connectionFailedCB.invoke(ex)
```
andreas-schroeder: here also, I think we must switch back to the graph stage threads with the asyncCallback. This however changed the amount of reconnects in the test.

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
```diff
@@ -176,18 +165,21 @@ private[jms] trait JmsConsumerConnector extends JmsConnector[JmsConsumerSession]
   protected def createSession(connection: jms.Connection,
                               createDestination: jms.Session => jms.Destination): JmsConsumerSession
 
-  override def openSessions(onConnectionFailure: jms.JMSException => Unit): Future[Seq[JmsConsumerSession]] =
-    openRecoverableConnection(startConnection = true, onConnectionFailure).flatMap { connection =>
-      val createDestination = jmsSettings.destination match {
-        case Some(destination) => destination.create
-        case _ => throw new IllegalArgumentException("Destination is missing")
-      }
```
andreas-schroeder: This check is now finally gone :)

jms/src/main/scala/akka/stream/alpakka/jms/JmsProducerStage.scala
```diff
@@ -72,26 +87,21 @@ private[jms] final class JmsProducerStage[A <: JmsMessage, PassThrough](settings
...
-                Future {
-                  jmsProducer.send(m.message)
-                  elem
-                }.onComplete(holder)(akka.dispatch.ExecutionContexts.sameThreadExecutionContext)
-              case p: PassThroughMessage[_, _] =>
-                val holder = new Holder[E](NotYetThere, futureCB, None)
-                inFlightMessagesWithProducer.enqueue(holder)
+                val holder = new Holder[E](NotYetThere)
+                inFlightMessages.enqueue(holder)
+                sendWithRetries(SendAttempt[E](m, holder))
```
andreas-schroeder: Extracted the send logic to a separate method that we can call, and storing the information we need for re-trying in the `SendAttempt` class

jms/src/main/scala/akka/stream/alpakka/jms/JmsProducerStage.scala
```diff
@@ -72,26 +87,21 @@ private[jms] final class JmsProducerStage[A <: JmsMessage, PassThrough](settings
...
+              case _: PassThroughMessage[_, _] =>
+                val holder = new Holder[E](NotYetThere)
+                inFlightMessages.enqueue(holder)
                 holder(Success(elem))
+                pushNextIfPossible()
```
andreas-schroeder: As we're already in the graph stage thread,  we can continue with handling the pass-through message directly. Also, the `holder.apply` method got stripped from the `futureCB`, as it got replaced with the `sendCompletedCB` async callback.

jms/src/main/scala/akka/stream/alpakka/jms/JmsProducerStage.scala
```diff
@@ -100,41 +110,82 @@ private[jms] final class JmsProducerStage[A <: JmsMessage, PassThrough](settings
...
+
+      private def sendWithRetries(send: SendAttempt[E]): Unit = {
+        import send._
+        val eventualSend = if (jmsProducers.nonEmpty) {
+          val jmsProducer: JmsMessageProducer = jmsProducers.dequeue()
+          Future(jmsProducer.send(message.message)).andThen {
+            case tried => sendCompletedCB.invoke((send, tried, jmsProducer))
+          }
+        } else {
+          nextTryOrFail(new RuntimeException("JmsProducer is not connected, skipping send attempt"), send)
```
andreas-schroeder: Potentially, the send attempt can happen while the producer is disconnected. Here, this is handled as a regular attempt that counts towards the max retries.

jms/src/main/scala/akka/stream/alpakka/jms/JmsProducerStage.scala
```diff
@@ -100,41 +110,82 @@ private[jms] final class JmsProducerStage[A <: JmsMessage, PassThrough](settings
...
+        outcome match {
+          case Success(_) =>
+            holder(Success(message))
+            pushNextIfPossible()
+          case Failure(t: JMSException) =>
+            nextTryOrFail(t, send)
+          case Failure(t) =>
+            holder(Failure(t))
+            handleFailure(t, holder)
+        }
```
andreas-schroeder: This here is the replacement for the previous `futureCB`.

jms/src/main/scala/akka/stream/alpakka/jms/JmsProducerStage.scala
```diff
@@ -100,41 +110,82 @@ private[jms] final class JmsProducerStage[A <: JmsMessage, PassThrough](settings
...
-          val holder = inFlightMessagesWithProducer.dequeue()
-          holder.jmsProducer.foreach(jmsProducers.enqueue) // put back jms producer to the pool.
+          val holder = inFlightMessages.dequeue()
           holder.elem match {
             case Success(elem) =>
               push(out, elem)
               pullIfNeeded() // Ask for the next element.
 
-            case Failure(NonFatal(ex)) => handleFailure(ex, holder)
+            case Failure(ex) => handleFailure(ex, holder)
```
andreas-schroeder: everything that is wrapped in a `Failure` is `NonFatal` (except you really try hard)

jms/src/main/scala/akka/stream/alpakka/jms/JmsProducerStage.scala
```diff
@@ -155,8 +206,7 @@ private[jms] object JmsProducerStage {
    *
    * To get a condensed view of what the Holder is about, have a look there too.
    */
-  class Holder[A](var elem: Try[A], val cb: AsyncCallback[Holder[A]], val jmsProducer: Option[JmsMessageProducer])
-      extends (Try[A] => Unit) {
+  class Holder[A](var elem: Try[A]) extends (Try[A] => Unit) {
```
andreas-schroeder: future callback is not needed, producer is handled outside of the message, so the holder got a lot simpler.

jms/src/main/scala/akka/stream/alpakka/jms/javadsl/JmsConsumer.scala
```diff
@@ -9,16 +9,15 @@ import javax.jms.Message
...
+import scala.collection.JavaConverters._
 
 object JmsConsumer {
 
   /**
    * Java API: Creates an [[JmsConsumer]] for [[javax.jms.Message]]
    */
   def create(settings: JmsConsumerSettings): akka.stream.javadsl.Source[Message, KillSwitch] =
-    akka.stream.javadsl.Source.fromGraph(new JmsConsumerStage(settings))
+    akka.stream.alpakka.jms.scaladsl.JmsConsumer.apply(settings).asJava
```
andreas-schroeder: Using the Scala code to avoid having to duplicate the `require`'s

jms/src/main/scala/akka/stream/alpakka/jms/scaladsl/JmsConsumer.scala
```diff
@@ -5,21 +5,22 @@
...
 object JmsConsumer {
 
   /**
    * Scala API: Creates an [[JmsConsumer]] for [[javax.jms.Message]] instances
    */
-  def apply(settings: JmsConsumerSettings): Source[Message, KillSwitch] =
-    Source.fromGraph(new JmsConsumerStage(settings))
+  def apply(settings: JmsConsumerSettings): Source[Message, KillSwitch] = {
+    require(settings.destination.isDefined, noConsumerDestination(settings))
+    Source.fromGraph(new JmsConsumerStage(settings, settings.destination.get))
```
andreas-schroeder: Now, every time a consumer or producer is created, the destination is checked and explicitly passed to the stages.

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsConnectorsSpec.scala
```diff
@@ -1180,6 +1181,49 @@ class JmsConnectorsSpec extends JmsSpec with MockitoSugar {
...
+      val restartTime = System.currentTimeMillis()
+      Thread.sleep(500)
+      cancellable.cancel()
+
+      val resultList = result.futureValue
+      def index(m: Map[String, Any]) = m("index").asInstanceOf[Long]
+      def time(m: Map[String, Any]) = m("time").asInstanceOf[Long]
+      resultList.filter(b => time(b) > restartTime) shouldNot be(empty)
+      resultList.sliding(2).forall(pair => index(pair.head) + 1 == index(pair.last)) shouldBe true
+    }
```
andreas-schroeder: This is for sure not the only test we need, more will follow.

andreas-schroeder:
One note: in this version the connection status callback logic of the consumer is racy (as the failing test luckily indicated). I've fixed that locally, and can push that tonight.

However, by now, I think that exposing the connection status via a materialized value and allow for polling that is the more straightforward solution.

ennru:
Yes, exposing the connection status needs to be something else than just futures as it may flip more than once. It could materialize a source streaming connect/disconnect events, but I guess that could be too much for most scenarios.
What use-case do you want to support with exposing the connection status?

ennru:
Yes, please revert to the safe-point you describe and continue on the re-connect and connectivity status in a separate PR.
I believe we need to discuss the "connectivity status" concept as such, we do not support it on any other connector today, but it might be relevant to more than just the JMS connector.

Regarding the API changes: we'll release Alpakka 1.0-M1 today which will contain the `flexiFlow` methods. But I'm not too concerned about changing their API in another milestone release.

andreas-schroeder:
@ennru revert is done, the Jms connectivity status feature is something I'll provide you as a separate issue.

andreas-schroeder:
* Fixed send retry test ( https://github.com/akka/alpakka/pull/1227/commits/4bf19f6487a67fae4ca75a31a201ed01fea8a422 who would have thought milliseconds are too coarse-grained 🙄)
* Added docs for retry settings
* Re-organised and improved docs (found and fixed two typos)

Which, finally, makes it ready for review.

andreas-schroeder:
by the way, this PR would also partially fix #1231 (and another part of the test flakiness would be fixed by #1252 , specifically commit https://github.com/akka/alpakka/commit/745a6ef7112c2a86abbbe46754ed9899e305a5b4 )

andreas-schroeder:
rebased on top of master

jms/src/main/scala/akka/stream/alpakka/jms/JmsProducerStage.scala
```diff
@@ -72,26 +83,21 @@ private[jms] final class JmsProducerStage[A <: JmsMessage, PassThrough](settings
...
         new InHandler {
 
-          override def onUpstreamFinish(): Unit = if (inFlightMessagesWithProducer.isEmpty) completeStage()
+          override def onUpstreamFinish(): Unit = if (inFlightMessages.isEmpty) completeStage()
 
           override def onPush(): Unit = {
             val elem: E = grab(in)
             elem match {
               case m: Message[_, _] =>
                 // fetch a jms producer from the pool, and create a holder object to capture the in-flight message.
```
2m: Comment got out of date.

andreas-schroeder:
@2m comment and test fixed.

2m:
LGTM, but I have not spent lots of time on this connector. @akara could you take a look at this PR as well?

jms/src/main/scala/akka/stream/alpakka/jms/Jms.scala
akara: Nice!

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
akara: Thanks for taking this out.

jms/src/main/scala/akka/stream/alpakka/jms/scaladsl/JmsConsumer.scala
akara: Hmm, `settings.destination.get` might tick off some static analysis tools as it is considered unsafe for the compiler. This is because the compiler does not check we have just called `settings.destination.isDefined` just the line above. It may be safer to use extraction (and the static analysis tools definitely like it better), as follows:

```scala
settings.destination match {
  case None => throw new IllegalArgumentException(noConsumerDestination(settings))
  case Some(dest) => Source.fromGraph(new JmsConsumerStage(settings, dest))
}
```

jms/src/main/scala/akka/stream/alpakka/jms/scaladsl/JmsConsumer.scala
```diff
@@ -65,8 +66,10 @@ object JmsConsumer {
    * @param settings The settings for the ack source.
    * @return Source for JMS messages in an AckEnvelope.
    */
-  def ackSource(settings: JmsConsumerSettings): Source[AckEnvelope, KillSwitch] =
-    Source.fromGraph(new JmsAckSourceStage(settings))
+  def ackSource(settings: JmsConsumerSettings): Source[AckEnvelope, KillSwitch] = {
+    require(settings.destination.isDefined, noConsumerDestination(settings))
+    Source.fromGraph(new JmsAckSourceStage(settings, settings.destination.get))
```
akara: Ditto here. Using match and extraction is a little safer. If you don't cover all cases, the compiler will complain.

jms/src/main/scala/akka/stream/alpakka/jms/scaladsl/JmsConsumer.scala
```diff
@@ -75,12 +78,26 @@ object JmsConsumer {
    * @param settings The settings for the tx source
    * @return Source of the JMS messages in a TxEnvelope
    */
-  def txSource(settings: JmsConsumerSettings): Source[TxEnvelope, KillSwitch] =
-    Source.fromGraph(new JmsTxSourceStage(settings))
+  def txSource(settings: JmsConsumerSettings): Source[TxEnvelope, KillSwitch] = {
+    require(settings.destination.isDefined, noConsumerDestination(settings))
+    Source.fromGraph(new JmsTxSourceStage(settings, settings.destination.get))
```
akara: And here too.

jms/src/main/scala/akka/stream/alpakka/jms/scaladsl/JmsConsumer.scala
```diff
@@ -75,12 +78,26 @@ object JmsConsumer {
...
+  }
 
   /**
    * Scala API: Creates a [[JmsConsumer]] for browsing messages non-destructively
    */
-  def browse(settings: JmsBrowseSettings): Source[Message, NotUsed] =
-    Source.fromGraph(new JmsBrowseStage(settings))
+  def browse(settings: JmsBrowseSettings): Source[Message, NotUsed] = {
+    require(settings.destination.isDefined, noBrowseDestination(settings))
+    Source.fromGraph(new JmsBrowseStage(settings, settings.destination.get))
```
akara: Ditto.

jms/src/test/scala/akka/stream/alpakka/jms/scaladsl/JmsConnectorsSpec.scala
```diff
@@ -873,42 +873,35 @@ class JmsConnectorsSpec extends JmsSpec with MockitoSugar {
...
+      when(producer.send(any[javax.jms.Destination], any[Message], anyInt(), anyInt(), anyLong()))
+        .thenAnswer(delayedSend)
 
       val in = (1 to 50).map(i => JmsTextMessage(i.toString))
       val jmsFlow = JmsProducer.flow[JmsTextMessage](JmsProducerSettings(factory).withQueue("test").withSessionCount(8))
 
       val result = Source(in).via(jmsFlow).toMat(Sink.seq)(Keep.right).run()
 
       result.futureValue shouldEqual in
+      delays.get() shouldBe 50
```
akara: `delays.get()` does not have any side-effect, right? Should it rather be spelled `delays.get shouldBe 50`?

jms/src/test/scala/akka/stream/alpakka/jms/JmsProducerRetrySpec.scala
```diff
@@ -0,0 +1,223 @@
...
+          .withQueue("test")
+          .withSendRetrySettings(
+            SendRetrySettings().withInitialRetry(10.millis).withMaxBackoff(10.millis).withMaxRetries(5)
+          )
+      )
+
+      val result = Source(List("one")).runWith(jms)
+
+      result.failed.futureValue shouldBe a[JMSException]
+      sendAttempts.get() shouldBe 6
```
akara: Similarly, `sendAttempts.get shouldBe 6`

jms/src/test/scala/akka/stream/alpakka/jms/JmsProducerRetrySpec.scala
```diff
@@ -0,0 +1,223 @@
...
+      val jms = JmsProducer.textSink(
+        JmsProducerSettings(factory)
+          .withQueue("test")
+          .withSendRetrySettings(SendRetrySettings().withMaxRetries(0))
+      )
+
+      val result = Source(List("one")).runWith(jms)
+
+      result.failed.futureValue shouldBe a[JMSException]
+      sendAttempts.get() shouldBe 1
```
akara: Ditto here.

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
```diff
@@ -136,33 +147,19 @@ private[jms] trait JmsConnector[S <: JmsSession] {
...
-          val delay = if (maxed) maxBackoff else waitTime(nextN)
-          if (delay >= maxBackoff) {
-            after(maxBackoff, system.scheduler) {
-              openConnectionWithRetry(startConnection, nextN, maxed = true)
-            }
-          } else {
-            after(delay, system.scheduler) {
-              openConnectionWithRetry(startConnection, nextN)
-            }
+          after(waitTime(nextN), system.scheduler) {
```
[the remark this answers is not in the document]
andreas-schroeder: The time is capped at `maxBackoff` in `waitTime`:
https://github.com/akka/alpakka/pull/1227/files#diff-01235c136811eb7422e859a2a1190b2fR112 and for the connect backoff, here: https://github.com/akka/alpakka/pull/1227/files#diff-01235c136811eb7422e859a2a1190b2fR90
Is this what you mean by rolling over?

jms/src/main/scala/akka/stream/alpakka/jms/JmsProducerStage.scala
```diff
@@ -100,41 +105,82 @@ private[jms] final class JmsProducerStage[A <: JmsMessage, PassThrough](settings
...
+          nextTryOrFail(send, RetrySkippedOnMissingConnection)
+        }
+      }
+
+      def nextTryOrFail(send: SendAttempt[E], ex: Throwable): Unit = {
+        import settings.sendRetrySettings._
+        import send._
+        if (maxRetries < 0 || attempt + 1 <= maxRetries) {
+          val nextAttempt = attempt + 1
+          scheduleOnce(send.copy(attempt = nextAttempt), waitTime(nextAttempt))
```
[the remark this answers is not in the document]
andreas-schroeder: yes, as I implemented it, I was also concerned with consistency. However, I found myself being inclined to use `TimerGraphStageLogic `, as I needed access the GraphStage state in the callback. Using `after` would have meant to use atomics for the producer pool (if I remember correctly, it has been a while).

jms/src/main/scala/akka/stream/alpakka/jms/JmsConnector.scala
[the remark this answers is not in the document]
andreas-schroeder: I got what you mean: as soon as the duration exceeds `Long.MaxValue` nanoseconds, multiplication of finite duration will fail with `java.lang.IllegalArgumentException`. `Math.pow(2, 64)` already exceeds `Long.MaxValue`, so we definitely need to take care of that. I'll fix this :+1:

andreas-schroeder:
Hi @2m, commit https://github.com/akka/alpakka/pull/1227/commits/c480a1b30e90d1ae2006515699af31ae97812367 now addresses @akara's last remaining point, and even the Travis-CI gods benevolently approved this last change. So I think this PR is good to go to master :)

2m:
Thank you everyone for great work! Pressing the big green button.

PR #1228, opened 2018-09-23
Google Cloud Pub/Sub: handle unsuccessful publish response codes

damianbl:
- instead of misleading `spray.json.DeserializationException: Object is missing required member 'messageIds'` return RuntimeException with the meaningful error message

### DIFF
--- a/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/HttpApi.scala
+++ b/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/HttpApi.scala
@@ -128,9 +128,13 @@ private trait HttpApi {
     for {
       request <- Marshal((HttpMethods.POST, url, request)).to[HttpRequest]
       response <- doRequest(request, maybeAccessToken)
-      publishResponse <- Unmarshal(response.entity).to[PublishResponse]
-    } yield publishResponse.messageIds
-  }
+    } yield {
+      response.status match {
+        case StatusCodes.OK => Unmarshal(response.entity).to[PublishResponse].map(_.messageIds)
+        case _ => throw new RuntimeException(s"Unexpected publish response. Code: [${response.status}]")
+      }
+    }
+  }.flatMap(identity)(materializer.executionContext)
 
   def getAccessToken(clientEmail: String, privateKey: PrivateKey, when: Instant)(
       implicit as: ActorSystem,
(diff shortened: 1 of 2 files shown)


google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/HttpApi.scala
```diff
@@ -128,9 +128,13 @@ private trait HttpApi {
     for {
       request <- Marshal((HttpMethods.POST, url, request)).to[HttpRequest]
       response <- doRequest(request, maybeAccessToken)
-      publishResponse <- Unmarshal(response.entity).to[PublishResponse]
-    } yield publishResponse.messageIds
-  }
+    } yield {
+      response.status match {
+        case StatusCodes.OK => Unmarshal(response.entity).to[PublishResponse].map(_.messageIds)
+        case _ => throw new RuntimeException(s"Unexpected publish response. Code: [${response.status}]")
```
ennru: Does the response contain anything else that might be useful? An error message possibly?
It might be good to introduce an exception class for this instead of using RTE.

google-cloud-pub-sub/src/test/scala/akka/stream/alpakka/googlecloud/pubsub/HttpApiSpec.scala
```diff
@@ -253,6 +253,45 @@ class HttpApiSpec extends FlatSpec with BeforeAndAfterAll with ScalaFutures with
...
+    )
+
+    val result =
+      TestHttpApi.publish(TestCredentials.projectId,
+                          "topic1",
+                          Some(accessToken),
+                          TestCredentials.apiKey,
+                          publishRequest)
+
+    assertThrows[RuntimeException] { result.futureValue }
```
ennru: It would be good to assert the exception more exact eg. by checking the message.

ennru:
Thank you for improving this!

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/HttpApi.scala
damianbl: Unfortunately it doesn't contain any more info. Just 404.

google-cloud-pub-sub/src/test/scala/akka/stream/alpakka/googlecloud/pubsub/HttpApiSpec.scala
damianbl: Agree. Let's improve it later.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/HttpApi.scala
damianbl: Yes, custom exception might be good. Let's improve it later.

PR #1230, opened 2018-09-24
add [openjdk11] to Travis build matrix

sullis:


2m:
We are using Travis CI stages which have a non trivial interaction with other properties defined at the top level of the `.travis.yml`, such as `scala`, `jdk` and others. As you can see from the generated job list in Travis CI page, only for the first job of the `test` stage the multiple `jdk` values were applied.

Therefore we wiil have to come up with another compact way of adding jdk11 testing.

sullis:
closing.

PR #1239, opened 2018-09-25
Print JUnit test names when they run

2m:


2m:
`-q` looks quite useful as well:

> Suppress stdout for successful tests.

Added that also.

PR #1242, opened 2018-09-26
Fix version warning hook element id

2m:


2m:
This is how it looks like:

2018-09-26-121754_2431x898_scrot

Issue #1231, opened 2018-09-24
Failed: JMS Connectors should sink disconnect exceptional completion


ennru:
https://travis-ci.org/akka/alpakka/jobs/432357452

```
[info] JmsConnectorsSpec:
[info] The JMS Connectors
[info] - should sink disconnect exceptional completion *** FAILED *** (90 milliseconds)
[info]   javax.jms.JMSException: Cannot send, channel has already failed: tcp://127.0.0.1:44374 was not an instance of akka.stream.alpakka.jms.ConnectionRetryException, but an instance of javax.jms.JMSException (JmsConnectorsSpec.scala:598)
```

ennru:
Again https://travis-ci.org/akka/alpakka/jobs/443305487

2m:
One more: https://travis-ci.org/akka/alpakka/jobs/446242926

ennru:
Fixed with #1252

PR #1252, opened 2018-09-30
Jms connection status

### LINKED ISSUE #1251: JMS: Expose connection status of consumer and producer
Now that the JMS producer and consumer stages perform retries (a potentially infinite amount of times), it is hard / impossible to know when the connectivity to the JMS broker is healthy and when it isn't. 

Before connection retries were implemented, it was possible to get the connection status (indirectly) from the stream's status:
- successfully created and running stream -> connection established and healthy
- stream completed with failure -> connection is unhealthy

In my setting, I used the JMS connectivity status as a part of the readiness check of the Kubernetes pod my app is running in (https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-readiness-probes). In this way, I can signal to Kubernetes that 
- when the pod is already running, the pod should not receive traffic from its load balancer.
- when the pod is a newly created one during a rolling deployment, this deployment needs to be halted until the pod becomes ready (which will never happen in case the JMS connectivity is mis-configured). This allows to prevent replacement of working pods by e.g. mis-configured ones.

So all in all, I think it is very helpful to know what the current connectivity status of the JMS consumer and producers is, and that it would be beneficial to expose it through the stage's materialized value.

andreas-schroeder:
This is my proposal for JMS connection status.

Fixes #1251


andreas-schroeder:
This PR extends #1227, don't merge this before the former is merged/approved.

andreas-schroeder:
Commit https://github.com/akka/alpakka/pull/1252/commits/745a6ef7112c2a86abbbe46754ed9899e305a5b4 fixes an issue in the test mentioned here: #1231 - However, a different error message.

andreas-schroeder:
rebased on top of master

2m:
I am wondering if exposing a boolean as a connectivity status is useful enough.

An alternative would be to expose a `Source[JmsConnectorState, ...]` instead. From that it would be possible for the user to create a stream that consumers state changes and updates an atomic reference if that is what is needed.

WDYT?

andreas-schroeder:
The boolean connectivity status would be enough for my use case, as I don't need immediate connectivity updates - having a delay of 1 second would be fine for me.

That said, I see having a stream of connectivity updates could also be beneficial if a fast response to connectivity changes is needed, so let's explore how this could look like.

How would you suggest expose the `Source`? As part of the materialized value, or by using a fan-out shape in the stage?

andreas-schroeder:
Hi @2m, I've spiked the change and one challenge is that the source needs to be pre-materialized, and that can only happen during `preStart`. Consequently, the source might not be there at the time the materialized value's `connected` method is called, and I either need to busy wait for the `preStart` to complete or return a `Future[Source[...]]`. I like to keep the API simple, but I guess that to stay in line with the non-blocking mantra of Akka, we need to go for the `Future` - what do you think?

2m:
How are you building the status source?

If it is, for example, a [`Source.queue`](https://github.com/akka/akka/blob/c89b4a188412b7184451e01bbf620d868f624bbe/akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala#L623) then you can [prematerialize](https://github.com/akka/akka/blob/c89b4a188412b7184451e01bbf620d868f624bbe/akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala#L92-L99) the source to get out the materialized value, which now can be passed into the graph stage.

Alternatively, to convert from `Future[Source[T, ...]]` to `Source[T, ...]` you can do:

```scala
val fs = Future[Source[T, ...]]
val s: Source[T, ...] = Source.fromFuture(fs).flatMapConcat(identity)
```

andreas-schroeder:
@2m here we go - I rebased on master and squashed everything into one commit. Have a look and tell me what you think.

I will add a couple more tests for connection status, and for correct handling of reconnection failures on each stage (creating connection, session, consumer/producer, and destination).

andreas-schroeder:
Here are some more tests. I was thinking to replace the status update `Stopping` with `Failed` and `Completed` respectively, but that is a bit more effort than just having `Stopping`. What do you think, @2m ?

andreas-schroeder:
So I went ahead and did the split of "stopping" into "failing" and "completing". I've also updated the docs and updated all factory methods in `JmsProducer` and `JmsConsumer`.

For the Java-API, this is a breaking change. 
For the Scala-API, this is a non-breaking change.

This PR is complete from my side, please have a look.

andreas-schroeder:
rebased once more on top of master.

andreas-schroeder:
Thanks @2m - also for your input, I think this PR wouldn't be in that shape without your ideas.

jms/src/main/scala/akka/stream/alpakka/jms/javadsl/package.scala
```diff
@@ -0,0 +1,28 @@
...
+import akka.stream.alpakka.jms.scaladsl.JmsConnectorState._
+import akka.stream.javadsl.Source
+
+package object javadsl {
+
+  @InternalApi private[javadsl] def transformConnectorState(
+      source: stream.scaladsl.Source[jms.scaladsl.JmsConnectorState, NotUsed]
+  ): Source[JmsConnectorState, NotUsed] =
+    source.map {
+      case Disconnected => JmsConnectorState.Disconnected
```
ennru: I guess the Scala states could contain the Java state and offer `asJava`.

andreas-schroeder: Ok, that introduce a dependency from the scaladsl to the javadsl, would that be fine for you?

ennru: Yes, even if the dependency goes the other direction for most cases.

andreas-schroeder: Implemented with https://github.com/akka/alpakka/pull/1252/commits/6a406b743c6209de04bd21e14985a1ea98e512e5

ennru:
Thank you for your continued efforts with the JMS connector.

andreas-schroeder:
Thank you for keeping everything nicely together! Really appreciate the efforts you and your team put into Alpakka.

Btw, this PR also fixed the test failures mentioned in issues #1031 and #1231.

sullis:
when is the next Milestone release?

PR #1256, opened 2018-10-02
add [openjdk11] to Travis build matrix

sullis:


ennru:
As we discussed 10 days ago in https://github.com/akka/alpakka/pull/1230, this approach does not work for the Alpakka build.

PR #1259, opened 2018-10-04
Google Pub/Sub: backoff after no messages fetched

### LINKED ISSUE #1254: PubSub subscriber source spins in tight loop, using 100% CPU
While testing the Google PubSub subscriber (non-GRPC version) using the [PubSub emulator](https://cloud.google.com/pubsub/docs/emulator), I noticed that both the emulator process and my subscriber process were each using 100% of one CPU core, although there was no data being published or read in PubSub.

I traced this back to [this line](https://github.com/akka/alpakka/blob/master/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/HttpApi.scala#L79) of the HttpApi class, where the `returnImmediately` parameter is set to true.

As per the [PubSub subscriber docs](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/pull), setting the `returnImmediately` parameter to true causes HTTP requests to return immediately if there is no data available (instead of blocking for a short time to wait for available data).

In the case of an idle PubSub topic, this seems to have the effect that the PubSub source simply spins in a tight loop, constantly polling the PubSub HTTP endpoint and not getting any data, thereby causing the 100% CPU usage.

This situation is obviously amplified when running with the PubSub emulator locally, as there is virtually no latency in requests (and conversely, when running locally against a "real" PubSub instance, this situation would be not at all visible due to network latency). However, it stands to reason that code running in GCP could also suffer from a similar situation due to the low latency of everything being co-located in GCP.

I've done a local test where I patched Alpakka to set the `returnImmediately` flag to `false`, and it appears to immediately resolve the issue (while not introducing any extra noticeable latency). 

Either setting `returnImmediately` to `false` instead of `true`, or making this an option (somewhat similar to how the [`pollInterval` is a parameter in the GRPC-based implementation](https://github.com/akka/alpakka/blob/master/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/scaladsl/GooglePubSub.scala#L46)) would resolve this issue.

gabrielreid:
Fixes #1254.

Adds a returnImmediately flag to the creation of a PubSub Source,
with the default still set to true for backwards compatibility.

### DIFF
--- a/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSource.scala
+++ b/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSource.scala
@@ -4,14 +4,16 @@
 
 package akka.stream.alpakka.googlecloud.pubsub
 
+import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.stream.{Attributes, Materializer, Outlet, SourceShape}
-import akka.stream.stage.{GraphStage, GraphStageLogic, OutHandler}
+import akka.stream.stage.{GraphStage, GraphStageLogic, OutHandler, TimerGraphStageLogic}
 
 import scala.util.{Failure, Success, Try}
 import GooglePubSubSource._
 
 import scala.collection.immutable
+import scala.concurrent.duration._
 
 @akka.annotation.InternalApi
 private final class GooglePubSubSource(projectId: String,
@@ -25,7 +27,7 @@ private final class GooglePubSubSource(projectId: String,
   override val shape: SourceShape[ReceivedMessage] = SourceShape(out)
 
   override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
-    new GraphStageLogic(shape) {
+    new TimerGraphStageLogic(shape) {
       private var state: State = Pending
 
       def fetch(implicit mat: Materializer): Unit = {
@@ -50,6 +52,9 @@ private final class GooglePubSubSource(projectId: String,
         state = Fetching
       }
 
+      override def onTimer(timerKey: Any): Unit =
+        fetch(materializer)
+
       private val callback = getAsyncCallback[(Try[PullResponse], Materializer)] {
         case (Failure(tr), _) =>
           failStage(tr)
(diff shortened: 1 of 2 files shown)


gabrielreid:
Also, small FYI, I've signed an ICLA

2m:
Thanks for the PR. However this now means that `httpApi.pull` will now block a thread if there are no messages immediately available. While this will solve the issue described in #1254 it introduces another one, which is blocking one thread out of the thread pool that is being used to run the whole stream.

Instead, we should keep `returnImmediately` set to `true`, but instead of calling `fetch` immediately when no messages are returned, the stage should set a timer to fire after a second or so. And on that timer event it should call `fetch` again:

https://github.com/akka/alpakka/blob/7404ed625228f46a48b8be1e7f4ed512662c91c4/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSource.scala#L67-L69

Could you change this PR to this?

gabrielreid:
@2m really sorry for dropping the ball on this. I've updated the PR as suggested, and updated the test case to specifically trigger this situation.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/GooglePubSubSource.scala
```diff
@@ -66,7 +67,7 @@ private final class GooglePubSubSource(projectId: String,
               }
             case Nil =>
               state = Fetching
-              fetch(mat)
+              mat.scheduleOnce(1.second, () => fetch(mat))
```
2m: This will schedule `fetch` to be ran on the execution context that is different from the graph stage context. Instead it should use `scheduleOnce` from the `TimerGraphStageLogic` ([more info in the docs here](https://doc.akka.io/docs/akka/current/stream/stream-customize.html#using-timers)).

It is important because inside of `fetch` it updates a mutable `state` variable. And we need to make sure that all accesses to that variable are done from the graph stage context.

gabrielreid: Thanks for taking the time to point me in the right direction -- greatly appreciated!

I've updated the PR to make use of TimerGraphStageLogic as suggested.

PR #1267, opened 2018-10-09
AWS SQS/SNS: Support for per message destinations

### LINKED ISSUE #1261: SNS/SQS per message destination
When we're dealing with a flow that sends messages to multiple destinations, we're forced to use Partition inside a GraphDSL to choose the right flow, when it could be as easy as:

```scala
lazy val snsFlow = SnsPublisher.generic() //it does not expect a topic arn

Source
  .single(new PublishRequest().withTopicArn(...)) //because it'll be decided by each element
  .via(snsFlow)
  //(...)
``` 

And the same logic can be applied to SQS as well.

gabfssilva:
Fixes #1261

### DIFF
--- a/sns/src/main/scala/akka/stream/alpakka/sns/impl/SnsPublishFlowStage.scala
+++ b/sns/src/main/scala/akka/stream/alpakka/sns/impl/SnsPublishFlowStage.scala
@@ -15,7 +15,7 @@ import com.amazonaws.services.sns.model.{PublishRequest, PublishResult}
  * INTERNAL API
  */
 @InternalApi
-private[sns] final class SnsPublishFlowStage(topicArn: String, snsClient: AmazonSNSAsync)
+private[sns] final class SnsPublishFlowStage(snsClient: AmazonSNSAsync)
     extends GraphStage[FlowShape[PublishRequest, PublishResult]] {
 
   private val in = Inlet[PublishRequest]("SnsPublishFlow.in")
@@ -48,7 +48,7 @@ private[sns] final class SnsPublishFlowStage(topicArn: String, snsClient: Amazon
 
       override def onPush(): Unit = {
         isMessageInFlight = true
-        val request = grab(in).withTopicArn(topicArn)
+        val request = grab(in)
         snsClient.publishAsync(request, asyncHandler)
       }
 
--- a/sns/src/main/scala/akka/stream/alpakka/sns/javadsl/SnsPublisher.scala
+++ b/sns/src/main/scala/akka/stream/alpakka/sns/javadsl/SnsPublisher.scala
@@ -33,7 +33,17 @@ object SnsPublisher {
    * creates a [[akka.stream.javadsl.Flow Flow]] to publish messages to a SNS topic using an [[com.amazonaws.services.sns.AmazonSNSAsync AmazonSNSAsync]]
    */
   def createPublishFlow(topicArn: String, snsClient: AmazonSNSAsync): Flow[PublishRequest, PublishResult, NotUsed] =
-    Flow.fromGraph(new SnsPublishFlowStage(topicArn, snsClient))
+    Flow
+      .fromFunction(new function.Function[PublishRequest, PublishRequest] {
+        override def apply(req: PublishRequest): PublishRequest = req.withTopicArn(topicArn)
+      })
+      .via(new SnsPublishFlowStage(snsClient))
+
+  /**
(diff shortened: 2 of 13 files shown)


sns/src/main/scala/akka/stream/alpakka/sns/impl/SnsPublishFlowStage.scala
```diff
@@ -48,7 +48,12 @@ private[sns] final class SnsPublishFlowStage(topicArn: String, snsClient: Amazon
 
       override def onPush(): Unit = {
         isMessageInFlight = true
-        val request = grab(in).withTopicArn(topicArn)
+        val request =
+          topicArn match {
+            case None => grab(in)
```
ennru: I wonder if it should fail for messages without `topicArn`?
The logic in Alpakka JMS is to have the default destination specified and use that in case a message does not have a per message destination.

sns/src/main/scala/akka/stream/alpakka/sns/javadsl/SnsPublisher.scala
```diff
@@ -33,7 +33,13 @@ object SnsPublisher {
    * creates a [[akka.stream.javadsl.Flow Flow]] to publish messages to a SNS topic using an [[com.amazonaws.services.sns.AmazonSNSAsync AmazonSNSAsync]]
    */
   def createPublishFlow(topicArn: String, snsClient: AmazonSNSAsync): Flow[PublishRequest, PublishResult, NotUsed] =
-    Flow.fromGraph(new SnsPublishFlowStage(topicArn, snsClient))
+    Flow.fromGraph(new SnsPublishFlowStage(Some(topicArn), snsClient))
+
+  /**
+   * creates a [[akka.stream.javadsl.Flow Flow]] to publish messages to a SNS topic using an [[com.amazonaws.services.sns.AmazonSNSAsync AmazonSNSAsync]]
```
ennru: Update the Scaladoc to reflect the difference between the two factory methods, please.

sns/src/main/scala/akka/stream/alpakka/sns/javadsl/SnsPublisher.scala
```diff
@@ -47,4 +53,9 @@ object SnsPublisher {
   def createPublishSink(topicArn: String, snsClient: AmazonSNSAsync): Sink[PublishRequest, CompletionStage[Done]] =
     createPublishFlow(topicArn, snsClient).toMat(Sink.ignore(), Keep.right[NotUsed, CompletionStage[Done]])
 
+  /**
+   * creates a [[akka.stream.javadsl.Sink Sink]] to publish messages to a SNS topic using an [[com.amazonaws.services.sns.AmazonSNSAsync AmazonSNSAsync]]
```
ennru: Same here.

sqs/src/main/scala/akka/stream/alpakka/sqs/impl/SqsFlowStage.scala
```diff
@@ -86,7 +86,11 @@ import scala.util.{Failure, Success, Try}
 
           override def onPush() = {
             inFlight += 1
-            val msg = grab(in).withQueueUrl(queueUrl)
+            val msg = queueUrl match {
+              case Some(url) => grab(in).withQueueUrl(url)
```
ennru: Same discussion as for SNS.

sns/src/main/scala/akka/stream/alpakka/sns/impl/SnsPublishFlowStage.scala
gabfssilva: Don't you think it is kind of confusing? I think we could validate if the PublishRequest/SendMessageRequest does not have a destionation, but, not sure if it'd be a good idea to use just one flow for both features.
Also, I thought we could rename the flows, fixedTopicFlow/fixedTopicSink and fixedQueueFlow/fixedQueueSink are good names for the old flows and sinks.

ennru: Looking at the existing code, I believe you can just push the setting of the `topicArn` to a `map` in the factory methods so the stages wouldn't know anything about destinations.

gabfssilva:
Not sure what you think, but, I tried to create a pattern between the SNS and SQS sinks and flows docs and also fixed some that were wrong. =P

PR #1170, opened 2018-08-27
Remove deprecated methods for 1.0

juanjoDiaz:
Deprecated methods were piling up. I know it's never nice for the user to remove them. However, I think that a 1.0 release should be a clean release without deprecated methods.

Wdyt?

PR #1268, opened 2018-10-09
Set correct since version for deprecations

ennru:
Some deprecations were created before we decided to call it version 1.0-M1.


juanjoDiaz:
I really think that all deprecation should be removed before the final 1.0 release as I proposed on #1170.
So it's a "clean" start.

Do you guys agree? Should I maintain #1170 up-to-date with master? or should I just close it?

2m:
We surely should remove deprecations introduced in older versions. But I am not sure about the deprecations introduced in `v0.20`. If we remove those methods, we only leave milestone versions for the users to migrate from old deprecated methods to the new ones. Which also might be just fine.

PR #1273, opened 2018-10-16
Disable AMQP automatic recovery by default

### LINKED ISSUE #1034: AMQP: Support auto-recovery
The RabbitMQ client library supports adding recovery listeners that can support recovery from broker or connection failures.
@JonathanKnight tried to use this in #332 but seems to have abandoned the PR.

### LINKED ISSUE #1270: AMQP: recovery problem after RabbitMQ restart
I have designed simple consumer which works fine for me. So I can put some messages into queue using RabbitMQ management UI and see corresponding log records "AMPQ message is consumed: [message]" in my service console output.
But when I simulate networking (connection error) issue by restarting RabbitMQ things became worse. Stream just dies silently and no exception is throw by AMPQ Source. 
Then when I try to put new messages in RabbitMQ management UI they are not consumed anymore because stream has died.
As the solution  I have tried to use RestartSource.withBackoff, but new issue appears. Source was restarted, but after this every 2nd message is not consumed, some messages just disappear. Moreover  I have noticed 2 active connections were shown in RabbitMQ Management UI instead of single connection which was displayed before RabbitMQ restart. So looks like old connection was not closed and some messages disappeared there.  I have tied to use .withAutomaticRecoveryEnabled(true) feature, but nothing changes. 
Guys, please see my consumer code 
[AMPQClientConsumer.scala.txt](https://github.com/akka/alpakka/files/2464926/AMPQClientConsumer.scala.txt)

Please advice.

juanjoDiaz:
Closes #1270
Closes #1034

### DIFF
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
@@ -67,8 +67,8 @@ final class AmqpDetailsConnectionProvider private (
     val handshakeTimeout: Option[Int] = None,
     val shutdownTimeout: Option[Int] = None,
     val networkRecoveryInterval: Option[Int] = None,
-    val automaticRecoveryEnabled: Option[Boolean] = None,
-    val topologyRecoveryEnabled: Option[Boolean] = None,
+    val automaticRecoveryEnabled: Boolean = false,
+    val topologyRecoveryEnabled: Boolean = false,
     val exceptionHandler: Option[ExceptionHandler] = None,
     val connectionName: Option[String] = None
 ) extends AmqpConnectionProvider {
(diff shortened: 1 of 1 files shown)


amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -67,8 +67,8 @@ final class AmqpDetailsConnectionProvider private (
     val handshakeTimeout: Option[Int] = None,
     val shutdownTimeout: Option[Int] = None,
     val networkRecoveryInterval: Option[Int] = None,
-    val automaticRecoveryEnabled: Option[Boolean] = None,
-    val topologyRecoveryEnabled: Option[Boolean] = None,
+    val automaticRecoveryEnabled: Option[Boolean] = Some(false),
+    val topologyRecoveryEnabled: Option[Boolean] = Some(false),
```
ennru: Shouldn't the Options be dropped in favour of flat Booleans?

juanjoDiaz: I left the Option just for consistency since all the other parameters are optional.
But I have no strong opinion on the matter.

Should I change them to plain Boolean?

ennru: Yes, please.

ennru:
Thank you!

pawelkaczor:
> Closes #1270 Closes #1034

@ennru 

According to https://doc.akka.io/docs/alpakka/current/amqp.html#connecting-to-server

> Warning
Please be aware that the basic usage of AmqpConnectionProvider like this AmqpUriConnectionProvider(s"amqp://$host:$port") has an issue with recovering connections, more details can be found in this issue

only AmqpUriConnectionProvider is affected. If so, why do you disable automatic recovery in AmqpDetailsConnectionProvider ?

PR #1275, opened 2018-10-17
AWS S3: Optional wrapping for ObjectMetadata in client.download()

### LINKED ISSUE #1165: AWS S3: Optional wrapping missing for ObjectMetadata in client.download() while it is present in client.getObjectMetadata()
When downloading objects from S3, [alpakka returns a source](https://github.com/akka/alpakka/blob/master/s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3Client.scala#L497) of the object's contents along with a future of the ObjectMetadata. If the object doesn't exist, the ObjectMetadata is still constructed and doesn't have anything discernible features that would indicate that the object didn't exist.

On the other hand, we can look up an ObjectMetaData through a [different part of the API](https://github.com/akka/alpakka/blob/master/s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3Client.scala#L245) and know if the object exists or not as we get an empty option if it does not.

jtjeferreira:
fixes #1165


s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -90,20 +90,23 @@ private[alpakka] final class S3Stream(settings: S3Settings)(implicit system: Act
...
-    val meta = future.map {
-      case (entity, headers) ⇒ computeMetaData(headers, entity)
-    }
-    (source, meta)
+
+    future
+      .flatMap {
+        case (entity, headers) =>
+          Future.successful(
+            Some((entity.dataBytes.mapMaterializedValue(_ => NotUsed), computeMetaData(headers, entity)))
```
2m: This could be `.map` instead of a `.flatMap` and then it would not need additional wrapping with `Future.successful`.

s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3Client.scala
```diff
@@ -481,11 +481,11 @@ final class S3Client(s3Settings: S3Settings, system: ActorSystem, mat: Materiali
               sse)
 
   private def toJava[M](
-      download: (akka.stream.scaladsl.Source[ByteString, M], Future[scaladsl.ObjectMetadata])
-  ): JPair[Source[ByteString, M], CompletionStage[ObjectMetadata]] = {
-    val (stream, meta) = download
-    JPair(stream.asJava, meta.map(metaDataToJava)(mat.executionContext).toJava)
-  }
+      download: Future[Option[(akka.stream.scaladsl.Source[ByteString, M], scaladsl.ObjectMetadata)]]
+  ): CompletionStage[Option[JPair[Source[ByteString, M], ObjectMetadata]]] =
```
ennru: That's a fair bit of unwrapping. The `Option` would need conversion, as well, though.

jtjeferreira: nice catch. Should be fixed now...

s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3SourceSpec.scala
```diff
@@ -36,7 +36,7 @@ class S3SourceSpec extends S3WireMockBase with S3ClientIntegrationSpec {
     mockDownload()
 
     //#download
-    val (s3Source: Source[ByteString, _], _) = s3Client.download(bucket, bucketKey)
+    val Some((s3Source: Source[ByteString, _], _)) = s3Client.download(bucket, bucketKey).futureValue
```
ennru: This code is part of the documentation and will look rather strange there. I guess this needs to be rewritten to show how to use the changed API.
(applies to the other snippets, as well)

jtjeferreira: Hi! I tried to make this snippets consistent with the `getObjectMetadata` snippets. Can you have a look?

PR #1287, opened 2018-10-25
AWS SQS: Source - Allow configuring visibility timeout

barakbaronk:
Hi, first time contributing here, so hopefully I didn't miss anything from the instructions.
In addition, I signed the CLA

**Feature**: 
Allow setting the visibility timeout parameter in the SQS `Receive` request.
this means that in each `Receive` request it will override the default visibility timeout from SQS queue configuration. if one will not set it, the default will be taken.

The requirement:
if one knows the maximum time it takes to process a message and want to set it in the application (and not in the SQS queue console) this change will reduce the number of calls. So, instead of 2 calls - `Receive` and `ChangeVisibilityTimeout`, we will just run the `Receive`

We, At the company I'm working for, started a code migration from our old SQS client to Alpakka.
during the migration we found out this issue, which required us to run 2 consecutive calls - receive the message and change the visibility timeout. I think it's an overhead we can reduce.

I added tests dedicate for this feature
Finally, I looked at the issue tracker and PRs and didn't see any comment about this feature.

### DIFF
--- a/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceSettings.scala
+++ b/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceSettings.scala
@@ -17,7 +17,7 @@ final class SqsSourceSettings private (
     val attributeNames: immutable.Seq[AttributeName],
     val messageAttributeNames: immutable.Seq[MessageAttributeName],
     val closeOnEmptyReceive: Boolean,
-    val visibilityTimeout: Option[Int]
+    val visibilityTimeout: Option[FiniteDuration]
 ) {
   require(maxBatchSize <= maxBufferSize, "maxBatchSize must be lower or equal than maxBufferSize")
   // SQS requirements
@@ -96,7 +96,7 @@ final class SqsSourceSettings private (
    *
    * Default: None - taken from the SQS queue configuration
    */
-  def withVisibilityTimeout(timeout: Int): SqsSourceSettings =
+  def withVisibilityTimeout(timeout: FiniteDuration): SqsSourceSettings =
     copy(visibilityTimeout = Some(timeout))
 
   private def copy(
@@ -106,7 +106,7 @@ final class SqsSourceSettings private (
       attributeNames: immutable.Seq[AttributeName] = attributeNames,
       messageAttributeNames: immutable.Seq[MessageAttributeName] = messageAttributeNames,
       closeOnEmptyReceive: Boolean = closeOnEmptyReceive,
-      visibilityTimeout: Option[Int] = visibilityTimeout
+      visibilityTimeout: Option[FiniteDuration] = visibilityTimeout
   ): SqsSourceSettings = new SqsSourceSettings(
     waitTimeSeconds,
     maxBufferSize,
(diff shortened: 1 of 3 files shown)


sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceSettings.scala
```diff
@@ -16,7 +16,8 @@ final class SqsSourceSettings private (
     val maxBatchSize: Int,
     val attributeNames: immutable.Seq[AttributeName],
     val messageAttributeNames: immutable.Seq[MessageAttributeName],
-    val closeOnEmptyReceive: Boolean
+    val closeOnEmptyReceive: Boolean,
+    val visibilityTimeout: Option[Int]
```
ennru: I'd rather see this as a `FiniteDuration`. Possibly representing `None` with `Duration.Zero` instead.

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceSettings.scala
```diff
@@ -113,6 +125,7 @@ final class SqsSourceSettings private (
     s"attributeNames=${attributeNames.mkString(",")}, " +
     s"messageAttributeNames=${messageAttributeNames.mkString(",")}, " +
     s"closeOnEmptyReceive=$closeOnEmptyReceive" +
```
ennru:
```suggestion
     s"closeOnEmptyReceive=$closeOnEmptyReceive," +
```

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceSettings.scala
```diff
@@ -113,6 +125,7 @@ final class SqsSourceSettings private (
...
+    s"visibilityTomeout=$visibilityTimeout" +
```
ennru:
```suggestion
    s"visibilityTimeout=${visibilityTimeout.toCoarsest}" +
```

sqs/src/main/scala/akka/stream/alpakka/sqs/impl/SqsSourceStage.scala
```diff
@@ -116,3 +119,11 @@ import scala.collection.JavaConverters._
       )
     }
 }
+
+object SqsSourceStage {
+
+  private implicit class ReceiveMessageExtension(request: ReceiveMessageRequest) {
```
ennru: I'd like to be it explicit when constructing the `ReceiveMessageRequest` instead.

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsSourceSettings.scala
barakbaronk: it might be a problem setting the default to `Zero`, since it's a valid value (see [doc](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html#terminating-message-visibility-timeout))
so if one does not want to use this feature, we won't override the default value in SQS queue console.

barakbaronk:
@ennru thanks!
fixed your comment. I left the settings parameter as `Option` since zero finite duration is a valid value.
I also removed the implicit class, but I had to change the receive request variable to `var`

ennru:
Great that you reached out to improving Alpakka SQS - I hope your migration towards streaming integrations works well.

PR #1290, opened 2018-10-26
use 'create=false' for ActiveMQ 'vm' transport uri

sullis:


sullis:
ActiveMQ docs: http://activemq.apache.org/vm-transport-reference.html

ennru:
What does this mean? Why is it useful?

PR #1293, opened 2018-10-26
Express the carry type

huntc:
This ensures that the client code usage is nicer, and tighter, when working with the carry object. The Java API is more affected that the Scala API. The Scala API is hardly affected at all, and where it is requires substituting `Command[_]` and `Event[_]` with `Command[Nothing]` and `Event[Nothing]` respectively (unless you have a carry, where then you provide its type instead of `Nothing`).

There's a problem here though, with Java and its non-ability to express parameterised classes given type erasure etc. etc. i.e.:

```java
Pair<
        SourceQueueWithComplete<Command<Object>>,
        Source<DecodeErrorOrEvent<Object>, NotUsed>>
    run =
        Source.<Command<Object>>queue(2, OverflowStrategy.dropHead())
            .via(mqttFlow)
            .toMat(BroadcastHub.of(DecodeErrorOrEvent.class), Keep.both())
            .run(materializer);
```

...is a problem given that I'm unable to create the `BroadcastHub.of` that I want. How do we work around this in Java? Ideally, you'd be able to state, `BroadcastHub.of(DecodeErrorOrEvent<Object>.class)`, but you can't - and I understand why. So, how do we cater for this?

Incidentally, [someone else had exactly this issue with Java Akka streams](https://groups.google.com/forum/#!searchin/akka-user/class$20cast%7Csort:date/akka-user/DVpN4OUCEjc/M9JLSyUlCgAJ).

I've got a feeling that I've worked around this before, and there was even an Akka utility to help me, I just don't recall where/what right now... Any help is appreciated. Thanks!

### DIFF
--- a/mqtt-streaming/src/test/java/docs/javadsl/MqttFlowTest.java
+++ b/mqtt-streaming/src/test/java/docs/javadsl/MqttFlowTest.java
@@ -91,26 +91,28 @@ public class MqttFlowTest {
     Flow<ByteString, ByteString, CompletionStage<Tcp.OutgoingConnection>> connection =
         Tcp.get(system).outgoingConnection("localhost", 1883);
 
-    Flow<Command<?>, DecodeErrorOrEvent, NotUsed> mqttFlow =
+    Flow<Command<Object>, DecodeErrorOrEvent<Object>, NotUsed> mqttFlow =
         Mqtt.clientSessionFlow(session).join(connection);
     // #create-streaming-flow
 
     // #run-streaming-flow
-    Pair<SourceQueueWithComplete<Command<?>>, CompletionStage<DecodeErrorOrEvent>> run =
-        Source.<Command<?>>queue(3, OverflowStrategy.fail())
-            .via(mqttFlow)
-            .drop(3)
-            .toMat(Sink.head(), Keep.both())
-            .run(materializer);
+    Pair<SourceQueueWithComplete<Command<Object>>, CompletionStage<DecodeErrorOrEvent<Object>>>
+        run =
+            Source.<Command<Object>>queue(3, OverflowStrategy.fail())
+                .via(mqttFlow)
+                .drop(3)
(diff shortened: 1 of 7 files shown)


ennru:
We recently got this contribution to the docs: [Pass through](https://developer.lightbend.com/docs/alpakka/current/patterns.html#passthrough)

huntc:
Thanks @ennru. I’ve long wanted to see Akka Streams itself provide for this scenario. Our situation though is that the MQTT connector already caters for carrying these elements.

Any advice on how to use BroadcastHub with generic types from Java?

huntc:
I should add that the PassThroughFlow example wouldn’t work for us because with have two flows: a command flow and an event flow. For example a publish command needs to provide a carry that is emitted with its correlating publish ack event. As well as there being two flows, there’s not guaranteed order in which event flow elements are received or you may get a publish ack for a later message and then get one for an earlier one, or it may be that a ping req is received before you get a publish ack etc.

huntc:
@ennru I've now fixed the Java API and updated the description of the PR accordingly.

mqtt-streaming/src/test/java/docs/javadsl/MqttFlowTest.java
```diff
@@ -149,7 +149,7 @@ public void establishServerBidirectionalConnectionAndSubscribeToATopic()
                       run =
                           Source.<Command<Object>>queue(2, OverflowStrategy.dropHead())
                               .via(mqttFlow)
-                              .toMat(BroadcastHub.of(DecodeErrorOrEvent.class), Keep.both())
+                              .toMat(BroadcastHub.of(DecodeErrorOrEvent.classOf()), Keep.both())
```
dwijnand: JFYI `BroadcastHub.of` doesn't even use the class value.  It's only there to help define the type parameter.  So you can actually pass `null` and let type inference define type parameter.  😄

PR #1297, opened 2018-10-29
Unix Domain Socket - Bump jnr-enxio and remove local fix

longshorej:
@huntc's https://github.com/jnr/jnr-enxio/pull/28 was merged and released, so we can remove now bump dependencies and remove the fix that was copied over here.


ennru:
Travis is not happy
```
[info] docs.scaladsl.UnixDomainSocketSpec *** ABORTED *** (276 milliseconds)
[info]   java.lang.NoClassDefFoundError: Could not initialize class jnr.enxio.channels.Native$SingletonHolder
[info]   at jnr.enxio.channels.Native.libc(Native.java:70)
```

```
[error] Test docs.javadsl.UnixDomainSocketTest.aUnixDomainSocketShouldReceiveWhatIsSent failed: java.lang.UnsatisfiedLinkError: could not load FFI provider jnr.ffi.provider.jffi.Provider, took 0.306 sec
[error]     at jnr.ffi.provider.InvalidProvider$1.loadLibrary(InvalidProvider.java:48)
```

See https://travis-ci.org/akka/alpakka/jobs/447906794

huntc:
@longshorej Are you seeing this error locally also?

2m:
Notices the same failure in other PR validation jobs as well, so created a ticket to track that: https://github.com/akka/alpakka/issues/1377

huntc:
@longshorej I rebased and pushed to your original branch. I've at least got the dependencies working sweetly.

huntc:
@ennru This PR is now (finally) ready

huntc:
I've just added another commit to this, which includes:

* fixed a bug where `expand` was being used where `mapConcat` should have been
* provided some useful debugging statements
* cleaned up some tests
* included a new test that presently fails (will raise a separate issue)

huntc:
CI is failing, but not because of this PR or the sub-project (actually mqtt-streaming... but I'll look at that again separately). My view is that this PR is ready. :-)

project/Dependencies.scala
```diff
@@ -383,7 +383,8 @@ object Dependencies {
 
   val UnixDomainSocket = Seq(
     libraryDependencies ++= Seq(
-      "com.github.jnr" % "jnr-unixsocket" % "0.19" // BSD/ApacheV2/CPL/MIT as per https://github.com/akka/alpakka/issues/620#issuecomment-348727265
+      "com.github.jnr" % "jffi" % "1.2.17" classifier "complete", // ApacheV2
```
huntc: This took me a very long time to work out - you need the `complete` jar given the way sbt/ivy downloads artifacts.

unix-domain-socket/src/main/scala/akka/stream/alpakka/unixdomainsocket/impl/UnixDomainSocketImpl.scala
```diff
@@ -251,9 +291,9 @@ private[unixdomainsocket] object UnixDomainSocketImpl {
 
     val sendSink = Sink.fromGraph(
       Flow[ByteString]
-        .expand { bytes =>
+        .mapConcat { bytes =>
```
huntc: This was a very silly mistake on my part. Fixed. :-)

unix-domain-socket/src/test/scala/docs/scaladsl/UnixDomainSocketSpec.scala
```diff
@@ -31,52 +31,71 @@ class UnixDomainSocketSpec
...
-        Files.createTempFile("UnixDomainSocketSpec1", ".sock").toFile
-      file.delete()
-      file.deleteOnExit()
+        dir.resolve("sock1").toFile
+
+      val received = Promise[ByteString]
+
+      val serverSideFlow = Flow[ByteString]
+        .buffer(1, OverflowStrategy.backpressure)
+        .wireTap(bytes => received.success(bytes))
```
huntc: The details of the flow itself are unimportant to the doc reader. I also changed the nature of the flow here so that it will not close downstream if there's still something to be sent.

unix-domain-socket/src/test/scala/docs/scaladsl/UnixDomainSocketSpec.scala
```diff
@@ -31,52 +31,71 @@ class UnixDomainSocketSpec
...
+          .runWith(Sink.ignore)
+        //#outgoingConnection
+        received.future.map(receiveBytes => assert(receiveBytes == sendBytes))
+        //#outgoingConnection
+      }
+      //#outgoingConnection
+    }
+
+    "send and receive more ten times the size of a buffer" ignore {
+      val BufferSizeBytes = 64 * 1024
```
huntc: This test presently fails so it is an area worth further investigation. Already discussed by #1349.

unix-domain-socket/src/test/scala/docs/scaladsl/UnixDomainSocketSpec.scala
```diff
@@ -31,52 +31,71 @@ class UnixDomainSocketSpec
 
   implicit val ma: ActorMaterializer = ActorMaterializer()
 
+  private val dir = Files.createTempDirectory("UnixDomainSocketSpec")
```
huntc: This leads to much less ceremony when creating sockets.

2m:
I see a `NullPointerException` and stalled test run in the Travis logs and I also get that when running Unix Domain Socket connector tests myself on the code from this PR.

Tests from current master run fine here. So there is still something in the PR that throws `NullPointerException` and stalls the tests. It seems that the exception is being thrown just after Java tests and Scala tests never startup.

huntc:
There was an NPE in the test run after I wrote my comment. But then I pushed another commit to fix it. I’ll look again. Cheers.

huntc:
OK, I pushed a commit to fix that NPE. Darn those nulls.

2m:
It still NPEs in Travis. Although not every time. We have added Scala 2.13 support recently so now it runs tests not two but three times (for every Scala versions). And that is enough tries to make it fail on some of the run.

huntc:
Hmm. I think that particular test may be a bit flaky. I think I'm going to have to drop JNR to move forward further.

huntc:
Actually it looks like an async close error on binding. I shall investigate further.

huntc:
Fingers crossed - no guarding against some IOExceptions - darn those exceptions. And to think, I used to think that exceptions were cool... :-P

huntc:
OK, UDS appears to now build ok. The two build failures are unassociated.

huntc:
@ennru @2m This is ready for a final review from my perspective. :-)

PR #1302, opened 2018-11-01
Elasticsearch: failStage should happen when shouldRetry returns false

### LINKED ISSUE #1300: Elasticsearch: Inverted retry logic
Currently ElasticsearchFlowStage _handleFailure_ method usage of _shouldRetry_ return value is inverted:
RetryNever retries indefinitely
RetryAtFixedRate fails immediately for any maxRetries values > 0

[this line](https://github.com/akka/alpakka/blob/ba8bee3cef55bd8be84bbb1babf67599cd194cb4/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchFlowStage.scala#L65) should say

`        if (!settings.retryLogic.shouldRetry(retryCount, List(exception.toString))) {`

sergeantwolf:
Fixes #1300 
note: I am considering adding a test for this case, however it proved to be not as straightforward as the fix itself 😄 
Opening PR in case it is acceptable to get the merge without it, will continue looking into creating a test in any case - as part of this PR or a separate one.


ennru:
Thank you! Looking forward to the test case.

PR #1304, opened 2018-11-04
MQTT streaming: Retire ask in favour of promise

huntc:
The ask pattern is overkill for situations where there is a strong coupling between one actor and another, as is the case between an MQTT session and its many command and event flows. For instance, if an MQTT session actor is lost for any reason, then its associated flows will also be failed. In this scenario, where a flow notifies the session of, say, an event having been received, it is effectively “calling” the session actor as per a function call, but using a message. The request is effectively synchronous in nature (although there is no blocking of course). The session actor must promise to yield a result, no matter what. Not yielding a result is a bug.

As such, we can use promises instead of the ask pattern for tightly coupled callers calling on actors. This then yields the advantage that request/reply timeouts do not have to be considered. Indeed, a strong motivation for the retirement of ask here is that it is difficult to determine a default timeout value. The performance of the MQTT session on a single core processor with memory constraints is quite different to that of a quad core with enormous amounts of memory. Timeouts perhaps make more sense when considering networks, where request/reply is required instead of a synchronous one.

Thanks to @longshorej for motivating this change and his thoughts on when to use the synchronous call pattern in place of ask.

### DIFF
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
@@ -11,8 +11,8 @@ import akka.actor.typed.scaladsl.{ActorContext, Behaviors}
 import akka.annotation.InternalApi
 import akka.stream.{Materializer, OverflowStrategy}
 import akka.stream.scaladsl.{BroadcastHub, Keep, Source, SourceQueueWithComplete}
-import akka.util.Timeout
 
+import scala.concurrent.Promise
 import scala.concurrent.duration.FiniteDuration
 import scala.util.control.NoStackTrace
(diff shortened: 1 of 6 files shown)


mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -253,7 +253,8 @@ import scala.util.{Failure, Success}
                     Subscriber(subscribeData, remote, data.subscriberPacketRouter, data.settings),
                     subscriberName
                   )
-                case _: Some[_] => // Ignored for existing subscriptions
+                case _: Some[_] =>
+                  remote.failure(new IllegalStateException("Duplicate subscribe: " + subscribe))
```
longshorej: :+1:

PR #1305, opened 2018-11-06
Fetch full history in Travis so version can always be determined

2m:


2m:
Failures were #1306 and #1307

Issue #1310, opened 2018-11-08
AWS SQS: All model classes should implement proper equality


RustedBones:
In the sqs connector, some model classes do not implement proper equality:
See [`MessageAction`](https://github.com/akka/alpakka/blob/fe367429ffd02160ea5c043ee0fec93f46dd8a6e/sqs/src/main/scala/akka/stream/alpakka/sqs/model.scala#L12) implementations.

This is quite inconvenient for testing. At the moment the workaround is to test equality on the `toString` results:
```scala
val source = TestSource.probe[Message]
val sink   = TestSink.probe[MessageAction]
val flow: Flow[Message, MessageAction]
val message: Message
val expected: MessageAction

val (publisher, subscriber) = source
      .viaMat(flow)(Keep.left)
      .toMat(sink)(Keep.both)
      .run()

subscriber.request(1)
publisher.sendNext(message)

subscriber.expectNext().toString() shouldBe expected.toString
```
Equals should be implemented or models should be converted to case classes

PR #1311, opened 2018-11-08
AWS SQS: Set proper equality for SQS model

RustedBones:
For review convenience, check diff only  on commit [
1447d1b](https://github.com/akka/alpakka/commit/1447d1be0e3f699a4bf86c6940495efbc4cd3e5b)


ennru:
Fixes #1310

ennru:
Thank you for reporting and fixing it!

PR #1315, opened 2018-11-08
MQTT streaming: Resilient client publications

huntc:
Client publications wouldn’t previously withstand a connection being lost, whereas server publications would. I’ve now aligned the client publication handling with the server publication handling.

In addition, session objects are now told to perform a publish command directly. The rationale is that the session object is the only thing that can guarantee the QoS requirements of a PUBLISH. I use the "tell" syntax of actors as I feel that it has the same meaning i.e. "communicate directly". For example:

```scala
clientSession ! Command(Publish(topic, ByteString("ohi")))
```

Prior to this work, PUBLISH commands were communicated with a command flow, which is often associated with a network connection. The problem with the approach was that PUBLISH commands could be lost if a network connection was lost i.e. the command is consumed, sent via TCP, TCP fails, and the command is lost. To substantiate this, TCP flows were often coupled with `RestartFlow`, which states, _"The restart process is inherently lossy, ..."_.

I've tested these changes with an external program and observed that publications recover from a loss of network connectivity.

### DIFF
--- a/docs/src/main/paradox/mqtt-streaming.md
+++ b/docs/src/main/paradox/mqtt-streaming.md
@@ -10,7 +10,7 @@ Further information on [mqtt.org](https://mqtt.org/).
 
 @@@ note { title="Paho Differences" }
 
-Alpakka contains @ref[another MQTT connector](mqtt.md) which is based on the Eclipse Paho client. Unlike the Paho version, this library has no dependencies other than those of Akka Streams i.e. it is entirely reactive. As such, there should be a significant performance advantage given its pure-Akka foundations, particularly in terms of memory usage given its diligent use of threads.
+Alpakka contains @ref[another MQTT connector](mqtt.md) which is based on the Eclipse Paho client. Unlike the Paho version, this library has no dependencies other than those of Akka Streams i.e. it is entirely reactive. As such, there should be a significant performance advantage given its pure-Akka foundations in terms of memory usage given its diligent use of threads.
 
 This library also differs in that it separates out the concern of how MQTT is connected. Unlike Paho, where TCP is assumed, this library can join in any flow. The end result is that by using this library, Unix Domain Sockets, TCP, UDP or anything else can be used to transport MQTT.
 
@@ -51,8 +51,12 @@ Scala
 Java
 : @@snip [snip](/mqtt-streaming/src/test/java/docs/javadsl/MqttFlowTest.java) { #run-streaming-flow }
 
-We drop the first 3 events received as they will be ACKs to our connect, subscribe and publish. The next event
-received is the publication to the topic we just subscribed to.
+Note that the `Publish` command is not offered to the command flow given MQTT QoS requirements. Instead, the 
+session is told to perform `Publish` given that it can retry continuously with buffering until a command 
+flow is established.
+
+We filter the events received as there will be ACKs to our connect, subscribe and publish. The collected event
+is the publication to the topic we just subscribed to.
 
 ## Flow through a server session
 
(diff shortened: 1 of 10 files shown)


mqtt-streaming/src/test/scala/docs/scaladsl/MqttFlowSpec.scala
```diff
@@ -51,19 +52,23 @@ class MqttFlowSpec
...
+        Publish(ControlPacketFlags.RETAIN | ControlPacketFlags.QoSAtLeastOnceDelivery, topic, ByteString("ohi"))
+      )
       //#run-streaming-flow
 
       events.futureValue match {
-        case Right(Event(Publish(_, `topic`, _, bytes), _)) => bytes shouldBe ByteString("ohi")
+        case Publish(_, `topic`, _, bytes) => bytes shouldBe ByteString("ohi")
         case e => fail("Unexpected event: " + e)
       }
     }
```
huntc: Other than the `session ! Command`, this is mostly about cleaning the test up and making it more resilient to out-of-expected-sequence `PubAck` messages.

mqtt-streaming/src/test/scala/docs/scaladsl/MqttFlowSpec.scala
```diff
@@ -98,15 +103,19 @@ class MqttFlowSpec
                 .toMat(BroadcastHub.sink)(Keep.both)
                 .run()
 
+              val subscribed = Promise[Done]
```
huntc: This promise is about improving the reliability of the test i.e. we don't publish messages on a topic until we know that a client has subscribed.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/javadsl/MqttSession.scala
```diff
@@ -23,6 +23,15 @@ import akka.stream.javadsl.Source
...
 abstract class MqttSession {
 
+  /**
+   * Tell the session to perform a command regardless of the state it is
+   * in. This is important for sending Publish messages in particular,
+   * as a connection may not have been established with a session.
+   * @param cp The command to perform
+   * @tparam A The type of any carry for the command.
+   */
+  def tell[A](cp: Command[A]): Unit
```
longshorej: Great idea!

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSession.scala
```diff
@@ -338,6 +356,12 @@ final class ActorMqttServerSession(settings: MqttSessionSettings)(implicit mat:
 
   import system.dispatcher
 
+  override def ![A](cp: Command[A]): Unit = cp match {
+    case Command(cp: Publish, carry) =>
+      serverConnector ! ServerConnector.PublishReceivedLocally(cp, carry)
+    case c: Command[_] => throw new IllegalStateException(c + " is not a server command that can be sent directly")
```
longshorej: Is there benefit in allowing `Subscribe`, `Unsubscribe` here as well? Or perhaps as it is now, resending those remains the client's responsibility?

huntc: I think those ones should remain the client responsibility. They are idempotent and the session isn’t persistent about their retries ie you really should have a connection established.

docs/src/main/paradox/mqtt-streaming.md
```diff
@@ -66,7 +70,9 @@ Java
 
 The resulting source's type shows how `Event`s are received and `Command`s are queued in reply. Our example
 acknowledges a connection, subscription and publication. Upon receiving a publication, it is re-published
-from the server so that any client that is subscribed will receive it.
+from the server so that any client that is subscribed will receive it. An addition detail is that we hold
```
ennru:
```suggestion
from the server so that any client that is subscribed will receive it. An additional detail is that we hold
```

docs/src/main/paradox/mqtt-streaming.md
```diff
@@ -66,7 +70,9 @@ Java
...
+off re-publishing until we have a subscription from the client. Note also how the session is told to perform
+`Publish` commands directly as they will be broadcast to all clients subscribed to the topic.
```
ennru:
```suggestion
`Publish` commands directly as they will be broadcasted to all clients subscribed to the topic.
```

PR #1317, opened 2018-11-09
Fixed Retry strategy

sderosiaux:
It seems the retry condition is reversed (if we should retry, we shouldn't give up).


ennru:
Thank you for identifying this, we already have #1302

Issue #1316, opened 2018-11-09
Failed: AvroParquetSinkTest.createNewParquetFile


ennru:
```
[info] Test docs.javadsl.AvroParquetSinkTest.createNewParquetFile started
[error] Test docs.javadsl.AvroParquetSinkTest.createNewParquetFile failed: java.lang.RuntimeException: file:/home/travis/build/akka/alpakka/javaTestFolder/test.parquet is not a Parquet file (too small length: 0), took 4.232 sec
[error]     at org.apache.parquet.hadoop.ParquetFileReader.readFooter(ParquetFileReader.java:514)
[error]     at org.apache.parquet.hadoop.ParquetFileReader.<init>(ParquetFileReader.java:689)
[error]     at org.apache.parquet.hadoop.ParquetFileReader.open(ParquetFileReader.java:595)
[error]     at org.apache.parquet.hadoop.ParquetReader.initReader(ParquetReader.java:152)
[error]     at org.apache.parquet.hadoop.ParquetReader.read(ParquetReader.java:135)
[error]     at docs.javadsl.AvroParquetSinkTest.checkResponse(AvroParquetSinkTest.java:98)
[error]     at docs.javadsl.AvroParquetSinkTest.createNewParquetFile(AvroParquetSinkTest.java:85)
```

https://travis-ci.org/akka/alpakka/jobs/452489678

ennru:
Resolved by #1319

PR #1319, opened 2018-11-11
Avro Parquet -  AvroParquetSinkTest.createNewParquetFile #1316

dannylesnik:
File was zero size because ParquetWriter was not closed. 

It was happened due to the one of two options:

-  Either  `Source.from(records).runWith(sink, materializer)` took more then 1 sec.
-  There was an uncaught exception in Sink.    

Anyway, I removed `Thread,sleep` method. At least we will get more clear exception next time.


ennru:
Fixes #1316

ennru:
Thank you for looking into this, it bothered me a lot to see this fail all the time.

dannylesnik:
Not a problem. This is just a reminder that no one should use `Thread.sleep` even for unit tests :)

PR #1320, opened 2018-11-11
WIP - DO NOT MERGE - Avoid a race condition when creating child actors

huntc:
This PR requires https://github.com/akka/alpakka/pull/1315 to be merged and then re-basing. For now, review just the [last commit](https://github.com/akka/alpakka/pull/1320/commits/f6477c4c33177f8c28226d35906ac40af0bbc060).

We observed an `InvalidActorNameException` when creating child actors under load. This could have been due to the following sequence when receiving a “Publish Received Locally” message:

1. prl received, producer actor created
2. producer actor terminates and sends a termination message
3. prl received before termination message is received, the parent actor creates another producer
4. the termination message is received and then attempts to create another producer with the same name

The solution is to explicitly track active consumers and producers rather than rely on another data structure such as `context.children`, which will be updated in response to other events.

### DIFF
--- a/docs/src/main/paradox/mqtt-streaming.md
+++ b/docs/src/main/paradox/mqtt-streaming.md
@@ -10,7 +10,7 @@ Further information on [mqtt.org](https://mqtt.org/).
 
 @@@ note { title="Paho Differences" }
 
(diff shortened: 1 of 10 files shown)


huntc:
Handled by https://github.com/akka/alpakka/pull/1327

PR #1321, opened 2018-11-11
WIP - DO NOT MERGE - Corrected misinterpretation of behaviour setup

huntc:
This PR requires https://github.com/akka/alpakka/pull/1315 to be merged and then re-basing. For now, review just the [last commit](https://github.com/akka/alpakka/pull/1321/commits/a4445d4a4b0bacdc40615a0ce8016d2a495ef3c2).

I had thought that `Behaviors.setup` would only get called once for an actor. This is not the case - it is called each time a behaviour is transitioned to.

There was only one place where my misunderstanding mattered, which was where a promise could be tried to be completed multiple times.


huntc:
Handled by https://github.com/akka/alpakka/pull/1327

PR #1327, opened 2018-11-13
Further MQTT streaming hardening

huntc:
Numerous hardening activities around MQTT streaming. Please visit each commit for a full description. However, in summary:

* Corrected misinterpretation of behaviour setup
* Avoid a race condition when creating child actors
* Handle a duplicate publish while consuming
* Connection packet id distinction which now allows more than one client to send the same packet id as another

### DIFF
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
@@ -292,7 +292,7 @@ import scala.util.{Failure, Success}
               val consumerName = ActorName.mkName(ConsumerNamePrefix + topicName + "-" + context.children.size)
               context.watchWith(
                 context.spawn(
-                  Consumer(publish, packetId, local, data.consumerPacketRouter, data.settings),
+                  Consumer(publish, None, packetId, local, data.consumerPacketRouter, data.settings),
                   consumerName
                 ),
                 ConsumerFree(publish.topicName)
@@ -311,6 +311,7 @@ import scala.util.{Failure, Success}
               context.watchWith(
                 context.spawn(
                   Consumer(prfr.publish,
+                           None,
                            prfr.publish.packetId.get,
                            prfr.local,
                            data.consumerPacketRouter,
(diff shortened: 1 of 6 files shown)


mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSession.scala
```diff
@@ -226,6 +233,10 @@ final class ActorMqttClientSession(settings: MqttSessionSettings)(implicit mat:
           reply.future.map {
             case Unsubscriber.ForwardUnsubAck(carry: Option[A] @unchecked) => Right(Event(cp, carry))
           }
+        case Right(cp @ Publish(flags, _, Some(packetId), _)) if flags.contains(ControlPacketFlags.DUP) =>
```
longshorej: Any implications here for a broken/malicious client that e.g. sets the DUP flag when it isn't really one?

huntc: In this instance, a publish will be dropped given that there's no packet id already registered.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -323,35 +326,34 @@ import scala.util.{Failure, Success}
...
+              serverConnected(
+                data.copy(pendingLocalPublications = data.pendingLocalPublications :+ (publish.topicName -> prl))
+              )
             }
           case (context, ProducerFree(topicName)) =>
             val i = data.pendingLocalPublications.indexWhere(_._1 == topicName)
             if (i >= 0) {
               val prl = data.pendingLocalPublications(i)._2
-              val producerName = ActorName.mkName(ProducerNamePrefix + topicName)
+              val producerName = ActorName.mkName(ProducerNamePrefix + topicName + context.children.size)
```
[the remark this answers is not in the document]
longshorej: Also, MQTT topics should be case sensitive. Does that match the semantics of Akka actor names, i.e. is actor `PUBLISHER` different than `PuBLiShEr`?

ennru:
Thank you!

Issue #1132, opened 2018-08-06
AMQP : support "confirm" Sink and Flow producers


jcare44:
Hi,
I think it would be great for the AMQP lib to handle publish confirms, both in a specific Sink and via a Flow. I like [the approach of the alpakka Kafka lib on that](https://doc.akka.io/api/akka-stream-kafka/0.22/akka/kafka/scaladsl/Producer$.html), and I think it would be great to implement something similar for AMQP.
What do you think of it ?

juanjoDiaz:
Can you elaborate a bit on what you are suggesting?

Do you mean having a passthrough materialized view? (I found myself needing that recently).

2m:
Yes, I aggree that it would be useful to have AMQP producer modeled as a Flow, where the input to the Flow would be `OutgoingMessage` and the flow would emit confirmations that the publish has been successful or not.

PR #1330, opened 2018-11-15
[AMQP] add publish flow & support for publish confirm

jcare44:
Hi :)
Following my feature proposition https://github.com/akka/alpakka/issues/1132, this PR aims to add a publish flow stage and the support for AMQP publish confirms.
I'm not sure if it match all your guidelines yet, so your comments are very welcomed :)

### DIFF
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
@@ -4,7 +4,10 @@
 
 package akka.stream.alpakka.amqp
 
+import java.time.Duration
+
 import akka.annotation.InternalApi
+import akka.util.JavaDurationConverters._
 
 import scala.collection.JavaConverters._
 import scala.collection.immutable
@@ -159,7 +162,7 @@ final class AmqpPublishConfirmConfiguration private (
 ) {
   override def toString: String =
     "AmqpPublishConfirmConfiguration(" +
-    s"confirmTimeout=$confirmTimeout" +
+    s"confirmTimeout=${confirmTimeout.toCoarsest}" +
     ")"
 }
 
@@ -167,14 +170,11 @@ object AmqpPublishConfirmConfiguration {
   def apply(confirmTimeout: FiniteDuration): AmqpPublishConfirmConfiguration =
     new AmqpPublishConfirmConfiguration(confirmTimeout)
 
-  def unapply(settings: AmqpPublishConfirmConfiguration): Option[FiniteDuration] =
-    Some(settings.confirmTimeout)
-
   /**
    * Java API
    */
-  def create(confirmTimeout: FiniteDuration): AmqpPublishConfirmConfiguration =
-    new AmqpPublishConfirmConfiguration(confirmTimeout)
+  def create(confirmTimeout: Duration): AmqpPublishConfirmConfiguration =
+    new AmqpPublishConfirmConfiguration(confirmTimeout.asScala)
 }
 
 final class AmqpReplyToSinkSettings private (
@@ -189,6 +189,12 @@ final class AmqpReplyToSinkSettings private (
 
   def withPublishConfirm(confirmTimeout: FiniteDuration): AmqpReplyToSinkSettings =
     copy(publishConfirm = Some(AmqpPublishConfirmConfiguration(confirmTimeout)))
+
+  /**
+   * Java API
+   */
+  def withPublishConfirm(confirmTimeout: Duration): AmqpReplyToSinkSettings =
+    copy(publishConfirm = Some(AmqpPublishConfirmConfiguration.create(confirmTimeout)))
 
   private def copy(connectionProvider: AmqpConnectionProvider = connectionProvider,
                    failIfReplyToMissing: Boolean = failIfReplyToMissing,
@@ -243,6 +249,9 @@ final class AmqpWriteSettings private (
   def withDeclarations(declarations: java.util.List[Declaration]): AmqpWriteSettings =
     copy(declarations = declarations.asScala.toIndexedSeq)
 
+  def withPublishConfirm(confirmTimeout: Duration): AmqpWriteSettings =
+    copy(publishConfirm = Some(AmqpPublishConfirmConfiguration.create(confirmTimeout)))
+
   private def copy(connectionProvider: AmqpConnectionProvider = connectionProvider,
                    exchange: Option[String] = exchange,
                    routingKey: Option[String] = routingKey,
@@ -251,11 +260,12 @@ final class AmqpWriteSettings private (
     new AmqpWriteSettings(connectionProvider, exchange, routingKey, declarations, publishConfirm)
 
   override def toString: String =
-    "AmqpSinkSettings(" +
+    "AmqpWriteSettings(" +
     s"connectionProvider=$connectionProvider, " +
     s"exchange=$exchange, " +
     s"routingKey=$routingKey, " +
     s"declarations=$declarations" +
+    s"publishConfirm=$publishConfirm" +
     ")"
 }
 
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
@@ -153,12 +153,28 @@ object TemporaryQueueSourceSettings {
     TemporaryQueueSourceSettings(connectionProvider, exchange)
 }
 
-final case class AmqpPublishConfirmSettings(confirmTimeout: Long)
+final class AmqpPublishConfirmConfiguration private (
+    val confirmTimeout: Long
+)
+
+object AmqpPublishConfirmConfiguration {
+  def apply(confirmTimeout: Long): AmqpPublishConfirmConfiguration =
+    new AmqpPublishConfirmConfiguration(confirmTimeout)
+
+  def unapply(settings: AmqpPublishConfirmConfiguration): Option[Long] =
+    Some(settings.confirmTimeout)
+
+  /**
+   * Java API
+   */
+  def create(confirmTimeout: Long): AmqpPublishConfirmConfiguration =
+    new AmqpPublishConfirmConfiguration(confirmTimeout)
+}
 
 final class AmqpReplyToSinkSettings private (
     val connectionProvider: AmqpConnectionProvider,
     val failIfReplyToMissing: Boolean = true,
-    val publishConfirm: Option[AmqpPublishConfirmSettings] = None
+    val publishConfirm: Option[AmqpPublishConfirmConfiguration] = None
 ) extends AmqpConnectorSettings {
   override final val declarations = Nil
 
@@ -166,11 +182,11 @@ final class AmqpReplyToSinkSettings private (
     copy(failIfReplyToMissing = failIfReplyToMissing)
 
   def withPublishConfirms(confirmTimeout: Long = 1000): AmqpReplyToSinkSettings =
-    copy(publishConfirm = Some(AmqpPublishConfirmSettings(confirmTimeout)))
+    copy(publishConfirm = Some(AmqpPublishConfirmConfiguration(confirmTimeout)))
 
   private def copy(connectionProvider: AmqpConnectionProvider = connectionProvider,
                    failIfReplyToMissing: Boolean = failIfReplyToMissing,
-                   publishConfirm: Option[AmqpPublishConfirmSettings] = publishConfirm) =
+                   publishConfirm: Option[AmqpPublishConfirmConfiguration] = publishConfirm) =
     new AmqpReplyToSinkSettings(connectionProvider, failIfReplyToMissing, publishConfirm)
 
   override def toString: String =
@@ -196,7 +212,7 @@ final class AmqpWriteSettings private (
     val exchange: Option[String] = None,
     val routingKey: Option[String] = None,
     val declarations: immutable.Seq[Declaration] = Nil,
-    val publishConfirm: Option[AmqpPublishConfirmSettings] = None
+    val publishConfirm: Option[AmqpPublishConfirmConfiguration] = None
 ) extends AmqpConnectorSettings {
 
   def withExchange(exchange: String): AmqpWriteSettings =
@@ -211,8 +227,8 @@ final class AmqpWriteSettings private (
   def withDeclarations(declarations: immutable.Seq[Declaration]): AmqpWriteSettings =
     copy(declarations = declarations)
 
-  def withPublishConfirms(confirmTimeout: Long = 1000): AmqpWriteSettings =
-    copy(publishConfirm = Some(AmqpPublishConfirmSettings(confirmTimeout)))
+  def withPublishConfirms(confirmTimeout: Long): AmqpWriteSettings =
+    copy(publishConfirm = Some(AmqpPublishConfirmConfiguration(confirmTimeout)))
 
   /**
    * Java API
@@ -224,7 +240,7 @@ final class AmqpWriteSettings private (
                    exchange: Option[String] = exchange,
                    routingKey: Option[String] = routingKey,
                    declarations: immutable.Seq[Declaration] = declarations,
-                   publishConfirm: Option[AmqpPublishConfirmSettings] = publishConfirm) =
+                   publishConfirm: Option[AmqpPublishConfirmConfiguration] = publishConfirm) =
     new AmqpWriteSettings(connectionProvider, exchange, routingKey, declarations, publishConfirm)
 
   override def toString: String =
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
@@ -153,7 +153,7 @@ object TemporaryQueueSourceSettings {
     TemporaryQueueSourceSettings(connectionProvider, exchange)
 }
 
-final case class AmqpPublishConfirmSettings(confirmTimeout: Long)
+final case class AmqpPublishConfirmSettings(confirmTimeout: Long = 1000)
 
 final class AmqpReplyToSinkSettings private (
     val connectionProvider: AmqpConnectionProvider,
@@ -213,6 +213,9 @@ final class AmqpWriteSettings private (
 
   def withPublishConfirms(confirmTimeout: Long = 1000): AmqpWriteSettings =
     copy(publishConfirm = Some(AmqpPublishConfirmSettings(confirmTimeout)))
+
+  def withPublishConfirms(): AmqpWriteSettings =
+    copy(publishConfirm = Some(AmqpPublishConfirmSettings()))
 
   /**
    * Java API
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
@@ -8,6 +8,7 @@ import akka.annotation.InternalApi
 
 import scala.collection.JavaConverters._
 import scala.collection.immutable
+import scala.concurrent.duration.FiniteDuration
 
 /**
  * Internal API
@@ -154,7 +155,7 @@ object TemporaryQueueSourceSettings {
 }
 
 final class AmqpPublishConfirmConfiguration private (
-    val confirmTimeout: Long
+    val confirmTimeout: FiniteDuration
 ) {
   override def toString: String =
     "AmqpPublishConfirmConfiguration(" +
@@ -163,16 +164,16 @@ final class AmqpPublishConfirmConfiguration private (
 }
 
 object AmqpPublishConfirmConfiguration {
-  def apply(confirmTimeout: Long): AmqpPublishConfirmConfiguration =
+  def apply(confirmTimeout: FiniteDuration): AmqpPublishConfirmConfiguration =
     new AmqpPublishConfirmConfiguration(confirmTimeout)
 
-  def unapply(settings: AmqpPublishConfirmConfiguration): Option[Long] =
+  def unapply(settings: AmqpPublishConfirmConfiguration): Option[FiniteDuration] =
     Some(settings.confirmTimeout)
 
   /**
    * Java API
    */
-  def create(confirmTimeout: Long): AmqpPublishConfirmConfiguration =
+  def create(confirmTimeout: FiniteDuration): AmqpPublishConfirmConfiguration =
     new AmqpPublishConfirmConfiguration(confirmTimeout)
 }
 
@@ -186,7 +187,7 @@ final class AmqpReplyToSinkSettings private (
   def withFailIfReplyToMissing(failIfReplyToMissing: Boolean): AmqpReplyToSinkSettings =
     copy(failIfReplyToMissing = failIfReplyToMissing)
 
-  def withPublishConfirms(confirmTimeout: Long): AmqpReplyToSinkSettings =
+  def withPublishConfirm(confirmTimeout: FiniteDuration): AmqpReplyToSinkSettings =
     copy(publishConfirm = Some(AmqpPublishConfirmConfiguration(confirmTimeout)))
 
   private def copy(connectionProvider: AmqpConnectionProvider = connectionProvider,
@@ -233,7 +234,7 @@ final class AmqpWriteSettings private (
   def withDeclarations(declarations: immutable.Seq[Declaration]): AmqpWriteSettings =
     copy(declarations = declarations)
 
-  def withPublishConfirms(confirmTimeout: Long): AmqpWriteSettings =
+  def withPublishConfirm(confirmTimeout: FiniteDuration): AmqpWriteSettings =
     copy(publishConfirm = Some(AmqpPublishConfirmConfiguration(confirmTimeout)))
 
   /**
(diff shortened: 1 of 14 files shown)


amqp/src/main/scala/akka/stream/alpakka/amqp/javadsl/AmqpPublishFlow.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Java API: Creates an [[AmqpPublishFlow]] that accepts (ByteString, passthrough) elements.
+   *
+   * This stage materializes to a CompletionStage<Done>, which can be used to know when the Flow completes, either normally
+   * or because of an amqp failure
+   */
+  def createSimple[O](settings: AmqpSinkSettings): Flow[Pair[ByteString, O], O, CompletionStage[Done]] =
+    Flow
+      .fromFunction(new Function[Pair[ByteString, O], (ByteString, O)] {
```
ennru: You may use `Flow.of[Pair[ByteString, O]]` and a regular `map` operator.

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpPublishFlow.scala
```diff
@@ -0,0 +1,40 @@
...
+
+import akka.Done
+import akka.stream.alpakka.amqp.{AmqpSinkSettings, OutgoingMessage}
+import akka.stream.alpakka.amqp.impl.AmqpPublishFlowStage
+import akka.stream.scaladsl.{Flow, Keep}
+import akka.util.ByteString
+
+import scala.concurrent.Future
+
+object AmqpPublishFlow {
```
ennru: I think `AmqpFlow` would be enough as publishing can be assumed to be the default.

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpPublishFlow.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * Scala API: Creates an [[AmqpPublishFlow]] that accepts (ByteString, passthrough) elements.
+   *
+   * This stage materializes to a `Future[Done]`, which can be used to know when the Flow completes, either normally
+   * or because of an amqp failure
+   */
+  def simple[O](settings: AmqpSinkSettings): Flow[(ByteString, O), O, Future[Done]] =
```
ennru: I believe we could use something more convenient as a tuple and somehow merge the pass-through support directly into `OutgoingMessage` (which I'd like to become `AmqpWriteMessage`). 
But leave that for now, that could become another API-changing PR.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -135,18 +135,41 @@ object TemporaryQueueSourceSettings {
     TemporaryQueueSourceSettings(connectionProvider, exchange)
 }
 
+final class AmqpPublishConfirmConfiguration private (
```
ennru: Does the confirm timeout require a configuration class for itself? 
Wouldn't `confirmTimeout: Option[FiniteDuration]` express all that is needed?

jcare44: I can do so if you prefer.
I thought it was more clear to have `publishConfirm: Option[AmqpPublishConfirmConfiguration] = None` in the Sink settings rather than simply `confirmTimeout: Option[FiniteDuration]` since the `Option` here represent the activation of the "confirm" feature. When I read `confirmTimeout: Option[FiniteDuration]`, I first think that it's the timeout that is optional, and not the "confirm" feature itself :/

jcare44:
Thank you for your review, and sorry for the delay.
Apart for the commented one, I should have addressed all your comments (and have rebased on master). 
If you have any more insight on this PR, please don't hesitate to add more comments :)

amqp/src/main/scala/akka/stream/alpakka/amqp/javadsl/AmqpPublishFlow.scala
jcare44: Well I tried it, and remembered why I had done so ^^
Your solution works fine on scala 2.12, but with scala 2.11 I would still need to create this `akka.japi.function.Function` in the map :/
So here, `.fromFunction` is just a more concise version of `.of(classOf[Pair[ByteString, O]]).map(new Function...)`

2m:
@jcare44, there has been some changes in the AMQP connector in the last week. Could you rebase this PR on the latest master?

jcare44:
rebased :)

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -153,22 +154,52 @@ object TemporaryQueueSourceSettings {
...
+  def apply(confirmTimeout: FiniteDuration): AmqpPublishConfirmConfiguration =
+    new AmqpPublishConfirmConfiguration(confirmTimeout)
+
+  def unapply(settings: AmqpPublishConfirmConfiguration): Option[FiniteDuration] =
+    Some(settings.confirmTimeout)
+
+  /**
+   * Java API
+   */
+  def create(confirmTimeout: FiniteDuration): AmqpPublishConfirmConfiguration =
```
ennru: For the Java API please use `java.time.Duration` and convert it with `akka.util.JavaDurationConverters._`.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -153,22 +154,52 @@ object TemporaryQueueSourceSettings {
...
+    "AmqpPublishConfirmConfiguration(" +
+    s"confirmTimeout=$confirmTimeout" +
+    ")"
+}
+
+object AmqpPublishConfirmConfiguration {
+  def apply(confirmTimeout: FiniteDuration): AmqpPublishConfirmConfiguration =
+    new AmqpPublishConfirmConfiguration(confirmTimeout)
+
+  def unapply(settings: AmqpPublishConfirmConfiguration): Option[FiniteDuration] =
```
ennru: Extractors are hard to evolve in a binary compatible way, please remove and access the value directly where needed.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -153,22 +154,52 @@ object TemporaryQueueSourceSettings {
     TemporaryQueueSourceSettings(connectionProvider, exchange)
 }
 
+final class AmqpPublishConfirmConfiguration private (
+    val confirmTimeout: FiniteDuration
+) {
+  override def toString: String =
+    "AmqpPublishConfirmConfiguration(" +
+    s"confirmTimeout=$confirmTimeout" +
```
ennru:
```suggestion
    s"confirmTimeout=${confirmTimeout.toCoarsest}" +
```

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -153,22 +154,52 @@ object TemporaryQueueSourceSettings {
...
+    val publishConfirm: Option[AmqpPublishConfirmConfiguration] = None
 ) extends AmqpConnectorSettings {
   override final val declarations = Nil
 
   def withFailIfReplyToMissing(failIfReplyToMissing: Boolean): AmqpReplyToSinkSettings =
     copy(failIfReplyToMissing = failIfReplyToMissing)
 
-  private def copy(connectionProvider: AmqpConnectionProvider = connectionProvider, failIfReplyToMissing: Boolean) =
-    new AmqpReplyToSinkSettings(connectionProvider, failIfReplyToMissing)
+  def withPublishConfirm(confirmTimeout: FiniteDuration): AmqpReplyToSinkSettings =
```
ennru: Add the corresponding Java API with `java.time.Duration`.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -202,6 +234,9 @@ final class AmqpWriteSettings private (
   def withDeclarations(declarations: immutable.Seq[Declaration]): AmqpWriteSettings =
     copy(declarations = declarations)
 
+  def withPublishConfirm(confirmTimeout: FiniteDuration): AmqpWriteSettings =
```
ennru: Add Java API.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -211,8 +246,9 @@ final class AmqpWriteSettings private (
...
                    exchange: Option[String] = exchange,
                    routingKey: Option[String] = routingKey,
-                   declarations: immutable.Seq[Declaration] = declarations) =
-    new AmqpWriteSettings(connectionProvider, exchange, routingKey, declarations)
+                   declarations: immutable.Seq[Declaration] = declarations,
+                   publishConfirm: Option[AmqpPublishConfirmConfiguration] = publishConfirm) =
+    new AmqpWriteSettings(connectionProvider, exchange, routingKey, declarations, publishConfirm)
 
   override def toString: String =
     "AmqpSinkSettings(" +
```
ennru: Add the new property in `toString`. (And change it to `AmqpWriteSettings` while at it.)

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpPublishFlowStage.scala
```diff
@@ -0,0 +1,104 @@
...
+
+  val in = Inlet[(WriteMessage, O)]("AmqpPublishConfirmFlow.in")
+  val out = Outlet[O]("AmqpPublishConfirmFlow.out")
+
+  override def shape: FlowShape[(WriteMessage, O), O] = FlowShape.of(in, out)
+
+  override protected def initialAttributes: Attributes =
+    Attributes
+      .name("AmqpPublishFlowStage")
+      .and(ActorAttributes.dispatcher("akka.stream.default-blocking-io-dispatcher"))
```
ennru:
```suggestion
      .and(ActorAttributes.IODispatcher)
```

docs/src/main/paradox/amqp.md
```diff
@@ -54,6 +57,23 @@ Scala
 Java
 : @@snip [snip](/amqp/src/test/java/docs/javadsl/AmqpDocsTest.java) { #create-sink }
 
+#### With a Flow
+
+Create a flow, that accepts a Tuple2 of @scaladoc[ByteString](akka.util.ByteString) and a PassThrough element (a String in the example). The @scaladoc[ByteString](akka.util.ByteString) will be forwarded to the AMQP server, and the PassThrough element passed downstream.
```
ennru:
```suggestion
Create a flow, that accepts a @scala[tuple]@java[`akka.japi.Pair`] of @scaladoc[ByteString](akka.util.ByteString) and a PassThrough element (a String in the example). The @scaladoc[ByteString](akka.util.ByteString) will be forwarded to the AMQP server, and the PassThrough element passed downstream.
```

docs/src/main/paradox/amqp.md
```diff
@@ -54,6 +57,23 @@ Scala
...
+
+Scala
+: @@snip [snip](/amqp/src/test/scala/docs/scaladsl/AmqpDocsSpec.scala) { #create-publish-flow }
+
+Java
+: @@snip [snip](/amqp/src/test/java/docs/javadsl/AmqpDocsTest.java) { #create-publish-flow }
+
+Note: in This example, we activate AMQP's publish confirms so that only truly published elements pass this stage. This is completely optional.
```
ennru:
```suggestion
Note: in This example, we activate AMQP's publish confirms so that only truly published elements pass this stage and backpressure applies via the AMQP broker. This is completely optional.
```

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpPublishFlowStage.scala
```diff
@@ -0,0 +1,104 @@
...
+
+  override def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, Future[Done]) = {
+    val promise = Promise[Done]()
```
ennru: How would the materialized value be used? It seems just to get the stream failure or completion.

jcare44:
Thank you for your feed back and sorry for the delay.
This commit should address your latest comments :)

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpPublishFlowStage.scala
jcare44: Yes you're right, it doesn't seem to provide much value here. 
I've switched it to a `GraphStage`

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpPublishFlowStage.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  override def shape: FlowShape[(WriteMessage, O), O] = FlowShape.of(in, out)
+
+  override protected def initialAttributes: Attributes =
+    Attributes
+      .name("AmqpPublishFlowStage")
+      .and(ActorAttributes.IODispatcher)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) with AmqpConnectorLogic {
```
fancywriter: Hi @jcare44 
I indeed like your implementation better than mine (and you also have updated the documentation which I didn't in #1584 )
However, small suggestion - I really thing that `AmqpConnectorLogic` has to be abstract class with constructor, not trait - it removes some code duplication. What do you think?

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpPublishFlowStage.scala
```diff
@@ -0,0 +1,78 @@
...
+              elem.routingKey.getOrElse(routingKey),
+              elem.mandatory,
+              elem.immediate,
+              elem.properties.orNull,
+              elem.bytes.toArray
+            )
+
+            settings.publishConfirm match {
+              case Some(publishConfirmConf) =>
+                Try(channel.waitForConfirmsOrDie(publishConfirmConf.confirmTimeout.toMillis)) match {
```
fancywriter: I think that `channel.addConfirmListener` could be better - without blocking. What do you think?
You can copy it from #1584

fancywriter:
I have small concern about `channel.waitForConfirmsOrDie` - according to the name it introduces blocking inside akka stream stage. Am I right?
@jcare44 @ennru

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpPublishFlowStage.scala
jcare44: Hi @fancywriter, thanks for your comments.
Hum yes it could be an abstract class, but I don't see much value to it. What would you mutualise in it? In your PR, I only saw the `override settings` moved as argument, and I don't see what more we could put in it :/
In the other hand, I think that it being a `trait` allows for more flexibility in the future, since you can compose it with other traits and still be open for an abstract class to be created if need be.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpPublishFlowStage.scala
jcare44: Well, I also thought of it at first, and it clearly as the advantage of being non-blocking and therefore faster, but I think the huge disadvantage of it is that it doesn't maintain order :/
With `waitForConfirmsOrDie`, it ensure order is maintained both in the flow and in amqp.

We'll see with usage, but I think use cases where `waitForConfirmsOrDie` is to slow are rather rare :/

What do you think ? Do you have such use cases ?

fancywriter: Well, right now I have a code base in proprietary project (which is not written akka/alpakka) where exactly confirm listener is used. I wanted to migrate it to alpakka, but don't want to change the logic, because don't want to break things.

Order is not always needed. Probably can be restored by **deliveryTag** parameter, but not sure here.

Or may be this thing could be configurable and we can add both implementations. (can create next PR with adding new implementation and keeping this once this one will be merged).

kstrek: If I can add one thing here - waiting for confirmation after each publish (as with `waitForConfirmsOrDie`) is a true performance killer, especially in combination with persistent messages (more on that [here](https://www.rabbitmq.com/confirms.html#publisher-confirms-latency)). As @fancywriter mentioned, with `channel.addConfirmListener` order can be restored (next deliveryTag can be retrieved before message is published - I even started my own implementation this way), but actually I think both approaches are valuable and have their own use cases. It would be perfect to have 3 different flow connectors: 

- synchronous (like with `waitForConfirmsOrDie`) - preserving order of messages **delivered** to the message broker
- asynchronous ordered (with `channel.addConfirmListener` and some kind of output buffer that dispatches messages downstream in order) - more performant while preserving order in messages that **leave the graph stage**
- asynchronous unordered (also with `channel.addConfirmListener`) - the most performant but without any ordering guarantees

In absence of namming pattern in Alpakka for similar situation (at least I don't see one), those could be named similarly as variants of `map` operator from vanilla Akka Streams, i.e. `AmqpFlow.simple`, `AmqpFlow.async` and `AmqpFlow.asyncUnordered`

I also agree that we could just finish synchronous variant here, and remaining ones in another PR (I will be happy to contribute as well) - just please keep in mind other variants while naming stuff in public api ;)

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
ennru: Ok, makes sense.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
ennru: With the reasoning of the `Option` above, this should take `AmqpPublishConfirmConfiguration` instead.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpPublishFlowStage.scala
ennru: @kstrek Sounds like a plan. I hope the variants can be implemented by variants of this graph stage logic.

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpFlow.scala
```diff
@@ -0,0 +1,38 @@
...
+
+object AmqpFlow {
+
+  /**
+   * Scala API: Creates an [[AmqpFlow]] that accepts (ByteString, passthrough) elements.
+   *
+   * This stage materializes to a `Future[Done]`, which can be used to know when the Flow completes, either normally
+   * or because of an amqp failure
+   */
+  def simple[O](settings: AmqpWriteSettings): Flow[(ByteString, O), O, NotUsed] =
```
ennru: I'm not sure if `simple` adds so much value.

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpFlow.scala
```diff
@@ -0,0 +1,38 @@
...
+   * Scala API:
+   *
+   * Connects to an AMQP server upon materialization and sends incoming messages to the server.
+   * Each materialized flow will create one connection to the broker. This stage sends messages to
+   * the queue named in the replyTo options of the message instead of from settings declared at construction.
+   *
+   * This stage materializes to a Future[Done], which can be used to know when the Flow completes, either normally
+   * or because of an amqp failure
+   */
+  def apply[O](settings: AmqpWriteSettings): Flow[(WriteMessage, O), O, NotUsed] =
```
ennru: Akka has added `withContext` support which composed with `Flow[(in, passThrough), (out, passThrough), ..]`, so it would be better to emit tuples. An alternative is to push the path-through into `WriteMessage` as many other connectors do.

Is there anything interesting that should be emitted as a `WriteResult`? One option could be to let the stage emit futures that signal confirmation so it becomes a user concern to fail the stream.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpPublishFlowStage.scala
jcare44: Sorry for the late reply. Yes that seems like a great idea! 
Did you see anything in the current PR that would hinder future addition? It looked fine for me, but I may have missed some.

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpFlow.scala
jcare44: Not sure either, here I simply followed what was done in other dsl (RPC & Sink for example).
I can remove it if you prefer.

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpFlow.scala
jcare44: Oh I didn't saw that. 
But to me it doesn't seem like a flow need a Tuple2 in and out to be compatible with that. The [`Flow.asFlowWithContext` method](https://doc.akka.io/docs/akka/current/stream/operators/Flow/asFlowWithContext.html) takes functions in order to transform from/into Tuple2, so with the current signature I think it should be compatible. Or did I miss something ?
([Some other examples here](https://github.com/akka/akka/pull/25951/files#diff-5bc226683709f15a4d491c27f88e98e7))

> Is there anything interesting that should be emitted as a WriteResult? 

Hum I don't think of anything that would make sense as `WriteResult` for this Flow :/. 

> One option could be to let the stage emit futures that signal confirmation so it becomes a user concern to fail the stream.

For the upcoming `async` and `asyncUnordered`, it could make sense to return a `Future[Unit]` (or `Try[Unit]`), but for this one, I'm afraid it could break the message order:
If we return a failed Future (or Try) downstream, I think it lets the publish stage take in the next element. So technically, couldn't the publish of the next element be sent before we fail the Flow ?

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpPublishFlowStage.scala
kstrek: @jcare44 I don't see anything in particular. Looks fine for me too.

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpFlow.scala
kstrek: I agree that it doesn't really make sense in this simple ("blocking") variant to pass futures downstream. Also, if we were to do that in `async` and `asyncUnordered` there would be no difference at all between those two - reordering is a result of out-of-order confirmations, and `async` is meant to internally wait for those confirmations and dispatch messages downstream in initial ("memorized") order. 
Future-emitting flow looks like yet another possible asynchronous variant where user could arbitrarily handle confirmations, but I'm not sure if there is any real use case for that.

ennru:
Just looking into this again, your latest change looks like a great idea. The different confirm options (no confirm, sync confirm, async ordered, async unordered) could be just different `AmqpPublishConfirmConfiguration`s. I suggest to make it non-optional and have the current state "no confirm" as the default. WDYT?

kstrek:
Well, I presume that your idea is based on the fact that we now have `AmqpPublishConfirmConfiguration` in `AmqpReplyToSink` and `AmqpSink`, and it would be nice to have different confirm options there as well 😉. 

Now I think it might be good idea to change the approach a little, and, instead of failing the stage on publication failure, pass the result of publication downstream, so that the user could decide what to do. With that there would be no need to modify `AmqpSink` - users interested in publication result would just use some variant of `AmpqFlow`. Using the same approach consequently for `AmqpReplyToSink`, however, would also require creating something like `AmqpReplyToFlow` with all it's variants.

I think it's cleaner to have different variants of flows created by different factory methods (like `AmpqFlow.simple(...)` and `AmpqFlow.asyncUnordered(...)`) as with that we have clear distinction: using specific factory method defines how your stage fundamentally works, and `AmqpPublishConfirmConfiguration` is used just to "tweak" it's inner operations (timeouts etc.).

ennru:
Next release is intended to be Alpakka 2.0. Do you think you can apply the suggested changes?

kstrek:
@ennru I created another pull request that covers the changes and adds other discussed types of flow connectors. Please take a look: #1920

ennru:
Thank you @jcare44 for kicking this off! 
As you seem to have left it behind, @kstrek suggest #1921 instead.

Superseded by #1921

PR #1341, opened 2018-11-25
MQTT streaming: Modal subscriptions

huntc:
Prior to this change, an MQTT client could only send one SUBSCRIBE for a given topic at a time without receiving an `IllegalStateException`. It is quite plausible that multiple SUBSCRIBE requests for the same topic are made in succession given connection failures. This change now stashes requests while a SUBSCRIBE is occurring on the client, in the same way that the server does.

Also, I was sending out a SUBSCRIBE request for each topic filter, whereas it should go out as just one request. This would have caused errors had the feature been leveraged by someone (we aren’t leveraging this feature presently). (same for UNSUBSCRIBE - doh!)


huntc:
@ennru Test failures appear disassociated with my changes/subproject.

huntc:
Thanks!

Issue #1111, opened 2018-07-17
Redis connector 


SandishKumarHN:

2m:
Great! Do you have any specific usecase in mind, that would greatly benefit from the streaming approach? Also, do you have any particular Redis client library in mind? AFAIK there are a couple of different async Scala and Java ones.

SandishKumarHN:
The use case would something like this "Taking advantage of Redis in memory storage engine to do list and set operations makes it an amazing platform to use for a message queue."
I'm thinking of using this https://github.com/lettuce-io/lettuce-core Redis Scala async client

riyadparvez:
@SandishKumarHN Any update on this issue?

SandishKumarHN:
@riyadparvez  working on it will create a PR by Dec 1st week

sourcekick:
Very interesting Issue! @SandishKumarHN Do you have something? I see some work is also underway and pending for reviews in https://github.com/akka/alpakka/pull/1350

hveiga:
Is there any progress on this? If not, I am willing to put some time on it to get it ready for a PR.

ennru:
No, we haven't seen any activity since the review of https://github.com/akka/alpakka/pull/1350

choffmeister:
Since none of the existing redis client implementations did what I wanted (I wanted to to be based on akka and to super TLS), I started my own implementation, purely based on akka streams (no custom actors and stuff).

https://github.com/choffmeister/microservice-utils/blob/master/microservice-utils-rate-limiter/src/main/scala/de/choffmeister/microserviceutils/ratelimiter/RedisClient.scala

The parsing is still incomplete (no support for parsing complex string replies), but does what I need for now (needed it for rate limiting).

ennru:
@choffmeister Thanks for sharing!

choffmeister:
@ennru I finished implementing the redis reply parsing and vastly improved the test coverage:

https://github.com/choffmeister/microservice-utils/blob/d84adf528258b417ffc3f5d2e3a5e05b1882080c/microservice-utils-redis/src/main/scala/de/choffmeister/microserviceutils/redis/RedisClient.scala

In this state this might be a candidate for alpakka. All other implementations I saw heavily used custom written actors for state management. This implementation is way simpler.

choffmeister:
[...]

PR #1350, opened 2018-11-27
Redis integration

dannylesnik:
Redis integratio.  Java/Scala DSL for Source, Sink and Flow including Pub/Sub functionality.

This is the skeleton for Redis Integration with Akka Streams.  it currently has some basic commands and pub/sub functionality which might play very well with Akka streams. Unit tests are running as well for each command.
If such integration is relevant, I will continue to work on it for next couple of weeks and will let you know when the work is done.


ennru:
Yes, a Redis connector would be highly appreciated. The Lightbend Alpakka team might be a bit slow on reviewing right now, though. 
Maybe someone else can help you pushing this forward? @juanjoDiaz @andreas-schroeder @akara  

Fixes #1111

andreas-schroeder:
@ennru I will take a look, I have used lettuce before, so it's kind of familiar to me.

redis/src/test/scala/docs/scaladsl/RedisFlowSpec.scala
```diff
@@ -0,0 +1,200 @@
...
+    "implement pub/sub for single topic and return notfication of single consumer" in {
+      val topic = "topic20"
+      RedisSource.subscribe(Seq(topic), pubSub).runWith(Sink.ignore)
+
+      val result = Source
+        .single("Bla")
+        .map(f => RedisPubSub(topic, f))
+        .via(RedisFlow.publish[String, String](1, redisClient.connectPubSub().async().getStatefulConnection))
+        .runWith(Sink.head[RedisOperationResult[RedisPubSub[String, String], Long]])
+      redisClient.connectPubSub().sync().unsubscribe(topic)
```
andreas-schroeder: This here could be avoided if the stage did unsubscribe on completion.

redis/src/main/scala/akka/streams/alpakka/redis/impl/AlpakkaPubSubListener.scala
```diff
@@ -0,0 +1,34 @@
...
+import akka.streams.alpakka.redis.{RedisPSubscribeResult, RedisPubSub}
+import io.lettuce.core.pubsub.RedisPubSubAdapter
+
+import scala.concurrent.Promise
+
+/**
+ * Internal API
+ */
+@InternalApi
+private[redis] class AlpakkaPubSubListener[K, V](
```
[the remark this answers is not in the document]
dannylesnik: Already removed.

redis/src/main/scala/akka/streams/alpakka/redis/javadsl/RedisFlow.scala
```diff
@@ -0,0 +1,59 @@
...
+  def append[K, V](
+      parallelism: Int,
+      connection: StatefulRedisConnection[K, V],
+      executionContext: ExecutionContext
+  ): Flow[RedisKeyValue[K, V], RedisOperationResult[RedisKeyValue[K, V], java.lang.Long], NotUsed] =
+    Flow
+      .fromGraph(
+        akka.streams.alpakka.redis.scaladsl.RedisFlow
+          .append(parallelism, connection)(executionContext)
+          .map(f => RedisOperationResult(f.output, f.result.map(t => java.lang.Long.valueOf(t.toString))))
```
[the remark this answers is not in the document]
dannylesnik: Indeed.

redis/src/main/scala/akka/streams/alpakka/redis/scaladsl/RedisFlow.scala
```diff
@@ -0,0 +1,78 @@
...
+      resultFuture.map(r => RedisOperationResult(redisSet, Success(r))).recover {
+        case ex: Throwable => RedisOperationResult(redisSet, Failure(ex))
+      }
+    }
+
+  def mset[K, V](parallelism: Int, connection: StatefulRedisConnection[K, V])(
+      implicit executionContext: ExecutionContext
+  ): Flow[Seq[RedisKeyValue[K, V]], RedisOperationResult[Seq[RedisKeyValue[K, V]], String], NotUsed] =
+    Flow[Seq[RedisKeyValue[K, V]]].mapAsync(parallelism) { sets =>
+      val map = sets.map(f => (f.key, f.value)).toMap.asJava
```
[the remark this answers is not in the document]
dannylesnik: 1. Can you please elaborate on "directly foreaching over the Seq"? foreach returns `Unit` and why this is better than `asJava` method.  
 
> We could have a flow in Impl that works with the Akka Scala API, but transports lettuce's expected types as stream items

2. This is exactly what I'm doing I have `scalaDsl.RedisFlow` as base implementation and all other Sinks and javadsl flow based on it. 

3. Not all Redis commands are good for items streaming,  mset or hmset works on bulks. It supposed to be bounded array of key values. For example I received response of key /values from one of previous stages. I transfer this collection of key values as single object. Therefore, the inpit for Stage supposed to be collection of items instead of stream of key/values.

redis/src/main/scala/akka/streams/alpakka/redis/scaladsl/RedisFlow.scala
```diff
@@ -0,0 +1,78 @@
...
+  def append[K, V](parallelism: Int, connection: StatefulRedisConnection[K, V])(
+      implicit executionContext: ExecutionContext
+  ): Flow[RedisKeyValue[K, V], RedisOperationResult[RedisKeyValue[K, V], Long], NotUsed] =
+    Flow[RedisKeyValue[K, V]].mapAsync(parallelism) { redisSet =>
+      connection
+        .async()
+        .append(redisSet.key, redisSet.value)
+        .toScala
+        .map(r => RedisOperationResult(redisSet, Success(r.longValue())))
+        .recover { case ex: Throwable => RedisOperationResult(redisSet, Failure(ex)) }
```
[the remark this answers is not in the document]
dannylesnik: I'll try to take a look I I can create and generic method.

redis/src/main/scala/akka/streams/alpakka/redis/scaladsl/RedisSource.scala
```diff
@@ -0,0 +1,38 @@
...
+import akka.stream.scaladsl.Source
+import akka.streams.alpakka.redis.RedisPubSub
+import io.lettuce.core.pubsub.StatefulRedisPubSubConnection
+import io.lettuce.core.pubsub.api.reactive.ChannelMessage
+
+object RedisSource {
+
+  def subscribe[K, V](topics: Seq[K],
+                      connection: StatefulRedisPubSubConnection[K, V]): Source[RedisPubSub[K, V], NotUsed] = {
+    connection.reactive().subscribe(topics: _*).subscribe()
```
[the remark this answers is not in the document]
dannylesnik: Very good point. Will implement it as part of the RedisFlow.

redis/src/main/scala/akka/streams/alpakka/redis/scaladsl/RedisFlow.scala
```diff
@@ -0,0 +1,78 @@
...
+import scala.collection.JavaConverters._
+import scala.compat.java8.FutureConverters._
+import scala.concurrent.ExecutionContext
+import scala.util.{Failure, Success}
+
+object RedisFlow {
+
+  def set[K, V](parallelism: Int, connection: StatefulRedisConnection[K, V])(
+      implicit executionContext: ExecutionContext
+  ): Flow[RedisKeyValue[K, V], RedisOperationResult[RedisKeyValue[K, V], String], NotUsed] =
```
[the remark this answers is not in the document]
dannylesnik: >  think parallelism and connection should be part of a settings object. This way, the API can evolve without breaking.

Yes. This is good point. Will do that. 

> I've seen that a couple other connectors require executionContext on flow creation - this is something I'd ask @ennru or @2m about their opinion on. I would rather have the stage use Akka Stream's dispatchers. How that would work is something you can see from ElasticSearch / Sqs connectors (using a stage producing Futures followed by mapAsync(parallelism)(identity)) For how to get the execution context in a stage, have a look at the JMS flow.

For getting Default Akka dispatcher I need to inherit GraphStageLogic and I don't see any reason for that except using Akka-Streams default dispatchers. I think it might be useful to use dedicated dispatcher for specific external IO operation and monitor it using Kamon.io or Cinnamon.   

> I'm also not sure about passing the connection into the flow operation vs. passing in the Redis client and creating a connection within the stage (and closing it once the stage completes). Can you elaborate why you chose the first option?

Doing that will be extremely expensive in terms of performance especially at high throughput in stage. Sometimes it might be cheaper to keep connection open than reopen it every time. Evolution of HTTP protocol from 1.0 to 1.1 and then to HTTP 2.0 is good example for that.

redis/src/test/scala/docs/scaladsl/RedisFlowSpec.scala
dannylesnik: Yes. will create From state for unsubscribe.

redis/src/main/scala/akka/streams/alpakka/redis/scaladsl/RedisFlow.scala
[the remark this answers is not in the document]
andreas-schroeder: 1. Sure:
```
val map = new java.util.HashMap[K, V]()
sets.foreach(f => map.put(f.key, f.value))
```
This would be to avoid the construction of an intermediate Scala map.

2. Sorry for being vague on my initial comment here. Yes, there is a base implementation in the `scaladsl` package that is used by the `javadsl` implementation and that is great. Looking at the implementation of e.g. `javadsl.RedisFlow.mset` as a whole, it converts the Java List to a Scala Seq first (https://github.com/akka/alpakka/pull/1350/files#diff-2bcc3be581d64d32cd8662ecab9163a8R49), and then converts the Seq to a Java Map (via a Scala Map). To me, it looks like the amount conversions can be reduced: I would suggest to have an internal mset Flow implementation that takes Java Maps, and use that in the public javadsl Flow (converting from a Java List to a Java Map), and also in the scaladsl Flow (converting from a Scala Seq to a Java Map). That way, the amount of collection transformation in the flow could be reduced.

3. mset and hmset are accept streams of item batches, sure. I also saw in the tests that the data type of the mset operation align with the output type of `grouped`, which is great.

redis/src/main/scala/akka/streams/alpakka/redis/scaladsl/RedisFlow.scala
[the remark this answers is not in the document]
andreas-schroeder: Sorry for being unclear in my comment on connections. I wouldn't recommend creating a new connection for each Redis operation - that would indeed reduce the max achievable throughput drastically. What I'm thinking about is lifting the burden of connection management from the user, and manage the connection in the Alpakka Redis connector. Looking at other connectors however (more specifically, the ElasticSearch one), it looks like passing the Redis connection into the stage and let the user take care of closing the connection is a pattern that others follow as well.

redis/src/main/scala/akka/streams/alpakka/redis/scaladsl/RedisFlow.scala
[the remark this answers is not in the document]
dannylesnik: 1. Frankly. I would be a little surprised if I saw such a collection creation using for-each implementation in any production code. 

> val map = new java.util.HashMap[K, V]()
sets.foreach(f => map.put(f.key, f.value))


Assuming that this code will run on Java 10 using G1 GC, I'm not sure that intermediate collection creation is such a big issue. It will be generated in JVM Eden space and will not survive first minor garbage collection. This is the the price that we are paying for working with immutable collections and datatypes. 
We have some production code which performs text analysis on documents using scala immutable collection and with right GC tuning I see 4-5 GigaBytes of String collections disappear from Eden immediately without causing any performance issue on the system. 

I'm really not sure that it worth to change implementation for skipping one immutable collection creation. @ennru  what do you think?

redis/src/main/scala/akka/streams/alpakka/redis/scaladsl/RedisSink.scala
```diff
@@ -0,0 +1,46 @@
...
+      implicit executionContext: ExecutionContext
+  ): Sink[RedisHSet[K, V], Future[Done]] =
+    RedisFlow.hset(parallelism, connection).toMat(Sink.ignore)(Keep.right)
+
+  def hmset[K, V](parallelism: Int, connection: StatefulRedisConnection[K, V])(
+      implicit executionContext: ExecutionContext
+  ): Sink[RedisHSet[K, V], Future[Done]] =
+    RedisFlow.hset(parallelism, connection).toMat(Sink.ignore)(Keep.right)
+
+}
```
akara: Do we actually need this file, or all the Sink signatures? Since all operations create a `RedisOperationResult` which contains the error indicators, using or providing this sink feels like adopting a norm of ignoring errors. No matter what happens, we all arrive at `Sink.ignore`.

redis/src/main/scala/akka/streams/alpakka/redis/scaladsl/RedisSource.scala
```diff
@@ -0,0 +1,60 @@
...
+import io.lettuce.core.pubsub.api.reactive.ChannelMessage
+import reactor.core.publisher.Flux
+
+import scala.collection.immutable.Seq
+
+object RedisSource {
+
+  def subscribe[K, V](topics: Seq[K],
+                      connection: StatefulRedisPubSubConnection[K, V]): Source[RedisPubSub[K, V], NotUsed] = {
+    connection.reactive().subscribe(topics: _*).subscribe()
```
akara: Kafka source materializes to `Control` and the JMS source materializes to `KillSwitch` to allow proper stop of subscription and proper stream shutdown. Shall we do something along those lines?

redis/src/test/scala/docs/scaladsl/RedisFlowSpec.scala
```diff
@@ -0,0 +1,200 @@
...
+
+      val recievedMessage = RedisSource.subscribe(Seq(topic), pubSub).runWith(Sink.head[RedisPubSub[String, String]])
+
+      Source
+        .single("Bla")
+        .map(f => RedisPubSub(topic, f))
+        .via(RedisFlow.publish[String, String](1, redisClient.connectPubSub().async().getStatefulConnection))
+        .runWith(Sink.head[RedisOperationResult[RedisPubSub[String, String], Long]])
+
+      val result = Await.result(recievedMessage, Duration(5, TimeUnit.SECONDS))
```
akara: Just from a code style standpoint, `Duration(5, TimeUnit.SECONDS)` can be replaced with just `5.seconds` if we have `import scala.concurrent.duration._`. I'd put that import upfront in the file as it is well used throughout. This is just a suggestion. If there are other reasons for using `Duration` directly? Please stick with the current style.

akara:
Another general question: If we get a source stage, subscribe, and `take(n)` where `n` is less than the published message, what will happen? Basically, we did not unsubscribe yet, but the stream has already stopped after `n` messages.

dannylesnik:
@akara In this API you can subscribe only in Source stage. Subscribing in Flow doesn't make any sense to me. In my latest commit I implemented unsubscribe as part of the flow Stream. So you csn subscribe in source, take n and unsubscribe as next stage.

redis/src/main/scala/akka/streams/alpakka/redis/model.scala
```diff
@@ -0,0 +1,145 @@
...
+ */
+
+package akka.streams.alpakka.redis
+
+import java.util.Optional
+import scala.collection.immutable.Seq
+import scala.util.{Failure, Success, Try}
+import scala.collection.JavaConverters._
+
+final class RedisFieldValue[K, V] private (val field: K, val value: V) {
```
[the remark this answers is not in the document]
dannylesnik: Yes. This because of javadsl.  In my previous contribution with Couchbase intrgration enru asked my to use this structures instead of case class. In general every object that can be created by user using javadsl can't be case class.

redis/src/main/scala/akka/streams/alpakka/redis/model.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  override def equals(other: Any): Boolean = other match {
+    case that: RedisFieldValue[K, V] =>
+      field == that.field &&
+      value == that.value
+    case _ => false
+  }
+
+  override def hashCode(): Int =
+    31 * field.hashCode() + 31 * value.hashCode()
```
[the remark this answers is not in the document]
dannylesnik: In my opinion every object that eventually could be put to HashMap or HashMap on any stage should have hashcode + equals implemented.

redis/src/main/scala/akka/streams/alpakka/redis/model.scala
```diff
@@ -0,0 +1,145 @@
...
+
+}
+
+final class RedisHMSet[K, V] private (val key: K, val values: Seq[RedisFieldValue[K, V]]) {
+
+  override def equals(other: Any): Boolean = other match {
+    case that: RedisHMSet[K, V] =>
+      key == that.key &&
+      values.length == that.values.length &&
+      values.zip(that.values).map { case (a, b) => a.equals(b) }.forall(_ == true)
```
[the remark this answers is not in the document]
dannylesnik: Agree. This is more compact. Fixed.

redis/src/main/scala/akka/streams/alpakka/redis/model.scala
```diff
@@ -0,0 +1,145 @@
...
+  override def equals(other: Any): Boolean = other match {
+    case that: RedisKeyValue[K, V] =>
+      key == that.key &&
+      value == that.value
+    case _ => false
+  }
+
+  override def hashCode(): Int =
+    31 * key.hashCode() + 31 * value.hashCode()
+}
```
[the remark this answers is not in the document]
dannylesnik: Unfortunately, supporting javadsl leads to some biolerplate code. Using case classes instead definitely would give as much more elegant code.

redis/src/main/scala/akka/streams/alpakka/redis/scaladsl/RedisSink.scala
dannylesnik: This is really good question and your comment makes sense but I prefer to give all possibles API for customer. So for example If user don't care about Redis operation result or don't mind  to lose some data in case master node is not available in the cluster, so he could use RedisSink instead of Flow. I did the same for Couchbase integration and mentioned this point in documentation.

redis/src/main/scala/akka/streams/alpakka/redis/scaladsl/RedisSource.scala
dannylesnik: I implemented unsubscribe as part of the flow, so user can unsubscribe when ever he wants/

redis/src/main/scala/akka/streams/alpakka/redis/model.scala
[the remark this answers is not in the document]
akara: You should still be able to use case classes while supporting javadsl. You just need to add the `create` method to the companion `object`. `apply`, `equals`, `hashCode` and `toString` can all be avoided :-) It is good to mark them `final`, too. You can see an example for such usages, here: https://github.com/akka/alpakka/blob/master/jms/src/main/scala/akka/stream/alpakka/jms/JmsMessages.scala#L65

But this is just a suggestion to keep the code smaller.

akara:
@dannylesnik Sorry for not being clear. I did not mean subscribing in a `Flow`. That definitely does not make sense. What I want to make sure is that the subscription of a source automatically terminates with the termination of the stream. There seems to be some ambiguity as the `unsubscribe` is separate and will terminate all subscriptions to a topic not pertaining to a particular stream. Basically, the scope seems to be global.

redis/src/main/scala/akka/streams/alpakka/redis/scaladsl/RedisSource.scala
```diff
@@ -0,0 +1,59 @@
...
+  def hget[K, V](key: K, field: K, connection: StatefulRedisConnection[K, V]): Source[RedisHSet[K, V], NotUsed] =
+    Source.fromPublisher(connection.reactive().hget(key, field)).map(f => RedisHSet(key, field, f))
+
+  def hmget[K, V](key: K,
+                  fields: Seq[K],
+                  connection: StatefulRedisConnection[K, V]): Source[RedisKeyValue[K, V], NotUsed] =
+    Source.fromPublisher(connection.reactive().hmget(key, fields: _*)).map(f => RedisKeyValue(f.getKey, f.getValue))
+
+  def hgetall[K, V](key: K,
+                    connection: StatefulRedisConnection[K, V]): Source[scala.Seq[RedisKeyValue[K, V]], NotUsed] =
```
2m: It would be nice to flatten this source to `Source[RedisKeyValue[K, V]], ...]`. Also, instead of using `Mono<Map<K, V>> hgetall(K key)` which brings whole `Map` at a time, it could use

```scala
Mono<Long> hgetall(KeyValueStreamingChannel<K, V> channel, K key);
```

which would stream keyvalues one by one as they arrive. But for that we will need to write a custom `KeyValueStreamingChannel` backed by a GraphStage.

redis/src/main/scala/akka/streams/alpakka/redis/scaladsl/RedisFlow.scala
```diff
@@ -0,0 +1,159 @@
...
+        .recover { case ex => RedisOperationResult(sets, Failure(ex)) }
+    }
+
+  def lpush[K, V](parallelism: Int, connection: StatefulRedisConnection[K, V])(
+      implicit executionContext: ExecutionContext
+  ): Flow[RedisKeyValues[K, V], RedisOperationResult[RedisKeyValues[K, V], Long], NotUsed] =
+    Flow[RedisKeyValues[K, V]].mapAsync(parallelism) { keyValues =>
+      val resultFuture = connection.async().lpush(keyValues.key, keyValues.values: _*).toScala
+      resultFuture.map(r => redis.RedisOperationResult(keyValues, Success(r.longValue()))).recover {
+        case ex: Throwable => ex.printStackTrace(); RedisOperationResult(keyValues, Failure(ex))
```
2m: `printStackTrace()` probably a leftover from debugging?

redis/src/main/scala/akka/streams/alpakka/redis/model.scala
```diff
@@ -0,0 +1,143 @@
...
+  def create[K, V](key: K, values: Array[V]): RedisKeyValues[K, V] = new RedisKeyValues(key, values)
+}
+
+object RedisKeyValue {
+
+  def apply[K, V](key: K, value: V): RedisKeyValue[K, V] = new RedisKeyValue(key, value)
+  def create[K, V](key: K, value: V) = new RedisKeyValue(key, value)
+}
+
+final case class RedisOperationResult[T, R] private (output: T, result: Try[R]) {
```
2m: As far as I can see, `output` always holds the request or the input to the operation. Should it be called `input` instead?

redis/src/main/scala/akka/streams/alpakka/redis/scaladsl/RedisSource.scala
```diff
@@ -0,0 +1,59 @@
...
+          .observeChannels()
+          .filter(new Predicate[ChannelMessage[K, V]] {
+            override def test(t: ChannelMessage[K, V]): Boolean = topics.contains(t.getChannel)
+          })
+      )
+      .map(f => RedisPubSub(f.getChannel, f.getMessage))
+  }
+
+  def get[K, V](key: K, connection: StatefulRedisConnection[K, V]): Source[RedisKeyValue[K, V], NotUsed] =
+    Source.fromPublisher(connection.reactive().get(key)).map(f => RedisKeyValue(key, f))
```
2m: This could be implemented using `RedisFlow` in the same way as all `RedisSink` operations are using `RedisFlow`.

redis/src/test/scala/docs/scaladsl/impl/SubscurbeListener.scala
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package docs.scaladsl.impl
+
+import java.util.concurrent.LinkedBlockingQueue
+
+import io.lettuce.core.pubsub.RedisPubSubListener
+
+class SubscurbeListener(subscribeQueue: LinkedBlockingQueue[String], unsubscribeQueue: LinkedBlockingQueue[String])
```
2m: Typo in the class name.

redis/src/main/scala/akka/streams/alpakka/redis/model.scala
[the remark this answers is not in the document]
2m: The main motivation behind not using case classes is binary compatibility concerns. We can not add or remove attributes to case classes and keep binary compatible API. This is mostly a problem when case classes are used for settings, because settings have a higher probability of changing.

However for general classes like `RedisKeyValue` we could leave it as a case class reducing the needed boilerplate. Also the `create` method for Java API in the companion object is great and is needed even for a case class.

redis/src/test/scala/docs/scaladsl/impl/SubscurbeListener.scala
dannylesnik: Fixed

redis/src/main/scala/akka/streams/alpakka/redis/scaladsl/RedisFlow.scala
dannylesnik: indeed :)

redis/src/main/scala/akka/streams/alpakka/redis/scaladsl/RedisSource.scala
dannylesnik: This is very correct approach, I"ll create implementation of hgetall, will ask to review it once it finished.

sourcekick:
@dannylesnik We are very interested in this and I would love to contribute. Can you finish your work on this and what would be the correct way for me to take it over, as in, from where should I fork?

@2m fyi

redis/src/main/scala/akka/streams/alpakka/redis/impl/RedisStreamingSourceStage.scala
```diff
@@ -0,0 +1,174 @@
...
+
+  def resultHandler(checkConcurrency: () => Boolean,
+                    successCallback: AsyncCallback[RedisKeyValue[K, V]],
+                    failureCallback: AsyncCallback[Exception]): RedisFuture[lang.Long] =
+    connection
+      .async()
+      .hmget(
+        new KeyValueStreamingChannel[K, V]() {
+          override def onKeyValue(key: K, value: V): Unit =
+            if (checkConcurrency()) {
```
2m: `checkConcurrency()` modifies GraphStage logic state. AFAICS this is called from a lettuce thread, right? In that case, to safely access and modify GtaphStage logic state this logic should go through Async GraphStage callback.

redis/src/main/scala/akka/streams/alpakka/redis/impl/RedisStreamingSourceStage.scala
```diff
@@ -0,0 +1,174 @@
...
+                                                 val connection: StatefulRedisConnection[K, V],
+                                                 val settings: RedisSubscriberSettings =
+                                                   RedisSubscriberSettings.Defaults)
+    extends RedisStreamingSourceStage[K, V] {
+
+  def resultHandler(checkConcurrency: () => Boolean,
+                    successCallback: AsyncCallback[RedisKeyValue[K, V]],
+                    failureCallback: AsyncCallback[Exception]): RedisFuture[lang.Long] =
+    connection
+      .async()
```
2m: I am not too familiar with lettuce API, but can we use `connection.reactive()` here and simplify this GraphStage?

2m:
@sourcekick you are free to continue from the work in this PR. It would be great if you could focus on getting some basic operations that you are finding useful working and have that PR merged. Then further work could be based on that.

redis/src/main/scala/akka/streams/alpakka/redis/impl/RedisStreamingSourceStage.scala
2m: I have tried it out, and it is possible to adapt `hmget` to akka-stream `Source` directly using the reactive lettuce client:

```scala
val source: Source[KeyValue[K, V], NotUsed] =
  Source.fromPublisher(connection.reactive().hmget(key, fields: _*))
```

redis/src/main/scala/akka/streams/alpakka/redis/impl/RedisStreamingSourceStage.scala
```diff
@@ -0,0 +1,174 @@
...
+                                                   override val settings: RedisSubscriberSettings =
+                                                     RedisSubscriberSettings.Defaults)
+    extends RedisStreamingSourceStage[K, V] {
+
+  def resultHandler(checkConcurrency: () => Boolean,
+                    successCallback: AsyncCallback[RedisKeyValue[K, V]],
+                    failureCallback: AsyncCallback[Exception]): RedisFuture[lang.Long] =
+    connection
+      .async()
+      .hgetall(
```
2m: `hgetall` can be adapted to akka-stream `Source` directly using lettuce reactive client:

```scala
val source: Source[KeyValue[K, V], NotUsed] =
  Source
    .fromPublisher(connection.reactive().hgetall(key))
    .mapConcat(_.entrySet().asScala.toIndexedSeq)
    .map(entry => KeyValue.just(entry.getKey, entry.getValue))
```

raboof:
Let's close this for now due to inactivity - if anyone wants to resume work on this feel free to re-open of course!

PR #1351, opened 2018-11-29
Solr: Use akka supervision strategies on error-handling

DmitryBMsk:
https://doc.akka.io/docs/akka/2.5.3/scala/stream/stream-error.html

### DIFF
--- a/solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
+++ b/solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
@@ -11,9 +11,10 @@ import org.apache.solr.client.solrj.SolrClient
 import org.apache.solr.client.solrj.impl.CloudSolrClient
 import org.apache.solr.client.solrj.response.UpdateResponse
 import org.apache.solr.common.SolrInputDocument
-
+import akka.stream.ActorAttributes.SupervisionStrategy
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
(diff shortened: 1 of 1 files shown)


solr/src/main/scala/akka/stream/alpakka/solr/SolrFlowStage.scala
```diff
@@ -241,10 +245,17 @@ private[solr] final class SolrFlowLogic[T, C](
...
   override def onPush(): Unit = {
     val messagesIn = grab(in)
+    try {
+      sendBulkToSolr(messagesIn)
+      tryPull()
+    } catch {
+      case NonFatal(ex) ⇒
+        decider(ex) match {
+          case Supervision.Stop ⇒ failStage(ex)
+          case _ ⇒ tryPull() // for resume and restart strategies tryPull
```
ennru: This will drop all messages in `messagesIn`, applying it in `send` could reduce the number of messages dropped.

ennru: Was that the intended behaviour?

DmitryBMsk: This fix allow to use restart or resume strategies  on Supervision.Decider like this:
https://gist.github.com/DmitryBMsk/1ce60ce8f565527f242ca6b4ab70132b
Otherwise any stream may fail during long update process in Solr instead of resuming/restarting.

ennru:
Thank you for adding this!

PR #1358, opened 2018-12-06
Scala 2.12.8

sullis:


ennru:
The compilation error is a regression in Scala 2.12.8: https://github.com/scala/bug/issues/11305

2m:
The linked bug fix is targeted for 2.12.9. Lets wait for that and skip 2.12.8.

2m:
ftp connector does not compile with Scala 2.13.0-M5, with the same error.

PR #1372, opened 2018-12-13
FTP: Upgrade dependencies

ennru:
Minor API change in the Apache SSHD server used for testing.


ennru:
I had hope this upgrade would improve on test failures, but it didn't https://travis-ci.org/akka/alpakka/jobs/467557674#L883

Issue #1353, opened 2018-11-30
XML: ParseEventMarker in wrong package


ennru:
It lives in `akka.stream.alpakka`, but should be in `akka.stream.alpakka.xml`.

PR #1373, opened 2018-12-13
Moving ParseEventMarker to akka.stream.alpakka.xml package

scoquelin:
Tackling #1353


ennru:
Thank you for fixing it.

PR #1375, opened 2018-12-16
Fix  "Overflow from watch service" error for using in events  based on  +500 files in watched directory

DmitryBMsk:
This fix will allow to use DirectoryChangesSource with large amout of arriving files(>500) ant prevent failstage with "Upstream failed, cause: RuntimeException: Overflow from watch service: " error.
See https://stackoverflow.com/questions/18586091/in-jdk7-watch-service-api-when-will-the-overflow-event-be-thrown

### DIFF
--- a/file/src/main/java/akka/stream/alpakka/file/impl/DirectoryChangesSource.java
+++ b/file/src/main/java/akka/stream/alpakka/file/impl/DirectoryChangesSource.java
@@ -22,6 +22,7 @@ import scala.concurrent.duration.FiniteDuration;
 import java.io.IOException;
 import java.nio.file.*;
 import java.util.ArrayDeque;
+import java.util.List;
 import java.util.Queue;
 import java.util.function.BiFunction;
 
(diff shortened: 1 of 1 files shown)


file/src/main/java/akka/stream/alpakka/file/impl/DirectoryChangesSource.java
```diff
@@ -157,22 +158,26 @@ private void doPoll() {
...
               // that may be ok for some scenarios but to make sure it does not pass unnoticed we
               // fail the stage
-              failStage(
-                  new RuntimeException("Overflow from watch service: '" + directoryPath + "'"));
-
+              //to prevent "overflow from watch service error" if there is an trigger on arriving  >~500 files
+              try {
+                Files.list(directoryPath)
+                    .filter(t -> t.toFile().isFile())
+                    .forEach(t -> buffer.add(combiner.apply(t, DirectoryChange.Creation))); //use "Creation" event by default
```
ennru: So, on overflow, all files in the directory will be reported as `Creation` even files that were there before, right?

DmitryBMsk: Yes, you are right.

DmitryBMsk:
@2m , agree with you. Using RecoverWith or RestartSource looks much better.  I am looking forward to make another PR to deal with "overflow" error. Thanks

2m:
No problem. Looking forward to that!

ennru:
Closing this now, we still hope to see the suggested additions to the docs.

PR #1369, opened 2018-12-12
Filter generated protobuf classes from scaladoc

2m:
Fixes #1367

PR #1379, opened 2018-12-17
WIP: Build paradox and api docs with sbt-site

2m:
WIP - do not merge before we switch doc deployment from techhub to gustav.

Builds on top of https://github.com/akka/alpakka/pull/1369


2m:
Changed doc symlink creation only for the stable release, so that doc server is not polluted with old snapshot docs.

PR #1383, opened 2018-12-18
Use correct branch on travis for whitesource

raboof:


2m:
Hm. I wonder why is that required? If Travis is building master commit, the output of `git describe --all` is already `heads/master`

raboof:
I think because travis does a `git checkout -qf 6cd97b4e7903c796b8a001ca36f6da053a69e43a` to make sure it actually builds the commit that triggered the build, not whatever is on 'master' at the time the build starts.

Issue #991, opened 2018-05-31
Failed: FtpsStageSpec FtpRemoveSink should remove a file


2m:
```
[info] FtpRemoveSink
[info] - should remove a file *** FAILED *** (612 milliseconds)
[info]   IOResult(1,Failure(javax.net.ssl.SSLException: bad record MAC)) was not equal to IOResult(1,Success(Done)) (CommonFtpStageSpec.scala:298)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.MatchersHelper$.indicateFailure(MatchersHelper.scala:346)
[info]   at org.scalatest.Matchers$AnyShouldWrapper.shouldBe(Matchers.scala:6864)
[info]   at akka.stream.alpakka.ftp.CommonFtpStageSpec$$anonfun$6$$anonfun$apply$mcV$sp$19.apply(CommonFtpStageSpec.scala:298)
[info]   at akka.stream.alpakka.ftp.CommonFtpStageSpec$$anonfun$6$$anonfun$apply$mcV$sp$19.apply(CommonFtpStageSpec.scala:290)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.WordSpecLike$$anon$1.apply(WordSpecLike.scala:1078)
[info]   at org.scalatest.TestSuite$class.withFixture(TestSuite.scala:196)
[info]   at akka.stream.alpakka.ftp.FtpsStageSpec.withFixture(CommonFtpStageSpec.scala:23)
```

https://travis-ci.org/2m/alpakka/jobs/386054111

2m:
There seems to be some spurious test failures in various specs caused by `javax.net.ssl.SSLException: bad record MAC` exception.

A recent one: https://travis-ci.org/akka/alpakka/jobs/393496260

2m:
https://travis-ci.org/akka/alpakka/jobs/396448017

2m:
https://travis-ci.org/akka/alpakka/jobs/403107035

2m:
https://travis-ci.org/akka/alpakka/jobs/438480372

```
[info] FtpMoveSink
[...]

PR #1386, opened 2018-12-19
Ignore mongodb classes from scaladoc

2m:
Another mystery scaladoc appearance.


2m:
Failure was https://github.com/akka/alpakka/issues/991

PR #1387, opened 2018-12-19
MQTT streaming: Connection hardening

huntc:
When connections are lost the client and server flows now shutdown appropriately. This was not always the case previously.

In addition, the stashing was artificially limited which could cause problems. Overflowing memory is now entirely a caller concern.

### DIFF
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/MqttSessionSettings.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/MqttSessionSettings.scala
@@ -29,7 +29,6 @@ object MqttSessionSettings {
  * Configuration settings for client and server usage.
  */
 final class MqttSessionSettings private (val maxPacketSize: Int = 4096,
-                                         val maxClientConnectionStashSize: Int = 100,
                                          val clientTerminationWatcherBufferSize: Int = 100,
                                          val commandParallelism: Int = 50,
                                          val eventParallelism: Int = 10,
@@ -183,14 +182,6 @@ final class MqttSessionSettings private (val maxPacketSize: Int = 4096,
     copy(receiveUnsubAckTimeout = receiveUnsubAckTimeout.asScala)
 
   /**
-   * For clients, the number of commands to buffer in memory while waiting for a connection ack. For servers,
-   * the number of commands to buffer in memory while waiting for a local connection ack command. Defaults
-   * to 100.
-   */
-  def withMaxClientConnectionStashSize(maxClientConnectionStashSize: Int): MqttSessionSettings =
-    copy(maxClientConnectionStashSize = maxClientConnectionStashSize)
-
-  /**
    * The maximum number of client termination event observers permitted. Defaults to 100 which should be
    * more than adequate for most situations.
    */
@@ -205,7 +196,6 @@ final class MqttSessionSettings private (val maxPacketSize: Int = 4096,
     copy(serverSendBufferSize = serverSendBufferSize)
 
   private def copy(maxPacketSize: Int = maxPacketSize,
-                   maxClientConnectionStashSize: Int = maxClientConnectionStashSize,
                    clientTerminationWatcherBufferSize: Int = clientTerminationWatcherBufferSize,
                    commandParallelism: Int = commandParallelism,
                    eventParallelism: Int = eventParallelism,
@@ -219,7 +209,6 @@ final class MqttSessionSettings private (val maxPacketSize: Int = 4096,
                    serverSendBufferSize: Int = serverSendBufferSize) =
     new MqttSessionSettings(
       maxPacketSize,
-      maxClientConnectionStashSize,
       clientTerminationWatcherBufferSize,
       commandParallelism,
       eventParallelism,
@@ -234,5 +223,5 @@ final class MqttSessionSettings private (val maxPacketSize: Int = 4096,
     )
 
   override def toString: String =
-    s"MqttSessionSettings(maxPacketSize=$maxPacketSize,maxClientConnectionStashSize=$maxClientConnectionStashSize,clientTerminationWatcherBufferSize=$clientTerminationWatcherBufferSize,commandParallelism=$commandParallelism,eventParallelism=$eventParallelism,receiveConnectTimeout=$receiveConnectTimeout,receiveConnAckTimeout=$receiveConnAckTimeout,receivePubAckRecTimeout=$receivePubAckRecTimeout,receivePubCompTimeout=$receivePubCompTimeout,receivePubRelTimeout=$receivePubRelTimeout,receiveSubAckTimeout=$receiveSubAckTimeout,receiveUnsubAckTimeout=$receiveUnsubAckTimeout,serverSendBufferSize=$serverSendBufferSize)"
+    s"MqttSessionSettings(maxPacketSize=$maxPacketSize,clientTerminationWatcherBufferSize=$clientTerminationWatcherBufferSize,commandParallelism=$commandParallelism,eventParallelism=$eventParallelism,receiveConnectTimeout=$receiveConnectTimeout,receiveConnAckTimeout=$receiveConnAckTimeout,receivePubAckRecTimeout=$receivePubAckRecTimeout,receivePubCompTimeout=$receivePubCompTimeout,receivePubRelTimeout=$receivePubRelTimeout,receiveSubAckTimeout=$receiveSubAckTimeout,receiveUnsubAckTimeout=$receiveUnsubAckTimeout,serverSendBufferSize=$serverSendBufferSize)"
 }
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
@@ -53,19 +53,21 @@ import scala.util.{Failure, Success}
 
   // Our FSM data, FSM events and commands emitted by the FSM
 
-  sealed abstract class Data(val consumerPacketRouter: ActorRef[RemotePacketRouter.Request[Consumer.Event]],
+  sealed abstract class Data(val stash: Seq[Event],
+                             val consumerPacketRouter: ActorRef[RemotePacketRouter.Request[Consumer.Event]],
                              val producerPacketRouter: ActorRef[LocalPacketRouter.Request[Producer.Event]],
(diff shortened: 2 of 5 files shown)


huntc:
@ennru @2m I'm having difficulty with BidiFlows and would really appreciate some help. The problem area is described by https://github.com/akka/akka/issues/17997#issuecomment-429670321. The particular test that fails here is at https://github.com/akka/alpakka/pull/1387/files#diff-4e77dbf4c34133b8ec8d5d84f75be18aR205 as the upstream of the BidiFlow completes, but the downstream doesn't.

huntc:
Following on from this, I've disabled the failing test until I can understand how BidiFlow is supposed to behave. @ennru This is now ready for a review.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/MqttSessionSettings.scala
```diff
@@ -182,14 +181,6 @@ final class MqttSessionSettings private (val maxPacketSize: Int = 4096,
...
     copy(receiveUnsubAckTimeout = receiveUnsubAckTimeout.asScala)
 
-  /**
-   * For clients, the number of commands to buffer in memory while waiting for a connection ack. For servers,
-   * the number of commands to buffer in memory while waiting for a local connection ack command. Defaults
-   * to 100.
-   */
-  def withMaxClientConnectionStashSize(maxClientConnectionStashSize: Int): MqttSessionSettings =
-    copy(maxClientConnectionStashSize = maxClientConnectionStashSize)
-
```
huntc: Better to let the call control the number of messages being sent rather than introducing an arbitrary limit.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -53,34 +53,37 @@ import scala.util.{Failure, Success}
 
   // Our FSM data, FSM events and commands emitted by the FSM
 
-  sealed abstract class Data(val consumerPacketRouter: ActorRef[RemotePacketRouter.Request[Consumer.Event]],
+  sealed abstract class Data(val stash: Seq[Event],
+                             val consumerPacketRouter: ActorRef[RemotePacketRouter.Request[Consumer.Event]],
                              val producerPacketRouter: ActorRef[LocalPacketRouter.Request[Producer.Event]],
                              val subscriberPacketRouter: ActorRef[LocalPacketRouter.Request[Subscriber.Event]],
                              val unsubscriberPacketRouter: ActorRef[LocalPacketRouter.Request[Unsubscriber.Event]],
                              val settings: MqttSessionSettings)
```
huntc: The stash is now common to all of the states here.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -183,17 +189,17 @@ import scala.util.{Failure, Success}
...
           data.producerPacketRouter,
           data.subscriberPacketRouter,
           data.unsubscriberPacketRouter,
           data.settings
         )
       )
-    case (_, prl: PublishReceivedLocally) if data.stash.size < data.settings.maxClientConnectionStashSize =>
-      disconnected(data.copy(stash = data.stash :+ prl))
+    case (_, e) =>
+      disconnected(data.copy(stash = data.stash :+ e))
```
huntc: I'm unsure why this was previously constrained in terms of what messages should be stashed. We could have easily dropped some accidentally.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -249,7 +256,9 @@ import scala.util.{Failure, Success}
             disconnect(context, data.connect.connectFlags, data.remote, data)
           case (context, ReceiveConnAckTimeout) =>
             disconnect(context, data.connect.connectFlags, data.remote, data)
-          case (_, e) if data.stash.size < data.settings.maxClientConnectionStashSize =>
+          case (context, ConnectionLost) =>
+            disconnect(context, data.connect.connectFlags, data.remote, data)
+          case (_, e) =>
```
huntc: We now handle a connection being lost within the client side. This had been forgotten.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ServerState.scala
```diff
@@ -596,6 +598,21 @@ import scala.util.{Failure, Success}
...
+                data.activeProducers,
+                data.pendingLocalPublications,
+                data.pendingRemotePublications,
+                data.consumerPacketRouter,
+                data.producerPacketRouter,
+                data.publisherPacketRouter,
+                data.unpublisherPacketRouter,
+                data.settings
+              )
+            )
```
huntc: Same for the server - we now move to the disconnected state if we have lost the connection.

mqtt-streaming/src/test/scala/docs/scaladsl/MqttSessionSpec.scala
```diff
@@ -103,20 +103,23 @@ class MqttSessionSpec
       val connect = Connect("some-client-id", ConnectFlags.None)
       val carry = "some-carry"
 
-      val result =
+      val (client, result) =
         Source
-          .single(Command(connect, carry))
+          .queue[Command[String]](1, OverflowStrategy.fail)
```
huntc: This and the next test are changed so that we don't complete the stream prematurely. Now that we process the connection being lost this and the next test would fail without these changes (and rightly so).

mqtt-streaming/src/test/scala/docs/scaladsl/MqttSessionSpec.scala
```diff
@@ -1025,8 +1063,7 @@ class MqttSessionSpec
             case Right(Event(cp: Publish, _)) if cp.topicName == publish.topicName =>
               publishReceived.success(Done)
           })
-          .drop(4)
-          .toMat(Sink.head)(Keep.left)
+          .toMat(Sink.ignore)(Keep.left)
```
huntc: We no longer require a result and thus the `head`.

huntc:
Actually, I do know how to complete the downstream if the upstream completes. It is just that I cannot seem to generalise it within the `Mqtt.clientSessionFlow`/`Mqtt.serverSessionFlow` implementations.

Here's how I can express the test where I have a concrete type parameter:

```scala
      import mat.executionContext
      val commandFlowCompleted = Promise[Done]
      val (client, result) = Source
        .queue[Command[Nothing]](1, OverflowStrategy.fail)
        .watchTermination() {
          case (e, done) =>
            done.foreach(commandFlowCompleted.success)
            e
        }
        .via(
          Mqtt
            .clientSessionFlow(session)
            .join(pipeToServer)
        )
        .merge(Source
                 .fromFuture(commandFlowCompleted.future)
                 .flatMapConcat(_ => Source.empty[Either[MqttCodec.DecodeError, Event[Nothing]]]),
               eagerComplete = true)
        .toMat(Sink.ignore)(Keep.both)
        .run()
```

huntc:
Thanks @2m. I’m constructing flows without the use of fromSinkAndSourceCoupled. I will look into the implementation of that to see if I can replicate it within my MQTT session flow factory functions.

Perhaps BidiFlow could benefit from introducing the coupling functionality also. It isn’t clear what the behaviour is supposed to be re completion. If you could spur the team to clarify that behaviour in the issue I referenced then that’d be great.

huntc:
@2m I have now generalised it on my own branch... but it involves an execution context for creating a promise that is shared between the two flows. This would mean that the resulting BidiFlow would only be able to be used just once. So, while it works, I'm not entirely satisfied. Any further thoughts? Again, in essence, I want a BidiFlow's downstream to complete when its upstream completes. Thanks, and here's what I've got so far:

```scala
  /**
   * Create a bidirectional flow that maintains client session state with an MQTT endpoint.
   * The bidirectional flow can be joined with an endpoint flow that receives
   * [[ByteString]] payloads and independently produces [[ByteString]] payloads e.g.
   * an MQTT server.
   *
   * @param session the MQTT client session to use
   * @return the bidirectional flow
   */
  def clientSessionFlow[A](
      session: MqttClientSession
  )(
      implicit mat: Materializer
  ): BidiFlow[Command[A], ByteString, ByteString, Either[MqttCodec.DecodeError, Event[A]], NotUsed] = {
    import mat.executionContext
    val commandFlowCompleted = Promise[Done]
    BidiFlow.fromFlows(
      session.commandFlow[A].watchTermination() {
        case (e, done) =>
          done.foreach(commandFlowCompleted.success)
          e
      },
      session
        .eventFlow[A]
        .merge(
          Source
            .fromFuture(commandFlowCompleted.future)
            .flatMapConcat(_ => Source.empty),
          eagerComplete = true
        )
    )
  }
```

2m:
Nice progress, but yes, such BidiFlow will only be possible to be materialized once. You could extend this idea and move the Promise creation to a custom graph stage createLogic method. Then it would be ensured that a new promise is created upon every materialization.

huntc:
Thanks. Yes, I had been thinking of a custom graph stage. Do you know of any custom graph stage examples of BidiFlow?

2m:
Yes, there are a couple in akka-http (for example [this](https://github.com/akka/akka-http/blob/ff1258ec2be4d658f143d1b7dd95d70d31744d1e/akka-http-core/src/main/scala/akka/http/impl/util/One2OneBidiFlow.scala)). Good way to find those is to search for `BidiShape`.

huntc:
@2m Fixed! Check it out: https://github.com/akka/alpakka/pull/1387/files#diff-83a6e8f5e901b242a2c35556d64896acR27

This is now ready for a final review and (hopefully!) a merge. cc. @ennru

mqtt-streaming/src/test/scala/docs/scaladsl/MqttSessionSpec.scala
```diff
@@ -194,6 +202,35 @@ class MqttSessionSpec
       server.expectMsg(disconnectBytes)
     }
 
+    "disconnect when connection lost" in {
```
huntc: This test principally represents what the PR is about.

2m:
Nice. The `CoupledTerminationBidi` has a potential to be moved to akka-stream-contrib (which eventually will be merged into Alpakka repo).

PR #1394, opened 2018-12-21
MQTT streaming: More connection failure hardening

huntc:
We've been testing the MQTT connector in a production environment, hence discovering a few bugs. The production environment consists of a very low powered CPU and so the scenarios being fixed are largely as a result of a completely different timing behaviour to that of much faster machines.

The collection of commits here largely address an oversight on my part where I had misunderstood `flatMapMerge` to complete on a `Source` completing. It doesn't, and it shouldn't, and I don't know what I was thinking. :-)

I've now introduced a kill switch so that when a disconnection occurs within a client or server session then their associated command flows are terminated. I've enhanced a test to prove this out.

I've made some inline comments of other little things that I've fixed.

Finally, I've done some observational testing with our code that uses this library. We will be continuing to perform these hardening tests in the near term.

### DIFF
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
@@ -276,8 +276,6 @@ import scala.util.{Failure, Success}
     timer =>
       if (data.keepAlive.toMillis > 0)
         timer.startSingleTimer("send-pingreq", SendPingReqTimeout, data.keepAlive)
-      else
-        data.remote.complete() // We'll never be sending pings so free up the command channel for other things
 
       Behaviors
         .receivePartial[Event] {
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
@@ -198,6 +198,8 @@ import scala.util.{Failure, Success}
           data.settings
         )
       )
+    case (_, ConnectionLost) =>
+      Behavior.same
     case (_, e) =>
       disconnected(data.copy(stash = data.stash :+ e))
   }
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
@@ -221,9 +221,12 @@ import scala.util.{Failure, Success}
     )
   }
 
+  private val ReceiveConnAck = "receive-connack"
+
   def serverConnect(data: ConnectReceived)(implicit mat: Materializer): Behavior[Event] = Behaviors.withTimers {
     timer =>
-      timer.startSingleTimer("receive-connack", ReceiveConnAckTimeout, data.settings.receiveConnAckTimeout)
+      if (!timer.isTimerActive(ReceiveConnAck))
+        timer.startSingleTimer(ReceiveConnAck, ReceiveConnAckTimeout, data.settings.receiveConnAckTimeout)
 
       Behaviors
         .receivePartial[Event] {
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSession.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSession.scala
@@ -9,7 +9,7 @@ import java.util.concurrent.atomic.AtomicLong
 
 import akka.{NotUsed, actor => untyped}
 import akka.actor.typed.scaladsl.adapter._
-import akka.stream.{ActorAttributes, Materializer, OverflowStrategy, Supervision}
+import akka.stream._
 import akka.stream.alpakka.mqtt.streaming.impl._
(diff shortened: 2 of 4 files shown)


mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -198,6 +198,8 @@ import scala.util.{Failure, Success}
           data.settings
         )
       )
+    case (_, ConnectionLost) =>
+      Behavior.same
```
huntc: I was accidentally stashing a lost connection while being in a disconnected state. No need - and indeed it can then lead to a valid connection being cancelled.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -221,9 +223,12 @@ import scala.util.{Failure, Success}
     )
   }
 
+  private val ReceiveConnAck = "receive-connack"
+
   def serverConnect(data: ConnectReceived)(implicit mat: Materializer): Behavior[Event] = Behaviors.withTimers {
     timer =>
-      timer.startSingleTimer("receive-connack", ReceiveConnAckTimeout, data.settings.receiveConnAckTimeout)
+      if (!timer.isTimerActive(ReceiveConnAck))
+        timer.startSingleTimer(ReceiveConnAck, ReceiveConnAckTimeout, data.settings.receiveConnAckTimeout)
```
huntc: I was accidentally resetting the timer when I was stashing. We now ensure that the timer is set only if it has not been already.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -273,8 +278,6 @@ import scala.util.{Failure, Success}
     timer =>
       if (data.keepAlive.toMillis > 0)
         timer.startSingleTimer("send-pingreq", SendPingReqTimeout, data.keepAlive)
-      else
-        data.remote.complete() // We'll never be sending pings so free up the command channel for other things
```
huntc: This was a hangover from an older approach, and is no longer valid - there can now be many uses of the `data.remote` `Source`, beyond `PINGREQ`.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSession.scala
```diff
@@ -162,8 +165,9 @@ final class ActorMqttClientSession(settings: MqttSessionSettings)(implicit mat:
                 case ClientConnector.ForwardPubRel(packetId) =>
                   PubRel(packetId).encode(ByteString.newBuilder).result()
               }.mapError {
-                case ClientConnector.PingFailed => ActorMqttClientSession.PingFailed
-              })
+                  case ClientConnector.PingFailed => ActorMqttClientSession.PingFailed
+                }
+                .watchTermination()((_, done) => done.foreach(_ => killSwitch.shutdown())))
             )
```
huntc: This is the crux of the change for the PR. When our connect stream ends it means that the session has moved to a disconnected state. We, therefore, need to kill upstream.

mqtt-streaming/src/test/scala/docs/scaladsl/MqttSessionSpec.scala
```diff
@@ -200,6 +200,8 @@ class MqttSessionSpec
       client.offer(Command(disconnect))
 
       server.expectMsg(disconnectBytes)
+
+      result.futureValue shouldBe Done
```
huntc: Prior to the kill switch change, the stream would not complete and this new assertion would fail. The kill switch ensures that the stream completes.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -221,9 +223,12 @@ import scala.util.{Failure, Success}
     )
   }
 
+  private val ReceiveConnAck = "receive-connack"
```
ennru: I'd like to recommend case objects as timer identifiers.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSession.scala
```diff
@@ -162,8 +165,9 @@ final class ActorMqttClientSession(settings: MqttSessionSettings)(implicit mat:
                 case ClientConnector.ForwardPubRel(packetId) =>
                   PubRel(packetId).encode(ByteString.newBuilder).result()
               }.mapError {
-                case ClientConnector.PingFailed => ActorMqttClientSession.PingFailed
-              })
+                  case ClientConnector.PingFailed => ActorMqttClientSession.PingFailed
```
ennru: Indentation looks weird, but the scalafmt check is happy.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
huntc: Why so? More space used...

ennru: Cheaper equality, never mind.

PR #1406, opened 2019-01-05
akka-http 10.1.7

sullis:


2m:
Thanks for the PR. A couple of tests failed where assertions depend on the order of Http response content which changed with the new Akka Http version.

Would you be up for fixing those testcases? Or even better yet refactoring testscases so they are order independent?

2m:
Superseded by #1422

Issue #832, opened 2018-03-12
Remaining apiKey params can be removed for google pub/sub


francisdb:
As far as I can see these are not used, probably leftover after #667

remove from methods
remove from docs/examples

PR #1407, opened 2019-01-08
google-cloud-pub-sub: remove redundant api key

stephennancekivell:
Fixes issue https://github.com/akka/alpakka/issues/832

The usage of the api key was removed in https://github.com/akka/alpakka/pull/683

Hopefully we can get this in before 1.0 as it will be a breaking change for the scala/java api.


ennru:
Thank you for cleaning this up.

stephennancekivell:
No worries. Thanks

Issue #1377, opened 2018-12-17
Failed: UnixDomainSocketSpec A Unix Domain Socket should receive what is sent


2m:
```
[info] UnixDomainSocketSpec:
[info] A Unix Domain Socket
[info] - should receive what is sent *** FAILED *** (656 milliseconds)
[info]   java.util.NoSuchElementException: head of empty stream
[info]   at akka.stream.scaladsl.Sink$$anonfun$head$1$$anonfun$apply$2$$anonfun$apply$3.apply(Sink.scala:174)
[info]   at akka.stream.scaladsl.Sink$$anonfun$head$1$$anonfun$apply$2$$anonfun$apply$3.apply(Sink.scala:174)
[info]   at scala.Option.getOrElse(Option.scala:121)
[info]   at akka.stream.scaladsl.Sink$$anonfun$head$1$$anonfun$apply$2.apply(Sink.scala:174)
[info]   at akka.stream.scaladsl.Sink$$anonfun$head$1$$anonfun$apply$2.apply(Sink.scala:174)
[info]   at scala.util.Success$$anonfun$map$1.apply(Try.scala:237)
[info]   at scala.util.Try$.apply(Try.scala:192)
[info]   at scala.util.Success.map(Try.scala:237)
[info]   at scala.concurrent.Future$$anonfun$map$1.apply(Future.scala:237)
```

https://travis-ci.org/akka/alpakka/jobs/467872221

2m:
Different testcase, same cause: https://travis-ci.org/akka/alpakka/jobs/475052701

huntc:
Sorry for the delay - I must get to this...

2m:
https://travis-ci.org/akka/alpakka/jobs/486887543

2m:
https://travis-ci.org/akka/alpakka/jobs/487365089

ennru:
https://travis-ci.org/akka/alpakka/jobs/491569285#L644

2m:
https://travis-ci.org/akka/alpakka/jobs/493140517

raboof:
https://travis-ci.com/akka/alpakka/jobs/283084979 looks a bit different now, but seems like the same problem:
```
[error] Test docs.javadsl.UnixDomainSocketTest.aUnixDomainSocketShouldReceiveWhatIsSent failed: java.util.concurrent.TimeoutException: null, took 10.444 sec
[...]

PR #1412, opened 2019-01-11
Change Alpakka URLs to point directly to its new home

ennru:


2m:
Failure was https://github.com/akka/alpakka/issues/1377

PR #1415, opened 2019-01-13
AWS SQS: use ElasticMQ in Docker container instead of embedded

### LINKED ISSUE #1154: AWS SQS: use ElasticMQ in Docker container instead of embedded
To minimize interference between the Akka-based ElasticMQ and the tests running, the integration test should use ElasticMQ's Docker image instead.

RustedBones:
Fixes #1154

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,7 +94,9 @@ jobs:
       - PRE_CMD="docker-compose up -d amazonsns"
       - CMD=+sns/testChanged
     - env: CMD=+solr/testChanged
-    - env: CMD=+sqs/testChanged
+    - env:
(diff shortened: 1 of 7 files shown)


RustedBones:
Seems that `ChangeMessageVisibilityBatch` is not yet implemented in [goaws](https://github.com/p4tin/goaws#snssqs-api-status).
I would suggest to change when the full API will be supported

RustedBones:
I know there is also the [localstack](https://github.com/localstack/localstack) project that aims to simulate all AWS services in a single docker image.

Should we consider it since it can be use for all alpakka AWS connectors?

ennru:
"localstack" would be interesting to explore, but not necessarily more useful than specific images for certain connectors.

PR #1418, opened 2019-01-14
Fail the client when no CONNACK or SUBACK

huntc:
Prior to this PR, a lack of a CONNACK would result in the client stream completing, but with no explanation. We now fail the client stream with an appropriate exception.

Also, the lack of a SUBACK would be silently ignored. This PR makes the lack of a SUBACK consistent with a lack of a CONNACK in that the client stream will be failed with an appropriate exception.

Also, a small improvement to the command logging so that we can capture any generated packet ids.

### DIFF
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
@@ -6,7 +6,7 @@ package akka.stream.alpakka.mqtt.streaming
 package impl
 
 import akka.NotUsed
-import akka.actor.typed.{ActorRef, Behavior, PostStop, Terminated}
+import akka.actor.typed.{ActorRef, Behavior, ChildFailed, PostStop, Terminated}
 import akka.actor.typed.scaladsl.{ActorContext, Behaviors}
 import akka.annotation.InternalApi
 import akka.stream.{Materializer, OverflowStrategy}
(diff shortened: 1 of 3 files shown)


2m:
One of the newly added test-cases failed:

```
[info] - should Connect and fail given no ack *** FAILED *** (3 seconds, 18 milliseconds)
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsg while waiting for ByteString(16, 26, 0, 4, 77, 81, 84, 84, 4, 0, 0, 60, 0, 14, 115, 111, 109, 101, 45, 99, 108, 105, 101, 110, 116, 45, 105, 100)
[info]   at scala.Predef$.assert(Predef.scala:170)
[info]   at akka.testkit.TestKitBase$class.expectMsg_internal(TestKit.scala:402)
[info]   at akka.testkit.TestKitBase$class.expectMsg(TestKit.scala:379)
[info]   at akka.testkit.TestKit.expectMsg(TestKit.scala:896)
[info]   at docs.scaladsl.MqttSessionSpec$$anonfun$1$$anonfun$apply$mcV$sp$9.apply(MqttSessionSpec.scala:193)
[info]   at docs.scaladsl.MqttSessionSpec$$anonfun$1$$anonfun$apply$mcV$sp$9.apply(MqttSessionSpec.scala:170)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
```

huntc:
Thanks for that. These two new tests should now be fixed.

Issue #1360, opened 2018-12-07
Failed: docs.javadsl.SqsPublishTest.sendMessageWithBatchesAsFlow


2m:
```
[error] Test docs.javadsl.SqsPublishTest.sendMessageWithBatchesAsFlow failed: java.util.concurrent.TimeoutException: null, took 2.25 sec
[error]     at java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1771)
[error]     at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1915)
[error]     at scala.concurrent.java8.FuturesConvertersImpl$CF.super$get(FutureConvertersImpl.scala:84)
[error]     at scala.concurrent.java8.FuturesConvertersImpl$CF.$anonfun$get$2(FutureConvertersImpl.scala:84)
[error]     at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:53)
[error]     at scala.concurrent.package$.blocking(package.scala:142)
[error]     at scala.concurrent.java8.FuturesConvertersImpl$CF.get(FutureConvertersImpl.scala:84)
[error]     at docs.javadsl.SqsPublishTest.sendMessageWithBatchesAsFlow(SqsPublishTest.java:253)
...
```

https://travis-ci.org/akka/alpakka/jobs/462223196

2m:
https://travis-ci.org/akka/alpakka/jobs/479853147

2m:
https://travis-ci.org/akka/alpakka/jobs/483188726

2m:
https://travis-ci.org/akka/alpakka/jobs/483908250

2m:
https://travis-ci.org/akka/alpakka/jobs/486971710

ennru:
https://travis-ci.org/akka/alpakka/jobs/491569281#L616

2m:
https://travis-ci.org/akka/alpakka/jobs/491602012

ennru:
This hasn't surfaced since over a year.
Closing.

PR #1422, opened 2019-01-15
Upgrade Akka Http to 10.1.7

2m:
Also fix tests that depended on JSON attribute ordering which changed with the new Akka Http version.

Supersedes #1406

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -11,7 +11,7 @@ object Dependencies {
   }
 
   val AwsSdkVersion = "1.11.476"
-  val AkkaHttpVersion = "10.1.5"
+  val AkkaHttpVersion = "10.1.7"
 
   val Common = Seq(
     // These libraries are added to all modules via the `Common` AutoPlugin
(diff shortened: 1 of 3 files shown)


project/Dependencies.scala
```diff
@@ -58,7 +58,7 @@ object Dependencies {
       "io.reactivex" % "rxjava-reactive-streams" % "1.2.1", //ApacheV2
       "org.reactivestreams" % "reactive-streams" % "1.0.2", //CC0	http://creativecommons.org/publicdomain/zero/1.0/
       "com.typesafe.play" %% "play-json" % "2.6.9" % Test, // MIT like: http://www.slf4j.org/license.html
-      "com.typesafe.akka" %% "akka-http" % "10.1.4" % Test, // MIT like: http://www.slf4j.org/license.html
+      "com.typesafe.akka" %% "akka-http" % AkkaHttpVersion % Test, // MIT like: http://www.slf4j.org/license.html
```
ennru: The license comment seems to have been copy and wasted. I wonder if these comments are useful at all - now that we show licenses in the dependencies listing in the docs.

2m: Good point. If we ever add building documentation from PRs, then it will be easy to see the licenses of the changed dependencies or new connectors. Then we will be able to remove these comments from build sources.

2m:
Unrelated failure was #1360

Issue #963, opened 2018-05-21
OrientDBTest.flow failed


ennru:
Failed after unrelated changes on 2018-05-21

```
[ERROR] [05/20/2018 05:03:49.370] [default-akka.actor.default-dispatcher-6] [akka://default/system/StreamSupervisor-1/flow-5-0-seqSink] Error during postStop in [akka.stream.alpakka.orientdb.OrientDBSourceStage@657e6235]: null
java.lang.NullPointerException
	at com.orientechnologies.orient.core.db.OPartitionedDatabasePool$PoolPartition.access$700(OPartitionedDatabasePool.java:418)
	at com.orientechnologies.orient.core.db.OPartitionedDatabasePool$DatabaseDocumentTxPooled.close(OPartitionedDatabasePool.java:479)
	at akka.stream.alpakka.orientdb.OrientDBSourceLogic.postStop(OrientDBSourceStage.scala:71)
	at akka.stream.impl.fusing.GraphInterpreter.finalizeStage(GraphInterpreter.scala:571)
	at akka.stream.impl.fusing.GraphInterpreter.afterStageHasRun(GraphInterpreter.scala:552)
	at akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:426)
	at akka.stream.impl.fusing.GraphInterpreterShell.runBatch(ActorGraphInterpreter.scala:588)
	at akka.stream.impl.fusing.GraphInterpreterShell.init(ActorGraphInterpreter.scala:558)
	at akka.stream.impl.fusing.ActorGraphInterpreter.tryInit(ActorGraphInterpreter.scala:679)
	at akka.stream.impl.fusing.ActorGraphInterpreter.preStart(ActorGraphInterpreter.scala:727)
	at akka.actor.Actor.aroundPreStart(Actor.scala:528)
	at akka.actor.Actor.aroundPreStart$(Actor.scala:528)
	at akka.stream.impl.fusing.ActorGraphInterpreter.aroundPreStart(ActorGraphInterpreter.scala:670)
	at akka.actor.ActorCell.create(ActorCell.scala:652)
	at akka.actor.ActorCell.invokeAll$1(ActorCell.scala:523)
	at akka.actor.ActorCell.systemInvoke(ActorCell.scala:545)
	at akka.dispatch.Mailbox.processAllSystemMessages(Mailbox.scala:283)
	at akka.dispatch.Mailbox.run(Mailbox.scala:224)
[...]

PR #1426, opened 2019-01-16
Couchbase: introduce upsertDocWithResult to handle write errors in-stream

ennru:
Add a `upsertDocWithResult` operator emitting `CouchbaseWriteResult` subclasses for users that need to handle write failures in-stream.

cc @dannylesnik

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ jobs:
       - PRE_CMD="docker-compose up -d cassandra"
       - CMD=+cassandra/testChanged
     - env:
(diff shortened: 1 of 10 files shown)


dannylesnik:
Fantastic, but in my opinion we should add such functionality for delete operation as well.

ennru:
Thank you for pointing that out. I've now added support for it with `deleteWithResult`. 
And it helped me to find out that `delete` didn't work as expected!

2m:
Unrelated failures were #991 and #963

PR #1430, opened 2019-01-17
Pass the argument from the method

2m:
Fixes `withSSLContext` being effectively a NOOP.


amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectionProvider.scala
```diff
@@ -224,7 +224,7 @@ final class AmqpSSLConfiguration private (val protocol: Option[String] = None,
     copy(protocol = Some(protocol), trustManager = Some(trustManager))
 
   def withSSLContext(context: Option[javax.net.ssl.SSLContext]): AmqpSSLConfiguration =
-    copy(protocol = protocol)
```
ennru: Wow, great find.

Issue #1435, opened 2019-01-18
google cloud pubsub: internal max messages should be configurable


stephennancekivell:
Internal to the pubsub Source there is a maxMessages parameter. It some scenarios you may want to tweak this number. It could be a new parameter on PubSubConfig. For example if each message is really big and your worried about memory usage. 

https://github.com/akka/alpakka/blob/master/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApi.scala#L77

stephennancekivell:
This is done. Thank you. Closing.

PR #1438, opened 2019-01-18
gce pubsub. When pulling make the options configurable.

stephennancekivell:
Addresses https://github.com/akka/alpakka/issues/1435

Previously  returnImmediately and maxMessages where hard coded in the api. Now they are exposed as options in the config.

### DIFF
--- a/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/impl/GooglePubSubSource.scala
+++ b/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/impl/GooglePubSubSource.scala
@@ -13,6 +13,7 @@ import akka.stream.stage.{GraphStage, GraphStageLogic, OutHandler, TimerGraphSta
 import akka.stream.{Attributes, Materializer, Outlet, SourceShape}
 
 import scala.collection.immutable
+import scala.concurrent.Future
 import scala.concurrent.duration._
 import scala.util.{Failure, Success, Try}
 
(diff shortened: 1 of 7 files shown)


google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/impl/GooglePubSubSource.scala
```diff
@@ -33,14 +36,21 @@ private[pubsub] final class GooglePubSubSource(projectId: String,
...
 
-        val req = if (httpApi.isEmulated) {
+        def pull(maybeAccessToken: Option[String]): Future[PullResponse] =
           httpApi
-            .pull(project = projectId, subscription = subscription, maybeAccessToken = None)
+            .pull(project = projectId,
+                  subscription = subscription,
+                  maybeAccessToken = maybeAccessToken,
+                  returnImmediately = returnImmediately,
+                  maxMessages = maxMessages)
```
2m: Not necessary to list the names of the arguments here. This is usually done when there are default values for arguments, and only some of them are passed from the call site.

stephennancekivell: yea, Its not technically necessary. But I thought it was a little nicer, easier to review. But sure, i'll take it out.

stephennancekivell:
Thanks @2m

2m:
Thanks for a quick turnaround!

PR #1440, opened 2019-01-20
Fixes around unsubscription, client connection, packet routing and stashing

huntc:
Lots more production testing and a few edge-case fixes.

> When reviewing this PR, it may be best to do so by commit. There's a bit of re-factoring, which makes the required changes a little noisy when viewed together.

### Cleaning up on unsubscription

I discovered that if a client unsubscribes then the server continues to produce publications with `DUP` if its previous one hadn't been acknowledged... This led me then to properly consider what should happen during the process of an `UNSUBSCRIBE`. In essence, it was previously possible for active consumers (client subscriptions) and active producers (server subscriptions) to continue post having unsubscribed from topics.

For a client scenario, this was relatively benign as, if a server didn't produce any more `PUBLISH` messages then it would eventually time-out the active consumer.

However, for a server scenario, this was a problem because the server's active producer could continue to produce `PUBLISH` messages.

Both sides now deliberately stop any active consumers and producers for an unsubscribed topic.

### Cleaning up on client connection

Prior to this PR, state was instead being cleaned up on disconnection given the clean session flag being set. This was entirely incorrect as state is required to be cleaned up when requesting a connection with the clean session flag set.

### Handling packet id routing

I’d wrongly assumed that flatMapMerge would support the supervision handling in place for the command flows. The event flows should be fine as they use mapAsync.

This oversight would cause legitimate situations where a PUBACK received locally could not be routed to its consumer as the consumer had already finished, which, in turn, would restart the flow.

### Client/server stashing and disconnectivity

For the client state, I know ensure that any stashed messages are carried through while in a disconnected state. In production, I observed a situation where a consumer completed but didn’t deregister from the internal data structure given that the actor was in a disconnected state.

For the server state, we now ensure that stashed messages are dispatched and that the remote source queue is completed - very similar to how the client behaves. I think we had a resource leak here before… Also, we ignore “connection lost” messages when we are in a disconnected state - we know that the connection is lost at this point, so we want to avoid failing in that scenario.

Both the client and server state data structures now look quite consistent with each other, which makes me feel warm and fuzzy inside.

Finally, there were some errors in the server reconnect test. A different connection id is required each time as should be the case when used in anger.

### DIFF
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
@@ -637,10 +637,6 @@ import scala.util.{Failure, Success}
           data.remote.fail(Unsubscriber.UnsubscribeFailed)
           disconnect(context, data.remote, data)
         case (context, _: Terminated) =>
-          val unsubscribableConsumers = for ((topicName, consumer) <- data.activeConsumers;
-                                             topicFilter <- data.topicFilters
-                                             if Topics.filter(topicFilter, topicName)) yield topicName -> consumer
-          unsubscribableConsumers.foreach { case (_, consumer) => context.stop(consumer) }
           data.stash.foreach(context.self.tell)
           serverConnected(
             ConnAckReceived(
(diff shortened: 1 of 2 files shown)


2m:
It is great to have these stability improvements coming!

PR #1456, opened 2019-01-27
AWS Lambda: use AWS SDK v2 instead of AWS SDK v1

sullis:


project/Dependencies.scala
```diff
@@ -11,8 +11,11 @@ object Dependencies {
   }
 
   val AwsSdkVersion = "1.11.476"
+  val AwsSdk2Version = "2.3.9"
   val AkkaHttpVersion = "10.1.7"
-  
+
+  val Java8CompatVersion = "0.9.0"
```
ennru: A short heads-up on this:
Can you get around upgrading? We do not want to override Akka's version of this dependency.

sullis: I'll fix this today.

2m:
Thank you @sullis for the PR. It would be great if you could take this PR one step further and remove the `AwsLambdaFlowStage` all-together. This stage is not needed anymore as the AWS SDK v2 now returns back `CompletionStage`s which can be integrated directly with `mapAsync`.

This approach has been already taken by SQS and SNS connectors. Take a look here for inspiration: 

https://github.com/akka/alpakka/blob/dcb8fe6c08298f1d515dd2ad173866869809f6c5/sns/src/main/scala/akka/stream/alpakka/sns/scaladsl/SnsPublisher.scala#L45-L49

sullis:
> Thank you @sullis for the PR. It would be great if you could take this PR one step further and remove the `AwsLambdaFlowStage` all-together. 

Nice!   I'll look at the code tonight.

sullis:
Work-in-progress.

project/Dependencies.scala
sullis: Done

sullis:
ready for review @2m @ennru

2m:
Great, thanks!

PR #1458, opened 2019-01-28
LogRotator with sink factory

### LINKED ISSUE #1452: File: LogRotatorSink can't use compression
The `LogRotatorSink` does not compose with `Compression.gzip` as the compression header and trailer will end up in the first and the last files, not every file.

To support compression in a good way `LogRotatorSink` could be extended to take an optional `Flow[ByteString, ByteString, NotUsed]` which will be passed into the flow for each file.

ennru:
* Add the `LogRotatorSink.withSinkFactory`
* Allow for non-`IOResult`-returning sinks
* Allow for other rotation value than `Path`
* Run Java snippets

Fixes #1452

### DIFF
--- a/file/src/main/scala/akka/stream/alpakka/file/javadsl/LogRotatorSink.scala
+++ b/file/src/main/scala/akka/stream/alpakka/file/javadsl/LogRotatorSink.scala
@@ -5,7 +5,6 @@
 package akka.stream.alpakka.file.javadsl
 
 import java.nio.file.{Path, StandardOpenOption}
-import java.util
 import java.util.Optional
 import java.util.concurrent.CompletionStage
 
@@ -20,8 +19,15 @@ import scala.collection.JavaConverters._
 import scala.concurrent.Future
 
 import scala.compat.java8.FutureConverters._
+import scala.compat.java8.OptionConverters._
 
 object LogRotatorSink {
+
+  /**
+   * Sink directing the incoming `ByteString`s to new files whenever `generatorFunction` returns a new value.
+   *
+   * @param f creates a function that triggers rotation by returning a value
+   */
   def createFromFunction(
       f: function.Creator[function.Function[ByteString, Optional[Path]]]
   ): javadsl.Sink[ByteString, CompletionStage[Done]] =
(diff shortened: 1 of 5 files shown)


file/src/main/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSink.scala
```diff
@@ -8,163 +8,197 @@ import java.nio.file.{OpenOption, Path, StandardOpenOption}
...
 object LogRotatorSink {
+
+  /**
+   * Sink directing the incoming `ByteString`s to new files whenever `generatorFunction` returns a new value.
+   *
+   * @param functionGeneratorFunction creates a function that triggers rotation by returning a value
+   * @param fileOpenOptions file options for file creation
+   */
   def apply(
       functionGeneratorFunction: () => ByteString => Option[Path],
```
2m: I see that this has not changed in this PR, but do you see a need for the additional wrapping with a function? Could this be `ByteString => Option[Path]` instead?

file/src/main/scala/akka/stream/alpakka/file/javadsl/LogRotatorSink.scala
```diff
@@ -5,43 +5,73 @@
...
 import scala.collection.JavaConverters._
+import scala.concurrent.Future
+
+import scala.compat.java8.FutureConverters._
+import scala.compat.java8.OptionConverters._
 
 object LogRotatorSink {
+
+  /**
+   * Sink directing the incoming `ByteString`s to new files whenever `generatorFunction` returns a new value.
```
2m: There is a mismatch of how the argument is named in the declaration of the method and the comment.

file/src/main/scala/akka/stream/alpakka/file/javadsl/LogRotatorSink.scala
```diff
@@ -5,43 +5,73 @@
...
-        .LogRotatorSink({ () =>
-          val fun = f.create()
-          elem =>
-            Option(fun(elem).orElse(null))
-        }, options.asScala.toSet)
+        .LogRotatorSink(asScala(f), fileOpenOptions.asScala.toSet)
+        .toCompletionStage()
+    )
+
+  def withSinkFactory[C, R](
```
2m: Missing API doc.

file/src/main/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSink.scala
```diff
@@ -8,163 +8,197 @@ import java.nio.file.{OpenOption, Path, StandardOpenOption}
...
+    val decider =
+      inheritedAttributes.get[SupervisionStrategy].map(_.decider).getOrElse(Supervision.stoppingDecider)
+    val logic = new Logic(promise, decider)
+    (logic, promise.future)
+  }
+
+  private final class Logic(promise: Promise[Done], decider: Decider) extends GraphStageLogic(shape) {
+    val pathGeneratorFunction: ByteString => Option[C] = functionGeneratorFunction()
+    var sourceOut: SubSourceOutlet[ByteString] = _
+    var fileSinkCompleted: immutable.Seq[Future[R]] = immutable.Seq.empty
```
2m: This now holds materialized values of generic sinks, therefor it should be renamed. `materializedValues`?

file/src/main/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSink.scala
```diff
@@ -8,163 +8,197 @@ import java.nio.file.{OpenOption, Path, StandardOpenOption}
...
+      try {
+        ret = pathGeneratorFunction(data)
+      } catch {
+        case ex: Throwable =>
+          failThisStage(ex)
+      }
+      ret
+    }
+
+    def fileSinkFutureCallbackHandler(future: Future[R])(h: Holder[R]): Unit =
```
2m: Method name was to specific, and now does not match what it does anymore.

file/src/test/java/docs/javadsl/LogRotatorSinkTest.java
```diff
@@ -82,13 +112,30 @@ public static void main(String... args) {
...
         Source.from(Arrays.asList("test1", "test2", "test3", "test4", "test5", "test6"))
             .map(ByteString::fromString)
             .runWith(LogRotatorSink.createFromFunction(pathGeneratorCreator), materializer);
+
+    // GZip compressing the data written
+    CompletionStage<Done> compressedCompletion =
+        source.runWith(
+            LogRotatorSink.withSinkFactory(
+                pathGeneratorCreator,
+                path ->
```
2m: I really like the API improvements. It is great to have the `path` reused that is returned from the `pathGeneratorCreator`.

file/src/main/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSink.scala
2m: Scratch that. I see that it is used to capture some state in the `fileLengthFunctionFactory` example below.

PR #1466, opened 2019-01-30
Remove grpc publisher/subscriber from factory arguments

2m:
Publisher/Subscriber is instead resolved from an Akka Extension that manages one instance of these resources per Actor System.

Otherwise a manually initialized publisher/subscriber can be injected using Stream Attributes.


2m:
Fixed the review feedback and some boy-scouting.

2m:
Unrelated failure was https://github.com/akka/alpakka/issues/1377

PR #1476, opened 2019-02-01
Add log rotator, compression, sftp example

2m:
Example that uses newly improved Log Rotator sink with compression and sftp.


2m:
I think you meant the `ftp-samples.md`. Added it there.

ennru:
I think it might make sense to link from the `LogRotationSink` docs, as well.

PR #1478, opened 2019-02-01
ElasticSearch null check for index and type name on source and flow

seglo:
# Pull Request Checklist

## Purpose

Add null checks for `indexName` on `ElasticsearchSource`.  Add null checks for `indexName` and `typeName` on `ElasticsearchFlow`

## Background Context

The `indexName` and `typeName` `String` params are used in `ElasticsearchFlowStage` when creating the indexing JSON request for Elasticsearch.  If the parameters are `null` and added as `JsStrings` to the request JSON (https://github.com/akka/alpakka/blob/v1.0-M2/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchFlowStage.scala#L157..L158) then an NPE will be thrown from deep inside spray-json when the request JSON is serialized.  This is a known issue in spray-json (https://github.com/spray/spray-json/issues/70).  The reason I passed `null` to these parameters was due to a different issue related to a known Scala bug (https://github.com/scala/bug/issues/1352) and initialization quirks in Scala ([Why is my abstract or overridden val null?](https://docs.scala-lang.org/tutorials/FAQ/initialization-order.html)), but nonetheless I think it would be prudent to add a `null` check to avoid the possible spray-json NPE in the first place because it can be challenging to troubleshoot.

I added a similar null check for `indexName` to `ElasticsearchSourceStage` because it's encoded into query path to Elasticsearch.  In this case `typeName` is an `Option` so any potential NPE would be easier to reason about in a stack trace.

### DIFF
--- a/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchFlowStage.scala
+++ b/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchFlowStage.scala
@@ -32,6 +32,8 @@ private[elasticsearch] final class ElasticsearchFlowStage[T, C](
     settings: ElasticsearchWriteSettings,
     writer: MessageWriter[T]
 ) extends GraphStage[FlowShape[WriteMessage[T, C], Future[Seq[WriteResult[T, C]]]]] {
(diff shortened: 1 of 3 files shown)


ennru:
Thanks, @seglo

PR #1480, opened 2019-02-05
Parallel subscriptions/unsubscriptions and SubAck notifications

huntc:
Prior to this PR, more than 10 control packets (configurable) could come in very quickly and prevent the client or server handling more. This created a problem if, say, 10 `PUBLISH` packets were received then followed by a `SUBACK`. The `SUBACK` would never be processed.

The changes here no longer restrict subscriptions and unsubscriptions for both the client and server from occurring only one at a time. This then frees the session to process packets without effectively being blocked.

An added bonus is that the code is somewhat simplified. :-)

Also new, is the introduction of the ability to be notified of server-side SubAck completion. It is important for the server to know when it is the right time to start publishing for a newly subscribed topic; too early and the publications will be dropped given that internal state hasn’t yet updated (sometimes!). This PR introduces the ability for certain commands, presently only server-side SubAck, to complete a process when the `SubAck` has been dispatched and internal state updated. A couple of flaky tests have been fixed as a consequence of this new feature.

Lastly, Producers are messages are kept around post an unsubscribe. Producers still need the opportunity of receiving PUBACKs and so forth, even if they’ve now been unsubscribed.

BONUS: a flaky test has been fixed.

### DIFF
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
@@ -341,12 +341,12 @@ import scala.util.{Failure, Success}
             remote.success(ForwardDisconnect)
             timer.cancel(SendPingreq)
             disconnect(context, data.remote, data)
-          case (context, SubscribeReceivedLocally(subscribe, subscribeData, remote)) =>
+          case (context, SubscribeReceivedLocally(_, subscribeData, remote)) =>
             context.watch(
               context.spawnAnonymous(Subscriber(subscribeData, remote, data.subscriberPacketRouter, data.settings))
             )
             serverConnected(data)
-          case (context, UnsubscribeReceivedLocally(unsubscribe, unsubscribeData, remote)) =>
+          case (context, UnsubscribeReceivedLocally(_, unsubscribeData, remote)) =>
             context.watch(
               context
                 .spawnAnonymous(Unsubscriber(unsubscribeData, remote, data.unsubscriberPacketRouter, data.settings))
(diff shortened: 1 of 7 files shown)


huntc:
[CI has failed quite legitimately](https://travis-ci.org/akka/alpakka/jobs/488908793#L684). This PR may expose the particular issue more given that the changes lead to greater throughput. However, the problem in the tests was there before.

What happens is that, when the server receives a subscription in these tests, it then signals that any publish received should now be sent out. The problem with this is that the internal state of the MQTT server session may not yet have updated and registered the topic filters that it will handle. Therefore, any local publish will be dropped.

While the error is strictly within the test, I'm left wondering if the server should provide a reliable means of retaining a publish i.e. publish only when the internal state has been updated. Perhaps a retained publish can be sent as a promise with `SubAck`. Thoughts @longshorej ?

longshorej:
I think it's only a problem with the test, right? In reality, the process of publishing would be decoupled from subscriptions. Might be good enough to just delay the publishing that occurs after SUBACK by a few seconds, i.e. scheduleOnce to complete the promise in three seconds or something like that.

huntc:
@longshorej I've incorporated the `SubAck` notification functionality as discussed. Here's an example:

```scala
val subscribed = Promise[Done]
source
  .runForeach {
    case Right(Event(_: Connect, _)) =>
      queue.offer(Command(ConnAck(ConnAckFlags.None, ConnAckReturnCode.ConnectionAccepted)))
    case Right(Event(cp: Subscribe, _)) =>
      queue.offer(Command(SubAck(cp.packetId, cp.topicFilters.map(_._2))))
      subscribed.success(Done)
      queue.offer(Command(SubAck(cp.packetId, cp.topicFilters.map(_._2)), Some(subscribed), None))
    case Right(Event(publish @ Publish(flags, _, Some(packetId), _), _))
        if flags.contains(ControlPacketFlags.RETAIN) =>
      queue.offer(Command(PubAck(packetId)))
      import mat.executionContext
      subscribed.future.foreach(_ => session ! Command(publish))
    case _ => // Ignore everything else
  }
```

I've made this WIP until you can review and test it out in our code base.

huntc:
Everything fixed up. The PR is now ready for a final review.

huntc:
Ready for review/merge.

huntc:
Yeah, they're very tricky! I've a new appreciation for Akka HTTP. :-)

PR #1489, opened 2019-02-12
mqtt-streaming: Prefer wrapping instead of reissuing packet ids

longshorej:
Packet ids now monotonically increase until they reach the maximum value, upon which they wrap around to the minimum. Care is taken to ensure that packet ids that are in use are not issued, as it is
technically possible, albeit rare.

This fixes a number of potential race conditions. For example, consider a PUBLISH -> PUBACK exchange with QoS1 with the old behavior:

1) Server retransmits an unacknowledged publish
2) Before client receives retransmission, it finally ACKs original publish.
3) Server receives ACK, and performs new publish, reusing the last id.
4) Client receives retransmission from # 1, and ACKs it.
5) Server receives ACK, and misinterprets it as an ack from # 3
6) Oh no!

By wrapping around, we've effectively eliminated the changes of this happening. This is similar to how Linux issues PIDs for processes to avoid a similar sort of problem.

/cc @huntc

### DIFF
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/RequestState.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/RequestState.scala
@@ -336,8 +336,8 @@ import scala.util.{Failure, Success}
    * In case some brokers treat 0 as no packet id, we set our min to 1
    * e.g. https://renesasrulz.com/synergy/synergy_tech_notes/f/technical-bulletin-board-notification-postings/8998/mqtt-client-packet-identifier-is-0-by-default-which-causes-azure-iot-hub-to-reset-connection
    */
-  private val MinPacketId = PacketId(1)
-  private val MaxPacketId = PacketId(0xffff)
+  val MinPacketId = PacketId(1)
+  val MaxPacketId = PacketId(0xffff)
 
   // Requests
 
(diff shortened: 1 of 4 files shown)


mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/RequestState.scala
```diff
@@ -336,8 +336,8 @@ import scala.util.{Failure, Success}
    * In case some brokers treat 0 as no packet id, we set our min to 1
    * e.g. https://renesasrulz.com/synergy/synergy_tech_notes/f/technical-bulletin-board-notification-postings/8998/mqtt-client-packet-identifier-is-0-by-default-which-causes-azure-iot-hub-to-reset-connection
    */
-  private val MinPacketId = PacketId(1)
-  private val MaxPacketId = PacketId(0xffff)
+  val MinPacketId = PacketId(1)
+  val MaxPacketId = PacketId(0xffff)
```
longshorej: These are enclosed in a `private[streaming]` already. I am using them in the tests, hence removing the private here.

huntc:
I’m wondering if we can simplify this even further. The idea is centered on maintaining a set where we track the allocated packet ids.

0. Fail if allocation set is full. 

1. Allocate by adding to the last number allocated. 

2. If exceeds max then set to min. 

3. Test if allocated is already in the allocated packet ids set. If so, add 1 and go to 2.

4. Add packet id to allocated packet ids set.

When deallocating, simply remove from the set.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/RequestState.scala
```diff
@@ -375,29 +401,53 @@ import scala.util.{Failure, Success}
...
+                nextRegistrations,
+                findNextPacketId(nextRegistrations, currentPacketId),
+                pendingRegistrations
+              )
+
+            case None =>
+              // all packet ids are taken, so we'll wait until one is unregistered
+              // to continue
+
+              main(registrantsByPacketId, nextPacketId, pendingRegistrations :+ register)
```
huntc: I'm still unsure as to the benefit of having a `pendingRegistrations` structure. It should be a rare condition and adds complexity.

huntc:
@ennru, @longshorej and I discussed further - this PR is now reviewed from our perspective and ready to go.

longshorej:
Rebased!

PR #1490, opened 2019-02-13
WIP - DO NO MERGE - Increment packet id monotonically

huntc:
[An alternate implementation to Jason's](https://github.com/akka/alpakka/pull/1489) that also strives to ensure those packet ids are unable to clash and makes requests easier to trace. Bonus: simplifies the existing implementation.

Jason and I need to discuss the pros and cons of each PR.

### DIFF
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/RequestState.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/RequestState.scala
@@ -338,6 +338,7 @@ import scala.util.{Failure, Success}
    */
   private val MinPacketId = PacketId(1)
   private val MaxPacketId = PacketId(0xffff)
+  private val MaxNrOfPacketId = MaxPacketId.underlying - MinPacketId.underlying
 
   // Requests
 
@@ -355,7 +356,7 @@ import scala.util.{Failure, Success}
    * Construct with the starting state
    */
   def apply[A]: Behavior[Request[A]] =
-    new LocalPacketRouter[A].main(Map.empty, MinPacketId)
+    new LocalPacketRouter[A].main(Map.empty, None)
 }
 
 /*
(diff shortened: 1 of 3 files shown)


mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/RequestState.scala
```diff
@@ -375,22 +376,25 @@ import scala.util.{Failure, Success}
...
+          else if (registrantsByPacketId.contains(id))
+            step(PacketId(id.underlying + 1))
+          else
+            id
+        val nextPacketId = step(lastPacketId.fold(MinPacketId)(l => PacketId(l.underlying + 1)))
         reply.success(Registered(nextPacketId))
-        main(registrantsByPacketId + (nextPacketId -> registrant), PacketId(nextPacketId.underlying + 1))
+        main(registrantsByPacketId + (nextPacketId -> registrant), Some(nextPacketId))
       case _: Register[A] =>
         Behaviors.same // We cannot allocate any more. This will eventually cause a timeout to occur on the requestor.
```
longshorej: Note that this means the `Promise` will never be completed, so that needs to be dealt with here. Additionally, any remaining `Promise`s need to be failed when this actor stops.

In general, I don't see the need to fail in this condition. I think it's better to track the registrants and fulfill the promise on the next `Unregister`. I'd say that's the main difference between the two PRs.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/RequestState.scala
```diff
@@ -375,22 +376,25 @@ import scala.util.{Failure, Success}
 
   // Processing
 
-  def main(registrantsByPacketId: Map[PacketId, ActorRef[A]], nextPacketId: PacketId): Behavior[Request[A]] =
+  def main(registrantsByPacketId: Map[PacketId, ActorRef[A]], lastPacketId: Option[PacketId]): Behavior[Request[A]] =
     Behaviors.receiveMessage {
-      case Register(registrant: ActorRef[A], reply) if nextPacketId.underlying <= MaxPacketId.underlying =>
+      case Register(registrant: ActorRef[A], reply) if registrantsByPacketId.size < MaxNrOfPacketId =>
+        @tailrec
+        def step(id: PacketId): PacketId =
```
longshorej: Note that if `registrantsByPacketId` ever ends up holding an invalid packet ID (due to bugs introduced by future changes), this will never terminate.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/RequestState.scala
```diff
@@ -375,22 +376,25 @@ import scala.util.{Failure, Success}
...
+          if (id.underlying == MaxPacketId.underlying)
+            step(MinPacketId)
```
longshorej: This means `MaxPacketId` will never be issued. In practice, not really a problem, but sort of quirky behavior IMO.

huntc:
Closing as I’m preferring the resiliency of the other approach.

Issue #1223, opened 2018-09-20
Make use of assertAllStagesStopped


ennru:
The later Akka versions expose `assertAllStagesStopped` helper in the testkit. Alpakka connectors should use it to ensure a proper shutdown of the stages.

ennru:
Trying it on `AmqpConnectorsSpec` makes several tests fail.

ennru:
It is available now since #1224 is merged.

ennru:
Most tests use it nowadays.

PR #1167, opened 2018-08-27
Don't ack when 'no ack required'

ennru:
This should fix #1110 if I understand things correctly.

PR #1501, opened 2019-02-15
Amqp: assert all stages stopped and fix source shutdown

### LINKED ISSUE #1500: AMQP: source doesn't shut down
`AmqpSource`s should shut down properly when downstream is finished, depending on the `NamedQueueSourceSettings.ackRequired` setting.

### Expected Behavior

If `NamedQueueSourceSettings.ackRequired` is false, the AMQP source stage should shut down immediately on downstream finish.

### Actual Behavior

The `NamedQueueSourceSettings.ackRequired` setting enables AMQP's `autoAck` if false, but the source stage is still awaiting acks or nacks for all messages (by counting outstanding acks) before it shuts down.

ennru:
## Fixes

Fixes #1500 

Part of #1223 

## Purpose

Make use of `assertAllStagesStopped` to ensure proper stream shutdown in test cases.
Add `ApiMayChange` on stream element classes as they have unfortunate names.

### DIFF
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpSourceStage.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpSourceStage.scala
@@ -8,7 +8,7 @@ import akka.Done
 import akka.annotation.InternalApi
 import akka.stream.alpakka.amqp._
 import akka.stream.alpakka.amqp.scaladsl.CommittableIncomingMessage
-import akka.stream.stage.{GraphStage, GraphStageLogic, OutHandler}
+import akka.stream.stage.{GraphStage, GraphStageLogic, OutHandler, StageLogging}
 import akka.stream.{Attributes, Outlet, SourceShape}
 import akka.util.ByteString
 import com.rabbitmq.client.AMQP.BasicProperties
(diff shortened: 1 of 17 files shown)


amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpSourceStage.scala
```diff
@@ -156,10 +160,12 @@ private[amqp] final class AmqpSourceStage(settings: AmqpSourceSettings, bufferSi
               pushMessage(queue.dequeue())
             }
 
-          override def onDownstreamFinish(): Unit = {
-            setKeepGoing(true)
-            if (unackedMessages == 0) super.onDownstreamFinish()
-          }
+          override def onDownstreamFinish(): Unit =
+            if (finishWithOutstandingAcks || unackedMessages == 0) super.onDownstreamFinish()
```
juanjoDiaz: I think that it's misleading to have a `unackedMessages` counter different to zero if messages are actually acknowledged.

It would be cleaner to do the check at line 168/174
(https://github.com/akka/alpakka/pull/1501/files#diff-79b48cb12e370d65fef5180dc16ab6aaR174)
like
```scala
      def pushMessage(message: CommittableIncomingMessage): Unit = {
        push(out, message)
        if (super.settings.ackRequired) unackedMessages += 1
      }
```

ennru: You are right. 
Does it have any effect in case `ack` or `nack` are called when `autoAck` is `true`? 
Those could be short-circuited I guess. Or the source should not even emit committable messages...

juanjoDiaz: True.
Probably you should do the same at when the counter is decreased: `if (super.settings.ackRequired) unackedMessages -= 1` or even ``if (unackedMessages > 0) unackedMessages -= 1` to ensure that it doesn't go negative.

ennru: I went a bit further and made ack/nack no-ops when using auto-ack.

juanjoDiaz: Looks great now.

However, I still think that the API is kind of broken (as discussed in #1167 ) since you can use the `withAckRequired` when creating a `committableSource` which doesn't make sense.

ennru: Yes, you're right. I'll put an API might change there.

ennru: I've created #1513

juanjoDiaz: Thanks!

Apart from that, the PR LGTM!
It fixes the actual issue and doesn't touch the current API.

PR #1506, opened 2019-02-15
Akka 2.5.21 and type inference changes

ennru:
Use the latest Akka release.

Type inference changes in a few places.

### DIFF
--- a/mqtt-streaming/src/test/scala/docs/scaladsl/MqttSessionSpec.scala
+++ b/mqtt-streaming/src/test/scala/docs/scaladsl/MqttSessionSpec.scala
@@ -54,7 +54,7 @@ class MqttSessionSpec
               .clientSessionFlow(session)
               .join(pipeToServer)
           )
-          .toMat(Sink.collection)(Keep.both)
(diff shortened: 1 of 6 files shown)


mqtt-streaming/src/test/scala/docs/scaladsl/MqttSessionSpec.scala
```diff
@@ -54,7 +54,7 @@ class MqttSessionSpec
               .clientSessionFlow(session)
               .join(pipeToServer)
           )
-          .toMat(Sink.collection)(Keep.both)
+          .toMat(Sink.seq)(Keep.both)
```
huntc: Why this change?

huntc: I see this issue remains also: https://github.com/akka/akka/issues/26305

ennru: Yes, that is the reason. I think this solution is the best for now.

Issue #494, opened 2017-09-27
Javadsl for MongoDb Connector


patriknw:
In https://github.com/akka/alpakka/issues/115 a MongoDB connector was added, but without Java API. It would be great to add `javadsl` (I don't know if that would require a different underlying implementation/driver).

stephennancekivell:
I'd recommend a different module for the a java connector, as there is a different driver that returns different document types to the scala one.

https://mongodb.github.io/mongo-java-driver/
http://mongodb.github.io/mongo-scala-driver/

ennru:
Or base both on the [Reactive Streams compliant driver](http://mongodb.github.io/mongo-java-driver-reactivestreams/)?

dbuschman7:
On the separate connector/driver issue. 
The Scala driver uses the "core" jars from the java driver so a separate connector would not be needed.

dotbg:
Looks like it was done in #1514

ennru:
That's right, thank you. Closing.

PR #1514, opened 2019-02-19
MongoDB: Towards Alpakka 1.0

### LINKED ISSUE #1082: MongoDB: become Alpakka 1.0 ready
Before changing the existing MongoDB connector, we should consider a new version to support both Java and Scala (see https://github.com/akka/alpakka/issues/494).

- [ ] Move internal code to `impl` package
- [ ] Mark internal code `@InternalApi` and `private[connector]`
- [ ] Move documentation example code to packages `docs.javadsl` and `docs.scaladsl`

2m:
## Fixes

Fixes #1082

## Purpose

Adds javadsl for API parity.

## Background Context

To reuse the implementation, MongoDB connector was switched from scala MongoDB driver to the Reactive Streams one.

BSON construction API can still be used from the scala driver, but that jar will need to be pulled in separately, as it is done in tests.

## Missing

* [x] doc updates
* [x] javadsl sink tests


2m:
Added missing Java tests and documentation updates.

docs/src/main/paradox/mongodb.md
```diff
@@ -31,113 +31,106 @@ The table below shows direct dependencies of this module and the second tab show
...
 
 Scala
-: @@snip [snip](/mongodb/src/test/scala/akka/stream/alpakka/mongodb/MongoSourceSpec.scala) { #init-connection }
+: @@snip [snip](/mongodb/src/test/scala/docs/scaladsl/MongoSourceSpec.scala) { #codecs }
+
+Java
+: @@snip [snip](/mongodb/src/test/java/docs/javadsl/MongoSourceTest.java) { #codecs }
 
-For codec support, add the registry to the database or the collection.
+Sources provided by this connector need a prepared collection to communicate with MongoDB server.
```
ennru:
```suggestion
Sources provided by this connector need a prepared collection to communicate with the MongoDB server.
```

docs/src/main/paradox/mongodb.md
```diff
@@ -31,113 +31,106 @@ The table below shows direct dependencies of this module and the second tab show
...
+To get a reference to collection, lets initialize a MongoDB connection and access the database.
```
ennru:
```suggestion
To get a reference to a collection, let's initialize a MongoDB connection and access the database.
```

mongodb/src/main/scala/akka/stream/alpakka/mongodb/javadsl/MongoFlow.scala
```diff
@@ -0,0 +1,147 @@
...
+   * A [[akka.stream.javadsl.Flow Flow]] that will insert documents into a collection.
+   *
+   * @param collection mongo db collection to insert to.
+   * @param options options to apply to the operation
+   */
+  def insertOne[T](collection: MongoCollection[T], options: InsertOneOptions): Flow[T, T, NotUsed] =
+    scaladsl.MongoFlow.insertOne(collection, options).asJava
+
+  /**
+   * A [[akka.stream.javadsl.Flow Flow]] that will insert batches documents into a collection.
```
ennru:
```suggestion
   * A [[akka.stream.javadsl.Flow Flow]] that will insert batches of documents into a collection.
```

mongodb/src/main/scala/akka/stream/alpakka/mongodb/javadsl/MongoFlow.scala
```diff
@@ -0,0 +1,147 @@
...
+   *
+   * @param collection mongo db collection to insert to.
+   */
+  def insertMany[T](collection: MongoCollection[T]): Flow[java.util.List[T], java.util.List[T], NotUsed] =
+    insertMany(collection, DefaultInsertManyOptions)
+
+  /**
+   * A [[akka.stream.javadsl.Flow Flow]] that will insert batches documents into a collection.
```
ennru:
```suggestion
   * A [[akka.stream.javadsl.Flow Flow]] that will insert batches of documents into a collection.
```

mongodb/src/main/scala/akka/stream/alpakka/mongodb/javadsl/MongoSink.scala
```diff
@@ -0,0 +1,129 @@
...
+  def insertOne[T](collection: MongoCollection[T]): Sink[T, CompletionStage[Done]] =
+    insertOne(collection, DefaultInsertOneOptions)
+
+  /**
+   * A [[akka.stream.javadsl.Sink Sink]] that will insert documents into a collection.
+   *
+   * @param collection mongo db collection to insert to.
+   * @param options options to apply to the operation
+   */
+  def insertOne[T](collection: MongoCollection[T], options: InsertOneOptions): Sink[T, CompletionStage[Done]] =
```
ennru: We could maybe drop the alternatives with options from the sink API. If users need those, they may use the Flow API. WDYT?

2m: I looked at it again, and the Sink variants are used quite extensively in tests. I say the shrinkage of the API surface is not worth the introduced discrepancy.

iozozturk:
Hi @ennru and @2m , I just realised that with this PR you removed all my work from original MongoDB connector and replaced it with some other library implementation. You guys opened the issue #1082 and without getting any reviews, @2m deleted whole implementation and replaced it and @ennru accepted the PR again without getting any community views. You didn't even mentioned or tag people to get notice. Why would I contribute anything if you are going to remove it in future at will?

Congrats! Decent community work!

seglo:
This was before my time, but I believe the intention was only to re-use new functionality that was available in the upstream Mongo driver that I assume wasn't available when the connector was first implemented/donated to Alpakka.

Since using the reactivestreams implementation significantly reduced the surface area of this connector it was easier to bring the Java DSL up to par at the same time, which is a requirement of all Alpakka connectors so they can be used by the Java community at large with the same features as the Scala community.

I'm certain there was no ill will about the original implementation!

ennru:
Hi @iozozturk 
Sorry if we didn't live up to the expectations you have for how open source should be managed.
We are still grateful that you kicked off the Alpakka MongoDB connector back in 2017 with #148. As reported in #494 there was no API for Java users and our ambition is to offer somewhat similar APIs for both Scala and Java.
So for Alpakka 1.0 we wanted to have the Java DSL as #1082 stated in July 2018. We've seen very little community engagement around Alpakka MongoDB (which I believe is partly due to [ReactiveMongo](http://reactivemongo.org/) being a great choice for Scala users) so that's why @2m as member of Lightbend's Alpakka team implemented it in February 2019 on top of the Reactive Streams compatible MongoDB driver.
It was never the intention to remove the code you contributed, this was just a step to make Alpakka MongoDB attractive for more users.
Sorry, that we did not explicitly reached out to you when we worked on this.

PR #1515, opened 2019-02-19
Elasticsearch: Make flows type signature simpler to prepare for `withContext`

ennru:
## Purpose

Change Elasticsearch flow signatures to single element flows.

## Background Context

Akka has introduced a new type of flow `withContext` which supports `passThrough`s to be passed hidden from the user.
Elasticsearch flows used to do have internal logic for batching which made them consume `WriteMessage`s but emit `Seq[WriteResult]`s which does not work together with context promotion.

The retry logic in `ElasticsearchFlowStage` does not work with context promotion, either.

This ended up in a major rewrite of `ElasticsearchFlowStage`.

## References

https://github.com/akka/akka/pull/25951

### DIFF
--- a/docs/src/main/paradox/elasticsearch.md
+++ b/docs/src/main/paradox/elasticsearch.md
@@ -118,7 +118,7 @@ Java
 
 | Parameter           | Default | Description                                                                                            |
 | ------------------- | ------- | ------------------------------------------------------------------------------------------------------ |
-| bufferSize          | 10      | `ElasticsearchSink` puts messages by one bulk request per messages of this buffer size.                |
+| bufferSize          | 10      | Flow and Sink batche messages to bulk requests when back-pressure applies.                             |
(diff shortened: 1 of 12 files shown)


docs/src/main/paradox/elasticsearch.md
```diff
@@ -118,15 +118,15 @@ Java
 
 | Parameter           | Default | Description                                                                                            |
 | ------------------- | ------- | ------------------------------------------------------------------------------------------------------ |
-| bufferSize          | 10      | `ElasticsearchSink` puts messages by one bulk request per messages of this buffer size.                |
+| bufferSize          | 10      | Flow and Sink batche messages to bulk requests when back-pressure applies.                             |
```
2m:
```suggestion
| bufferSize          | 10      | Flow and Sink batch messages to bulk requests when back-pressure applies.                             |
```

PR #1518, opened 2019-02-20
Always use latest JDK

2m:
## Purpose

The `~` tells jabba, to install the specified or later, if available, version.


2m:
Failures were the usual suspects.

PR #1520, opened 2019-02-20
WIP Sagemaker, shared AWS module

nkconnor:
This follows #1456 and the existing Lambda module. Since the general flow is almost identical, I've created an `aws-shared` module with an abstract flow stage that can be used in multiple AWS connectors. It has an abstract def for

```
  /**
    * Client async invocation method
    * @param request
    * @param handler
    * @return
    */
  def clientInvoke(request: In): CompletableFuture[Out]
```
this makes the Sagemaker integration small:

```
@InternalApi private[sagemaker]
final class SagemakerInvokeFlowStage(sagemakerClient: SageMakerRuntimeAsyncClient)(parallelism: Int)
  extends AsyncHandlerFlowStage[InvokeEndpointRequest, InvokeEndpointResponse](parallelism) {
  /**
    *
    * @param request
    * @return
    */
  override def clientInvoke(request: InvokeEndpointRequest): CompletableFuture[InvokeEndpointResponse] =
    sagemakerClient.invokeEndpoint(request)
}
```

Similarly, Lambda (and probably other AWS modules) could benefit from a shared flow stage.

* [ x ] Have you read through the [contributor guidelines]
* [ x ] Have you signed the [Lightbend CLA](https://www.lightbend.com/contribute/cla)?
* [ no ] Have you updated the documentation?
* [ no ] Have you added tests for any changed functionality?

I can work towards docs and tests; but wanted to drop this here first in case someone could provide feedback.

### DIFF
--- a/aws-sagemaker/src/main/scala/akka/stream/alpakka/aws/sagemaker/impl/SagemakerInvokeFlowStage.scala
+++ b/aws-sagemaker/src/main/scala/akka/stream/alpakka/aws/sagemaker/impl/SagemakerInvokeFlowStage.scala
@@ -1,0 +1,29 @@
+/*
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.aws.sagemaker.impl
+
+import akka.annotation.InternalApi
+import akka.stream.alpakka.aws.AsyncHandlerFlowStage
+
+import java.util.concurrent.CompletableFuture
+
+import software.amazon.awssdk.services.sagemakerruntime.SageMakerRuntimeAsyncClient
+import software.amazon.awssdk.services.sagemakerruntime.model.{InvokeEndpointRequest, InvokeEndpointResponse}
+
+/**
+ * INTERNAL API
+ */
+@InternalApi private[sagemaker] final class SagemakerInvokeFlowStage(sagemakerClient: SageMakerRuntimeAsyncClient)(
+    parallelism: Int
+) extends AsyncHandlerFlowStage[InvokeEndpointRequest, InvokeEndpointResponse](parallelism) {
+
+  /**
+   *
+   * @param request
+   * @return
+   */
+  override def clientInvoke(request: InvokeEndpointRequest): CompletableFuture[InvokeEndpointResponse] =
+    sagemakerClient.invokeEndpoint(request)
+}
--- a/aws-sagemaker/src/main/scala/akka/stream/alpakka/aws/sagemaker/javadsl/SagemakerInvokeFlow.scala
+++ b/aws-sagemaker/src/main/scala/akka/stream/alpakka/aws/sagemaker/javadsl/SagemakerInvokeFlow.scala
@@ -1,0 +1,24 @@
+/*
+ * Copyright (C) 2016-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.aws.sagemaker.javadsl
+
+import akka.NotUsed
+import akka.stream.alpakka.aws.sagemaker.impl.SagemakerInvokeFlowStage
+import akka.stream.javadsl.Flow
+import software.amazon.awssdk.services.sagemakerruntime.SageMakerRuntimeAsyncClient
+import software.amazon.awssdk.services.sagemakerruntime.model.{InvokeEndpointRequest, InvokeEndpointResponse}
+
+object SagemakerInvokeFlow {
+
+  /**
+   * Java API: creates a [[SagemakerInvokeFlowStage]] for a AWS Lambda function invocation
+   * using [[software.amazon.awssdk.services.sagemakerruntime.SageMakerRuntimeAsyncClient]]
+   */
+  def apply(parallelism: Int)(
+      implicit sagemakerRuntimeClient: SageMakerRuntimeAsyncClient
+  ): Flow[InvokeEndpointRequest, InvokeEndpointResponse, NotUsed] =
+    Flow.fromGraph(new SagemakerInvokeFlowStage(sagemakerRuntimeClient)(parallelism))
+
+}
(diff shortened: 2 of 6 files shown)


ennru:
Hi @nkconnor,
Welcome to Alpakka!
Thank you for creating a connector for AWS Sagemaker. We're busy finishing releases right now, but will hopefully have time to give you feedback on it next week.

2m:
Hi @nkconnor, thanks for the PR. Looking at how the SQS and SNS connectors were updated to AWS SDK v2, I think the stage, that you extracted to the common module, could be skipped entirely.

The client invocations in the AWS SDK v2 return `CompletionStage`s which integrate with `mapAsync` directly.

Take a look at how the SNS connector looks currently: https://github.com/akka/alpakka/blob/master/sns/src/main/scala/akka/stream/alpakka/sns/scaladsl/SnsPublisher.scala#L45 and let me know if you could take a similar approach in the sagemaker connector.

nkconnor:
Thanks @2m ... I'll update SGM to match. Also wondering if Lambda SDK v2 #1456 should follow the same.

another note, on the SNS method you referenced they are using `mapAsyncUnordered`. That probably will lead to confusion for clients as I typically think of the default for flows as ordered. maybe better split up as `publishFlowUnordered` and `publishFlow`

gabfssilva:
> another note, on the SNS method you referenced they are using `mapAsyncUnordered`. That probably will lead to confusion for clients as I typically think of the default for flows as ordered. maybe better split up as `publishFlowUnordered` and `publishFlow`

The idea to use `mapAsyncUnordered` is to improve performance, since SNS has no order guarantees whatsoever.

2m:
>  Also wondering if Lambda SDK v2 #1456 should follow the same.

Yes, that is correct. I have added a comment on the Lambda PR: https://github.com/akka/alpakka/pull/1456#issuecomment-467145203

nkconnor:
```
object SagemakerInvokeFlow {

  /**
    * Scala API: creates a [[akka.stream.scaladsl.Flow Flow]] to invoke an SGM endpoint
    * using [[software.amazon.awssdk.services.sagemakerruntime.SageMakerRuntimeAsyncClient]]
    */
  def async(concurrency: Int)(
             implicit sgmClient: SageMakerRuntimeAsyncClient
           ): Flow[InvokeEndpointRequest, InvokeEndpointResponse, NotUsed] =
      Flow[InvokeEndpointRequest]
        .mapAsync(concurrency)(sgmClient.invokeEndpoint(_).toScala)

  /**
    * Scala API: creates a [[akka.stream.scaladsl.Flow Flow]] to invoke an SGM endpoint
    * using [[software.amazon.awssdk.services.sagemakerruntime.SageMakerRuntimeAsyncClient]]
    */
  def asyncUnordered(concurrency: Int)(
    implicit sgmClient: SageMakerRuntimeAsyncClient
  ): Flow[InvokeEndpointRequest, InvokeEndpointResponse, NotUsed] =
    Flow[InvokeEndpointRequest]
    .mapAsyncUnordered(concurrency)(sgmClient.invokeEndpoint(_).toScala)

}
```

How does this look (scaladsl)?

The thinness of it makes me wonder about the value of adding it to the alpakka repo. The most sane DSL additions I can think of at this time would be cutting away some more Java SDK boilerplate by including contentType specific methods (JSON.. as string?, PB).

2m:
I think we can make the API a bit more succint:

```scala
object Sagemaker {
  def flow(...) = ...
  def flowUnordered(...) = ...
}
```

We have a couple of connectors that are very thin layers on top of client libraries. So I think it is fine to have a sagemaker connector as well. Especially if you see some improvements that can be done regarding boilerplate required by the AWS SDK.

nkconnor:
Keep in mind the Sagemaker "Service" API has almost 100 methods. And it has been growing rapidly.

**Amazon SageMaker Service**
AddTags
CreateAlgorithm
CreateCodeRepository
CreateCompilationJob
CreateEndpoint
CreateEndpointConfig
CreateHyperParameterTuningJob
CreateLabelingJob
CreateModel
CreateModelPackage
CreateNotebookInstance
CreateNotebookInstanceLifecycleConfig
CreatePresignedNotebookInstanceUrl
CreateTrainingJob
CreateTransformJob
**.....**
UpdateNotebookInstance
UpdateNotebookInstanceLifecycleConfig
UpdateWorkteam

**Amazon SageMaker Runtime**
InvokeEndpoint

2m:
Okay. That surely makes it cumbersome to have a method for every request. But is it the case that every service request either inherits from or is composed in the `InvokeEndpointRequest`? Then having only one flow would be enough for every service request.

One approach to deal with this was taken in the Alpakka DynamoDb connector where a typeclass for every request was written, which connected in a typesystem a request with a response. And then a single flow was able to handle any request. That solution was pretty neat from Scala API, but Java API is still verbose.

nkconnor:
>But is it the case that every service request either inherits from or is composed in the InvokeEndpointRequest?

Actually the service I'm implementing in this PR is the only one composed with InvokeEndpointRequest. This is part of the SagemakerRuntime, a separate SDK from SagemakerService which has the numerous endpoints listed above.

SagemakerRuntime SDK (this PR) currently only has 1 method.

Just thinking that distinguising the runtime from future addition of the service SDK may be valuable ... so `object SagemakerRuntime` and `object SagemakerService..`.

This could be handled by package naming as well `io.akka.alpakka.sagemaker.runtime` and `io.akka.alpakka.sagemaker.service`

The AWS SDK uses `com.amazonaws.services.sagemaker` and `com.amazonaws.services.sagemakerruntime`

seglo:
@nkconnor Hi. Is there any interest in continuing this effort?

nkconnor:
@seglo I was waiting for feedback on the naming convention above, but, it’s been so long that I’m not working on this problem any longer.

seglo:
A little stale. Closing for now. Comment if you would like to pick up this work.

Issue #828, opened 2018-03-11
Build for Scala 2.13 milestone


ennru:
Depends on:
* [x] https://github.com/sullis/jms-testkit/issues/4
* [ ] akka/akka-grpc#506 (google-cloud-pub-sub-grpc, depending on https://github.com/scalapb/ScalaPB/issues/540)
* [ ] https://github.com/slick/slick/issues/1958
* [x] https://github.com/pauldijou/jwt-scala/issues/115 (google-fcm, google-cloud-pub-sub)
* [ ] https://github.com/hseeberger/akka-http-json/issues/253 (ironmq)
* [ ] https://jira.mongodb.org/browse/SCALA-506 (mongodb)
* [ ] matsluni/aws-spi-akka-http/issues/3 (sns, sqs)

ennru:
akka-stream will soon become available for Scala 2.13-M4.

2m:
The `aws-spi-akka-http` dependency `s3mock` depends on Alpakka S3 in the test scope. It has come full circle. We will have to start publishing 2.13-M5 artifacts for the Alpakka connectors that are not blocked by any dependencies so others can catch up.

matsluni:
@2m, I am the author of `aws-spi-akka-http`. Would it makes sense to drop `s3mock` for something which is not dependent on alpakka? I guess that would simplify things?!

@gabssilva suggested @localstack.
Do you know any other alternatives?

2m:
Most of the Alpakka connectors (S3 included) have a few Scala dependencies and are not blocked for support of Scala 2.13. So we will be releasing artifacts for those that are built with Scala 2.13 quite soon. This will unblock `s3mock`. However if you see a viable alternative, like localstack, then it might make things easier in the future, when subsequent Scala 2.13 releases are out.

PR #1530, opened 2019-02-26
Build with Scala 2.13

2m:
## Fixes

Towards #828.

## Purpose

This enables cross building for Scala 2.13.0-M5. Connectors that have either missing dependencies or are problematic to cross-build, have cross-publishing disabled.


hbase/src/main/scala/akka/stream/alpakka/hbase/HTableSettings.scala
```diff
@@ -76,8 +76,5 @@ object HTableSettings {
                 tableName: TableName,
                 columnFamilies: java.util.List[String],
                 converter: java.util.function.Function[T, java.util.List[Mutation]]): HTableSettings[T] =
-    HTableSettings(conf,
-                   tableName,
-                   columnFamilies.asScala.to[immutable.Seq],
-                   converter.asScala(_).asScala.to[immutable.Seq])
+    HTableSettings(conf, tableName, columnFamilies.asScala.toIndexedSeq, converter.asScala(_).asScala.toIndexedSeq)
```
raboof: Might be nice to have the newlines here?

xml/src/main/scala/akka/stream/alpakka/xml/model.scala
```diff
@@ -48,6 +48,11 @@ private class MapOverTraversable[A, K, V](source: Traversable[A], fKey: A => K,
 
   override def iterator: Iterator[(K, V)] = source.toIterator.map(a => (fKey(a), fValue(a)))
 
+  def remove(key: K): scala.collection.immutable.Map[K, V] =
+    throw new UnsupportedOperationException("Convert to map from iterator to support removing elements.")
+
+  override def updated[V1 >: V](key: K, value: V1): scala.collection.immutable.Map[K, V1] =
+    throw new UnsupportedOperationException("Convert to map from iterator to support updating elements.")
```
helena: I always wonder about code like this - declaring one return type but only throwing

project/Dependencies.scala
```diff
@@ -3,7 +3,9 @@ import Keys._
 
 object Dependencies {
 
-  val ScalaVersions = Seq("2.12.7", "2.11.12")
+  val Scala212 = "2.12.7"
+  val Scala213 = "2.13.0-M5"
+  val ScalaVersions = Seq(Scala212, "2.11.12", Scala213)
```
raboof: Might be nice to also add a constant for 211 for consistency ;)

project/Dependencies.scala
```diff
@@ -342,8 +343,13 @@ object Dependencies {
...
     libraryDependencies ++= Seq(
-      "com.github.matsluni" %% "aws-spi-akka-http" % "0.0.4" excludeAll ExclusionRule(organization = "com.typesafe.akka"), // ApacheV2
-      "software.amazon.awssdk" % "sns" % AwsSdk2Version excludeAll(ExclusionRule(organization = "software.amazon.awssdk", name = "netty-nio-client"), ExclusionRule(organization = "io.netty")), // ApacheV2
+      "com.github.matsluni" %% "aws-spi-akka-http" % "0.0.4" excludeAll ExclusionRule(
+        organization = "com.typesafe.akka"
+      ), // ApacheV2
+      "software.amazon.awssdk" % "sns" % AwsSdk2Version excludeAll (ExclusionRule(
+        organization = "software.amazon.awssdk",
+        name = "netty-nio-client"
+      ), ExclusionRule(organization = "io.netty")), // ApacheV2
```
helena: Perhaps keep the wrapping the same, license on the similar relation to what's being licensed in these two, for convention reasons, since they are now different.

xml/src/main/scala/akka/stream/alpakka/xml/model.scala
```diff
@@ -48,6 +48,11 @@ private class MapOverTraversable[A, K, V](source: Traversable[A], fKey: A => K,
 
   override def iterator: Iterator[(K, V)] = source.toIterator.map(a => (fKey(a), fValue(a)))
 
+  def remove(key: K): scala.collection.immutable.Map[K, V] =
+    throw new UnsupportedOperationException("Convert to map from iterator to support removing elements.")
```
raboof: `Traversable` is not really an iterator, since it should be traversable multiple times (otherwise it'd only be a `TraversableOnce`) - couldn't we do something like `new MapOverTraversable(source.filter(k => fKey(k) != key), fKey, fValue)` here?

xml/src/main/scala/akka/stream/alpakka/xml/model.scala
raboof: Here we'd have to actually convert all elements of the traversable... perhaps indeed best to not support that since it might be surprising performance-wise?

2m: Yea. I agree. I'll implement the method or will try to get rid of the `MapOverTraversable` all together.

project/Dependencies.scala
2m: Yea, scalafmt reformatted it this way...

hbase/src/main/scala/akka/stream/alpakka/hbase/HTableSettings.scala
2m: Its scalafmt. :)

xml/src/main/scala/akka/stream/alpakka/xml/model.scala
2m: I decided to remove `MapOverTraversable` and go with collection conversions. Favoring no exceptions rather than slightly faster implementation.

2m:
Addressed the feedback. Tickets for connectors that have cross-publishing disabled incoming.

2m:
Fixes #828

PR #1545, opened 2019-02-28
Add maxOpenRequests param in DynamoConfig

### LINKED ISSUE #1542: Can't specify maxOpenRequests in DynamoClient
### Short description

Currently both `maxConnections` and `maxOpenRequests` in Dynamo are both always set to the `parallelism` setting, and it isn't possible to set them individually.

### Details

In the code here:
https://github.com/akka/alpakka/blob/a0a5c698033b0bd0e9007dc953b6faabd799e0b8/dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/impl/DynamoClientImpl.scala#L37-L39

both values are always the same as `parallelism` and there doesn't seem to be a way to set them individually. I tried manually setting `akka.http.host-connection-pool.max-open-requests` in my config but it doesn't seem to do anything, and the code above takes precedence.

My use case is that on app startup, initial requests can take a long time (due to JVM warmup, etc) and in this case, the requests pile up in the queue up to a certain limit, and further requests are cancelled immediately rather than being enqueued and eventually completing. After some time, the JVM has warmed up and the requests take a normal amount of time with no backpressure issues.

Essentially, I would like to be able to set `maxOpenRequests` to a larger value to handle this startup case, but I don't want `maxConnections` to be an equally large number.

Or is there another recommended way to do this? I could have my own handler that catches this `BufferOverflowException` and attempts to re-queue it, but it seems weird to reimplement queueing when it's already configurable in akka-http.

williamho:
# Pull Request Checklist

## Fixes

Fixes #1542

## Purpose

Allows the `maxOpenRequests` of the underlying host connection pool to
be configured separately from the `maxConnections` value (as specified
by the `parallelism` value in the config).

## Background Context

To preserve old behavior, I made it an optional field, where if it's
absent, then it will use the `parallelism` value as before. However the
one thing I'm not sure of is whether I should have the `Option` exposed
in the public interface (since this is meant to be used in Java as
well). Glad to accept any suggestions on how to handle optional values
like this.

## References

#1542 - Can't specify maxOpenRequests in DynamoClient

### DIFF
--- a/dynamodb/src/main/resources/reference.conf
+++ b/dynamodb/src/main/resources/reference.conf
@@ -14,6 +14,10 @@ akka.stream.alpakka.dynamodb {
   # Max number of in flight requests from the AwsClient - must be a power of 2
   parallelism = 32
 
+  # Optional max number of outstanding requests to the underlying host connection pool.
+  # If unspecified, will be the same as the parallelism value. Must be a power of 2
+  #maxOpenRequests =
+
   # Optional Credentials. Used to define static credentials rather than use the DefaultAWSCredentialsProviderChain
   #credentials {
   #  access-key-id = "dummy-access-key"
--- a/dynamodb/src/main/resources/reference.conf
+++ b/dynamodb/src/main/resources/reference.conf
@@ -16,7 +16,7 @@ akka.stream.alpakka.dynamodb {
 
   # Optional max number of outstanding requests to the underlying host connection pool.
   # If unspecified, will be the same as the parallelism value. Must be a power of 2
-  #maxOpenRequests =
+  #max-open-requests =
 
   # Optional Credentials. Used to define static credentials rather than use the DefaultAWSCredentialsProviderChain
   #credentials {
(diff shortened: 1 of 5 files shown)


2m:
Thanks for the PR. It is looking very good. Regarding exposing `Option` in the public API - it is fine. It serves as a getter for Scala users. However you need to add a getter for Java users called `getGaxOpenRequests` which will return an `Optional`.

williamho:
Thanks, that makes sense. Now that I'm thinking about it though, while `maxOpenRequests` works for me, I can imagine that in the future, someone else will want to be able to set other fields on the underlying `ConnectionPoolSettings`.

Rather than adding just one configurable setting, would it be better to just accept the whole `ConnectionPoolSettings` as an optional param? This way it has maximum flexibility. Not sure if that would leak too much of the implementation details to the public API though.

williamho:
Actually, disregard my last comment. I forgot that the other connection pool settings can still be set with the usual `akka.http.host-connection-pool` config.

I've updated the PR to allow getting/setting with java Optionals, as well as using hyphens in the config for consistency with other config fields

2m:
Awesome stuff. Thanks a lot!

PR #1560, opened 2019-03-05
S3: add deleteByPrefix method

danelkotev:
## Purpose

Adding deleteByPrefix method which deletes all keys starting with a given prefix (similar to delete folder)


2m:
Great, thanks for the fixups!

PR #1561, opened 2019-03-06
Mqtt streaming: assertAllStageStopped

ennru:
Make use of `assertAllStagesStopped` in all tests running streams so that everything is shut down properly.

- [x] Needs some explanations in the docs

### DIFF
--- a/mqtt-streaming/src/test/scala/docs/scaladsl/MqttSessionSpec.scala
+++ b/mqtt-streaming/src/test/scala/docs/scaladsl/MqttSessionSpec.scala
@@ -1379,7 +1379,7 @@ class MqttSessionSpec
       server.watchCompletion().foreach(_ => session.shutdown())
     }
 
-    "close when no ping request received" ignore { // assertAllStagesStopped { // https://github.com/akka/akka/issues/17997#issuecomment-429670321
+    "close when no ping request received" ignore { // assertAllStagesStopped { // https://github.com/akka/alpakka/issues/1563
       val session = ActorMqttServerSession(settings)
 
       val client = TestProbe()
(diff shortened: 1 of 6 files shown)


mqtt-streaming/src/test/scala/docs/scaladsl/MqttSessionSpec.scala
```diff
@@ -763,12 +794,19 @@ class MqttSessionSpec
       session ! Command(publish)
 
       server.expectMsg(publishBytes)
+      // This reply triggers an error showing in the logs which hinders proper shutdown
+      //   7   │ 2019-03-06 11:28:23,035 ERROR [mqtt-spec-akka.actor.default-dispatcher-3] [akka.actor.OneForOneStrategy]  56 (of class java.lang.Integer)
+      //   8   │ scala.MatchError: 56 (of class java.lang.Integer)
+      //   9   │     at akka.stream.impl.fusing.GraphInterpreter.$anonfun$toSnapshot$4(GraphInterpreter.scala:662)
```
ennru: This indicates some error in an MQTT graph stage - I haven't gone deeper, yet.
cc @huntc

huntc: Thanks! I shall look at this first thing.

huntc:
I've been having a look at this, but not really coming up with anything concrete.

However, I note that the log states:

```
scala.MatchError: 56 (of class java.lang.Integer)
	at akka.stream.impl.fusing.GraphInterpreter.$anonfun$toSnapshot$4(GraphInterpreter.scala:662)
```

Here's the code at that line:

```scala
case x if (x | InClosed | OutClosed) == (InClosed | OutClosed) ⇒ ConnectionSnapshot.Closed
```

`56` represents `InClosed | OutClosed | OutReady`. I'm unfamiliar with the internals, but are we getting a match error here due to the match asserting that the stage cannot be both `OutClosed` and `OutReady`? Clearly, it is both... Should the test instead be:

```scala
case x if (x & (InClosed | OutClosed)) == (InClosed | OutClosed) ⇒ ConnectionSnapshot.Closed
```

ennru:
@2m and I discussed this, we've seen similar errors in Alpakka Kafka tests and we suspect it is connected to the use of TestProbes. Most likely not an error in MQTT streaming.

ennru:
I created #1563 to track that bit.

ennru:
@huntc Would you mind reviewing/approving this?

ennru:
Approved by @huntc -- I'll merge it now.

Issue #1549, opened 2019-03-01
Failed: MqttFlowSpec mqtt server flow should receive a bidirectional connection and a subscription to a topic


2m:
```
[info] MqttFlowSpec:
[info] mqtt client flow
[info] - should establish a bidirectional connection and subscribe to a topic (488 milliseconds)
[info] mqtt server flow
[info] - should receive a bidirectional connection and a subscription to a topic *** FAILED *** (5 seconds, 131 milliseconds)
[info]   A timeout occurred waiting for a future to complete. Queried 51 times, sleeping 100000000 nanoseconds between each query. (MqttFlowSpec.scala:151)
[info]   org.scalatest.concurrent.Futures$FutureConcept$$anon$1:
[info]   at org.scalatest.concurrent.Futures$FutureConcept.tryTryAgain$1(Futures.scala:538)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValueImpl(Futures.scala:550)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValueImpl$(Futures.scala:479)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:275)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue(Futures.scala:476)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue$(Futures.scala:475)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:275)
[info]   at docs.scaladsl.MqttFlowSpec.$anonfun$new$4(MqttFlowSpec.scala:151)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
```

https://travis-ci.org/akka/alpakka/jobs/500172922

huntc:
Hmm. I thought I had fixed this... will have to look again.

ennru:
New failure for this test during `assertAllStagesStopped`.

https://travis-ci.org/akka/alpakka/jobs/505595719#L701

2m:
https://travis-ci.com/akka/alpakka/jobs/203502744

2m:
[...]

PR #1564, opened 2019-03-07
Update akka-grpc to 0.6

2m:
## Purpose

Updates to the just released akka-grpc 0.6.

Had to drop 2.11 for Google Pub/Sub gRPC because akka-grpc does not support it anymore https://github.com/akka/akka-grpc/issues/520

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -71,7 +71,7 @@ lazy val alpakka = project
     ScalaUnidoc / unidoc / fullClasspath := {
       (ScalaUnidoc / unidoc / fullClasspath).value
         .filterNot(_.data.getAbsolutePath.contains("protobuf-java-2.5.0.jar"))
-	.filterNot(_.data.getAbsolutePath.contains("guava-26.0-android.jar"))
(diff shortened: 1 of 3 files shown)


ennru:
Build failure for 2.11
```
[error] sbt.librarymanagement.ResolveException: unresolved dependency: com.lightbend.akka#akka-stream-alpakka-google-cloud-pub-sub-grpc_2.11;1.0-M3+24-2aa67921: not found
```
But 2.13 compilation works. How can that be?

2m:
The `++` does the right thing, where it does not run the specified command on the projects that do not have specified Scala version in `crossScalaVersions`. 

The validation problem here with 2.11 was that Alpakka Google Pub/Sub gRPC is not available for 2.11 anymore, but doc-examples is still built with 2.11 and it depends on Pub/Sub connector.

Added a fix to stop compiling doc-examples with 2.11.

ennru:
Failures were #1549 and #1578

PR #1565, opened 2019-03-07
Mqtt streaming: convenience API for the casual user

ennru:
## Purpose

Add an API on top of MQTT streaming for the casual MQTT user so that migration off the Paho client becomes reasonable.

- [x] Introduce `MqttSource` factory methods for subscribing to MQTT
- [ ] Introduce `MqttFlow` factory methods for writing to MQTT

This is in very early development and I need help to get eg. acknowledging into a usable state. //cc @huntc


huntc:
First thoughts...

I’ve deliberately left ack-ing to the user as it isn’t about when the MQTT client receives the ack, but more about when the code that calls the MQTT client has *processed* the ack.

@longshorej Having implemented a complex app around the API we have, I think you’d also be able to add your opinion on the utility of higher level abstractions.

Parting thought for now, “less is more”. :-)

huntc:
PS I think it is great to be investigating this though!

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSource.scala
```diff
@@ -0,0 +1,81 @@
...
+// TODO anything we should do if no ConnAck arrives "in time"?
+//            case Right(event @ Event(_: ConnAck, _)) =>
+//              event
+              case Right(event @ Event(_: SubAck, _)) =>
+                subscribed.trySuccess(Done)
+                event
+              case Right(event) =>
+                event
+            }
+            .toMat(BroadcastHub.sink)(Keep.both)
```
huntc: You can also get a source to the queue by using prefixAndTail with an arg of 0. This is in place of BroadcastHub. I dunno which’d be more efficient though. I’ve used both forms.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/SetupStage.scala
```diff
@@ -0,0 +1,160 @@
...
+package akka.stream.alpakka.mqtt.streaming.impl
+
+import akka.annotation.InternalApi
+import akka.stream._
+import akka.stream.scaladsl.{Flow, Keep, Sink, Source}
+import akka.stream.stage._
+
+import scala.concurrent.{Future, Promise}
+
+private final class SetupSinkStage[T, M](factory: ActorMaterializer => Attributes => Sink[T, M])
```
huntc: It’d be great to have doc describing the motivation for these types.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSource.scala
```diff
@@ -0,0 +1,81 @@
...
+          Mqtt
+            .clientSessionFlow(mqttClientSession)
+            .join(Tcp(system).outgoingConnection("localhost", 1883))
+
+        val subscribed = Promise[Done]()
+
+        val (commands: SourceQueueWithComplete[Command[Nothing]], subscription: Source[Event[Nothing], NotUsed]) =
+          Source
+            .queue[Command[Nothing]](10, OverflowStrategy.fail)
+            .via(mqttFlow)
```
huntc: You’ll definitely need to consider restart flow here. Connections drop - a lot!

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSource.scala
```diff
@@ -0,0 +1,81 @@
...
+        implicit val system: ActorSystem = materializer.system
+        implicit val logging: LoggingAdapter = Logging.getLogger(system, this)
+
+        val mqttClientSession = ActorMqttClientSession(settings)
+        val mqttFlow: Flow[Command[Nothing], Either[MqttCodec.DecodeError, Event[Nothing]], NotUsed] =
+          Mqtt
+            .clientSessionFlow(mqttClientSession)
+            .join(Tcp(system).outgoingConnection("localhost", 1883))
+
+        val subscribed = Promise[Done]()
```
huntc: FYI You may get partial subscriptions ie the subscription response contains info on what was actually subscribed to.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSource.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  def atMostOnce(settings: MqttSessionSettings,
+                 clientId: String,
+                 subscriptions: immutable.Seq[(String, ControlPacketFlags)]): Source[Publish, Future[Done]] =
+    Setup
+      .source { implicit materializer => implicit attributes =>
+        implicit val system: ActorSystem = materializer.system
+        implicit val logging: LoggingAdapter = Logging.getLogger(system, this)
+
+        val mqttClientSession = ActorMqttClientSession(settings)
```
huntc: I think it is super important to keep the session parameterised to this function. Session life cycles endure beyond connections.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSource.scala
```diff
@@ -0,0 +1,81 @@
...
+                subscribed.trySuccess(Done)
+                event
+              case Right(event) =>
+                event
+            }
+            .toMat(BroadcastHub.sink)(Keep.both)
+            .run()
+
+        commands.offer(Command(Connect(clientId, ConnectFlags.CleanSession)))
+        commands.offer(Command(Subscribe(subscriptions)))
```
huntc: Use prepend on the flow.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSource.scala
```diff
@@ -0,0 +1,81 @@
...
+
+        val publishSource: Source[Publish, Future[Done]] = subscription
+          .collect {
+            case Event(publish @ Publish(flags, _, Some(packetId), _), _)
+                if flags.contains(ControlPacketFlags.RETAIN) =>
```
huntc: Why are you testing for retain? That’s for the server side mostly...

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/SetupStage.scala
ennru: Soon to become part of Akka... https://github.com/akka/akka/pull/26477

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSource.scala
ennru: Should all `Publish` events trigger a `PubAck` than?
I'll try to let an ack-handle be part of the emitted elements so that acking can be user-controlled for at-least-once semantics.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSource.scala
huntc: To add to this: when sending a Publish with a QoS > 0, the client session keeps track of whether the ACKs have taken place and will retry until they have. This concern spans connections.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSource.scala
huntc: QoS 0 doesn’t require an Ack. QoS 1 does. QoS 2 is quite a dance requiring 2.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSource.scala
ennru: You mean the materialized future should expose information from the `SubAck`? The `ControlPacketFlags`? 
Or should it even expect several `SubAck`s?

ennru:
I started looking into a flow which would behave as https://doc.akka.io/docs/alpakka/current/mqtt.html#publish-and-subscribe-in-a-single-flow
What I don't get is the way messages are sent to the broker. Why are they sent to the session and can't be offered to the source queue thingy?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSource.scala
huntc: Further thinking on this... if you're managing the connection within the convenience function, which you are, then you don't need to pass the session around after all. Nice one.

huntc:
> https://github.com/akka/alpakka/pull/1565#issuecomment-472083768

In the case of your API here, given that you're managing the session and the connection internal to your implementation, there is no distinction between a session and a connection to the user. Therefore, from your API's perspective, you don't have to worry about this.

However, to answer your question, sessions persist beyond connections. You can publish a message even if there's no connection. In this instance, if the publish has a QoS > 0 then it will continuously try and publish it until it receives its ack(s), which requires a connection.

mqtt-streaming/src/test/java/docs/javadsl/MqttSourceTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+    Thread.sleep(2 * 1000);
+    // #at-least-once
+
+    // stop the subscription
+    killSwitch.shutdown();
+    // #at-least-once
+
+    // TODO this assert fails
+    //    assertThat(
```
ennru: I have been fighting this a couple of hours, for some reason, the Java API doesn't really work :-(

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/MqttHighLevelApi.scala
```diff
@@ -0,0 +1,246 @@
...
+@ApiMayChange
+trait MqttTransportSettings {
+  def connectionFlow()(implicit system: ActorSystem): Flow[ByteString, ByteString, _]
+}
+
+@ApiMayChange
+final class MqttTcpTransportSettings private (
+    val host: String,
+    val port: Int
+) extends MqttTransportSettings {
```
huntc: No TLS considerations? I haven't yet connected to a single broker in the wild that doesn't require TLS.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/MqttHighLevelApi.scala
```diff
@@ -0,0 +1,246 @@
...
+    new MqttSubscriptions(subscriptions.asScala.toMap)
+
+  /** Add this subscription to the map of subscriptions configured already. */
+  def addSubscription(topic: String, qos: ControlPacketFlags): MqttSubscriptions =
+    new MqttSubscriptions(this.subscriptions.updated(topic, qos))
+
+  def addAtLeastOnce(topic: String): MqttSubscriptions =
+    new MqttSubscriptions(subscriptions.updated(topic, ControlPacketFlags.QoSAtLeastOnceDelivery))
+
+  def controlPacket: Subscribe = Subscribe(subscriptions.toList)
```
huntc: Other QoS levels?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/HighLevelMqttSource.scala
```diff
@@ -0,0 +1,158 @@
...
+  def atMostOnce(
+      mqttClientSession: MqttClientSession,
+      transportSettings: MqttTransportSettings,
+      restartSettings: MqttRestartSettings,
+      connectionSettings: MqttConnectionSettings,
+      subscriptions: MqttSubscribe
+  ): Source[Publish, Future[immutable.Seq[(String, ControlPacketFlags)]]] = {
+    type Out = Publish
+
+    val sendAcknowledge: SourceQueueWithComplete[Command[Nothing]] => PartialFunction[Event[Nothing], Out] =
```
huntc: Shouldn't you type param the `[Nothing]`?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSource.scala
```diff
@@ -0,0 +1,82 @@
...
+ */
+@ApiMayChange
+object MqttSource {
+
+  /**
+   * High-level API to subscribe to MQTT topics with at-most-once semantics.
+   */
+  @ApiMayChange
+  def atMostOnce(
+      mqttClientSession: MqttClientSession,
```
huntc: Given that you're controlling the entire connection, you can also control the session and not require the user to pass it in.

mqtt-streaming/src/test/scala/docs/scaladsl/MqttSourceSpec.scala
```diff
@@ -0,0 +1,355 @@
...
+
+      val mqttClientSession: MqttClientSession = ActorMqttClientSession(MqttSessionSettings())
+
+      val subscriptions = MqttSubscriptions(topic, ControlPacketFlags.QoSAtLeastOnceDelivery)
+      val (subscribed, received) = MqttSource
+        .atMostOnce(mqttClientSession,
+                    transportSettings,
+                    MqttRestartSettings(),
+                    MqttConnectionSettings(clientId),
+                    subscriptions)
```
huntc: Now seeing this, I wonder if there's an opportunity to conflate restart settings with connection settings. They are related.

mqtt-streaming/src/test/scala/docs/scaladsl/MqttSourceSpec.scala
```diff
@@ -0,0 +1,355 @@
...
+      }
+
+      publishFlow.watchCompletion().foreach { _ =>
+        mqttClientSession.shutdown()
+      }
+    }
+
+  }
+  private def sleepToReceiveAll(): Unit =
+    sleep(2.seconds, "to make sure we don't get more than expected")
```
huntc: I've always found that sleeping in tests leads to flaky CI scenarios. Is there a better way?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/HighLevelMqttSource.scala
```diff
@@ -0,0 +1,158 @@
...
+
+    createOnMaterialization[Out](mqttClientSession,
+                                 transportSettings,
+                                 restartSettings,
+                                 connectionSettings,
+                                 subscriptions,
+                                 sendAcknowledge)
+  }
+
+  def atLeastOnce[Out](
```
huntc: I'm unsure what the conventions should be, but we've been using type params as `A`, `B`, `C` etc. Perhaps we should be consistent even if it means changing the other code. WDYT?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/MqttHighLevelApi.scala
ennru: That's supported via `MqttTransportSettings` which is just a factory interface for connection flows.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/MqttHighLevelApi.scala
ennru: There is `addSubscription`.

huntc: Ah yes. Thanks.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/HighLevelMqttSource.scala
ennru: This is "just internal", we can change it when it becomes necessary.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/MqttHighLevelApi.scala
huntc: I guess for consistency though you could have `addAtMostOnce` and `addExactlyOnce`...?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSource.scala
ennru: Ok.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/HighLevelMqttSource.scala
huntc: OK.

mqtt-streaming/src/test/scala/docs/scaladsl/MqttSourceSpec.scala
ennru: I thought about it but wanted to keep the API with full flexibility untouched, maybe now if I move the `MqttClientSession` to be handled.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/HighLevelMqttSource.scala
ennru: I can rename it. For type-params which have an intention I often use "real" names.

huntc: Your call of course. I certainly don't mind `Out` - in fact, I quite like it... just wondering if there should be consistency or even whether it matters.

ennru:
Thank you for having another look at this, I know it is not perfect. I want an API for "the casual user" which isn't aware of the MQTT protocol internals (just like me).
When the back-pressure for sending to MQTT becomes ready, I'd like to see such a pre-packaged flow for sending messages, as well.

ennru:
Regarding the `PubAck`, are you saying I'm doing it entirely wrong?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/HighLevelMqttSource.scala
ennru: I like your ambition when working with Alpakka, though we have to accept we can't reach consistency on that level, I'm afraid.

huntc:
> Regarding the PubAck, are you saying I'm doing it entirely wrong?

No, just that I can't quite understand it. :-) Maybe I should take another look though.

`PubAck` is sent when the application is ready to receive another message; similar to reactive streams and requesting more elements. Therefore, making it automatic yields no real benefit (if that's what you're doing). Perhaps you can provide me an example of `PubAck` in action.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/HighLevelMqttSource.scala
huntc: I'm a very ambitious person. ;-)

ennru:
So `PubAck` conflates demand, and acknowledgement of "message received"?
The auto-`PubPack` is in the at-most-once flow, when subscribing with at-least-once the `Publish` packets don't contain a `packetId` so there is no chance to `PubAck`. AFAIU.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/HighLevelMqttSource.scala
```diff
@@ -0,0 +1,158 @@
...
+      restartSettings: MqttRestartSettings,
+      connectionSettings: MqttConnectionSettings,
+      subscriptions: MqttSubscribe
+  ): Source[Publish, Future[immutable.Seq[(String, ControlPacketFlags)]]] = {
+    type Out = Publish
+
+    val sendAcknowledge: SourceQueueWithComplete[Command[Nothing]] => PartialFunction[Event[Nothing], Out] =
+      commands => {
+        case Event(publish @ Publish(_, _, Some(packetId), _), _) =>
+          commands.offer(Command(PubAck(packetId)))
```
longshorej: I don't think you want to `PubAck` here - see [this table](https://docs.oasis-open.org/mqtt/mqtt/v5.0/cos02/mqtt-v5.0-cos02.html#_Toc1477420) for expected replies (none in QoS0)

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/HighLevelMqttSource.scala
```diff
@@ -0,0 +1,158 @@
...
+        .map {
+          case Left(decodeError) =>
+            throw new RuntimeException(decodeError.toString)
+          case Right(event @ Event(s: SubAck, _)) =>
+            val subscriptionAnswer = subscribePacket.topicFilters.map(_._1).zip(s.returnCodes)
+            subscribed.trySuccess(subscriptionAnswer)
+            event
+          case Right(event) =>
+            event
+        }
```
longshorej: In similar client code that we've written, we track `Connected`s for two purposes:

1) If we receive a `Connected` and we haven't received a `SUBACK`, our connection was likely restarted and the pending `SUBSCRIBE` was potentially dropped. Sometimes (rarely) this logic can result in two `SUBSCRIBE`s being sent to the broker, but that is benign.
2) If we receive a `Connected`, and we already have received a `SUBACK`, and `ConnAckFlags` doesn't contain `SessionPresent`, then we've reconnected to a broker that doesn't know about our session, and thus we fail the flow.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/HighLevelMqttSource.scala
```diff
@@ -0,0 +1,158 @@
...
+          .join(transport.connectionFlow())
+      }
+    }
+    val subscribed = Promise[immutable.Seq[(String, ControlPacketFlags)]]()
+
+    val subscribePacket = subscribe.controlPacket
+    val initCommands = immutable.Seq(
+      Command(connect.controlPacket),
+      Command(subscribePacket)
+    )
```
longshorej: Perhaps these should be prepended to the factory flow (L113-L115) so that they're sent on every connect? e.g. 

```scala
    val mqttFlow: Flow[Command[Nothing], Either[MqttCodec.DecodeError, Event[Nothing]], NotUsed] = {
      import restartSettings._
      RestartFlow.onFailuresWithBackoff(minBackoff, maxBackoff, randomFactor, maxRestarts) { () =>
        Flow[Command[Nothing]]
          .prepend(
            Source(
              immutable.Seq(
                Command(connect.controlPacket),
                Command(subscribePacket)
               )
            )
          )
          .via(
            Mqtt
              .clientSessionFlow(mqttClientSession)
              .join(transport.connectionFlow())
           )
      }
    }
```

ennru: Yes, thank you.

ennru:
I'm trying to understand if I should hide the `MqttClientSession` inside the `MqttSource` and how to use the `connectionId` in the best way.
But `testOnly *.MqttSourceSpec -- -z unacked` works sometimes and fails sometimes.

The log shows unhandled `ConnectionLost` and `Unregister` messages which seem to hinder proper shutdown as it fails in `assertAllStagesStopped`:
```
2019-05-10 14:47:15,856 INFO  [MqttSourceSpec-akka.actor.default-dispatcher-3] [akka.event.slf4j.Slf4jLogger]  Slf4jLogger started
2019-05-10 14:47:16,331 DEBUG [MqttSourceSpec-akka.actor.default-dispatcher-3] [docs.scaladsl.MqttSourceSpec]  sleeping 2 seconds to make sure we don't get more than expected
2019-05-10 14:47:16,503 DEBUG [MqttSourceSpec-akka.actor.default-dispatcher-16] [docs.scaladsl.MqttSourceSpec]  [client 1 received] Element: one-14:47:16.103
2019-05-10 14:47:16,505 DEBUG [MqttSourceSpec-akka.actor.default-dispatcher-2] [docs.scaladsl.MqttSourceSpec]  acked one-14:47:16.103
2019-05-10 14:47:16,509 DEBUG [MqttSourceSpec-akka.actor.default-dispatcher-2] [docs.scaladsl.MqttSourceSpec]  [client 1 received] Element: two-14:47:16.103
2019-05-10 14:47:16,510 DEBUG [MqttSourceSpec-akka.actor.default-dispatcher-2] [docs.scaladsl.MqttSourceSpec]  acked two-14:47:16.103
2019-05-10 14:47:16,511 DEBUG [MqttSourceSpec-akka.actor.default-dispatcher-14] [docs.scaladsl.MqttSourceSpec]  [client 1 received] Element: three-14:47:16.103
2019-05-10 14:47:18,335 DEBUG [MqttSourceSpec-akka.actor.default-dispatcher-16] [docs.scaladsl.MqttSourceSpec]  [client 1 received] Downstream finished.
2019-05-10 14:47:18,347 INFO  [MqttSourceSpec-akka.actor.default-dispatcher-2] [akka.actor.RepointableActorRef]  Message [akka.stream.alpakka.mqtt.streaming.impl.ClientConnector$ConnectionLost] without sender to Actor[akka://MqttSourceSpec/user/client-connector-1#-1279107321] was not delivered. [1] dead letters encountered. If this is not an expected behavior, then [Actor[akka://MqttSourceSpec/user/client-connector-1#-1279107321]] may have terminated unexpectedly, This logging can be turned off or adjusted with configuration settings 'akka.log-dead-letters' and 'akka.log-dead-letters-during-shutdown'.
2019-05-10 14:47:18,347 INFO  [MqttSourceSpec-akka.actor.default-dispatcher-2] [akka.actor.RepointableActorRef]  Message [akka.stream.alpakka.mqtt.streaming.impl.ClientConnector$ConnectionLost] without sender to Actor[akka://MqttSourceSpec/user/client-connector-1#-1279107321] was not delivered. [2] dead letters encountered. If this is not an expected behavior, then [Actor[akka://MqttSourceSpec/user/client-connector-1#-1279107321]] may have terminated unexpectedly, This logging can be turned off or adjusted with configuration settings 'akka.log-dead-letters' and 'akka.log-dead-letters-during-shutdown'.
2019-05-10 14:47:18,350 DEBUG [MqttSourceSpec-akka.actor.default-dispatcher-2] [docs.scaladsl.MqttSourceSpec]  sleeping 2 seconds to make sure we don't get more than expected
2019-05-10 14:47:18,367 DEBUG [MqttSourceSpec-akka.actor.default-dispatcher-2] [docs.scaladsl.MqttSourceSpec]  [client 2 received] Element: three-14:47:16.103
2019-05-10 14:47:18,368 DEBUG [MqttSourceSpec-akka.actor.default-dispatcher-2] [docs.scaladsl.MqttSourceSpec]  [client 2 received] Element: four-14:47:16.103
2019-05-10 14:47:18,368 DEBUG [MqttSourceSpec-akka.actor.default-dispatcher-3] [docs.scaladsl.MqttSourceSpec]  [client 2 received] Element: five-14:47:16.103
2019-05-10 14:47:20,351 DEBUG [MqttSourceSpec-akka.actor.default-dispatcher-16] [docs.scaladsl.MqttSourceSpec]  [client 2 received] Downstream finished.
2019-05-10 14:47:20,861 INFO  [MqttSourceSpec-akka.actor.default-dispatcher-15] [akka.actor.RepointableActorRef]  Message [akka.stream.alpakka.mqtt.streaming.impl.RemotePacketRouter$Unregister] without sender to Actor[akka://MqttSourceSpec/user/client-consumer-packet-id-allocator-0#906428085] was not delivered. [3] dead letters encountered. If this is not an expected behavior, then [Actor[akka://MqttSourceSpec/user/client-consumer-packet-id-allocator-0#906428085]] may have terminated unexpectedly, This logging can be turned off or adjusted with configuration settings 'akka.log-dead-letters' and 'akka.log-dead-letters-during-shutdown'.
2019-05-10 14:47:25,859 ERROR [MqttSourceSpec-akka.actor.default-dispatcher-15] [akka.actor.OneForOneStrategy]  67108865
java.lang.ArrayIndexOutOfBoundsException: 67108865
	at 
```

The last bit is from `assertAllStagesStopped` https://github.com/akka/akka/issues/26902.

huntc:
The connection id should be distinct for each connection made. Does that help?

ennru:
No, not really, but that is just me being slow, I guess. 
Doesn't that mean we should generate a new connection ID automatically on every connect? Eg. within a `RestartSource`

huntc:
Yes

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSource.scala
ennru: Ok, I'll try that instead now.

ennru:
And the `connectionId` should be unique per connection within a `MqttClientSession`. Wouldn't it be desirable to make the session create those connection ids?

huntc:
The connection ids pertain to the flow that’s materialised in relation to a connection, like a TCP connection, not a session. There can be many connections over the lifetime of one session.

ennru:
Yes, I got that. I'm looking for a good way to create unique connection identifiers and as they should be unique within a session it looks like a good place to hide a counter.

My test using acking works well locally, but it received an already acked message on in a new session and new connection when running on Traivs. Ideas?

https://travis-ci.com/akka/alpakka/jobs/200663311#L1400

huntc:
UUIDs should be fine, although I think we use the client’s port as the id. @longshorej?

Not sure what’s going on with that test. I’d have to take a closer look. Is this an existing test?

huntc:
I just took a look. When just use an AtomicLong counter for the client connection id. From memory, the server uses the client port.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/HighLevelMqttSource.scala
ennru: It might not be intended use, but in a test "At-most-once - should receive subscribed messages only once (AtLeastOnceDelivery)", I use a QoS1 subscription on this `atMostOnce` source. When I don't do this acking it doesn't progress after the first element...

huntc: I may not be understanding, but QoS 0 isn’t guaranteed to receive anything at all...

ennru:
I had a look at this historic PR and decided to give it a refresh instead of closing it.
I still think it would be worthwhile to offer a high-level API which makes MQTT streaming usable for people without deep knowledge of the protocol. My knowledge of it is quite shallow as well...

huntc:
It’s been a while! Happy to help. I’ve got a particular interest in simplifying the intervals also (separate PR). Btw this lib is being heavily used by us so it is working quite well.

ennru:
I won't have time to get back to this, I'm afraid.

PR #1577, opened 2019-03-12
MQTT streaming: SourceQueue backpressure

longshorej:
This adds a mechanism to `mqtt-streaming` to use `Source.queue` with `OverflowStrategy.backpressure` and wait for the backpressure signal before handling more messages.

I think we could probably increase from 1 to say 4 or 8, reducing latency at the expense of the cost of the memory required to buffer those events, but curious to get opinions on that.

While we're at it, it also removes `BroadcastHub` usage when there's only a single consumer, which should be a bit more performant.

Some funky things with the tests, in particular doc ones, so this is a draft for now.

/cc @huntc who vowed to pick up where I've left off :smiley:


mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/QueueOfferState.scala
```diff
@@ -0,0 +1,47 @@
...
+              throw new IllegalStateException(s"Failed to offer to queue: $other")
+
+            case Left(failure) =>
+              throw failure
+          }
+
+        case (_, other) =>
+          waitForQueueOfferCompleted(behavior, stash :+ other)
+      }
+      .orElse(behavior) // handle signals immediately
```
longshorej: One thing I'm not clear on are the semantics here. Note that on L43, we catch all events that flow. Thus, the intention is that all signals go to `behavior` but nothing else (until we hit L34 in the future)

If `behavior`'s signal handler (i.e. the behavior being run via `orElse`) returns `Behavior.same`, I'd like to clarify that it doesn't change the behavior of the actor to itself, i.e. `waitForQueueOfferCompleted`  (L26-L46) remains the behavior. Hopefully my question makes sense..

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -195,17 +201,28 @@ import scala.util.{Failure, Success}
...
       val (queue, source) = Source
-        .queue[ForwardConnectCommand](1, OverflowStrategy.dropHead)
-        .toMat(BroadcastHub.sink)(Keep.both)
-        .run()
-      remote.success(source)
+        .queue[ForwardConnectCommand](1, OverflowStrategy.backpressure)
+        .prefixAndTail(0)
+        .map(_._2)
+        .toMat(Sink.head)(Keep.both)
+        .run
```
huntc: @ennru @2m We'd be interested in your views on this pattern as opposed to using `BroadcastHub`. We are wondering if it might be more efficient to use `prefixAndTail(0)` rather than a `BroadcastHub` where we know there is only one consumer - possibly less machinery. Note that this `prefixAndTail(0)` pattern is also used in the Unix Domain Socket code. WDYT?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/QueueOfferState.scala
huntc: I *think* the behaviour would indeed remain the same i.e. stay as per the behaviour of `waitForQueueOfferCompleted`. But I've not yet proved this.

However, given L46 with the `orElse`, I'm not seeing the signal handler of `behavior` being applied. When I run the tests I see a `DeathPactException` which means that there is no signal handler for `Terminated`, but there is such a signal handler for `behavior`...

huntc:
I've been looking at this in some detail and ended posing [a question to the Lightbend forum](https://discuss.lightbend.com/t/akka-typed-behavior-orelse-appears-not-to-propagate-termination-signals/3643). I've also tidied up a couple of things around the edge, but nothing specifically to do with this PR.

@patriknw 'hoping you may be able to enlighten me re my question. :-)

mqtt-streaming/src/test/scala/docs/scaladsl/MqttSessionSpec.scala
```diff
@@ -1037,14 +1037,14 @@ class MqttSessionSpec
...
           )
-          .toMat(Sink.ignore)(Keep.both)
+          .toMat(Sink.headOption)(Keep.both)
           .run()
 
       val connect = Connect("some-client-id", ConnectFlags.None)
 
       client.offer(Command(connect))
 
-      result.failed.futureValue shouldBe a[WatchedActorTerminatedException]
```
huntc: This exhibits the core of my change here - this termination exception isn't something that the caller should be concerned with. It is internal to the session.

mqtt-streaming/src/test/scala/docs/scaladsl/MqttSessionSpec.scala
```diff
@@ -1271,7 +1271,7 @@ class MqttSessionSpec
             case Right(Event(cp: Unsubscribe, _)) if cp.topicFilters == unsubscribe.topicFilters =>
               unsubscribeReceived.success(Done)
           })
-          .toMat(Sink.seq)(Keep.both)
+          .toMat(Sink.ignore)(Keep.left)
           .run()
```
huntc: The `result` wasn't being used, so I removed it.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSession.scala
```diff
@@ -538,6 +544,9 @@ final class ActorMqttServerSession(settings: MqttSessionSettings)(implicit mat:
                 case c: Command[A] => throw new IllegalStateException(c + " is not a server command")
               }
             )
+            .recover {
+              case _: WatchedActorTerminatedException => ByteString.empty
```
ennru: Great to clear those from the logs when they shouldn't concern anyone.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
jrudolph: I'd suggest to use the more idiomatic `BroadcastHub` until that turns out to be an actual performance issue. But I can see how broadcast hub introduces a bit of extra baggage but it's probably not a problem.

huntc: Everything is a problem in our resource constrained environment! ;-) we should of course benchmark it. 

Putting the question another way, do you see anything wrong with this approach in principle?

longshorej: I put together a silly comparison and found that at least in terms of running time `BroadcastHub` performs slightly better (~6% faster): https://github.com/longshorej/garbage/tree/master/akka-broadcast-comparison

```
akka-broadcast-comparison WARMUP: 2 * 100x
akka-broadcast-comparison ************************
akka-broadcast-comparison [broadcastHub] time (averaged over 100 times): 30.486356810000004ms
akka-broadcast-comparison [prefixAndTail] time (averaged over 100 times): 27.15035694ms
akka-broadcast-comparison [broadcastHub] time (averaged over 100 times): 25.54272718ms
akka-broadcast-comparison [prefixAndTail] time (averaged over 100 times): 26.1548963ms
akka-broadcast-comparison ************************
akka-broadcast-comparison RUN: 1000x
akka-broadcast-comparison ************************
akka-broadcast-comparison [broadcastHub] time (averaged over 1000 times): 25.498494859999997ms
akka-broadcast-comparison [prefixAndTail] time (averaged over 1000 times): 27.157771376ms
akka-broadcast-comparison ... finished with exit code 0
```

This is timing how long it takes to offer 10k items to the queue and consume them via a `Sink.ignore`. The result indicated is a mean average over 1k executions.

The first warmup having `prefixAndTail` slightly ahead is just due to everything JITing. By the second warmup, `BroadcastHub` performs a bit better.

Of course, YMMV given the variance in microbenchmarks. But it to me supports what @jrudolph is saying, i.e. let's use `BroadcastHub`.

longshorej: Here's a `/usr/bin/time` comparision for peak RSS, again slight advantage in most metrics to BroadcastHub. Also note again this is a pretty dumb benchmark :stuck_out_tongue: 

### BroadcastHub

```
/usr/bin/time -v java -jar /home/longshorej/work/garbage/akka-broadcast-comparison/target/scala-2.12/akka-broadcast-comparison-assembly-0.1.0-SNAPSHOT.jar broadcastHub
************************
WARMUP: 2 * 100x
************************
[broadcastHub] time (averaged over 100 times): 32.08024019ms
[broadcastHub] time (averaged over 100 times): 27.110958380000003ms
************************
RUN: 1000x
************************
[broadcastHub] time (averaged over 1000 times): 25.825699086ms
	Command being timed: "java -jar /home/longshorej/work/garbage/akka-broadcast-comparison/target/scala-2.12/akka-broadcast-comparison-assembly-0.1.0-SNAPSHOT.jar broadcastHub"
	User time (seconds): 111.35
	System time (seconds): 3.40
	Percent of CPU this job got: 354%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:32.33
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 235636
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 0
	Minor (reclaiming a frame) page faults: 34031
	Voluntary context switches: 857104
	Involuntary context switches: 170048
	Swaps: 0
	File system inputs: 0
	File system outputs: 0
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0
```

## prefixAndTail

```
/usr/bin/time -v java -jar /home/longshorej/work/garbage/akka-broadcast-comparison/target/scala-2.12/akka-broadcast-comparison-assembly-0.1.0-SNAPSHOT.jar prefixAndTail
************************
WARMUP: 2 * 100x
************************
[prefixAndTail] time (averaged over 100 times): 34.995869400000004ms
[prefixAndTail] time (averaged over 100 times): 27.3833828ms
************************
RUN: 1000x
************************
[prefixAndTail] time (averaged over 1000 times): 27.194013631ms
	Command being timed: "java -jar /home/longshorej/work/garbage/akka-broadcast-comparison/target/scala-2.12/akka-broadcast-comparison-assembly-0.1.0-SNAPSHOT.jar prefixAndTail"
	User time (seconds): 116.08
	System time (seconds): 3.65
	Percent of CPU this job got: 351%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:34.02
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 236244
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 0
	Minor (reclaiming a frame) page faults: 36818
	Voluntary context switches: 1029486
	Involuntary context switches: 193720
	Swaps: 0
	File system inputs: 0
	File system outputs: 0
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0
```

huntc: BroadcastHub it is then! Thanks for the quick test.

ennru: Have you considered `Sink.asPublisher`? Might be interesting to measure how that compares.

2m: Yes. And with `Sink.asPublisher[T](fanout: Boolean)` you can control wether you want to support fanout or not.

longshorej: Ah, that's quite a bit faster for the microbenchmark, which makes sense as it doesn't need to wait for an offer signal:

```
fanOut = false

[publisher] time (averaged over 1000 times): 3.76545015ms
```
```
fanOut = true
[publisher] time (averaged over 1000 times): 8.502225306ms
```

I'm not sure how we can apply it to the MQTT usage though, as we are offering items one at a time occasionally. Perhaps we can use `fanOut = true` and run a `Source.single` through it in each place where we currently offer. I think that means we lose any form of backpressure on the outgoing elements though.

longshorej:
FYI - This is waiting on an Akka release with https://github.com/akka/akka/pull/26524 given its reliance on `Behavior#orElse`.

longshorej:
Rebased this to `master`, still waiting on an Akka release.

huntc:
Apparently, there may be an Akka release next week.

ennru:
I know you are keen to get this in now that Akka 2.5.22 is available.
Please hang in until we do Alpakka 1.0.1.

ennru:
This is still a draft, anything more we need to solve?

longshorej:
Ah, that was oversight! This is ready for review/merge.

longshorej:
Thanks @ennru - I'll get this updated tomorrow.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -463,10 +488,20 @@ import scala.util.{Failure, Success}
...
-            serverConnected(data)
+            import context.executionContext
+
+            data.remote
+              .offer(ForwardPublish(publish, None))
+              .onComplete(result => context.self.tell(QueueOfferCompleted(ByteString.empty, result.toEither)))
+
+            QueueOfferState.waitForQueueOfferCompleted(
+              serverConnected(data),
+              Seq.empty
```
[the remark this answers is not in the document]
longshorej: done

project/Dependencies.scala
```diff
@@ -10,12 +10,12 @@ object Dependencies {
 
   val AkkaVersion = sys.env.get("AKKA_SERIES") match {
     case Some("2.4") => sys.error("Akka 2.4 is not supported anymore")
-    case _ => "2.5.21"
+    case _ => "2.5.22"
   }
 
   val AwsSdkVersion = "1.11.476"
   val AwsSdk2Version = "2.5.20"
-  val AkkaHttpVersion = "10.1.7"
```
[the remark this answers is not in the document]
longshorej: thanks - removed

longshorej:
Okay, this is all set.

longshorej:
or not - MiMA failures. I *think* it's related to the new field `MqttSessionSettings#clientSendBufferSize`, but note that  `MqttSessionSettings` has a private constructor. Can I just whitelist the change?

ennru:
Yes, that is why we require the constructors to be private.

ennru:
I took the liberty to add the MiMa rule.

ennru:
@longshorej @huntc 	Just to be sure, as I guess you've run this code for a while now -- are we good to make it part of Alpakka 1.1.0 next week?

huntc:
Yeah, let’s get this in! Thanks

2m:
Rebased from the latest master, as the MiMa previous artifact version was derived to be `1.0.1` (because this PR was started before `1.0.2` release).

ennru:
Thank you for this important improvement.

PR #1582, opened 2019-03-14
Fixes wrong QoS condition in publish message flow in mqtt-streaming c…

jcroig:
…onnector.

# Pull Request Checklist

* [ ] Have you added tests for any changed functionality?

## Fixes

Fixes wrong QoS condition in publish message flow in mqtt-streaming connector. There's no functional change. Actually old code was working properly even when the check was incorrect by chance since the boolean check evaluates true for both contants `ControlPacketFlags.QoSAtMostOnceDelivery` and 
`ControlPacketFlags.QoSExactlyOnceDelivery` when message QoS is 2. However the code is misleading and doesn't reflect its real intention which is verify the QoS is "exactly once delivery" instead of "at most once delivery"

## Purpose

Substitute wrong QoS check by the correct one so that code express its intention beyond following the MQTT spec.

## Background Context

Simple change it is only a wrong constant value and used the right one.

## References

http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html

### DIFF
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/RequestState.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/RequestState.scala
@@ -125,7 +125,7 @@ import scala.util.{Failure, Success}
             local.success(ForwardPubAck(data.publishData))
             Behaviors.stopped
           case PubRecReceivedFromRemote(local)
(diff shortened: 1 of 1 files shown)


ennru:
Thank you, great to see more people trying the MQTT streaming connector!

jcroig:
My pleasure!

PR #1584, opened 2019-03-15
support confirm amqp flow, resolve #1132

### LINKED ISSUE #1132: AMQP : support "confirm" Sink and Flow producers

fancywriter:
Sorry, didn't notice already existing PR #1330, but did some things on my free time, wanted to share.
But the implementation is a little bit different, it uses confirm listener instead of waiting. I would prefer this option.

Feel free to share your thoughts about this draft.


ennru:
We don't have the time to look into this right now, what about if @jcare44 and @fancywriter come up with a combined proposal?

PR #1585, opened 2019-03-15
Use setup stage and attrbutes for external resources

### LINKED ISSUE #1575: Update reference connector to use setup stage

2m:
## Fixes

Fixes #1575

## Purpose

Reference connector is updated to use the Setup stage and attributes for external resource resolution, as it is already implemented in S3, DynamoDb and other connectors.


reference/src/main/scala/akka/stream/alpakka/reference/Resource.scala
```diff
@@ -20,9 +21,75 @@ import akka.util.ByteString
...
+}
+
+object Resource {
+  def apply(settings: ResourceSettings) = new Resource(settings)
+
+  def create(settings: ResourceSettings) = Resource(settings)
+}
+
+/**
+ * Settings required for the Resource should be extracted to a separate clas.
```
ennru:
```suggestion
 * Settings required for the Resource should be extracted to a separate class.
```

PR #1592, opened 2019-03-19
Termination signal handling

huntc:
There were a couple of behaviours that were not handling the termination signal where it was possible to receive it. This PR fixes that.

I recommend reviewing without white space changes as the formatter wants to format.


huntc:
An mqtt-streaming test failed, but I doubt anything to do with this change: https://travis-ci.com/akka/alpakka/jobs/185870624. This appears to be more of https://github.com/akka/alpakka/issues/1549#issuecomment-472345024.

PR #1595, opened 2019-03-19
mqtt-streaming: Various fixes

longshorej:
A number of commits here that improve the resilience of the mqtt-streaming connector. #1577 remains important as well, but will require an Akka release.

##### mqtt-streaming: Require a connection id to be specified for client flows 

This allows the MQTT session to better track which events are relevant, given that there can be races when old connections are torn down and new ones are established.

##### mqtt-streaming: Fix a bug where actor state was closed over 

This could cause duplicate publications to never be sent to new connections

##### mqtt-streaming: Republish messages on reconnect only by default 

Previously, messages for QoS1/2 were only republished on an interval after not receiving an ack.

It is more conventional to instead republish everything only on connect, and indeed to be compliant for MQTT 5, that is the only time this is allowed.

To accommodate this, the timeouts default to 0, but the previous behavior can still be restored by changing the default producer timeout settings.

### DIFF
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
@@ -11,6 +11,7 @@ import akka.actor.typed.scaladsl.{ActorContext, Behaviors}
 import akka.annotation.InternalApi
 import akka.stream.{Materializer, OverflowStrategy}
 import akka.stream.scaladsl.{BroadcastHub, Keep, Sink, Source, SourceQueueWithComplete}
+import akka.util.ByteString
 
 import scala.concurrent.Promise
 import scala.concurrent.duration.FiniteDuration
@@ -98,6 +99,7 @@ import scala.util.{Failure, Success}
         settings
       )
   final case class ConnectReceived(
+      connectionId: ByteString,
       connect: Connect,
       connectData: ConnectData,
       remote: SourceQueueWithComplete[ForwardConnectCommand],
@@ -124,6 +126,7 @@ import scala.util.{Failure, Success}
         settings
       )
   final case class ConnAckReceived(
+      connectionId: ByteString,
       connectFlags: ConnectFlags,
       keepAlive: FiniteDuration,
       pendingPingResp: Boolean,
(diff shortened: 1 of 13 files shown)


longshorej:
This is ready for a review @ennru @2m when you have a moment.

/cc @huntc

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -151,32 +154,58 @@ import scala.util.{Failure, Success}
...
+
+  final case class PublishReceivedFromRemote(override val connectionId: ByteString,
+                                             publish: Publish,
+                                             local: Promise[Consumer.ForwardPublish.type])
+      extends Event(connectionId)
+
+  final case class ConsumerFree(topicName: String) extends Event(ByteString.empty)
+
+  final case class PublishReceivedLocally(publish: Publish, publishData: Producer.PublishData)
+      extends Event(ByteString.empty)
```
longshorej: These and related aren't associated with a connection id, so their connection id is an empty `ByteString`. This is similar to how the server does things.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -283,17 +319,25 @@ import scala.util.{Failure, Success}
...
+                               data.settings.receiveConnAckTimeout)
       Behaviors
         .receivePartial[Event] {
-          case (context, ConnAckReceivedFromRemote(connAck, local))
+          case (context, connect @ ConnectReceivedLocally(connectionId, _, _, _))
+              if connectionId != data.connectionId =>
+            context.self ! connect
+            disconnect(context, data.remote, data)
+          case (_, event) if event.connectionId.nonEmpty && event.connectionId != data.connectionId =>
+            Behaviors.same
```
longshorej: This case is responsible for letting empty (thus session bound) or matching connection ids fall through to other cases.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -458,20 +515,20 @@ import scala.util.{Failure, Success}
             } else {
               serverConnected(data.copy(activeProducers = data.activeProducers - topicName))
             }
-          case (_, ReceivedProducerPublishingCommand(command)) =>
-            command.runWith(Sink.foreach {
-              case Producer.ForwardPublish(publish, packetId) => data.remote.offer(ForwardPublish(publish, packetId))
-              case Producer.ForwardPubRel(_, packetId) => data.remote.offer(ForwardPubRel(packetId))
```
longshorej: Closing over old state -- if the client ever reconnected, `data.remote` would be from an old connection and thus never actually republished.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -226,8 +257,13 @@ import scala.util.{Failure, Success}
               )
             )
           } else {
+            data.activeProducers.values.foreach { producer =>
+              producer ! Producer.ReceiveConnect
+            }
```
longshorej: Main change on the client to let the producers know about new `ReceiveConnect` so that they can republish.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ServerState.scala
```diff
@@ -390,6 +390,9 @@ import scala.util.{Failure, Success}
             data.stash.foreach(context.self.tell)
             timer.cancel(ReceiveConnAck)
 
+            data.activeProducers.values
+              .foreach(_ ! Producer.ReceiveConnect)
```
longshorej: Similarly on the server.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSession.scala
```diff
@@ -152,7 +152,9 @@ final class ActorMqttClientSession(settings: MqttSessionSettings)(implicit mat:
 
   private val pingReqBytes = PingReq.encode(ByteString.newBuilder).result()
 
-  override def commandFlow[A]: CommandFlow[A] =
+  override def commandFlow[A](connectionId: ByteString): CommandFlow[A] = {
+    val reply = Promise[Source[ClientConnector.ForwardConnectCommand, NotUsed]]
+
```
huntc: I'm unsure why this `reply` declaration has been moved...

huntc: That said, moving it is benign.

huntc:
@ennru @2m All LGTM. Thanks @longshorej !

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSession.scala
longshorej: Ah, this was a leftover as I was iterating on things. Thanks for noticing. I shall fix in a bit.

longshorej: This is fixed.

ennru:
Can you add a MiMa filter for everything in the `impl` package, please? 
Than we'll see how much the public API changed.

huntc:
> Can you add a MiMa filter for everything in the impl package, please? Than we'll see how much the public API changed.

I've not used MiMa before, and I've been looking a way for each project to be able to conveniently exclude their `impl` resident types. Do you know how that can be achieved with ease?

ennru:
Fair enough. I'll add it later today.

ennru:
Now we're down to the API breaking changes we want to see

* method `clientSessionFlow(akka.stream.alpakka.mqtt.streaming.javadsl.MqttClientSession): akka.stream.javadsl.BidiFlow` in object `akka.stream.alpakka.mqtt.streaming.javadsl.Mqtt` does not have a correspondent in current version
* method `clientSessionFlow(akka.stream.alpakka.mqtt.streaming.scaladsl.MqttClientSession): akka.stream.scaladsl.BidiFlow` in object `akka.stream.alpakka.mqtt.streaming.scaladsl.Mqtt` does not have a correspondent in current version

as those got a new parameter introduced.

longshorej:
Yeah, that's expected as each flow needs a connection id now (similar to the server). This module is considered an "API may change" one still, right? Given that we're still going through hardening exercises that seems reasonable to me.

huntc:
The lack of a connection id was an oversight on my part initially. 

It is all because of this notion of a session with MQTT. A session must distinguish connections. We were doing this on the server side and exposing it in the api, but we weren’t exposing it for the client api. The client session needs to understand which connection is sending a message as there can be a race between an active one shutting down and a new one being established.

Incidentally, your higher level APIs can hide this complexity given that you’re managing the connection entirely. A good candidate for a client connection id is the client port that the TCP socket has (not ther server port!). 

HTH.

longshorej:
I added the requested deprecated methods, but MiMa is still complaining about something. Any advice @ennru?

ennru:
Looks like you removed the MiMa exclusions file, the version I provided is required for it to accept the other changes.

longshorej:
Ah thanks, that must have happened in the rebase. I've readded it.

ennru:
Thank you @longshorej and @huntc for your continued work in this! Great to see other contributors helping out, as well.

PR #1600, opened 2019-03-20
FTP: fix default sftp overwrite behaviour

### LINKED ISSUE #1599: SFTP: overwrite existing file doesn't work
### Versions used 
Alpakka version: 1.0-RC1

### Expected Behavior

Uploading with `Sftp.toPath(path, settings, append = false)` should overwrite existing file.

### Actual Behavior

It doesn't work. New data is written at the start of the existing file.

### Relevant logs

It's perhaps due to missing `TRUNC` mode passed to the handler:

https://github.com/akka/alpakka/blob/v1.0-RC1/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/SftpOperations.scala#L107

doc:
```
net.schmizz.sshj.sftp.OpenMode.Trunc

  /**
     * Forces an existing file with the same name to be truncated to zero length when creating a file by specifying
     * {@link OpenMode#CREAT}. {@link OpenMode#CREAT} MUST also be specified if this flag is used.
     */
    TRUNC(0x00000010)
```

Fix:
` val openModes = Set(WRITE, CREAT) ++ (if (append) Set(APPEND) else Set(TRUNC))`

nukky94:
## Fixes

Fixes #1599 

## Purpose

Add missing `TRUNC` mode if `append` is false.
It should overwrite file if exists.

### DIFF
--- a/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/SftpOperations.scala
+++ b/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/SftpOperations.scala
@@ -104,7 +104,7 @@ private[ftp] trait SftpOperations { _: FtpLike[SSHClient, SftpSettings] =>
 
   def storeFileOutputStream(name: String, handler: Handler, append: Boolean): Try[OutputStream] = Try {
     import OpenMode._
-    val openModes = Set(WRITE, CREAT) ++ (if (append) Set(APPEND) else Set())
+    val openModes = Set(WRITE, CREAT) ++ (if (append) Set(APPEND) else Set(TRUNC))
     val remoteFile = handler.open(name, openModes.asJava)
     val os = new remoteFile.RemoteFileOutputStream() {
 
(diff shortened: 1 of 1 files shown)


nukky94:
I don't know why the build failed prematurely on Travis CI

ennru:
There are a few flaky tests in the Ftp module https://github.com/akka/alpakka/issues?utf8=%E2%9C%93&q=is%3Aopen+label%3Ap%3Aftp+label%3Afailed

ennru:
Speaking of tests, would you be able to add a test exercising the change you suggest?

nukky94:
There is already a test: https://github.com/akka/alpakka/blob/v1.0-RC1/ftp/src/test/scala/akka/stream/alpakka/ftp/CommonFtpStageSpec.scala#L208

ennru:
But that test did not catch that the overwrite behaviour was wrong. Another assert on that test might be enough to make sure all FTP sinks behave correctly.

PR #1603, opened 2019-03-23
MQTT Streaming: Fix keepalive

### LINKED ISSUE #1602: MQTT Streaming: keepalive might not work as expected
According to the MQTT spec the keepalive is "the maximum time interval that is permitted to elapse between the point at which the Client finishes transmitting one MQTT Control Packet and the point it starts sending the next". 

In my understanding (and observation) the current implementation resets the keepalive timer whenever an event is received from remote e.g. https://github.com/akka/alpakka/blob/master/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala#L366. In situation where no other publishing is done on the client side the next ping is sent after the keepalive timeout has already expired on the broker's side. The broker might terminate the connection then.

mwegrz:
# Pull Request Checklist

## Fixes

Fixes #1602

### DIFF
--- a/mqtt-streaming/src/main/mima-filters/1.0-RC1.backwards.excludes
+++ b/mqtt-streaming/src/main/mima-filters/1.0-RC1.backwards.excludes
@@ -1,0 +1,2 @@
+# Allow changes to impl
+ProblemFilters.exclude[Problem]("akka.stream.alpakka.mqtt.streaming.impl.*")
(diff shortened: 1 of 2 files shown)


mwegrz:
@huntc thank you for the quick review. Changed the parameter name and rebased as you suggested.

ennru:
I added the required file to exclude the `impl` package from binary compatibility checks.

PR #1604, opened 2019-03-24
SQS: Revise stream element data model

### LINKED ISSUE #625: No support for SQS Fifo queue?
When I try to use a FIFO queue, I get the following error:

> The request must contain the parameter MessageGroupId. (Service: AmazonSQS; Status Code: 400; Error Code: MissingParameter; Request ID: 628341eb-8a17-5841-a110-1374cae6b8a1)

If I simply change the `sqsEndpoint` to a non-fifo SQS queue, everything works as expected. I don't see how to even pass the message group ID, so it looks like it is unsupported at the moment, is this correct?

RustedBones:
# Pull Request Checklist

## Fixes

Fixes #625

## Purpose

Improve SQS batch operations

## Background Context

After the AWS SDK update, changes in the AWS models can make the API simpler and cleaner:
- No need to propagate FifoIdentifiers anymore
- Separation between request metadata (HTTP request info) and content metadata

### DIFF
--- a/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsModel.scala
+++ b/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsModel.scala
@@ -148,7 +148,7 @@ sealed abstract class SqsResult {
 }
 
 object SqsResult {
-  private[sqs] val EmptyResponseMetadata: SqsResponseMetadata = SqsResponseMetadata.create(
+  private[sqs] val EmptySqsResponseMetadata: SqsResponseMetadata = SqsResponseMetadata.create(
     DefaultAwsResponseMetadata.create(java.util.Collections.emptyMap())
   )
 }
@@ -258,7 +258,7 @@ object SqsAckResult {
 
     override type Result = NotUsed.type
 
-    override def responseMetadata: SqsResponseMetadata = SqsResult.EmptyResponseMetadata
+    override def responseMetadata: SqsResponseMetadata = SqsResult.EmptySqsResponseMetadata
 
     override def result: Result = NotUsed
 
@@ -356,7 +356,7 @@ object SqsAckResultEntry {
 
     override type Result = NotUsed.type
 
-    override def responseMetadata: SqsResponseMetadata = SqsResult.EmptyResponseMetadata
+    override def responseMetadata: SqsResponseMetadata = SqsResult.EmptySqsResponseMetadata
 
     override def result: Result = NotUsed
 
--- a/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsModel.scala
+++ b/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsModel.scala
@@ -126,11 +126,13 @@ object MessageAction {
 }
 
 /**
- * Messages returned by a SqsFlow.
+ * Messages returned by a SqsPublishFlow.
  *
- * @param message the SQS message.
+ * @param responseMetadata the SQS response metadata (AWS request ID, ...)
+ * @param metadata the SQS result metadata.
+ *                 This is either the SQS response itself or the SQS result entry in case of batching
  */
-final class SqsPublishResult[T <: SdkPojo] @InternalApi private[sqs] (
+final class SqsPublishResult[+T <: SdkPojo] @InternalApi private[sqs] (
     val responseMetadata: SqsResponseMetadata,
     val metadata: T
 ) {
@@ -154,7 +156,15 @@ final class SqsPublishResult[T <: SdkPojo] @InternalApi private[sqs] (
   override def hashCode(): Int = java.util.Objects.hash(responseMetadata, metadata)
 }
 
-final class SqsAckResult[T <: SdkPojo] @InternalApi private[sqs] (
+/**
+ * Messages returned by a SqsAckFlow.
+ *
+ * @param responseMetadata the SQS response metadata (AWS request ID, ...)
+ * @param metadata the SQS result metadata.
+ *                 This is either the SQS response itself or the SQS result entry in case of batching
+ * @param messageAction the action processed by the SqsAckFlow
+ */
+final class SqsAckResult[+T <: SdkPojo] @InternalApi private[sqs] (
     val responseMetadata: Option[SqsResponseMetadata],
     val metadata: Option[T],
     val messageAction: MessageAction
@@ -172,7 +182,7 @@ final class SqsAckResult[T <: SdkPojo] @InternalApi private[sqs] (
   def getResponseMetadata: java.util.Optional[SqsResponseMetadata] = responseMetadata.asJava
 
   /** Java API */
-  def getMetadata: java.util.Optional[T] = metadata.asJava
+  def getMetadata: java.util.Optional[_ <: T] = metadata.asJava
 
   /** Java API */
   def getMessageAction: MessageAction = messageAction
--- a/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsModel.scala
+++ b/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsModel.scala
@@ -148,7 +148,7 @@ sealed abstract class SqsResult {
 /**
  * Messages returned by a SqsPublishFlow
  */
-final class PublishResult @InternalApi private[sqs] (
+final class SqsPublishResult @InternalApi private[sqs] (
     val request: SendMessageRequest,
     response: SendMessageResponse
 ) extends SqsResult {
@@ -160,10 +160,10 @@ final class PublishResult @InternalApi private[sqs] (
   override def result: SendMessageResponse = response
 
   override def toString =
-    s"""PublishResult(request=$request,result=$result)"""
+    s"""SqsPublishResult(request=$request,result=$result)"""
 
   override def equals(other: Any): Boolean = other match {
-    case that: PublishResult =>
+    case that: SqsPublishResult =>
       java.util.Objects.equals(this.request, that.request) &&
       java.util.Objects.equals(this.result, that.result)
     case _ => false
@@ -175,7 +175,7 @@ final class PublishResult @InternalApi private[sqs] (
 /**
  * Messages returned by a SqsPublishFlow.grouped or batched
  */
-final class PublishResultEntry @InternalApi private[sqs] (
+final class SqsPublishResultEntry @InternalApi private[sqs] (
     val request: SendMessageRequest,
     override val result: SendMessageBatchResultEntry,
     override val responseMetadata: SqsResponseMetadata
@@ -184,10 +184,10 @@ final class PublishResultEntry @InternalApi private[sqs] (
   override type Result = SendMessageBatchResultEntry
 
   override def toString =
-    s"""PublishResultEntry(request=$request,result=$result)"""
+    s"""SqsPublishResultEntry(request=$request,result=$result)"""
 
   override def equals(other: Any): Boolean = other match {
-    case that: PublishResultEntry =>
+    case that: SqsPublishResultEntry =>
       java.util.Objects.equals(this.request, that.request) &&
       java.util.Objects.equals(this.result, that.result)
     case _ => false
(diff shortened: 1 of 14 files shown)


ennru:
Thank you for modernising this with the latest possibilities from the AWS SDK.
As we now want to promise binary compatibility between releases, this PR failed the MiMa check. Do you see a way to keep the RC1 API around while improving this?

(run `mimaReportBinaryIssues` to get the MiMa report locally)

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsModel.scala
```diff
@@ -125,70 +126,50 @@ object MessageAction {
 }
 
 /**
- * Additional identifiers for FIFO messsages
+ * Messages returned by a SqsFlow.
  *
- * @see [https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-additional-fifo-queue-identifiers.html]
+ * @param message the SQS message.
```
ennru: Please fix left-over to describe what we have now.

sqs/src/main/scala/akka/stream/alpakka/sqs/scaladsl/SqsPublishFlow.scala
```diff
@@ -107,40 +95,24 @@ object SqsPublishFlow {
         sqsClient
           .sendMessageBatch(req)
           .toScala
-          .map(req -> _)(sameThreadExecutionContext)
       }
-      .mapAsync(settings.concurrentRequests) {
-        case (request, response) if !response.failed().isEmpty =>
+      .map {
```
ennru: Great to remove the unused Future-wrapping.

sqs/src/main/scala/akka/stream/alpakka/sqs/scaladsl/SqsAckFlow.scala
```diff
@@ -30,7 +31,7 @@ object SqsAckFlow {
    */
   def apply(queueUrl: String, settings: SqsAckSettings = SqsAckSettings.Defaults)(
       implicit sqsClient: SqsAsyncClient
-  ): Flow[MessageAction, SqsAckResult[SqsResponse], NotUsed] =
+  ): Flow[MessageAction, SqsAckResult[SdkPojo], NotUsed] =
```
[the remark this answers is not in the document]
RustedBones: This improves the consistency for Batch actions.
In this case we want to give the metadata entry associated to the single `MessageAction` and not the response which was sent back from the whole batch.

As mentioned in #625, in the batch `SqsResponse`, we don't have any way to identify the specific metadata entry for the `MessageAction` because we don't know the batch id. (we may be able to do it by comparing md5 hashes of the messages but that's not nice)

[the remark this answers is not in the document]
ennru: But here we don't have batches. We can use `SdkPojo` for the batch actions.

ennru:
This PR requires a rebase to incorporate changes from #1606.

RustedBones:
Thanks @ennru. 
By going through the PR once more, I realize I may got rid of the `FifoMessageIdentifiers` a bit too early. The goal of this class was to propagate the `GroupId` and `DeduplicationId` to downstream in case the user wants to do some special operations depending on those values (like for the message, we propagate it downstream though SQS only gives the md5 as a response).

In order to pass all information to downstream, we can encapsulate the `SendMessageRequest` into the `SqsPublishResult`. Then we can safely remove the `FifoMessageIdentifiers` and this makes it more symmetric like the `SqsAckFlow` which also contains the `MessageAction`.

ennru:
Let's give this the time it needs and release Alpakka 1.0 anyway. I've created #1621 to mark the API as "may change".

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsModel.scala
```diff
@@ -125,82 +126,78 @@ object MessageAction {
...
+    s"""SqsPublishResult(responseMetadata=$responseMetadata,metadata=$metadata,request=$request)"""
 
   override def equals(other: Any): Boolean = other match {
-    case that: FifoMessageIdentifiers =>
-      java.util.Objects.equals(this.sequenceNumber, that.sequenceNumber) &&
-      java.util.Objects.equals(this.messageGroupId, that.messageGroupId) &&
-      java.util.Objects.equals(this.messageDeduplicationId, that.messageDeduplicationId)
+    case that: SqsPublishResult[T] =>
+      java.util.Objects.equals(this.responseMetadata, that.responseMetadata) &&
+      java.util.Objects.equals(this.metadata, that.metadata)
```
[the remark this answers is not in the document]
RustedBones: Good catch. Should have triggered an error in the test but I made an error there.

I'm still wondering why in `alpakka` we didn't go the `akka-http` way for the model: Java abstract classes/interface with the java api in a `javadsl` package and scala case classes in `scaladsl` extending the java api. We'll then not have to case of the `toString` and `equal`. Is it because of `copy` ?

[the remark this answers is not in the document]
ennru: Yes. Copy's generated default parameters make it hard to evolve case classes in a binary-compatible way.

ennru:
Picking this up again.

To reduce the risk for "`ToManyTypeParametersException`" I'd like to propose to introduce specific subclasses of `SqsPublishResult`:

```scala
final class SqsPublishSingleResult @InternalApi private[sqs] (
    responseMetadata: SqsResponseMetadata,
    metadata: SendMessageResponse,
    request: SendMessageRequest
) extends SqsPublishResult[SendMessageResponse](responseMetadata, metadata, request)

final class SqsPublishBatchResult @InternalApi private[sqs] (
    responseMetadata: SqsResponseMetadata,
    metadata: SendMessageBatchResultEntry,
    request: SendMessageRequest
) extends SqsPublishResult[SendMessageBatchResultEntry](responseMetadata, metadata, request)
```

Names can be improved probably, but the API looks simpler with this.

WDYT?

RustedBones:
Are we risking a `ToManyTypeParametersException` since the constructor visibility is package?

If we do this we should also do the same with `SqsAckResult` with `ChangeMessageVisibilityResponse`, `ChangeMessageVisibilityBatchResultEntry`,  `DeleteMessageResponse`, `DeleteMessageBatchResultEntry` and `SqsAckResult[Nothing]` (returned in a case of `MessageAction.Ignore`). That make all the type definitions a bit tedious IMHO.

ennru:
Yes, you're right.

RustedBones:
Thinking about it again, it's probably better to give a sealed abstract class along with the possible implementations for the SQS model

Right now, the output type from the `SqsAckFlow.grouped` is a `SqsAckResult[SdkPojo]` which is contains too many potential implementations due to the `SdkPojo` (not only used for SQS)  and would oblige the users to always provide a default case when pattern matching to be exhaustive. Same problem applies with `SqsResponse` in a smaller scale.

In addition, it would also be nice to provide unapply on the SQS model  for the scala users

ennru:
Yes, that is what I am looking for. `SdkPojo` is too broad and we have just two specialisations of it.

RustedBones:
I've tried to get rid of the type parameter from the publish and ack results to give a clean API:
```scala
SqsPublishFlow.apply: Flow[SendMessageRequest, PublishResult, NotUsed]
SqsPublishFlow.grouped: Flow[SendMessageRequest, PublishResultEntry, NotUsed]

SqsAckFlow.apply: Flow[MessageAction, SqsAckResult, NotUsed]
SqsAckFlow.grouped: Flow[MessageAction, SqsAckResultEntry, NotUsed]
``` 

`Ignore` message action is now filtered by the ack flows and don't forward any message downstream

In all the `SqsResult` there is an accessor to the `SqsResponseMetadata` but this field is not taken in account for the `equals` and `toString` as AWS does it on its own models.

sqs/src/test/java/docs/javadsl/SqsAckTest.java
```diff
@@ -92,27 +96,31 @@ public void testAcknowledge() throws Exception {
...
     SqsAsyncClient awsClient = mock(SqsAsyncClient.class);
     when(awsClient.deleteMessage(any(DeleteMessageRequest.class)))
-        .thenReturn(CompletableFuture.completedFuture(DeleteMessageResponse.builder().build()));
+        .thenReturn(CompletableFuture.completedFuture(response));
 
-    List<Message> messages = createMessages();
     Source<Message, NotUsed> source = Source.fromIterator(messages::iterator);
-
-    CompletionStage<List<SqsAckResult<SqsResponse>>> stage =
+    CompletionStage<List<SqsAckResult>> stage =
```
ennru: Nice.

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsModel.scala
```diff
@@ -125,89 +124,213 @@ object MessageAction {
...
 
   /** Java API */
-  def getMessageDeduplicationId: java.util.Optional[String] = messageDeduplicationId.asJava
+  def getResult: Result = result
+}
+
+/**
+ * Messages returned by a SqsPublishFlow
+ */
+final class PublishResult @InternalApi private[sqs] (
```
ennru: No `Sqs` prefix? Was that intentional?

RustedBones: The `MessageActions` are not prefixed with `Sqs`.

But you're right, I'm inconsistent with the naming: the super class are prefixed but not the implementations. I should either remove or prepend it everywhere.

RustedBones:
I've added `Ignore` to the `SqsAck`.
Instead of using `Option` for the `result` and `requestMetadata`, I return `NotUsed` and an empty metadata object.

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsModel.scala
```diff
@@ -35,14 +36,14 @@ object MessageAction {
   }
 
   final object Delete {
-    def apply(message: Message): MessageAction = new Delete(message)
+    def apply(message: Message): Delete = new Delete(message)
```
ennru: I'm looking into the MiMa differences and this change shows up. It should stay as `MessageAction` which makes the generics a bit easier for Java users.

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsModel.scala
```diff
@@ -53,7 +54,7 @@ object MessageAction {
   }
 
   final object Ignore {
-    def apply(message: Message): MessageAction = new Ignore(message)
+    def apply(message: Message): Ignore = new Ignore(message)
```
ennru: Same as for `Delete`, keep `MessageAction`.

sqs/src/main/scala/akka/stream/alpakka/sqs/testkit/MessageFactory.scala
```diff
@@ -3,53 +3,46 @@
...
+                              responseMetadata: SqsResponseMetadata = EmptySqsResponseMetadata): SqsDeleteResultEntry =
+    new SqsDeleteResultEntry(messageAction, result, responseMetadata)
+
+  def createChangeMessageVisibilityResultEntry(
+      messageAction: MessageAction.ChangeMessageVisibility,
+      result: ChangeMessageVisibilityBatchResultEntry,
+      responseMetadata: SqsResponseMetadata = EmptySqsResponseMetadata
+  ): SqsChangeMessageVisibilityResultEntry =
+    new SqsChangeMessageVisibilityResultEntry(messageAction, result, responseMetadata)
 }
```
ennru: Please add `SqsIgnoreResult` and `SqsIgnoreResultEntry` factories here.

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsModel.scala
RustedBones: The Java users should use `MessageAction.delete(message: Message): MessageAction` instead `apply`.

ennru: Of course. Thank you.

PR #1614, opened 2019-03-28
mqtt-streaming: Server shouldn't stash PUBLISH for uninterested clients

longshorej:
Server shouldn't stash PUBLISH for uninterested clients

This fixes a memory leak on the server, where when a client disconnects, all PUBLISH messages -- even those that the client hasn't subscribed to -- are stashed.

Instead, if it's a PUBLISH message, we now drop it.

/cc @huntc


mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSession.scala
```diff
@@ -494,12 +494,18 @@ final class ActorMqttServerSession(settings: MqttSessionSettings)(implicit mat:
...
+                      .foreach(_.complete(result.map(_ => Done)))
+                  }
+
                   Source
                     .fromFuture(reply.future.map(_ => cp.encode(ByteString.newBuilder).result()))
                     .recover {
                       case _: RemotePacketRouter.CannotRoute => ByteString.empty
                     }
-                    .wireTap(_ => completed.foreach(_.success(Done)))
+
```
huntc: Nice. Captures the failure also.

PR #1620, opened 2019-04-02
Test for git changes via shell script

ennru:
## Purpose

Make the decision if a project should be built in the shell to cut built time.

## Background Context

With the increasing number of projects, the Alpakka build is running too long time.


.travis.yml
```diff
@@ -40,78 +32,139 @@ jobs:
...
+      - CMD=+text/testChanged
+    - env:
+      - DIR=udp
+      - CMD=+udp/testChanged
+    - env:
+      - DIR=unix-domain-socket
+      - CMD=+unix-domain-socket/testChanged
+    - env:
+      - DIR=xml
+      - CMD=+xml/testChanged
```
2m: Since the `CMD` always starts with the `DIR`, maybe we could reuse that in the `buildChanged.sh` and do not repeat it as much in the yaml?

ennru: OK, let's see how Travis resolves this default value.
Even better would be to set the build step's name and use that (but I didn't find any docs for that).

PR #1629, opened 2019-04-03
mqtt-streaming: Fix a race condition in server publishing

longshorej:
The MQTT streaming API allows the server to supply a `Promise` when it offers a `SUBACK`. The contract here is that when the `Promise` is completed, it is safe for the server to begin sending `PUBLISH`es as all of the internal machinery is ready.

However, due to a race condition, `PUBLISH` commands sent in response to this `Promise` being completed could arrive before the server's `ClientConnection` has messaged itself the `Subscribed` instance. If this happens, the `PUBLISH` is ignored.

This introduces another promise that the `ClientConnection` is responsible for completing when it receives the `Subscribe` message in response. It then completes this `Promise`, which in turn completes the user's supplied `Promise`, thus eliminating the race and ensuring that
`PUBLISH`es are not ignored.

I've marked this as a draft for now until I've done further testing and figured out if it's feasible to unit test this.

### DIFF
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ServerState.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ServerState.scala
@@ -335,7 +335,7 @@ import scala.util.{Failure, Success}
       extends Event
   final case class SubscribeReceivedFromRemote(subscribe: Subscribe, local: Promise[Publisher.ForwardSubscribe.type])
       extends Event
-  final case class Subscribed(subscribe: Subscribe) extends Event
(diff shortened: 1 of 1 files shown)


huntc:
Thanks for identifying this bug.

I've got an alternate fix to this which has a little less change. In essence, the responsibility for promising the forwarding of the `SUBACK` is moved to the `ClientConnection` given that `ClientConnection`'s state is required to be updated as a pre-condition. WDYT?

longshorej:
Yeah, that looks better. I realize now though there's a leak we have to deal with.

* The `completed` promise optionally supplied with the `SUBACK` will always be completed, so no need to worry about that.
* The `subscribed` promise setup with `SubscribeReceivedFromRemote` must be failed if the SUBACK timeout occurs, otherwise each failed subscribe will leak memory.

longshorej:
I've updated this PR to have @huntc's commit (amended commit message), and another commit to fix the memory leak.

PR #1631, opened 2019-04-04
Mqtt streaming: Discontinue Scala 2.11

ennru:
## Purpose

Remove Scala 2.11 from cross Scala versions for MQTT streaming.

## Background Context

As MQTT streaming is built on top of Akka Typed which is not published for Scala 2.11 anymore, Alpakka won't publish Scala 2.11 artifacts for this anymore.


huntc:
Excellent.

PR #1636, opened 2019-04-06
Adds a Stdin source for easy string input to an Akka stream from stdin

Ewan-Keith:
## Purpose

Adds a new Source object `StdinSource` that when pulled, reads the next string entered at stdin and pushes it downstream.

## Background Context

It's trivial using existing Akka stream components to print the contents of a stream out to standard output (e.g. `Flow[A].map {x => println(x); x}`). However I haven't found any similar means of using existing functionality to build a continuous stream of input from standard input that is then pushed downstream.

This capability is not likely one that is required in production ready streams, but it could prove useful for interactive testing/exploration of streams under construction. It could also provide a means to construct simple command line tools using Akka Streams, although I'm unclear on whether this would be useful or even recommended.

The docs for this new component are not written yet, if this seems a useful addition to the Alpaka tool set I will right up basic documentation.


2m:
Hi @Ewan-Keith, thanks for the PR!

Akka Streams provide tools to convert to/from Java Input/Output streams. Have you seen [`StreamConverters.{as,from}{Input,Output}Stream`](https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/scaladsl/StreamConverters.scala#L48)?

Those factories compose with `System.in` and `System.out` (take a look [here](https://github.com/2m/sssio/blob/master/src/main/scala/Sssio.scala#L87) for such use case in the wild).

If `StreamConverters.fromOutputStream(() => System.out)` does achieve everything that you want, then having that as a separate connector might not be necessary. But this use-case might still need an example in the Alpakka docs, so such combination is easy to discover.

Ewan-Keith:
I hadn't seen the `StreamConverters` methods, those seem to provide exactly what I was looking for, so definitely no need for this additional connector.

There's an issue on the main Akka repo still open for adding use cases for these, https://github.com/akka/akka/issues/25468, I'll want to read up on the stream converters anyway so can have a look at writing something up as part of that process.

Thanks for the help, in my view there's no need for this addition given the above so closing the PR.

2m:
Great. Adding examples for those operators would be perfect!

Issue #920, opened 2018-05-02
AWS S3 Connector: Support for proxy authentication


jongray:
Could the S3 connector proxy support be extended to support HTTP(S) proxy authentication (or could there be a facility to plug an appropriately configured ClientTransport into the S3Settings)?

I tried a host of the form `username:password@host` in the Proxy() but with no success.

ennru:
Proxy support from #1639 was released in Alpakka 1.0.1.

Issue #1017, opened 2018-06-12
AWS S3: Implement "proper" proxy support


Scalahansolo:
Way back when, with the help of the community I implemented the initial (less than ideal) version of proxy support in the S3 module in alpakka. Now that akka-http supports proxy support, as well as authed proxy support I think that we should look into upgrading alpakka to use the authed proxy support. 

I could take a crack at an initial re-work, but based on how many of the tests ended up getting implemented, its going to be a fairly significant re-write of tests as well as the actual S3Stream/HttpRequests codebase.

Wanted to reach out and see if anyone was interested in working with me with this work?

ennru:
We'll upgrade to use Akka Http 10.1.2 with #970 

cc @sfali

Scalahansolo:
I was just now looking back on this ticket to go back through and re work this. I would like to take some time to work on this as I really think this is a better way forward. However, like my comment back in June, this change is going to non-backwards compatible and involve a fairly significant re work in the tests. Before I embark down that path, is this something that would be wanted? @ennru?

2m:
We are currently working towards Alpakka 1.0 after which we will keep API compatibility between major versions. So now is a good time to work on API braking changes.

Scalahansolo:
Is there any timeline for 1.0? So I have a rough idea how much time I have to get these changes done?

floekke:
Any news about this issue?

ennru:
Fixed with #1639

PR #1639, opened 2019-04-09
AWS  S3 - forward proxy support

floekke:
## Purpose

Akka HTTP has support for forward proxies via its [Pluggable Client Transports](https://doc.akka.io/docs/akka-http/current/client-side/client-transport.html). The functionality is used in this update in order to support forward proxies when integrating with S3.

## References

References #1017

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
@@ -747,7 +747,7 @@ import akka.util.ByteString
       implicit val settings = resolveSettings()
 
       requests
-        .via(superPool[MultipartCopy](settings, sys))
+        .via(superPool[MultipartCopy])
         .map {
           case (Success(r), multipartCopy) =>
             val entity = r.entity
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
@@ -5,6 +5,7 @@
 package akka.stream.alpakka.s3.impl
 
 import java.time.{Instant, LocalDate, ZoneOffset}
+import java.net.InetSocketAddress
 
 import akka.actor.ActorSystem
 import akka.annotation.InternalApi
@@ -14,10 +15,17 @@ import scala.collection.immutable.Seq
 import scala.concurrent.Future
 import scala.util.{Failure, Success}
 import akka.{Done, NotUsed}
-import akka.http.scaladsl.Http
+import akka.http.scaladsl.{ClientTransport, Http}
 import akka.http.scaladsl.model.StatusCodes.{NoContent, NotFound, OK}
-import akka.http.scaladsl.model.headers.{`Content-Length`, `Content-Type`, ByteRange, CustomHeader}
+import akka.http.scaladsl.model.headers.{
+  `Content-Length`,
+  `Content-Type`,
+  BasicHttpCredentials,
+  ByteRange,
+  CustomHeader
+}
 import akka.http.scaladsl.model.{headers => http, _}
+import akka.http.scaladsl.settings.{ClientConnectionSettings, ConnectionPoolSettings}
 import akka.http.scaladsl.unmarshalling.{Unmarshal, Unmarshaller}
 import akka.stream.alpakka.s3.BucketAccess.{AccessDenied, AccessGranted, NotExists}
 import akka.stream.{ActorMaterializer, Attributes, Materializer}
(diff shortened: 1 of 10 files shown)


ennru:
Sorry for being so quiet. We've been traveling and working deep down in the Alpakka Kafka connector for a while now.
We'll soon give you some feedback...

floekke:
> Sorry for being so quiet. We've been traveling and working deep down in the Alpakka Kafka connector for a while now.
> We'll soon give you some feedback...

No problem :)

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -618,12 +647,13 @@ import akka.util.ByteString
...
   )(parallelism: Int) =
-    Setup.source { implicit mat => _ =>
+    Setup.source { implicit mat => implicit attr =>
       import mat.executionContext
       implicit val sys = mat.system
+      implicit val settings = resolveSettings()
 
       requests
-        .via(Http().superPool[MultipartCopy]())
+        .via(superPool[MultipartCopy](settings, sys))
```
2m: `settings` and `sys` are implicits, so no need to pass those explicitly here.

s3/src/main/scala/akka/stream/alpakka/s3/settings.scala
```diff
@@ -90,7 +183,8 @@ final class S3Settings private (
     val s3RegionProvider: AwsRegionProvider,
     val pathStyleAccess: Boolean,
     val endpointUrl: Option[String],
-    val listBucketApiVersion: ApiVersion
+    val listBucketApiVersion: ApiVersion,
+    val forwardProxy: Option[ForwardProxy]
```
2m: Do you think we could reuse `proxy` which is already in the settings?

floekke: As I understand it and from experience using the library, the existing proxy, is meant for the case where you don't want to integrate with AWS S3 but a service that implements the S3 API.  You could still have a need of using a forward proxy together with this service. E.g. company policy might dictate the need for using a forward proxy, while connecting to some third-party S3 compatible service. Leaving the old proxy settings alone means backwards compatibility, but the name is probably not ideal, since 'proxy' without any qualifications is usually a forward proxy.. Though introducing a new 'forwardProxy' should help in disambiguating :)

2m: I see. Yea, the current `proxy` setting controls the request `Uri`, very similar to the `endpoint-url` setting as well. There is some overlap between these two and it could be cleaned, but that's for another time.

`ForwardProxy` sounds good to me then.

2m:
Also fixes #920

s3/src/main/scala/akka/stream/alpakka/s3/settings.scala
```diff
@@ -276,7 +380,8 @@ object S3Settings {
       s3RegionProvider = regionProvider,
       pathStyleAccess = pathStyleAccess,
       endpointUrl = endpointUrl,
-      listBucketApiVersion = apiVersion
+      listBucketApiVersion = apiVersion,
+      forwardProxy = None // TODO: load from config, see akka.http.client.proxy from https://doc.akka.io/docs/akka-http/current/client-side/client-transport.html
```
2m: The config path should be `alpakka.s3.forward-proxy` then.

ennru:
Please keep backwards-compatibility in mind when renaming configuration.

2m:
I'll add config support and doc improvements to this PR.

floekke:
> I'll add config support and doc improvements to this PR.

Sounds good! I've been very busy, so this has been low priority. I'll be able to continue soon though.

2m:
No problem! I also want to do a little bit of clean-up, by deprecating the current `proxy` setting, because that and the `endpoint-url` seem to do very similar things.

2m:
Pushed deprecation and loading `forward-proxy` settings from config.

s3/src/main/scala/akka/stream/alpakka/s3/settings.scala
```diff
@@ -68,6 +68,99 @@ object Proxy {
...
+  def withUsername(username: String) = copy(username = username)
+  def withPassword(password: String) = copy(password = password)
+
+  private def copy(username: String = username, password: String = password) =
+    new ForwardProxyCredentials(username, password)
+
+  override def toString =
+    "ForwardProxyCredentials(" +
+    s"username=$username," +
+    s"password=$password" +
```
ennru: Hide the password in `toString`.

s3/src/main/scala/akka/stream/alpakka/s3/settings.scala
```diff
@@ -68,6 +68,99 @@ object Proxy {
...
+final class ForwardProxy private (val host: String, val port: Int, val credentials: Option[ForwardProxyCredentials]) {
+
+  /** Java API */
+  def getHost: String = host
+
+  /** Java API */
+  def getPort: Int = port
+
+  /** Java API */
+  def getCredentials: Option[ForwardProxyCredentials] = credentials
```
ennru: Choose `j.u.Optional` for the Java API.

s3/src/main/scala/akka/stream/alpakka/s3/settings.scala
```diff
@@ -85,18 +178,20 @@ object ApiVersion {
 
 final class S3Settings private (
     val bufferType: BufferType,
-    val proxy: Option[Proxy],
+    @deprecated("Please use endpointUrl instead", since = "1.0.1") val proxy: Option[Proxy],
```
ennru: This could be defined as a fixed value non-constructor field now, right?

s3/src/main/scala/akka/stream/alpakka/s3/settings.scala
```diff
@@ -293,15 +409,17 @@ object S3Settings {
       s3RegionProvider: AwsRegionProvider,
       pathStyleAccess: Boolean,
       endpointUrl: Option[String],
-      listBucketApiVersion: ApiVersion
+      listBucketApiVersion: ApiVersion,
+      forwardProxy: Option[ForwardProxy]
```
ennru: New fields shouldn't be added here.

2m: Ah, yes of course. Deprecated the old ones and added a new one without `proxy` and with `forwardProxy`.

ennru: Don't offer a new one. The use `withForwardProxy` should be enough.

2m: The new one is still needed, as some attributes do not have default values. So I have changed the new one to only take those that do not have default values.

PR #1642, opened 2019-04-10
MQTT streaming: Fixes wrong Exception thrown when client ping timeouts

jcroig:
## Purpose

When mqtt client doesn't send ping request within the expected period (given by the connect keep alive parameter) stream should failed with `ServerConnector.PingFailed` exception but it fails with `ClientConnection.PingFailed` exception. That makes that ActorMqttServerSession command flow doesn't translate the failure to the public api error ActorMqttServerSession.PingFailed (MqttSession.scala::482)

## References

References #xxxx

## Background Context

There was an ignored failing unit test due to this bug that however was pointing to a different issue. After this PR this unit test is not failing anymore so I activate it again.


mqtt-streaming/src/test/scala/docs/scaladsl/MqttSessionSpec.scala
```diff
@@ -1514,7 +1515,9 @@ class MqttSessionSpec
       server.watchCompletion().foreach(_ => session.shutdown())
     }
 
-    "close when no ping request received" ignore { // assertAllStagesStopped { // https://github.com/akka/alpakka/issues/1563
+    "close when no ping request received" in assertAllStagesStopped {
+      implicit val patienceConfig = PatienceConfig(scaled(Span(3000, Millis)), scaled(Span(15, Millis)))
```
huntc: Can you please explain why we need this patience config? Is there a way to avoid it? I’m not overly concerned, but we’ve not had to use such config so far.

jcroig: The reason is that the test is using 1 second keep alive parameter so using default PatienceConfig of 150 milliseconds as max value is not enough for the ping timeout to trigger. Therefore I'm adding 3 seconds max time before I see the stream failing because of the ping timeout or otherwise failing.

huntc: How about reducing the ping timeout instead?

huntc:
Oh I just realised that the MQTT turnout is the granularity of seconds... so, don’t worry. A comment on the patience config as per your previous explanation would be great. Also, you may want to use .dilated” on the patience config time.

jcroig:
I'll add a little comment on that. 

Regarding patience config, I've used the same pattern that the default one which is used everywhere else, meaning the implicit one defined in `PatienceConfiguration` trait that is:

```implicit def patienceConfig: PatienceConfig = defaultPatienceConfig```

where `private val defaultPatienceConfig = PatienceConfig()`

Default values for PatienceConfig are:

```
final case class PatienceConfig(timeout: Span = scaled(Span(150, Millis)), interval: Span = scaled(Span(15, Millis)))
```

Instead of using `.dilated` to define the timeout and interval it uses `scaled` over a span. Note that this is also scaling the span but based on a different parameter than `.dilated`. Because this parameter will control the configuration of all PatienceConfig used everywhere I though it will be good to make this one behave as any other else used in any test.

What do you think?

huntc:
@jcroig I've just had a look for other uses of patience config within Alpakka. This appears to fall in line.

I don't think we need to do any more, and thanks for putting the comment in.

ennru:
Thank you for fixing the small details to make it even better!

PR #1644, opened 2019-04-12
Read from Kafka write to Elasticsearch example

ennru:
## Purpose

Add a new self-contained example which reads messages from a Kafka topic and writes these to Elasticsearch.
Make use of the new `withContext` API in Akka Streams.
Use [Testcontainers](https://www.testcontainers.org/) to start Kafka and Elasticsearch.


ennru:
This could be further improved when Alpakka Kafka has pre-baked `withContext` support.

See https://github.com/akka/alpakka-kafka/pull/780

ennru:
Failure was #1417

Issue #1657, opened 2019-04-18
Ftp - ls does not emit directories in default case


tgambet:
### Version used

Akka version: 2.5.21
Alpakka FTP version: 1.0.0

### Current behavior

Listing a directory with `Ftp.ls(path, ftpSettings)` doesn't emit any `FtpFile` for directories (`isDirectory` is always false) while  `Ftp.ls(path, ftpSettings, _ => false)` does.

### Expected

`Ftp.ls(path, ftpSettings)` should emit `FtpFile`s for all directories that it finds while traversing the directory tree.

2m:
Thanks for the ticket. The default value for `branchSelector` is `f => true`, which means it will go down every directory but it does not emit directories themselves in the stream.

It does make sense to emit directories together with files. But when implementing this, we should also give the ability to get the current behaviour.

tgambet:
I believe that the stream should contain both files and directories no matter what `branchSelector` is. It is unexpected that the branch selector changes the behavior of the command regarding what type of files are streamed. 
One can always get the current behavior by filtering the source : `Ftp.ls(...).filter(!_.isDirectory)`

2m:
That is a fair point. Do you think you could work on a PR to fix this?

tgambet:
Yes, I can try.

tgambet:
PR ready for review

PR #1658, opened 2019-04-19
FTP: emit traversed directories from Ftp.ls

tgambet:
## Purpose

This PR is about making `Ftp.ls` emit directories it encounters no matter if they were traversed or not.
The previous behavior being that you only get directories that were not entered, which is unexpected.

## References

References #1657

## Changes

`FtpGraphStageLogic` keeps a record of traversed directories and emit them first on pull.
For instance the path `/foo/bar/sample` will result in the emission of three `FtpFile`, one for `foo`, one for `bar`, and one for `sample`, in that order.

Additionally, the test-related class `FtpBaseSupport` had to be fixed to remove empty directories when cleaning files, and tests have been reviewed to match the new behavior of `listFiles`.

### DIFF
--- a/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/CommonFtpOperations.scala
+++ b/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/CommonFtpOperations.scala
@@ -23,7 +23,7 @@ private[ftp] trait CommonFtpOperations {
   type Handler = FTPClient
 
   def listFiles(basePath: String, handler: Handler): immutable.Seq[FtpFile] = {
-    val path = if (!basePath.isEmpty && basePath.head != '/') s"/$basePath" else basePath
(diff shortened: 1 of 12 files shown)


ennru:
This is quite a change in behaviour which we can't have in patch releases. The new behaviour would need to be opt-in.

tgambet:
Ok. It's too bad because I still think that the current behavior is not intuitive : if your banch selector resolves to true for a certain directory then you don't get it in the stream.

Then how about a `emitTraversedDirectories` boolean on `ls` to trigger the behavior I suggest, and that would be false by default ? Would that be acceptable ?

ennru:
Yes, that would be great.

ennru:
Thank you for this useful addition.
A snapshot will become available. See https://doc.akka.io/docs/alpakka/current/other-docs/snapshots.html

tgambet:
You're welcome. Thanks for the pointer.

PR #1664, opened 2019-04-25
:arrow_up: Geode 1.9.0

cheleb:
## Purpose
Geode dependencies upgrade.

## References

References #xxxx

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -84,7 +84,7 @@ services:
       ADDED_FLAGS: "--tls=1"
   geode:
     container_name: geode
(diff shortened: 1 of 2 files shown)


ennru:
With Alpakka being 1.0 we need to be more restrictive on dependency upgrades. We need to be sure the upgrades in turn are binary compatible to be applied within patch releases.
In Geode's case it brings in Jackson, JAXB, and a lot of other dependencies common to many projects. We can't risk an upgrade pulling in a new version of a dependency.

We'll defer this to Alpakka 1.1

ennru:
Thank you.

Issue #1347, opened 2018-11-26
Kinesis: Use AWS SDK 2.0


ennru:
See #372

francescoperera:
I will try to take a stab at it

ennru:
Any takers for this work. Kinesis is the last Alpakka connector on the old AWS SDK...

regiskuckaertz:
Hi! I started working on this, as we need it in house. I'll go through your guidelines before raising a PR 😃

PR #1667, opened 2019-04-26
AWS Kinesis KCL streams support

aserrallerios:
Adds the KCL Source and record checkpointer Flow/Sink.

Updated version of the old PR: https://github.com/akka/alpakka/pull/434

### DIFF
--- a/docs/src/main/paradox/kinesis.md
+++ b/docs/src/main/paradox/kinesis.md
@@ -8,22 +8,12 @@ For more information about Kinesis please visit the [Kinesis documentation](http
 
 Another Kinesis connector which is based on the Kinesis Client Library is available.
 
-The KCL Source can read from several shards and rebalance automatically when other Workers are started or stopped. It also handles record sequence checkpoints.
-
-Please read more about it at [GitHub aserrallerios/kcl-akka-stream](https://github.com/aserrallerios/kcl-akka-stream).
-@@@
-
-
-@@@ note { title="Alternative connector 2" }
-
-Another Kinesis connector which is based on the Kinesis Client Library is available.
-
 This library combines the convenience of Akka Streams with KCL checkpoint management, failover, load-balancing, and re-sharding capabilities.
 
 Please read more about it at [GitHub StreetContxt/kcl-akka-stream](https://github.com/StreetContxt/kcl-akka-stream).
 @@@
 
-@@@ note { title="Alternative connector 3" }
+@@@ note { title="Alternative connector 2" }
 
 Another Kinesis connector which is based on the Kinesis Client Library 2.x is available.
 
@@ -54,7 +44,7 @@ The table below shows direct dependencies of this module and the second tab show
 Sources and Flows provided by this connector need a `KinesisAsyncClient` instance to consume messages from a shard.
 
 @@@ note
-The `KinesisAsyncClient` instance you supply is thread-safe and can be shared amongst multiple `GraphStages`. 
+The `KinesisAsyncClient` instance you supply is thread-safe and can be shared amongst multiple `GraphStages`.
 As a result, individual `GraphStages` will not automatically shutdown the supplied client when they complete.
 It is recommended to shut the client instance down on Actor system termination.
 @@@
@@ -104,10 +94,10 @@ objects by calling [GetRecords](http://docs.aws.amazon.com/kinesis/latest/APIRef
 
 ### Kinesis Put via Flow or as Sink
 
-The 
-@scala[@scaladoc[KinesisFlow](akka.stream.alpakka.kinesis.scaladsl.KinesisFlow$) (or @scaladoc[KinesisSink](akka.stream.alpakka.kinesis.scaladsl.KinesisSink$))] 
-@java[@scaladoc[KinesisFlow](akka.stream.alpakka.kinesis.javadsl.KinesisFlow$) (or @scaladoc[KinesisSink](akka.stream.alpakka.kinesis.javadsl.KinesisSink$))] 
-publishes messages into a Kinesis stream using its partition key and message body. It uses dynamic size batches and can perform several requests in parallel. These features are necessary to achieve the best possible write throughput to the stream. The Flow outputs the result of publishing each record.
+The
+@scala[@scaladoc[KinesisFlow](akka.stream.alpakka.kinesis.scaladsl.KinesisFlow$) (or @scaladoc[KinesisSink](akka.stream.alpakka.kinesis.scaladsl.KinesisSink$))]
+@java[@scaladoc[KinesisFlow](akka.stream.alpakka.kinesis.javadsl.KinesisFlow$) (or @scaladoc[KinesisSink](akka.stream.alpakka.kinesis.javadsl.KinesisSink$))]
+publishes messages into a Kinesis stream using its partition key and message body. It uses dynamic size batches, can perform several requests in parallel and retries failed records. These features are necessary to achieve the best possible write throughput to the stream. The Flow outputs the result of publishing each record.
 
 @@@ warning
 Batching has a drawback: message order cannot be guaranteed, as some records within a single batch may fail to be published. That also means that the Flow output may not match the same input order.
@@ -146,6 +136,46 @@ Scala
 Java
 : @@snip [snip](/kinesis/src/test/java/docs/javadsl/KinesisSnippets.java) { #error-handling }
 @@@
+
+# AWS KCL Scheduler Source & checkpointer
+
+The KCL Source can read from several shards and rebalance automatically when other Schedulers are started or stopped. It also handles record sequence checkpoints.
+
+For more information about KCL please visit the [official documentation](http://docs.aws.amazon.com/streams/latest/dev/developing-consumers-with-kcl-v2.html).
+
+## Usage
+
+The KCL Scheduler Source needs to create and manage Scheduler instances in order to consume records from Kinesis Streams.
+
+In order to use it, you need to provide a Scheduler builder and the Source settings:
+
+Scala
+: @@snip (/kinesis/src/test/scala/docs/scaladsl/KclSnippets.scala) { #scheduler-settings }
+
+Java
+: @@snip (/kinesis/src/test/java/docs/javadsl/KclSnippets.java) { #scheduler-settings }
+
+Then the Source can be created as usual:
+
+Scala
+: @@snip (/kinesis/src/test/scala/docs/scaladsl/KclSnippets.scala) { #scheduler-source }
+
+Java
+: @@snip (/kinesis/src/test/java/docs/javadsl/KclSnippets.java) { #scheduler-source }
+
+## Committing records
+
+The KCL Scheduler Source publishes messages downstream that can be committed in order to mark progression of consumers by shard. This process can be done manually or using the provided checkpointer Flow/Sink.
+
+In order to use the Flow/Sink you must provide additional checkpoint settings:
+
+Scala
+: @@snip (/kinesis/src/test/scala/docs/scaladsl/KclSnippets.scala) { #checkpoint }
+
+Java
+: @@snip (/kinesis/src/test/java/docs/javadsl/KclSnippets.java) { #checkpoint }
+
+Note that checkpointer Flow may not maintain the input order of records of different shards.
 
 ## Kinesis Firehose Streams
 
@@ -201,9 +231,9 @@ Java
 : @@snip [snip](/kinesis/src/test/java/docs/javadsl/KinesisFirehoseSnippets.java) { #flow-sink }
 
 @@@ warning
-As of version 2, the library will not retry failed requests. See @ref[AWS Retry Configuration](aws-shared-configuration.md) how to configure it for the @javadoc[FirehoseAsyncClient](software.amazon.awssdk.services.firehose.FirehoseAsyncClient). 
+As of version 2, the library will not retry failed requests. See @ref[AWS Retry Configuration](aws-shared-configuration.md) how to configure it for the @javadoc[FirehoseAsyncClient](software.amazon.awssdk.services.firehose.FirehoseAsyncClient).
 
-This means that you may have to inspect individual responses to make sure they have been successful: 
+This means that you may have to inspect individual responses to make sure they have been successful:
 
 Scala
 : @@snip [snip](/kinesis/src/test/scala/docs/scaladsl/KinesisFirehoseSnippets.scala) { #error-handling }
--- a/kinesis/src/main/scala/akka/stream/alpakka/kinesis/CommittableRecord.scala
+++ b/kinesis/src/main/scala/akka/stream/alpakka/kinesis/CommittableRecord.scala
@@ -1,0 +1,98 @@
+/*
+ * Copyright (C) 2016-2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.kinesis
+
+import java.time.Instant
+
+import akka.annotation.InternalApi
+import akka.stream.alpakka.kinesis.CommittableRecord.{BatchData, ShardProcessorData}
+import software.amazon.kinesis.exceptions.ShutdownException
+import software.amazon.kinesis.lifecycle.ShutdownReason
+import software.amazon.kinesis.retrieval.KinesisClientRecord
+import software.amazon.kinesis.retrieval.kpl.ExtendedSequenceNumber
+
+abstract class CommittableRecord @InternalApi private[kinesis] (
+    val record: KinesisClientRecord,
+    val batchData: BatchData,
+    val processorData: ShardProcessorData
+) {
+
+  val sequenceNumber: String = record.sequenceNumber()
+  val subSequenceNumber: Long = record.subSequenceNumber()
+
+  /**
+   * Returns the ShutdownReason of the related
+   * [[software.amazon.kinesis.processor.ShardRecordProcessor]],
+   * if any.
+   */
+  def shutdownReason: Option[ShutdownReason]
+
+  /**
+   * Helper method that tells the caller if it's safe to invoke
+   * `forceCheckpoint` or not. It doesn't guarantee that invocations
+   * to either `tryToCheckpoint` or `forceCheckpoint` will succeed.
+   */
+  def canBeCheckpointed: Boolean =
+    !shutdownReason.contains(ShutdownReason.LEASE_LOST)
+
+  /**
+   * Safe checkpoint method, that will only attempt to checkpoint
+   * it the lease has not been lost and will capture expected
+   * Exceptions (that may still occur due to unavoidable race
+   * conditions). The method will still throw critical Exceptions.
+   *
+   * This method will potentially perform IO actions. Wrap accordingly
+   * in an IO Data Type if needed.
+   *
+   * See [[software.amazon.kinesis.processor.RecordProcessorCheckpointer]]
+   */
+  def tryToCheckpoint(): Unit =
+    if (canBeCheckpointed) {
+      try forceCheckpoint()
+      catch {
+        case _: ShutdownException => ()
+      }
+    }
+
+  /**
+   * Basic checkpoint method, the caller should decide if it's safe
+   * to invoke it. The method will throw any internal Exception.
+   *
+   * This method will potentially perform IO actions. Wrap accordingly
+   * in an IO Data Type if needed.
+   *
+   * See [[software.amazon.kinesis.processor.RecordProcessorCheckpointer]]
+   */
+  def forceCheckpoint(): Unit
+
+}
+
+object CommittableRecord {
+
+  // Only makes sense to compare Records belonging to the same shard
+  // Records that have been batched by the KPL producer all have the
+  // same sequence number but will differ by subsequence number
+  implicit val orderBySequenceNumber: Ordering[CommittableRecord] =
+    Ordering[(String, Long)].on(cr ⇒ (cr.sequenceNumber, cr.subSequenceNumber))
+
+  /**
+   * See [[akka.stream.alpakka.kinesis.impl.ShardProcessor]]
+   */
+  final class ShardProcessorData(
+      val shardId: String,
+      val recordProcessorStartingSequenceNumber: ExtendedSequenceNumber,
+      val pendingCheckpointSequenceNumber: ExtendedSequenceNumber
+  )
+
+  /**
+   * See [[akka.stream.alpakka.kinesis.impl.ShardProcessor]]
+   */
+  final class BatchData(
+      val cacheEntryTime: Instant,
+      val cacheExitTime: Instant,
+      val isAtShardEnd: Boolean,
+      val millisBehindLatest: Long
+  )
+}
--- a/kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisSchedulerErrors.scala
+++ b/kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisSchedulerErrors.scala
@@ -1,0 +1,12 @@
+/*
+ * Copyright (C) 2016-2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.kinesis
+
+object KinesisSchedulerErrors {
+
+  sealed class KinesisSchedulerError(err: Throwable) extends Throwable(err)
+  final case class SchedulerUnexpectedShutdown(cause: Throwable) extends KinesisSchedulerError(cause)
+
+}
--- a/kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisSchedulerSettings.scala
+++ b/kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisSchedulerSettings.scala
@@ -1,0 +1,61 @@
+/*
+ * Copyright (C) 2016-2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.kinesis
+
+import scala.concurrent.duration._
+
+final class KinesisSchedulerSourceSettings private (val bufferSize: Int, val backpressureTimeout: FiniteDuration) {
+  require(
+    bufferSize >= 1,
+    "Buffer size must be greater than 0; use size 1 to disable stage buffering"
+  )
+  def withBufferSize(size: Int): KinesisSchedulerSourceSettings =
+    KinesisSchedulerSourceSettings(size, backpressureTimeout)
+  def withBackpressureTimeout(timeout: java.time.Duration): KinesisSchedulerSourceSettings =
+    KinesisSchedulerSourceSettings(bufferSize, FiniteDuration.apply(timeout.toMillis, MILLISECONDS))
+}
+final class KinesisSchedulerCheckpointSettings private (val maxBatchSize: Int, val maxBatchWait: FiniteDuration) {
+  require(
+    maxBatchSize >= 1,
+    "Batch size must be greater than 0; use size 1 to commit records one at a time"
+  )
+
+  def withMaxBatchSize(size: Int): KinesisSchedulerCheckpointSettings =
+    KinesisSchedulerCheckpointSettings(size, maxBatchWait)
+  def withMaxBatchWait(duration: java.time.Duration): KinesisSchedulerCheckpointSettings =
+    KinesisSchedulerCheckpointSettings(maxBatchSize, FiniteDuration.apply(duration.toMillis, MILLISECONDS))
+}
+
+object KinesisSchedulerSourceSettings {
+
+  def apply(bufferSize: Int, backpressureTimeout: FiniteDuration): KinesisSchedulerSourceSettings =
+    new KinesisSchedulerSourceSettings(bufferSize, backpressureTimeout)
+  def apply: KinesisSchedulerSourceSettings = defaults
+
+  val defaults: KinesisSchedulerSourceSettings = KinesisSchedulerSourceSettings(1000, 1.minute)
+
+  /**
+   * Java API
+   */
+  def create(bufferSize: Int, backpressureTimeout: java.time.Duration): KinesisSchedulerSourceSettings =
+    apply(bufferSize, FiniteDuration.apply(backpressureTimeout.toMillis, MILLISECONDS))
+
+}
+
+object KinesisSchedulerCheckpointSettings {
+
+  def apply(maxBatchSize: Int, maxBatchWait: FiniteDuration): KinesisSchedulerCheckpointSettings =
+    new KinesisSchedulerCheckpointSettings(maxBatchSize, maxBatchWait)
+  def apply: KinesisSchedulerCheckpointSettings = defaults
+
+  val defaults: KinesisSchedulerCheckpointSettings = KinesisSchedulerCheckpointSettings(1000, 10.seconds)
+
+  /**
+   * Java API
+   */
+  def create(maxBatchSize: Int, maxBatchWait: java.time.Duration): KinesisSchedulerCheckpointSettings =
+    apply(maxBatchSize, FiniteDuration.apply(maxBatchWait.toMillis, MILLISECONDS))
+
+}
--- a/kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
+++ b/kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
@@ -1,0 +1,107 @@
+/*
+ * Copyright (C) 2016-2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.kinesis.impl
+
+import java.util.concurrent.Semaphore
+
+import akka.annotation.InternalApi
+import akka.stream.alpakka.kinesis.KinesisSchedulerErrors.SchedulerUnexpectedShutdown
+import akka.stream.alpakka.kinesis.{CommittableRecord, KinesisSchedulerSourceSettings}
+import akka.stream.stage._
+import akka.stream.{ActorAttributes, Attributes, Outlet, SourceShape}
+import software.amazon.kinesis.coordinator.Scheduler
+import software.amazon.kinesis.processor.{ShardRecordProcessor, ShardRecordProcessorFactory}
+
+import scala.annotation.tailrec
+import scala.collection.mutable
+import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.util.{Failure, Success, Try}
+
+/**
+ * Internal API
+ */
+@InternalApi
+private[kinesis] object KinesisSchedulerSourceStage {
+
+  final case class NewRecord(cr: CommittableRecord)
+  final case object Pump
+  final case object Complete
+  final case class SchedulerShutdown(result: Try[_])
+
+}
+
+/**
+ * Internal API
+ */
+@InternalApi
+private[kinesis] class KinesisSchedulerSourceStage(
+    settings: KinesisSchedulerSourceSettings,
+    schedulerBuilder: ShardRecordProcessorFactory => Scheduler
+)(implicit ec: ExecutionContext)
+    extends GraphStageWithMaterializedValue[SourceShape[CommittableRecord], Future[Scheduler]] {
+
+  private val out = Outlet[CommittableRecord]("Records")
+  override def shape: SourceShape[CommittableRecord] = new SourceShape[CommittableRecord](out)
+
+  override protected def initialAttributes: Attributes =
+    super.initialAttributes.and(ActorAttributes.IODispatcher)
+
+  override def createLogicAndMaterializedValue(
+      inheritedAttributes: Attributes
+  ): (GraphStageLogic, Future[Scheduler]) = {
+    val matValue = Promise[Scheduler]()
+    new Logic(matValue) -> matValue.future
+  }
+
+  final class Logic(matValue: Promise[Scheduler]) extends GraphStageLogic(shape) with StageLogging with OutHandler {
+    setHandler(out, this)
+
+    import KinesisSchedulerSourceStage._
+    import settings._
+
+    // We're transmitting backpressure from the Outlet to the Scheduler using a Semaphore instance
+    // semaphore.acquire ~> callback ~> push downstream ~> semaphore.release
+    private[this] val backpressureSemaphore = new Semaphore(bufferSize)
+    private[this] val buffer = mutable.Queue.empty[CommittableRecord]
+    private[this] var schedulerOpt: Option[Scheduler] = None
+
+    override def preStart(): Unit = {
+      val scheduler = schedulerBuilder(new ShardRecordProcessorFactory {
+        override def shardRecordProcessor(): ShardRecordProcessor =
+          new ShardProcessor(newRecordCallback)
+      })
+      schedulerOpt = Some(scheduler)
+      Future(scheduler.run()).onComplete(result => callback.invoke(SchedulerShutdown(result)))
+      matValue.success(scheduler)
+    }
+    private val callback: AsyncCallback[Any] = getAsyncCallback(awaitingRecords)
+    private val newRecordCallback: CommittableRecord => Unit = {
+      backpressureSemaphore.tryAcquire(backpressureTimeout.length, backpressureTimeout.unit)
+      record => callback.invoke(NewRecord(record))
+    }
+    override def onPull(): Unit = awaitingRecords(Pump)
+    override def onDownstreamFinish(): Unit = awaitingRecords(Complete)
+    @tailrec
+    private def awaitingRecords(in: Any): Unit = in match {
+      case NewRecord(record) =>
+        buffer.enqueue(record)
+        awaitingRecords(Pump)
+      case Pump =>
+        if (isAvailable(shape.out) && buffer.nonEmpty) {
+          push(shape.out, buffer.dequeue())
+          backpressureSemaphore.release()
+          awaitingRecords(Pump)
+        }
+      case SchedulerShutdown(Success(_)) | Complete =>
+        buffer.clear()
+        completeStage()
+      case SchedulerShutdown(Failure(e)) =>
+        buffer.clear()
+        failStage(SchedulerUnexpectedShutdown(e))
+    }
+    override def postStop(): Unit =
+      schedulerOpt.foreach(scheduler => Future(if (!scheduler.shutdownComplete()) scheduler.shutdown()))
+  }
+}
--- a/kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/ShardProcessor.scala
+++ b/kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/ShardProcessor.scala
@@ -1,0 +1,108 @@
+/*
+ * Copyright (C) 2016-2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.kinesis.impl
+
+import java.util.concurrent.Semaphore
+
+import akka.annotation.InternalApi
+import akka.stream.alpakka.kinesis.CommittableRecord
+import akka.stream.alpakka.kinesis.CommittableRecord.{BatchData, ShardProcessorData}
+import software.amazon.kinesis.lifecycle.ShutdownReason
+import software.amazon.kinesis.lifecycle.events._
+import software.amazon.kinesis.processor.{RecordProcessorCheckpointer, ShardRecordProcessor}
+import software.amazon.kinesis.retrieval.KinesisClientRecord
+
+import scala.collection.JavaConverters._
+
+@InternalApi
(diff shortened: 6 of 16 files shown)


aserrallerios:
Check it out @julianhowarth

aserrallerios:
FTP test are failing. All clear in Kinesis.

aserrallerios:
I've improved the shard termination, as discussed in https://github.com/aserrallerios/kcl-akka-stream/issues/13#issuecomment-446895988 with @julianhowarth.

project/Dependencies.scala
```diff
@@ -262,6 +262,10 @@ object Dependencies {
   val Kinesis = Seq(
     libraryDependencies ++= Seq(
       "com.amazonaws" % "aws-java-sdk-kinesis" % AwsSdkVersion, // ApacheV2
```
ennru: Hmm, so we'd depend on both SDK versions. Can we fix #1347 before merging this?

kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisSchedulerSettings.scala
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright (C) 2016-2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.kinesis
+
+import scala.concurrent.duration._
+
+case class KinesisSchedulerSourceSettings(bufferSize: Int, backpressureTimeout: FiniteDuration) {
```
ennru: Settings classes should be non-case classes to support evolving better. 
See [contributor advice](https://github.com/akka/alpakka/blob/master/contributor-advice.md#evolving-apis-with-binary-compatibility)

kinesis/src/main/scala/akka/stream/alpakka/kinesis/ShardProcessor.scala
```diff
@@ -0,0 +1,81 @@
...
+import software.amazon.kinesis.lifecycle.ShutdownReason
+import software.amazon.kinesis.lifecycle.events._
+import software.amazon.kinesis.processor.ShardRecordProcessor
+import software.amazon.kinesis.retrieval.KinesisClientRecord
+import software.amazon.kinesis.retrieval.kpl.ExtendedSequenceNumber
+
+import scala.collection.JavaConverters._
+import scala.concurrent.ExecutionContext
+
+private[kinesis] class ShardProcessor(
```
ennru: This should live in an `impl` package, I assume.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala
```diff
@@ -0,0 +1,108 @@
...
+import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.util.control.Exception
+import scala.util.{Failure, Success}
+
+object KinesisSchedulerSource {
+
+  def apply(
+      schedulerBuilder: ShardRecordProcessorFactory => Scheduler,
+      settings: KinesisSchedulerSourceSettings = KinesisSchedulerSourceSettings.defaultInstance
+  )(implicit schedulerExecutor: ExecutionContext): Source[CommittableRecord, Scheduler] =
```
ennru: A bit surprising to expect an execution context here, this would either need a full-blown `GraphStage` or the trick with the `SetupStage` we use in other Alpakka modules.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala
```diff
@@ -0,0 +1,108 @@
...
+    Source
+      .queue[CommittableRecord](settings.bufferSize, OverflowStrategy.backpressure)
+      .watchTermination()(Keep.both)
+      .mapMaterializedValue {
+        case (queue, watch) =>
+          val semaphore = new Semaphore(1, true)
```
ennru: Wow, haven't seen real blocking for a while, can this be solved in a non-blocking way?

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala
```diff
@@ -0,0 +1,108 @@
...
+  def checkpointRecordsFlow(
+      settings: KinesisSchedulerCheckpointSettings = KinesisSchedulerCheckpointSettings.defaultInstance
+  ): Flow[CommittableRecord, KinesisClientRecord, NotUsed] =
+    Flow[CommittableRecord]
+      .groupBy(MAX_KINESIS_SHARDS, _.shardId)
+      .groupedWithin(settings.maxBatchSize, settings.maxBatchWait)
+      .via(GraphDSL.create() { implicit b ⇒
+        import GraphDSL.Implicits._
+
+        val `{` =
```
ennru: Use alphanumeric names, please.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/CommittableRecord.scala
```diff
@@ -0,0 +1,46 @@
...
+package akka.stream.alpakka.kinesis
+
+import akka.Done
+import software.amazon.kinesis.lifecycle.ShutdownReason
+import software.amazon.kinesis.retrieval.KinesisClientRecord
+import software.amazon.kinesis.retrieval.kpl.ExtendedSequenceNumber
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class CommittableRecord(
```
ennru: This is a stream element type, right? Would factory methods simplify this API?

kinesis/src/main/scala/akka/stream/alpakka/kinesis/CommittableRecord.scala
```diff
@@ -0,0 +1,46 @@
...
+    val shardId: String,
+    val recordProcessorStartingSequenceNumber: ExtendedSequenceNumber,
+    val millisBehindLatest: Long,
+    val record: KinesisClientRecord,
+    recordProcessor: ShardProcessor,
+    checkpointer: KinesisClientRecord => Unit
+)(implicit executor: ExecutionContext) {
```
ennru: A bit unexpected to have an execution context within stream elements.

aserrallerios: Yeah, this looks very ugly. Should I just move it to the `tryToCheckpoint` method of this class?

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala
aserrallerios: Probably, but I have not found a better and/or simpler alternative than this implementation.

The thing is that the `Scheduler` lives in its own thread (it doesn't return control to the calling thread once you invoke `run()`) and we need to transmit backpressure to it, so I think that the callback needs to block somehow.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala
aserrallerios: Can you point me to an specific example?

ennru: S3 makes heavy use of it. https://github.com/akka/alpakka/blob/master/s3/src/main/scala/akka/stream/alpakka/s3/impl/SetupStage.scala
It will soon be available in Akka Streams by default https://github.com/akka/akka/pull/26477

kinesis/src/main/scala/akka/stream/alpakka/kinesis/CommittableRecord.scala
aserrallerios: Not sure about this comment. This class it's only instantiated from the `impl` code. Is it ok if I just make the constructor private?

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala
```diff
@@ -0,0 +1,113 @@
...
+        join.out.map(_._2).mapConcat(identity).map(_.record) ~> result
+
+        FlowShape(split.in, result.out)
+      })
+      .mergeSubstreams
+      .withAttributes(ActorAttributes.supervisionStrategy {
+        case _: ShutdownException =>
+          Resume
+        case _ => Stop
+      })
```
aserrallerios: @ennru can you help me with this stage?

This stage is just a Flow that "checkpoints" Kinesis records in an optimized and shard-aware fashion. A "checkpoint" is actually a synchronous IO call, that's why I wrap it around a Future (`CommittableRecord.tryToCheckpoint()`).

Do you think that it makes sense to remove the EC parameter also from this stage? If yes, what approach would you take?

aserrallerios: An option is to internally keep a reference of the EC that comes from the `ActorMaterializer` in each `CommittableRecord` (what I did before the latest change, but instead of using user's EC, use the materializer one.

Would that be ok?

kinesis/src/main/scala/akka/stream/alpakka/kinesis/CommittableRecord.scala
ennru: Yes, make it `@InternalApi private[kinesis]`. You may want to consider unit testing support like in [the reference connector](https://github.com/akka/alpakka/blob/master/reference/src/main/scala/akka/stream/alpakka/reference/testkit/MessageFactory.scala).

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala
ennru: Great that you found it, looks good.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/SetupStage.scala
```diff
@@ -0,0 +1,164 @@
...
+
+import akka.annotation.InternalApi
+import akka.stream._
+import akka.stream.scaladsl.{Flow, Keep, Sink, Source}
+import akka.stream.stage._
+
+import scala.concurrent.{Future, Promise}
+
+@InternalApi
+private[kinesis] final class SetupSinkStage[T, M](factory: ActorMaterializer => Attributes => Sink[T, M])
```
ennru: The setup-stages are now available as standard operators in Akka. Please change to use those.

aserrallerios:
Used akka's `Source.setup` and removed the code that `Async.await` on every `queue.offer`.

aserrallerios:
I had to do a change:

After the usage of `Source.setup`, the inner materialized value is wrapped in a `Future[_]`. As the inner MV was already a `Future[Scheduler]`, the final MV now becomes `Future[Future[Scheduler]]`.

In Scala 2.12 I can use `Future.flatten` to flatten it without `ExecutionContext`, but in older versions I cannot.

So, I will drop the MV (use `NotUsed`). If someone finds a way to conciliate a `Future[_]` MV with `SetupStage`, please, tell me and I will add the MV again.

aserrallerios:
mqtt/jms tests failed.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala
```diff
@@ -0,0 +1,94 @@
...
+      schedulerBuilder: ShardRecordProcessorFactory => Scheduler,
+      settings: KinesisSchedulerSourceSettings = KinesisSchedulerSourceSettings.defaultInstance
+  ): Source[CommittableRecord, NotUsed] =
+    Source
+      .setup { (mat, _) =>
+        import mat.executionContext
+        Source
+          .fromGraph(new KinesisSchedulerSourceStage(settings, schedulerBuilder))
+      }
+      .mapMaterializedValue(_ => NotUsed)
```
ennru: This will do the trick to retain the Scheduler future.
```scala
      .mapMaterializedValue(_.flatMap(identity)(ExecutionContexts.sameThreadExecutionContext))
```

docs/src/main/paradox/kinesis.md
```diff
@@ -135,6 +125,46 @@ Scala
 Java
 : @@snip [snip](/kinesis/src/test/java/docs/javadsl/KinesisSnippets.java) { #flow-sink }
 
+# AWS KCL Scheduler Source & checkpointer
+
+The KCL Source can read from several shards and rebalance automatically when other Schedulers are started or stopped. It also handles record sequence checkpoints.
+
+For more information about KCL please visit the [official documentation](http://docs.aws.amazon.com/streams/latest/dev/developing-consumers-with-kcl.html).
```
ennru: The version 2 link seems to be https://docs.aws.amazon.com/streams/latest/dev/developing-consumers-with-kcl-v2.html

project/Dependencies.scala
```diff
@@ -277,6 +277,8 @@ object Dependencies {
   val Kinesis = Seq(
     libraryDependencies ++= Seq(
         "com.amazonaws" % "aws-java-sdk-kinesis" % AwsSdkVersion, // ApacheV2
+        "software.amazon.awssdk" % "kinesis" % AwsSdk2Version, // ApacheV2
```
ennru: Having AWS SDK 1 and 2 on the dependency list isn't great. Before merging this we should fix #1347

kinesis/src/main/scala/akka/stream/alpakka/kinesis/Errors.scala
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (C) 2016-2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.kinesis
```
ennru: It might make sense to put this into a different package (eg. `kinesisscheduler` or `kinesiskcl`) so that is is kept separate just as `kinesisfirehose`.

aserrallerios:
Anything I can do to get this merged?

ennru:
Thank you for the ping.
Now that the whole Kinesis connector on AWS SDK v2 it would be great to warm this up.
Please resolve the conflicts, I'll have another look.

aserrallerios:
Done. Build fails elsewhere.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala
```diff
@@ -0,0 +1,110 @@
...
+    apply(schedulerBuilder, settings)
+      .groupBy(MAX_KINESIS_SHARDS, _.shardId)
+
+  def checkpointRecordsFlow(
+      settings: KinesisSchedulerCheckpointSettings = KinesisSchedulerCheckpointSettings.defaultInstance
+  ): Flow[CommittableRecord, KinesisClientRecord, NotUsed] =
+    Flow[CommittableRecord]
+      .groupBy(MAX_KINESIS_SHARDS, _.shardId)
+      .groupedWithin(settings.maxBatchSize, settings.maxBatchWait)
+      .via(GraphDSL.create() { implicit b ⇒
```
ennru:
```suggestion
      .via(GraphDSL.create() { implicit b =>
```

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala
```diff
@@ -0,0 +1,110 @@
...
+   * shards in coordination with other KCL Scheduler instances connected to the same stream. Only a single
+   * KCL Scheduler instance can own the lease of a shard at any given time. The Kinesis service assigns each
+   * shard lease among the available instances.
+   *
+   * Every message emitted downstream belongs to a shard. Messages can be committed to save the progress of
+   * the client consuming that shard.
+   *
+   * @param schedulerBuilder
+   * @param settings
+   * @return
```
ennru: Remove undocumented tags.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/CommittableRecord.scala
```diff
@@ -0,0 +1,46 @@
...
+import akka.Done
+import akka.annotation.InternalApi
+import akka.stream.alpakka.kinesis.impl.ShardProcessor
+import software.amazon.kinesis.lifecycle.ShutdownReason
+import software.amazon.kinesis.retrieval.KinesisClientRecord
+import software.amazon.kinesis.retrieval.kpl.ExtendedSequenceNumber
+
+import scala.util.Try
+
+class CommittableRecord @InternalApi private[kinesis] (
```
ennru: `final`?

kinesis/src/main/scala/akka/stream/alpakka/kinesis/CommittableRecord.scala
```diff
@@ -0,0 +1,46 @@
...
+  val subSequenceNumber: Long = record.subSequenceNumber()
+
+  def recordProcessorShutdownReason: Option[ShutdownReason] =
+    recordProcessor.shutdown
+
+  def canBeCheckpointed: Boolean =
+    !recordProcessorShutdownReason.contains(ShutdownReason.LEASE_LOST)
+
+  def tryToCheckpoint(): Try[Done] =
+    Try(checkpointer(record)).map(_ => Done)
```
ennru: Are these internal API?

kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisSchedulerErrors.scala
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (C) 2016-2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.kinesis
+
+object KinesisSchedulerErrors {
+
+  sealed class KinesisSchedulerError(err: Throwable) extends Throwable(err)
+  case class SchedulerUnexpectedShutdown(cause: Throwable) extends KinesisSchedulerError(cause)
```
ennru: `final`

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Internal API
+ */
+@InternalApi
+private[kinesis] object KinesisSchedulerSourceStage {
+
+  private[kinesis] final case class NewRecord(cr: CommittableRecord)
+  private[kinesis] final case object Pump
+  private[kinesis] final case object Complete
```
ennru: These can be unqualified `private`.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
```diff
@@ -0,0 +1,111 @@
...
+
+}
+
+/**
+ * Internal API
+ */
+@InternalApi
+private[kinesis] class KinesisSchedulerSourceStage(
+    settings: KinesisSchedulerSourceSettings = KinesisSchedulerSourceSettings.defaultInstance,
```
ennru:
```suggestion
    settings: KinesisSchedulerSourceSettings,
```

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
```diff
@@ -0,0 +1,111 @@
...
+    extends GraphStageWithMaterializedValue[SourceShape[CommittableRecord], Future[Scheduler]] {
+
+  private val out = Outlet[CommittableRecord]("Records")
+  override def shape: SourceShape[CommittableRecord] = new SourceShape[CommittableRecord](out)
+
+  override def createLogicAndMaterializedValue(
+      inheritedAttributes: Attributes
+  ): (GraphStageLogic, Future[Scheduler]) = {
+    val promise = Promise[Scheduler]()
+    (new GraphStageLogic(shape) with StageLogging with OutHandler {
```
ennru: It might read better if you make this a named internal class.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
```diff
@@ -0,0 +1,111 @@
...
+)(implicit ec: ExecutionContext)
+    extends GraphStageWithMaterializedValue[SourceShape[CommittableRecord], Future[Scheduler]] {
+
+  private val out = Outlet[CommittableRecord]("Records")
+  override def shape: SourceShape[CommittableRecord] = new SourceShape[CommittableRecord](out)
+
+  override def createLogicAndMaterializedValue(
+      inheritedAttributes: Attributes
+  ): (GraphStageLogic, Future[Scheduler]) = {
+    val promise = Promise[Scheduler]()
```
ennru: `matValue`?

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
```diff
@@ -0,0 +1,111 @@
...
+          completeStage()
+        case (_, NewRecord(record)) =>
+          buffer.enqueue(record)
+          self.ref ! Pump
+        case (_, Pump) =>
+          if (isAvailable(shape.out) && buffer.nonEmpty) {
+            push(shape.out, buffer.dequeue())
+            semaphore.release()
+            self.ref ! Pump
+          }
```
ennru: Move the most used cases first in the `match`.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
```diff
@@ -0,0 +1,111 @@
...
+        Future(scheduler.run()).onComplete {
+          case Failure(ex) =>
+            self.ref ! SchedulerUnexpectedShutdown(ex)
+          case Success(_) =>
+            self.ref ! Complete
+        }
+        promise.success(scheduler)
+      }
+
+      private def awaitingRecords(in: (ActorRef, Any)): Unit = in match {
```
ennru: match on `in._2`?

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
```diff
@@ -0,0 +1,111 @@
...
+      override def preStart(): Unit = {
+        self = getStageActor(awaitingRecords)
+        scheduler = schedulerBuilder(new ShardRecordProcessorFactory {
+          override def shardRecordProcessor(): ShardRecordProcessor =
+            new ShardProcessor(record => {
+              semaphore.tryAcquire(backpressureTimeout.length, backpressureTimeout.unit)
+              self.ref ! NewRecord(record)
+            })
+        })
+        Future(scheduler.run()).onComplete {
```
ennru: Alternatively, you could send the `Try` to `self.ref` instead of mapping it.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisSchedulerSettings.scala
```diff
@@ -0,0 +1,44 @@
...
+final class KinesisSchedulerCheckpointSettings(val maxBatchSize: Int, val maxBatchWait: FiniteDuration) {
+  require(
+    maxBatchSize >= 1,
+    "Batch size must be greater than 0; use size 1 to commit records one at a time"
+  )
+}
+
+object KinesisSchedulerSourceSettings {
+
+  val defaultInstance = new KinesisSchedulerSourceSettings(1000, 1.minute)
```
ennru: Return the default from parameterless `apply` and `create` methods.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/ShardProcessor.scala
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    // This implementation will try to checkpoint every Record with the original
+    // checkpointer. Other option would be to keep a reference of the latest
+    // checkpointer passed to this instance using any of these methods:
+    // * processRecords
+    // * shutdownRequested
+    // * shardEnded
+    val checkpoint = (record: KinesisClientRecord) =>
+      processRecordsInput.checkpointer().checkpoint(record.sequenceNumber(), record.subSequenceNumber())
```
ennru: Would that be something the user would want to decide?

kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisSchedulerSettings.scala
```diff
@@ -0,0 +1,44 @@
...
+    bufferSize >= 1,
+    "Buffer size must be greater than 0; use size 1 to disable stage buffering"
+  )
+}
+final class KinesisSchedulerCheckpointSettings(val maxBatchSize: Int, val maxBatchWait: FiniteDuration) {
+  require(
+    maxBatchSize >= 1,
+    "Batch size must be greater than 0; use size 1 to commit records one at a time"
+  )
+}
```
ennru: Please add `with...` methods (eg. `withMaxBatchWait(java.time.Duration)`).

kinesis/src/main/scala/akka/stream/alpakka/kinesis/CommittableRecord.scala
aserrallerios: It wraps a call to the internal checkpoint API, yes.

https://github.com/akka/alpakka/pull/1667/files#diff-877d82ee076716a8e876b7a6551328d0R47-R49

kinesis/src/test/scala/docs/scaladsl/KclSnippets.scala
```diff
@@ -0,0 +1,104 @@
...
+
+      //#Fan-out retrievalConfig - will incur additional AWS costs
+      val fanoutRetrievalConfig = configsBuilder.retrievalConfig
+      //#Fan-out retrievalConfig
+
+      //#Non-fan-out retrievalConfig i.e. equivalent of KCL 1 client
+      val retrievalConfig =
+        configsBuilder.retrievalConfig
+          .retrievalFactory(
+            new SynchronousBlockingRetrievalFactory(streamName, kinesisClient, new SimpleRecordsFetcherFactory, 1000)
```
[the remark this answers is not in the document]
aserrallerios: I removed unnecessary code from snippets.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisSchedulerSettings.scala
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright (C) 2016-2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.kinesis
+
+import scala.concurrent.duration._
+
+final class KinesisSchedulerSourceSettings(val bufferSize: Int, val backpressureTimeout: FiniteDuration) {
```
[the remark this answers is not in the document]
aserrallerios: Please check changes in KinesisSchedulerSettings.

kinesis/src/test/scala/docs/scaladsl/KclSnippets.scala
```diff
@@ -0,0 +1,104 @@
...
+        configsBuilder.retrievalConfig
+      )
+    }
+  //#scheduler-settings
+
+  //#scheduler-source
+  implicit val executor =
+    ExecutionContext.fromExecutorService(Executors.newFixedThreadPool(1000))
+
+  KinesisSchedulerSource(builder, schedulerSourceSettings).to(Sink.ignore)
```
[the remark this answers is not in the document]
aserrallerios: Added a some stream logic.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisSchedulerSettings.scala
aserrallerios: Should I add a `create` method in each companion obj here?

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/ShardProcessor.scala
aserrallerios: Do you mean using the "original" checkpointer vs use the "latest" checkpointer?

I think it's a low-level decission that the library should take for the user. It looks to work both ways, although that can change in the future. I'll try to improve the code a bit btw.

aserrallerios:
Please have a look now. The documentation improvements are still missing.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
```diff
@@ -0,0 +1,105 @@
...
+@InternalApi
+private[kinesis] class KinesisSchedulerSourceStage(
+    settings: KinesisSchedulerSourceSettings,
+    schedulerBuilder: ShardRecordProcessorFactory => Scheduler
+)(implicit ec: ExecutionContext)
+    extends GraphStageWithMaterializedValue[SourceShape[CommittableRecord], Future[Scheduler]] {
+
+  private val out = Outlet[CommittableRecord]("Records")
+  override def shape: SourceShape[CommittableRecord] = new SourceShape[CommittableRecord](out)
+
```
ennru: Make sure this runs on the IO dispatcher:
```
  override protected def initialAttributes: Attributes =
    super.initialAttributes.and(ActorAttributes.IODispatcher)
```

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
```diff
@@ -0,0 +1,105 @@
...
+    private[this] val semaphore = new Semaphore(bufferSize)
+    private[this] val buffer = mutable.Queue.empty[CommittableRecord]
+    private[this] var schedulerOpt: Option[Scheduler] = None
+    private[this] var self: StageActor = _
+
+    override def preStart(): Unit = {
+      self = getStageActor(awaitingRecords)
+      val newRecordCallback: CommittableRecord => Unit = {
+        semaphore.tryAcquire(backpressureTimeout.length, backpressureTimeout.unit)
+        self.ref ! NewRecord(_)
```
ennru: This is OK, but most other Alpakka connectors use async callbacks instead.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/javadsl/KinesisSchedulerSource.scala
```diff
@@ -0,0 +1,72 @@
...
+  ): Source[CommittableRecord, CompletionStage[Scheduler]] =
+    scaladsl.KinesisSchedulerSource
+      .apply(schedulerBuilder.build, settings)
+      .mapMaterializedValue(_.toJava)
+      .asJava
+
+  def create(
+      schedulerBuilder: SchedulerBuilder
+  ): Source[CommittableRecord, CompletionStage[Scheduler]] =
+    create(schedulerBuilder, KinesisSchedulerSourceSettings.defaultInstance)
```
ennru: Is it worth to have these overloads? Passing the default settings in the examples will do.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala
```diff
@@ -0,0 +1,103 @@
...
+   * shards in coordination with other KCL Scheduler instances connected to the same stream. Only a single
+   * KCL Scheduler instance can own the lease of a shard at any given time. The Kinesis service assigns each
+   * shard lease among the available instances.
+   *
+   * Every message emitted downstream belongs to a shard. Messages can be committed to save the progress of
+   * the client consuming that shard.
+   */
+  def apply(
+      schedulerBuilder: ShardRecordProcessorFactory => Scheduler,
+      settings: KinesisSchedulerSourceSettings = KinesisSchedulerSourceSettings.defaultInstance
```
ennru: Default parameters hinder bin-compatible evolvement. Passing the defaults explicitly is better.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala
```diff
@@ -0,0 +1,103 @@
...
+                case record if record.canBeCheckpointed =>
+                  record
+                    .tryToCheckpoint()
+                    .recover({
+                      case _: ShutdownException => Done
+                    })
+                    .get
+                case _ => Done
+              }
+              .addAttributes(Attributes(ActorAttributes.IODispatcher))
```
ennru: If you move this to a `ShardProcessor` object, the semaphore and the IO dispatcher show together and are easier to understand.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala
```diff
@@ -0,0 +1,103 @@
...
+      .via(GraphDSL.create() { implicit b =>
+        import GraphDSL.Implicits._
+        val split =
+          b.add(scaladsl.Broadcast[immutable.Seq[CommittableRecord]](2))
+        val join = b.add(Zip[Done, immutable.Seq[CommittableRecord]])
+        val result = b.add(Flow[KinesisClientRecord])
+
+        split
+          .out(0)
+          .map(_.max)
```
ennru: Might be good to point to the ordering defined in `CommittableRecord`.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala
```diff
@@ -0,0 +1,103 @@
...
+                    .tryToCheckpoint()
+                    .recover({
+                      case _: ShutdownException => Done
+                    })
+                    .get
+                case _ => Done
+              }
+              .addAttributes(Attributes(ActorAttributes.IODispatcher))
+          ) ~> join.in0
+        split.out(1) ~> join.in1
```
ennru: If you need to drop down into the DSL, use it all the way.
```
      split.out(0) ~> checkpoint ~> join.in0
      split.out(1) ~> join.in1
      
      join.out ~> flatten ~> result
```

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala
```diff
@@ -0,0 +1,103 @@
...
+    apply(schedulerBuilder, settings)
+      .groupBy(MAX_KINESIS_SHARDS, _.processorData.shardId)
+
+  def checkpointRecordsFlow(
+      settings: KinesisSchedulerCheckpointSettings = KinesisSchedulerCheckpointSettings.defaultInstance
+  ): Flow[CommittableRecord, KinesisClientRecord, NotUsed] =
+    Flow[CommittableRecord]
+      .groupBy(MAX_KINESIS_SHARDS, _.processorData.shardId)
+      .groupedWithin(settings.maxBatchSize, settings.maxBatchWait)
+      .via(GraphDSL.create() { implicit b =>
```
ennru: Put the DSL in a `val` so you can give it a describing name.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala
aserrallerios: Are you sure we want to remove default parameters? They are all over the Kinesis API...

aserrallerios:
The checkpoint coordination is tricky. Let me explain.

The issue is that when the `ShardProcessor` notifies that the shard has ended, you *must* checkpoint before returning control to the caller. That must happen synchronously, of course.

```
    /**
     * The checkpointer used to record that the record processor has completed the shard.
     *
     * The record processor <b>must</b> call {@link RecordProcessorCheckpointer#checkpoint()} before returning from
     * {@link ShardRecordProcessor#shardEnded(ShardEndedInput)}. Failing to do so will trigger the Scheduler to retry
     * shutdown until a successful checkpoint occurs.
     */
```

As we are checkpointing in an asynchronous way, that needs to be coordinated using another method: we lock the `shardEnded` method until all "regular" records have been checkpointed using the stream mechanism.

I'll put a comment in semaphore, but the `shardEnded` method *must* block, otherwise we're "closing" the shard before the "regular" records are consumed.

An alternative implementation to that is to give a grace period, to the final `shardEnded` checkpoint, but that'd potentially still leave records unconsumed.

I'm open to suggestions here.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/javadsl/KinesisSchedulerSource.scala
aserrallerios: Seen this pattern elsewhere in Kinesis. Maybe worth to consolidate globally in a later PR.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
aserrallerios: Please check the async callbacks logic now, is a bit less fair than before as I turned some of the intra-stage calls synchronous. I expect this to have higher performance. 

This can still be changed to everything asynchronous to maximize fairness.

aserrallerios:
Made all the checkpoint stuff a bit more explicit. Added comments and documentation to methods.

Only `kinesis.md` part is missing.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala
ennru: Yes, please remove them on these new methods.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisSchedulerSettings.scala
```diff
@@ -0,0 +1,65 @@
...
+  def apply(maxBatchSize: Int, maxBatchWait: java.time.Duration): KinesisSchedulerCheckpointSettings =
+    KinesisSchedulerCheckpointSettings(maxBatchSize, FiniteDuration.apply(maxBatchWait.toMillis, MILLISECONDS))
+  def apply: KinesisSchedulerCheckpointSettings = KinesisSchedulerCheckpointSettings(1000, 10.seconds)
+
+  val defaultInstance: KinesisSchedulerCheckpointSettings = KinesisSchedulerCheckpointSettings.apply
+
+  /**
+   * Java API
+   */
+  def create(maxBatchSize: Int, maxBatchWait: FiniteDuration): KinesisSchedulerCheckpointSettings =
```
ennru: Same here.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/KinesisSchedulerSettings.scala
```diff
@@ -0,0 +1,65 @@
...
+    KinesisSchedulerSourceSettings(bufferSize, FiniteDuration.apply(backpressureTimeout.toMillis, MILLISECONDS))
+  def apply: KinesisSchedulerSourceSettings = KinesisSchedulerSourceSettings(1000, 1.minute)
+
+  val defaultInstance: KinesisSchedulerSourceSettings = KinesisSchedulerSourceSettings.apply
+
+  /**
+   * Java API
+   */
+  def create(bufferSize: Int, backpressureTimeout: FiniteDuration): KinesisSchedulerSourceSettings =
+    new KinesisSchedulerSourceSettings(bufferSize, backpressureTimeout)
```
ennru: No reason for `java.time.Duration` in `apply`, but in the Java API.
You should create the instance in `val defaultInstance` and return that from `apply`.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/ShardProcessor.scala
```diff
@@ -0,0 +1,98 @@
...
+            lastRecord = processRecordsInput.isAtShardEnd && index + 1 == numberOfRecords
+          )
+        )
+    }
+  }
+
+  final class InternalCommittableRecord(record: KinesisClientRecord, batchData: BatchData, lastRecord: Boolean)
+      extends CommittableRecord(record, batchData, shardData) {
+    private def checkpoint(): Unit = {
+      checkpointer.checkpoint(record.sequenceNumber(), record.subSequenceNumber())
```
ennru: Is there any risk the `checkpointer` could be `null` here?

kinesis/src/main/scala/akka/stream/alpakka/kinesis/javadsl/KinesisSchedulerSource.scala
ennru: Yes, this is the same as the default arguments in the Scala DSL. I think it is better to make passing the default instance look as nice as possible (maybe call it just `defaults`?).

ennru:
@aserrallerios Any chance to push this the last bit?

aserrallerios:
Yes, let me see if I can get through it today.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/ShardProcessor.scala
aserrallerios: No, the `checkpointer` is gonna be referenced as soon as the `ShardProcessor` starts processing records (`processRecords` method is invoked). `checkpointer` reference can change but the variable is never de-referenced.

aserrallerios:
@ennru can you please remind me the suggestions regarding the `kinesis.md` file. I'm not able to find them.

project/Dependencies.scala
```diff
@@ -280,20 +280,20 @@ object Dependencies {
...
-          ExclusionRule("software.amazon.awssdk", "apache-client"),
-          ExclusionRule("software.amazon.awssdk", "netty-nio-client")
-        ),
-        "software.amazon.awssdk" % "firehose" % AwsSdk2Version excludeAll // ApacheV2
+        ExclusionRule(organization = "com.typesafe.akka"),
+        "org.mockito" % "mockito-core" % mockitoVersion % Test // MIT
+      ) ++ Seq(
+        "software.amazon.awssdk" % "kinesis" % AwsSdk2Version, // ApacheV2
+        "software.amazon.awssdk" % "firehose" % AwsSdk2Version, // ApacheV2
+        "software.amazon.awssdk" % "dynamodb" % AwsSdk2Version, // ApacheV2
```
ennru: DynamoDB is only used in tests, right?

aserrallerios: Is used by the KCL library, it's a transitive dependency. Probably I should remove it.

ennru: Ah, OK.
Might be better to pull it transitively, so please remove it.

ennru:
I believe the docs are OK.
We switched to Scalatest 3.1.0 which renames a few things. Would be great if you could rebase and fix those. (But if you are too short on time, I can do that.)

aserrallerios:
Rebased and squashed (I can push the old history if needed).

ennru:
Long-time coming! Thank you for your work on this.

seglo:
Hi @aserrallerios. We're observing a transiently failing error for a Kinesis test case.  You can find runtime details here in #2219.  The test in question utilizes a lot of `Thread.sleep's` and asynchronous code in `Futures` that are never awaited upon.  When I run locally it can fail ~50% of the time.  Can you take a look at this test?  I spent some time troubleshooting it this afternoon, but could not determine when additional checkpointing was called (it must be from the AWS libs).

aserrallerios:
Will have a look later today. Thanks!

seglo:
Thanks a lot @aserrallerios!

Issue #1663, opened 2019-04-24
FTP: run tests against FTP server in Docker


ennru:
### Short description

The embedded Apache FTP server used for tests has some limitations, instead, we could use an FTP server in Docker to test with.

References #1660 

This might solve some of the spurious test failures we see with FTP https://github.com/akka/alpakka/issues?utf8=%E2%9C%93&q=is%3Aopen+label%3Ap%3Aftp+label%3Afailed

ennru:
Fixed with #1668

PR #1668, opened 2019-04-26
FTP: run tests against (S)FTP server in Docker

tgambet:
## Purpose

The purpose of this PR is to move tests from using an embedded FTP and SFTP server to using docker.

## References

References #1663

### Test files

All test classes related to setting up the embedded servers have been removed or simplified.
Some files have been renamed to remove the confusion between FPT and SFTP specific files.

There are now only two main classes that tests can extend: `BaseFtpSupport` for FTP and FTPS support and `BaseSftpSupport` for SFTP support.

Issues encountered:
* the SFTP server doesn't allow writing to the user root folder so `BaseSftpSpec` and java equivalents had to be adapted to run tests inside an `upload/` subfolder transparently.
* the SFTP server does not accept keys that are publicly readable for security reasons, which is an issue for the current use case where the keys are committed to github. To prevent having to chmod the keys every time, I had to mount an `init.sh` script which overwrites the servers's self-generated keys on each startup and sets the correct rights.
* writing to the server outputstream and getting an `IOResult` doesn't guarantee that a file has indeed been fully written to disk yet, especially now that the server doesn't run on the jvm. To make tests pass more consistently I added an `eventually` clause whenever a file content must be checked.
* the SFTP server does not allow authentication with an invalid username regardless of whether we are using a key or not so the credentials passed to `SftpSettings` have to take that into account.
* testing connection failure by stopping the server is not possible anymore. Might be worth investigating using a killswitch to test correct reporting of failure?

### Production files

* `CommonFtpOperations` has been changed to take into account the server timezone. 
While testing against docker I found out that there was a two hours discrepancy between the system time and the time reported in the `FtpFile` (I'm in GMT+2). It turns out that the underlying `FTPClient` gives a `Calendar` instance for the last modified time which is subject to timezones. Setting the timezone to UTC fixes the issue but this should be reviewed more thoroughly.
* `FtpIOGraphStage` calls `completePendingCommand` where needed. See #1669.


tgambet:
Relevant Travis logs:
```
[info] Detected changes in directories: [, ftp, project]
[error] java.lang.RuntimeException: Could not create directory /home/travis/build/akka/alpakka/ftp/target/streams/$global/dependencyPositions/$global/streams
```
Not sure what's going on but it happens consistently. @ennru could you take a look please?

On my local environment running `sbt ftp/testChanged` results in a `FileNotFoundException` as scalafmt is trying to format a deleted file. I gave up trying to debug that. Help would be appreciated.

ennru:
I'm not sure about the Travis problem, will have a look tomorrow.
Your local problem should be solved by a `clean` in `ftp`.

tgambet:
Thank you, that did the trick.

ftp/src/test/java/akka/stream/alpakka/ftp/CommonStageTest.java
```diff
@@ -17,14 +17,15 @@
...
+import java.time.Instant;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
-interface CommonFtpStageTest extends FtpSupport, AkkaSupport {
+interface CommonStageTest extends BaseSupport, AkkaSupport {
```
ennru: Any particular reason for the rename? (not renaming would keep a smaller diff)

ftp/src/test/scala/akka/stream/alpakka/ftp/CommonFtpStageSpec.scala
```diff
@@ -278,56 +286,60 @@ trait CommonFtpStageSpec extends BaseSpec with Eventually {
...
+          getFtpFileContents(fileName).length shouldBe >(0)
         }
 
       val fileName = "sample_io_connection"
       val infiniteSource = Source.repeat(ByteString(0x00))
 
       val future = infiniteSource.runWith(storeToPath(s"/$fileName", append = false))
       waitForUploadToStart(fileName)
-      stopServer()
+      // stopServer()
```
ennru: You may drop these now. Yay.

ftp/src/test/scala/akka/stream/alpakka/ftp/BaseFtpsSpec.scala
```diff
@@ -29,7 +30,7 @@ trait BaseFtpsSpec extends FtpsSupportImpl with BaseSpec {
                                     emitTraversedDirectories: Boolean): Source[FtpFile, NotUsed] =
     Ftps.ls(basePath, settings, branchSelector, emitTraversedDirectories)
 
-  protected def retrieveFromPath(path: String): Source[ByteString, Future[IOResult]] =
+  protected def retrieveFromPath(path: String, fromRoot: Boolean = false): Source[ByteString, Future[IOResult]] =
```
ennru: I didn't see any use of the `fromRoot` parameter. Could it be dropped?

tgambet: It is used [here](https://github.com/akka/alpakka/pull/1668/files/ae5e67eadc9e710ef10e46154fba71373d5ce760#diff-95180c276c56d6b228c82b00d495d22eR185).

This is related to the fact that we can't write to the root folder of the SFTP server. In that test the path comes from ftpFiles and starts with `upload/` already (in the case of SFTP specs). So `upload/` [shouldn't be added again](https://github.com/akka/alpakka/pull/1668/files/ae5e67eadc9e710ef10e46154fba71373d5ce760#diff-9da90ed2137d0a1874b20e40e6b8d2a5R33) by `retrieveFromPath`. I tried to make things transparent from the test implementer perspective but I couldn't find a better solution than to leak an implementation detail here. You might have a better idea though.

This is also why [testing the path inside an FtpFile](https://github.com/akka/alpakka/pull/1668/files/ae5e67eadc9e710ef10e46154fba71373d5ce760#diff-95180c276c56d6b228c82b00d495d22eR145) is tricky since it may contain an `upload/` part or not depending on whether we are running an FTP test or an SFTP one.

ftp/src/test/java/akka/stream/alpakka/ftp/CommonStageTest.java
tgambet: I initially removed some 'Ftp' from trait names that were common to FTP and SFTP, mostly for me to understand better the overall file organization. I agree that it is not especially needed here, I'll revert.

ftp/src/test/scala/akka/stream/alpakka/ftp/CommonFtpStageSpec.scala
tgambet: You mean drop the whole "fail and report the exception in the result status if connection fails" test? It is currently ignored and I thought you might want to implement it another way.

ennru: Aha, I didn't notice it is ignored. I meant the `stopServer()` / `startServer()` comments. Leave it as it is.

ftp/src/test/scala/akka/stream/alpakka/ftp/BaseFtpsSpec.scala
ennru: OK.

ftp/src/test/java/akka/stream/alpakka/ftp/CommonStageTest.java
tgambet: Wait, did you also mean revert BaseSupport -> FtpSupport?

ennru: No. :-)

ennru:
Thank you once more for this effort, I really appreciate working with you.

tgambet:
Awesome, my pleasure.

PR #1670, opened 2019-04-29
Lazy I/O thread start

huntc:
Lazily start the IO thread only when a selector is used. This saves on a thread being used if there is never any Unix Domain Socket binding or connection.

I've noticed the following stacktrace as a consequence of this change, although I think it is benign for our circumstance due to (I think) being associated with shutdown:

```
Exception in thread "unix-domain-socket-io" java.lang.NullPointerException
	at jnr.enxio.channels.KQSelector$Descriptor.access$000(KQSelector.java:75)
	at jnr.enxio.channels.KQSelector.handleCancelledKeys(KQSelector.java:222)
	at jnr.enxio.channels.KQSelector.poll(KQSelector.java:150)
	at jnr.enxio.channels.KQSelector.select(KQSelector.java:145)
	at akka.stream.alpakka.unixdomainsocket.impl.UnixDomainSocketImpl$.akka$stream$alpakka$unixdomainsocket$impl$UnixDomainSocketImpl$$nioEventLoop(UnixDomainSocketImpl.scala:78)
	at akka.stream.alpakka.unixdomainsocket.impl.UnixDomainSocketImpl$$anon$1.run(UnixDomainSocketImpl.scala:365)
	at java.lang.Thread.run(Thread.java:748)
```

There's a separate ticket on JNR that I created for the above: https://github.com/jnr/jnr-enxio/pull/31

### DIFF
--- a/unix-domain-socket/src/main/scala/akka/stream/alpakka/unixdomainsocket/impl/UnixDomainSocketImpl.scala
+++ b/unix-domain-socket/src/main/scala/akka/stream/alpakka/unixdomainsocket/impl/UnixDomainSocketImpl.scala
@@ -358,13 +358,15 @@ private[unixdomainsocket] abstract class UnixDomainSocketImpl(system: ExtendedAc
   private implicit val materializer: ActorMaterializer = ActorMaterializer()(system)
   import system.dispatcher
 
-  private val sel = NativeSelectorProvider.getInstance.openSelector
-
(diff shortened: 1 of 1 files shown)


huntc:
Closing this as it actually isn't required and it introduces complexity. If the extension isn't loaded then the IO thread shouldn't start.

PR #1673, opened 2019-05-02
SOLR: Exclude message with PassThrough

giena:
This PR allows to pass a kafka offset in the passthrough, even if we don't want to update solr cloud.

You can see an analogy with this: https://doc.akka.io/docs/alpakka-kafka/current/atleastonce.html#excluding-messages

Failure to deserialize a kafka message is a particular case of conditional message processing.
It is also likely that we would have no message to produce to SolR when we encounter messages that fail to deserialize.
The solr flow will not let us pass through the corresponding committable offset without doing a request to solr.

Even if i know you are currently working on stream context, i did not change anything concerning the pass-through parameter. I think this is two different things.

### DIFF
--- a/solr/src/main/scala/akka/stream/alpakka/solr/impl/SolrFlowStage.scala
+++ b/solr/src/main/scala/akka/stream/alpakka/solr/impl/SolrFlowStage.scala
@@ -31,7 +31,7 @@ private[solr] final class SolrFlowStage[T, C](
     collection: String,
     client: SolrClient,
     settings: SolrUpdateSettings,
-    messageBinder: Option[T => SolrInputDocument]
+    messageBinder: T => SolrInputDocument
 ) extends GraphStage[FlowShape[immutable.Seq[WriteMessage[T, C]], immutable.Seq[WriteResult[T, C]]]] {
 
   private val in = Inlet[immutable.Seq[WriteMessage[T, C]]]("messages")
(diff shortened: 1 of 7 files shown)


solr/src/main/scala/akka/stream/alpakka/solr/scaladsl/SolrFlow.scala
```diff
@@ -140,7 +140,28 @@ object SolrFlow {
...
+          Some(binder)
+        )
+      )
+
+  /**
+   * Do not stream element to Solr. Just pass the pass-through data.
+   *
+   * @tparam PT pass-through type
+   */
+  def passThrough[T, PT](
```
ennru: I do not understand the use-case for this flow.

giena: Yes you're right. I correct it.

solr/src/main/scala/akka/stream/alpakka/solr/impl/SolrFlowStage.scala
```diff
@@ -149,21 +149,25 @@ private final class SolrFlowLogic[T, C](
     responses.find(_.getStatus != 0).getOrElse(responses.head)
   }
 
+  private def passThrough(current: immutable.Seq[WriteMessage[T, C]]): Unit =
+    if (log.isDebugEnabled) log.debug(s"passThrough writes $current")
```
ennru: This logging doesn't make much sense. Just remove this method.

Issue #1672, opened 2019-05-02
AWS SQS: Batch stages fail before emitting all successful items  


RustedBones:
### Versions used 

1.0.0

Akka version: 

2.5

### Expected Behavior

Considering the following batch response: 
```java
BatchResponse(Successful(A, C), Failed(B))
```
I expect the stream to emit the downstream the `A` and `C` items before failing.
Indeed, users may rely on those for proper functioning, eg:
- Publish result to avoid duplicating message in the queue
- Delete ack to clean resources

### Actual Behavior

On receive of the above `BatchResponse`, the stage fails straight, dropping the successful `A` and `C` result entries.

### Reproducible Test Case

I'll add a unit test on a PR to demonstrate

gabfssilva:
Not an expert about Akka Streams performance here, but, using mapConcat this way, we transform every request into another Source, this way we could decrease the performance drastically, doesn't it?

PR #1675, opened 2019-05-02
AWS SQS: Emit all successful batch item entries before failing stream

RustedBones:
## Purpose

Emit successful sqs batch result entry before failing

## References

References #1672

## Changes

- `SqsPublishFlow.batch`
- `SqsAckFlow.batch`

## Background Context

Wrap the result entries into a `Try`.
Collect first all successful results from a batch response and then append the failures.
`mapConcat` emits `Try` of result, unwrapped in a later map stage. 

To be discussed:
```
SqsPublishFlow.batch: Flow[Iterable[SendMessageRequest], List[SqsPublishResult[SendMessageBatchResponse]], NotUsed]
```
This API does not allow to properly tell for a batch which items were successful and witch are failed.

At the moment the old behavior is kept: If only one item fails, the stream fails. For this 'low level API', I would suggest to wrap every `SendMessageBatchResponse` in a `Try`.

### DIFF
--- a/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsBatchException.scala
+++ b/sqs/src/main/scala/akka/stream/alpakka/sqs/SqsBatchException.scala
@@ -6,21 +6,20 @@ package akka.stream.alpakka.sqs
 
 import akka.annotation.InternalApi
 
-final class SqsBatchException @InternalApi private[sqs] (val batchSize: Int, message: String)
-    extends Exception(message) {
+import scala.collection.JavaConverters._
 
-  @InternalApi
-  private[sqs] def this(batchSize: Int, cause: Throwable) {
-    this(batchSize, cause.getMessage)
-    initCause(cause)
-  }
-
-  @InternalApi
-  private[sqs] def this(batchSize: Int, message: String, cause: Throwable) {
-    this(batchSize, message)
-    initCause(cause)
-  }
+final class SqsBatchException[T <: AnyRef] @InternalApi private[sqs] (val errors: List[SqsResultErrorEntry[T]])
+    extends Exception(s"SQS batch operation failed with ${errors.size} errors") {
 
   /** Java API */
-  def getBatchSize: Int = batchSize
+  def getErrors: java.util.List[SqsResultErrorEntry[T]] = errors.asJava
+
+  override def toString: String = s"SqsBatchException(errors=$errors)"
+
+  override def equals(other: Any): Boolean = other match {
+    case that: SqsBatchException[T] => java.util.Objects.equals(this.errors, that.errors)
+    case _ => false
+  }
+
+  override def hashCode(): Int = java.util.Objects.hash(errors)
 }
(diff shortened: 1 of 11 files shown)


gabfssilva:
Not an expert about Akka Streams performance here, but, using mapConcat this way, we transform every request into another Source, this way we could decrease the performance drastically, does not it?

RustedBones:
`mapContat` just takes as output an iterator to `faltten` the stream.
In this case instead of taking an iterator from list as it was before from the `mapConcat(identity)`, I create an iterator from the scala `Stream` collection so the execution of `.map(_.get)` is executed lazily.

I'm not concerned about performance but I think reliability can be improved by moving the `Try` unwrapping after the `mapConcat` phase in an explicit pattern match

ennru:
The discussion of the stream element type fits well in with your other PR #1604.
We could bring these two things together to find the best solution.

RustedBones:
Would you prefer to rebase this commit on top of the other PR ?

ennru:
Yeah, let's finish #1604 first.

ennru:
Are you ready to continue this on top of Alpakka SQS 1.0.1?

RustedBones:
Sure :)
There's still small things to improve but I think I should probably wait a bit before requesting some new breaking changes :sweat_smile:

ennru:
There is a chance to carefully break APIs for Alpakka 2.0 now...

RustedBones:
Batch operation now emit a `SqsBatchResult[T]` containing successful and failed entries.

`SqsResult` has been extended so that  both success and error entries fit in `SqsResultEntry`

`SqsBatchException` contains all error entries

sqs/src/main/scala/akka/stream/alpakka/sqs/SqsBatchException.scala
```diff
@@ -6,21 +6,20 @@ package akka.stream.alpakka.sqs
...
-final class SqsBatchException @InternalApi private[sqs] (val batchSize: Int, message: String)
-    extends Exception(message) {
+import scala.collection.JavaConverters._
 
-  @InternalApi
-  private[sqs] def this(batchSize: Int, cause: Throwable) {
-    this(batchSize, cause.getMessage)
-    initCause(cause)
-  }
+final class SqsBatchException[T <: AnyRef] @InternalApi private[sqs] (val errors: List[SqsResultErrorEntry[T]])
```
ennru: A bit strange to carry elements in the exception.

RustedBones:
On a second read of the PR, I noticed that this solution is not complete:
When `SqsAckFlow` is used, the [merge](https://github.com/akka/alpakka/blob/4782c0b07cdeb549b65e6d89d993f666f19e19fb/sqs/src/main/scala/akka/stream/alpakka/sqs/scaladsl/SqsAckFlow.scala#L93) operatin may lead to failure without flushing all successful items. Some of them may indeed still be in the another upstream sub-flow when we fail.

So I'm not sure if this whole complication is worth, since it doesn't bring any grantee in the end. Just local improvement.

ennru:
Good point. So we can't easily give better failure behaviour for the general case.

ennru:
Thank you for trying to improve this.
We now know there is no simple solution to it.
I'll close this PR.

PR #1678, opened 2019-05-04
Align local date in s3 stream signing key to UTC

zingkg:
## Purpose

Align LocalDate.now() function in S3Stream.scala's signing key with UTC. This is because the credential scope in AWS should be in UTC.

## Changes

* Align LocalDate.now() to be LocalDate.now(UTC)

## Background Context

AWS is expecting UTC time zone for credential scope.


ennru:
Thank you for fixing this!

Issue #1054, opened 2018-06-28
InfluxDB connector


ennru:
InfluxDB fits very well in many IoT use cases where Alpakka can contribute a lot. Would be great if people with InfluxDB needs or knowledge stepped up and contribute to get a working connector.

fsanaulla:
I can help with it.

PR #1680, opened 2019-05-05
Influxdb connector

gkatzioura:
## Purpose

This is the Alpakka connector for InfluxDB

## References

References #1054

## Changes

Added the connector for InfluxDB. Added InfluxDB on docker compose configuration.

## Background Context

Commits are based on the source code of existing connectors like Elasticsearch and OrientDB.

Sources are created through queries. Inserts can be done both through models using the InfluxDB mapper or through Points. Batch operations are possible and enabled.


gkatzioura:
I get a failure both on the compatibility check (it also happens on master). Any help on fixing it it will be greatly appreciated

```java
[error] (awslambda / mimaReportBinaryIssues) akka-stream-alpakka-awslambda: Binary compatibility check failed!
[error] (influxdb / mimaPreviousClassfiles) sbt.librarymanagement.ResolveException: unresolved dependency: com.lightbend.akka#akka-stream-alpakka-influxdb_2.12;1.0-M3: not found
[error] (dynamodb / mimaReportBinaryIssues) akka-stream-alpakka-dynamodb: Binary compatibility check failed!
[error] (reference / mimaPreviousClassfiles) sbt.librarymanagement.ResolveException: unresolved dependency: com.lightbend.akka#akka-stream-alpakka-reference_2.12;1.0-M3: not found
```

ennru:
Thank you for this PR.

You may switch off the MiMa check for the new module with `.disablePlugins(MimaPlugin)`. The other problems you mention don't show for me or on Travis.

ennru:
Please add the new module to `.travis` as well.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  final class InfluxDBRecordLogic extends InfluxDBLogic {
+
+    override protected def write(messages: immutable.Seq[InfluxDBWriteMessage[T, C]]): Unit =
+      messages.foreach {
+        case InfluxDBWriteMessage(point: Point, _) => {
+          influxDB.write(point)
+          println(point.toString)
```
adkafka: Should we remove this `println`?

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
```diff
@@ -0,0 +1,86 @@
...
+      tryPull(in)
+    }
+  }
+
+  final class InfluxDBRecordLogic extends InfluxDBLogic {
+
+    override protected def write(messages: immutable.Seq[InfluxDBWriteMessage[T, C]]): Unit =
+      messages.foreach {
+        case InfluxDBWriteMessage(point: Point, _) => {
+          influxDB.write(point)
```
adkafka: I'm a little bit confused about how we are using this [`influxdb-java`](https://github.com/influxdata/influxdb-java) library. I apologize if these questions are basic, but I am not familiar with this library.

When we call `influxDB.write(point)`, what is that actually doing? Is it just adding the message to an internal buffer? Does it perform a synchronous write?

How does it handle write failures? Are we guaranteed that the previous write succeeded when we emit this element?

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
```diff
@@ -0,0 +1,86 @@
...
+    override def onPull(): Unit = if (!isClosed(in) && !hasBeenPulled(in)) pull(in)
+
+    override def onPush(): Unit = {
+      val messages = grab(in)
+      if (messages.nonEmpty) {
+
+        influxDB.enableBatch(BatchOptions.DEFAULTS)
+        write(messages)
+        val writtenMessages = messages.map(m => new InfluxDBWriteResult(m, Option.empty))
+        influxDB.close()
```
adkafka: Is it inefficient to call "close" here? Can we re-use the thread pool between elements?

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
gkatzioura: Sorry totally got away :/

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
gkatzioura: When we call to write a point and batch is enabled, then the write calls shall be batched until close is called.
Therefore in this case the messages are being written to the internal buffer.
Once close is called, then the messages are being flashed.

adkafka: So we need to call `close` in order to actually make the HTTP request to Influx? And when `close` returns, do we know if Influx has received the request (did it return an HTTP 201)?

gkatzioura: Exactly, once close is called the actual request is being made.
If there is an error during the request the driver throws an Exception.
The exception might be an InfluxDBException which has to do with the state or an InfluxDBIOException which has to do with the IO.

gkatzioura: Edit: my previous comment is misleading. Apart from the close operation which flushes the messages, there is also an interval that writes the messages to the database.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
gkatzioura: Batching can be enabled once. Therefore if two threads try to enable batch simultaneously an exception shall be thrown. Instead you enable batch once and then all write requests (originating from various threads) go through a buffer (link blocking queue) and the messages are flush through an interval  (default 1000 milliseconds) or when close is called.

gkatzioura: @adkafka

With regards to batching the reason for adding it, was to sent all the data at once instead of making write requests to InfluxDB one by one. Therefore the batch of writes on the Alpakka side would be processed as a batch on the client side too. Thus I am not actually using the batching functionality for the long term, instead I am using it in order to sent all the writes at once. 

There are various options with regards to the driver’s batching functionality.

1) Don’t add any batching through the driver at all. This will lead to writes being executed one by one.

2) Let the user handle the batching. Instead on enabling or closing the batching client side, the user will do so. He can use the same or various InfluxDB instances on the streams implemented and the user is going to decide when batching shall be enabled/disabled.

This might have some side effects. When batching is enabled (until the close operation is called). The data are going to be put in a queue and they are going to be processed either once a scheduler is called or when close is called. This shall lead to write requests from different flows (onPush) to be actually placed on a queue and get processed by a thread not managed by the Akka environment.

3) Another option is to enable batching per onPush calls (current solution). Instead of passing the InfluxDB instance, the flow shall create an instance from the configuration passed (just like the elastic and oriented implementation where the client is created on the spot) to execute the batches once off and then close.

Facilitating all of the options might make it bloated. Which one do you think is better (or if there is a better proposal)?

gkatzioura:
@adkafka 

With regards to the driver unfortunately I have not yet used the [https://github.com/paulgoldbaum/scala-influxdb-client](scala-influxdb-client).

The reason why I picked the [https://github.com/influxdata/influxdb-java](influxdb-java) is mainly because it is the official driver and I am also one of the contributors (implemented the QueryBuilder and the InfluxDBMapper).

There are also some features on this driver such as the QueryBuilder which make it possible to write complex queries efficiently using java.

Should there be a solution more suitable to Alpakka, I am more than happy to contribute and get more value for the implementation. For example a more Scala friendly query builder of a mapper for Scala only.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
adkafka: I am still a bit confused about what happens when we call `close`. It appears to me that one of the two is occurring:

1. The thread will *block* while it makes the HTTP call. If Influx accepts the write, the `close` function will return, therefore we can emit the messages we have written. If the write fails, an exception will be thrown, and we can handle that in the decider.
2. The thread does not block, and the write occurs asynchronously. 

If 1) is correct, we need to make sure we are handling the thread blocking appropriately. This is outside my knowledge range, maybe @ennru can help you more around this. But we don't want one thread with a blocking call to cause thread starvation on other threads. If we need to use blocking, we should run this on a separate dispatcher that is optimized for blocking calls. I understand the Influx Library has its own thread pool solution, but we still call `close` on our thread pool, so we have to be careful.

If 2) is correct, we don't have "at least once delivery", because we emit the messages before we are sure that Influx has acknowledged them. 

Can you help clarify which of these is occurring?

gkatzioura: 1) is correct.

I will explain the flow and why option 1 is correct.
Batching is enabled (influxDB.enableBatch()). 
By enabling batching all the points submitted for a write will be added to a queue.
By enabling batching a scheduler is configured, to run (by default every 1 second) and offload some of the messages in the queue to InfluxDB. It is clear that this step occurs in another thread managed by the driver.
Once batching is stop ( influxDB.close()) the scheduler is stopped and the points left on the queue are going to be written to the database (all of the batched in one request). This step is happening by the same thread that calls the close operation. Therefore the batch of points are written by the main thread.

Why '1)' is correct, and how the writes occur in the main thread?
In the connector code once batch is enabled the only action that happens is to add the messages to the queue and once they are added the batch operation is closed. Thus the point are send to persist through the main thread. The write and close steps occur before the scheduler is called at all which leads to the actions to take place on the main thread.

However considering the complexity of the above, which might lead to misconceptions and false usage, I shall make a commit which implements directly the actual functionality needed to sent all of the writes once off without using the drivers batch functionality (therefore without the need to deal and be aware of the internal mechanisms of the driver).

influxdb/src/main/resources/reference.conf
```diff
@@ -0,0 +1,9 @@
+akka.stream.alpakka.influxdb{
+
+  #batch write size
```
mingraham: Missing a space after `#` and have an extra space on line 7 before the `=`.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/InfluxDBSettings.scala
```diff
@@ -0,0 +1,33 @@
...
+import scala.concurrent.duration.TimeUnit
+
+object InfluxDBSettings {
+  val Default = new InfluxDBSettings(batchSize = 10, TimeUnit.MILLISECONDS)
+
+  def apply(): InfluxDBSettings = Default
+
+}
+
+final class InfluxDBSettings private (
```
mingraham: Curious on why we don't want to make this one a `final case class`?

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBSourceStage.scala
```diff
@@ -0,0 +1,128 @@
...
+                                                     influxDB: InfluxDB,
+                                                     query: Query,
+                                                     outlet: Outlet[T],
+                                                     shape: SourceShape[T])
+    extends GraphStageLogic(shape)
+    with OutHandler {
+
+  setHandler(outlet, this)
+
+  var dataRetrieved: Option[QueryResult] = Option.empty
```
mingraham: Would probably be more intuitive to users if we used `None` instead of `Option.empty`. Perhaps just a preference, though.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBSourceStage.scala
```diff
@@ -0,0 +1,128 @@
...
+    resultMapperHelper.cacheClassFields(clazz)
+
+    val queryResult = influxDB.query(query)
+    if (!queryResult.hasError) {
+      dataRetrieved = Option(queryResult)
+    }
+  }
+
+  override def onPull(): Unit =
+    if (dataRetrieved.isEmpty)
```
mingraham: Would be more Scala-esque to go with pattern matching here, and it would allow us to not have to call `.get` on an `Option` (even though we already checked it was non-empty).

```
dataRetrieved match {
  case Some(queryResult) =>
    ...
  case None =>
    ...
}
```

mingraham: Applies in other situations, as well.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/javadsl/InfluxDBFlow.scala
```diff
@@ -0,0 +1,52 @@
...
+import akka.stream.alpakka.influxdb.{InfluxDBSettings, InfluxDBWriteMessage, InfluxDBWriteResult}
+import org.influxdb.InfluxDB
+
+object InfluxDBFlow {
+
+  def create[T](
+      settings: InfluxDBSettings,
+      influxDB: InfluxDB
+  ): akka.stream.javadsl.Flow[InfluxDBWriteMessage[T, NotUsed], InfluxDBWriteResult[T, NotUsed], NotUsed] =
+    akka.stream.alpakka.influxdb.scaladsl.InfluxDBFlow.create(settings)(influxDB).asJava
```
mingraham: Could clean a lot of these types up by importing them. Would improve readability.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/scaladsl/InfluxDBFlow.scala
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Scala API to create InfluxDB flows.
+ */
+object InfluxDBFlow {
+
+  def create[T](settings: InfluxDBSettings)(
+      implicit influxDB: InfluxDB
+  ): Flow[InfluxDBWriteMessage[T, NotUsed], InfluxDBWriteResult[T, NotUsed], NotUsed] =
+    Flow[InfluxDBWriteMessage[T, NotUsed]]
+      .batch(settings.batchSize, immutable.Seq(_)) { case (seq, wm) => seq :+ wm }
```
mingraham: If desired, I believe you can just do:
```
.batch(settings.batchSize, immutable.Seq(_))(_ :+ _)
```

influxdb/src/main/scala/akka/stream/alpakka/influxdb/scaladsl/InfluxDBSource.scala
```diff
@@ -0,0 +1,43 @@
...
+ */
+object InfluxDBSource {
+
+  /**
+   * Java API: creates an [[akka.stream.alpakka.influxdb.impl.InfluxDBRawSourceStage]] from a given statement.
+   */
+  def apply(influxDB: InfluxDB, query: Query): Source[QueryResult, NotUsed] =
+    Source.fromGraph(
+      new InfluxDBRawSourceStage(
+        query,
```
mingraham: In a few places, I see that we are breaking out the arguments of the constructor. This is great if the line is too long and/or we want to name the arguments to improve readability/understandability, but in this case where we don't name the arguments and where it could fit on one line, I think it would be better to collapse this to one line.

Makes the code easier to comprehend, IMO.

influxdb/src/test/java/docs/javadsl/Cpu.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.time.Instant;
+import org.influxdb.annotation.Column;
+
+public class Cpu {
+
+  @Column(name = "time")
+  private Instant time;
+
+  @Column(name = "host", tag = true)
```
mingraham: Should the `name` be `"hostname"` here, to match the name of the variable it is decorating?

influxdb/src/test/java/docs/javadsl/InfluxDBSourceTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+  @After
+  public void cleanUp() {
+    cleanDatabase(influxDB, DATABASE_NAME);
+    StreamTestKit.assertAllStagesStopped(materializer);
+  }
+
+  @Test
+  public void streamQueryResult() throws Exception {
+    Query query = new Query("SELECT*FROM cpu", DATABASE_NAME);
```
mingraham: Missing spaces before and after the `*`

mingraham: Applies elsewhere.

influxdb/src/test/java/docs/javadsl/InfluxDBTest.java
```diff
@@ -0,0 +1,144 @@
...
+        queryResult.get(0).getResults().get(0).getSeries().get(0).getValues().size();
+
+    Assert.assertEquals(4, resultSize);
+  }
+
+  private List<InfluxDBWriteMessage<Point, NotUsed>> points(QueryResult queryResult) {
+    List<InfluxDBWriteMessage<Point, NotUsed>> points = new ArrayList<>();
+
+    for (QueryResult.Result result : queryResult.getResults()) {
+      for (QueryResult.Series serie : result.getSeries()) {
```
mingraham: Typo `serie`. Should be `series` I think

influxdb/src/test/java/docs/javadsl/TestUtils.java
```diff
@@ -0,0 +1,79 @@
...
+
+  public static void cleanDatabase(final InfluxDB influxDB, final String databaseName) {
+    influxDB.query(new Query("DROP MEASUREMENT cpu", databaseName));
+  }
+
+  public static void dropDatabase(InfluxDB influxDB, final String databaseName) {
+    influxDB.query(new Query("DROP DATABASE " + databaseName));
+  }
+
+  public static Point resultToPoint(QueryResult.Series serie, List<Object> values) {
```
mingraham: Typo `serie`, should be `series`.

influxdb/src/test/scala/docs/scaladsl/InfluxDBSourceSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+import org.influxdb.dto.Query
+
+class InfluxDBSourceSpec
+    extends WordSpec
+    with MustMatchers
+    with BeforeAndAfterEach
+    with BeforeAndAfterAll
+    with ScalaFutures {
+
+  val databaseName = "InfluxDBSourceSpec"
```
mingraham: Nit-pick: due to this being a constant, we can follow [the Scala standard](https://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods) of UpperCamelCasing (and probably using `final`, since this is a class):
```

```final val DatabaseName = ...```

File: influxdb/src/test/scala/docs/scaladsl/InfluxDBSpec.scala [hunk 16]
```diff
@@ -0,0 +1,105 @@
...
+      val f2 = InfluxDBSource.typed(classOf[InfluxDBSpecCpu], InfluxDBSettings(), influxDB, query).runWith(Sink.seq)
+
+      f2.futureValue.length mustBe 4
+    }
+
+    "consume and publish measurements" in assertAllStagesStopped {
+      val query = new Query("SELECT*FROM cpu", databaseName);
+
+      val f1 = InfluxDBSource(influxDB, query)
+        .map { queryReqult =>
```
mingraham: Seems to be a typo here: `queryReqult`. Would be simpler to just use the following instead of the `.map` and `.mapConcat`.
```
.mapConcat(resultToPoints)
```

influxdb/src/test/scala/docs/scaladsl/InfluxDBSpec.scala
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    def resultToPoints(queryResult: QueryResult): List[InfluxDBWriteMessage[Point, NotUsed]] = {
+      val points = for {
+        results <- queryResult.getResults.asScala
+        serie <- results.getSeries.asScala
+        values <- serie.getValues.asScala
+      } yield
+        (
+          InfluxDBWriteMessage(resultToPoint(serie, values))
```
mingraham: `serie` typo

influxdb/src/main/scala/akka/stream/alpakka/influxdb/scaladsl/InfluxDBFlow.scala
```diff
@@ -21,51 +21,31 @@ object InfluxDBFlow {
...
   ): Flow[InfluxDBWriteMessage[T, NotUsed], InfluxDBWriteResult[T, NotUsed], NotUsed] =
     Flow[InfluxDBWriteMessage[T, NotUsed]]
       .batch(settings.batchSize, immutable.Seq(_))(_ :+ _)
-      .via(
-        new impl.InfluxDBFlowStage[T, NotUsed](
-          Option(clazz),
-          influxDB
-        )
-      )
+      .via(new impl.InfluxDBFlowStage[T, NotUsed](Option(clazz), influxDB))
```
mingraham: Nit (feel free to ignore): usually is nice to instantiate `Option`s with `Some` instead of `Option`. I think I saw this elsewhere in the PR, so would be good to remain consistent. 👍

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
ennru: If any of the operations called are known to be blocking, we'd want to make the stage logic run on the IO dispatcher.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/scaladsl/InfluxDBFlow.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  def typedWithPassThrough[T, C](clazz: Class[T], settings: InfluxDBSettings)(
+      implicit influxDB: InfluxDB
+  ): Flow[InfluxDBWriteMessage[T, C], InfluxDBWriteResult[T, C], NotUsed] =
+    Flow[InfluxDBWriteMessage[T, C]]
+      .batch(settings.batchSize, immutable.Seq(_))(_ :+ _)
+      .via(new impl.InfluxDBFlowStage[T, C](Some(clazz), influxDB))
+      .mapConcat(identity)
+
+  def createWithContext[T, C](settings: InfluxDBSettings)(
```
ennru: Great that you prepared for `withContext`, we'll see that more and more.
I wonder if we should default to using tuples instead of the pass-through field in the messages.

influxdb/src/main/java/org/influxdb/impl/InfluxDBResultMapperHelper.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (C) 2016-2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package org.influxdb.impl;
```
ennru: Why does this class live in that package?

influxdb/src/main/resources/reference.conf
```diff
@@ -0,0 +1,18 @@
+akka.stream.alpakka.influxdb{
```
ennru: `alpakka.influxdb` is enough.

influxdb/src/main/java/org/influxdb/impl/InfluxDBResultMapperHelper.java
gkatzioura: This has to do with some package-private methods the influxdb-java driver contains which are used. I can extract them and put them in another class if needed.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
gkatzioura: I updated the code and removed any parts that have to do with the drivers batching operation based on external threads . The blocking operations are isolated and thus the logic shall be set to run on the IO dispatcher.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
gkatzioura: The latest commit has removed any driver thread-based batching implementations and sends a batch of measurements using and http request. Since it is a blocking operation the logic shall be moved on the IODispatcher. 
Therefore the streams environment shall have full control over the request and the batch size.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/scaladsl/InfluxDBFlow.scala
gkatzioura: Good! I suppose something like this

```scala


def typedWithContext[T, C](clazz: Class[T], settings: InfluxDBSettings)(
    implicit influxDB: InfluxDB
): Flow[((InfluxDBWriteMessage[T],NotUsed), C), (InfluxDBWriteResult[T], C), NotUsed] =
  Flow[((InfluxDBWriteMessage[T],NotUsed), C)]
    .map {
      case (wm, pt) =>
        InfluxDBWriteMessage(wm.point, pt)
    }
    .via(typedWithPassThrough(clazz, settings))
    .map { wr =>
      (wr, wr.writeMessage.passThrough)
    }

```

Is there any implementation that does so?

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
```diff
@@ -58,28 +61,70 @@ private[influxdb] class InfluxDBFlowStage[T, C](
   final class InfluxDBRecordLogic extends InfluxDBLogic {
 
     override protected def write(messages: immutable.Seq[InfluxDBWriteMessage[T, C]]): Unit =
-      messages.foreach {
-        case InfluxDBWriteMessage(point: Point, _) => {
-          influxDB.write(point)
+      messages
+        .filter {
+          case InfluxDBWriteMessage(point: Point, _, _, _) => {
```
mingraham: `point` and `others` actually aren't needed in these two `case`s, since we aren't using them in the subsequent blocks. We can just replace those with `_`.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
```diff
@@ -58,28 +61,70 @@ private[influxdb] class InfluxDBFlowStage[T, C](
...
+              case Some(databaseName) => Some(databaseName)
+              case None => Some(mapperHelper.retentionPolicy(im.point))
+            }
+          )
+        })
+        .map(
+          wm =>
+            toBatchPoints(wm._1._1,
+                          wm._1._2,
+                          wm._2.map(im => im.withPoint(mapperHelper.convertModelToPoint(im.point).asInstanceOf[T])))
```
mingraham: `im => im` -> `_`

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
```diff
@@ -58,28 +61,70 @@ private[influxdb] class InfluxDBFlowStage[T, C](
...
-
-    override def preStart(): Unit =
-      influxDBMapper = new InfluxDBMapper(influxDB)
+    private val mapperHelper: InfluxDBResultMapperHelper = new InfluxDBResultMapperHelper
 
     override protected def write(messages: immutable.Seq[InfluxDBWriteMessage[T, C]]): Unit =
-      messages.foreach {
-        case InfluxDBWriteMessage(typeMetric: Any, _) =>
-          influxDBMapper.save(typeMetric)
+      messages
```
mingraham: One suggestion for readability around flows is keeping the flow declarations succinct. If we define methods elsewhere that handle the `groupBy` logic and the `map` logic, we can write this like:
```
messages
  .groupBy(groupByDbRp)
  .map(toBatchPoints)
  .foreach(influxDB.write)

def groupByDbRp(im: InfluxDBWriteMessage[T, C]) = ...
def toBatchPoints(wm: Thing) = ...
```
I think this helps communicate to users/readers the intent of the flow better.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
gkatzioura: Yes indeed I tried to use the underscore however im is needed for the 

```java
mapperHelper.convertModelToPoint(im.point).asInstanceOf[T])
```

mingraham: Ah, my apologies, missed that. 👍

influxdb/src/main/scala/akka/stream/alpakka/influxdb/scaladsl/InfluxDBFlow.scala
ennru: We've added initial support in Alpakka Elasticsearch and have a usage example [in the documentation](https://doc.akka.io/docs/alpakka/current/examples/elasticsearch-samples.html#example-read-from-a-kafka-topic-and-publish-to-elasticsearch). It is implemented on top of the pass-through just as you did.
The upcoming Akka release 2.5.23 will fix a few things around the Java `withContext` API. We'll upgrade Alpakka immediately when that release is out.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/InfluxDBSettings.scala
```diff
@@ -0,0 +1,33 @@
...
+import scala.concurrent.duration.TimeUnit
+
+object InfluxDBSettings {
+  val Default = new InfluxDBSettings(batchSize = 10, TimeUnit.MILLISECONDS)
+
+  def apply(): InfluxDBSettings = Default
+
+}
+
+final case class InfluxDBSettings private (
```
ennru: To enable binary-compatible evolution, we can't use case classes for settings. Even though the constructor is private, scalac creates `apply` with the fields and `unapply`. We've been using [Kaze Class](https://github.com/ktoso/kaze-class) a lot to convert case classes.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/InfluxDBWriteMessage.scala
```diff
@@ -0,0 +1,49 @@
...
+  // Java-api - without passThrough
+  def create[T](point: T): InfluxDBWriteMessage[T, NotUsed] =
+    InfluxDBWriteMessage(point, NotUsed)
+
+  // Java-api - with passThrough
+  def create[T, C](point: T, passThrough: C) =
+    InfluxDBWriteMessage(point, passThrough)
+}
+
+final case class InfluxDBWriteMessage[T, C](point: T,
```
ennru: Make this just a regular class with a private constructor, please.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/InfluxDBWriteMessage.scala
```diff
@@ -0,0 +1,49 @@
...
+                                            passThrough: C,
+                                            databaseName: Option[String] = None,
```
ennru: A bit surprising for me to see `databaseName` per element. Isn't that a candidate to have in the settings? But I guess that is an InfluxDB specific thing.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
```diff
@@ -0,0 +1,131 @@
...
+        }
+      )
+
+    def convertToBatchPoints(wm: ((Some[String], Some[String]), immutable.Seq[InfluxDBWriteMessage[T, C]])) =
+      toBatchPoints(wm._1._1,
+                    wm._1._2,
+                    wm._2.map(im => im.withPoint(mapperHelper.convertModelToPoint(im.point).asInstanceOf[T])))
+  }
+
+  private def toBatchPoints(databaseName: Option[String],
```
ennru: Shouldn't this method live in `InfluxDBLogic`?

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
```diff
@@ -0,0 +1,131 @@
...
+                            retentionPolicy: Option[String],
+                            seq: Seq[InfluxDBWriteMessage[T, C]]) = {
+
+    val builder = databaseName match {
```
ennru: Looking into `BatchPoints` indicates you could just do `databaseName.orNull`.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
```diff
@@ -0,0 +1,131 @@
...
+      case Some(databaseName) => BatchPoints.database(databaseName)
+      case None => BatchPoints.builder()
+    }
+
+    if (retentionPolicy.isDefined) builder.retentionPolicy(retentionPolicy.get)
+
+    def convert(messages: Seq[InfluxDBWriteMessage[T, C]]): BatchPoints =
```
ennru: Add `@tailrec`.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
```diff
@@ -0,0 +1,131 @@
...
+import org.influxdb.impl.InfluxDBResultMapperHelper
+
+import scala.collection.immutable
+import org.influxdb.BatchOptions
+import org.influxdb.dto.{BatchPoints, Point}
+
+/**
+ * INTERNAL API
+ */
+private[influxdb] class InfluxDBFlowStage[T, C](
```
ennru: Add `@InternalApi`.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
```diff
@@ -0,0 +1,131 @@
...
+  private def toBatchPoints(databaseName: Option[String],
+                            retentionPolicy: Option[String],
+                            seq: Seq[InfluxDBWriteMessage[T, C]]) = {
+
+    val builder = databaseName match {
+      case Some(databaseName) => BatchPoints.database(databaseName)
+      case None => BatchPoints.builder()
+    }
+
+    if (retentionPolicy.isDefined) builder.retentionPolicy(retentionPolicy.get)
```
ennru:
```suggestion
    retentionPolicy.foreach(p => builder.retentionPolicy(p))
```

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
```diff
@@ -0,0 +1,131 @@
...
+  final class InfluxDBRecordLogic extends InfluxDBLogic {
+
+    override protected def write(messages: immutable.Seq[InfluxDBWriteMessage[T, C]]): Unit =
+      messages
+        .filter {
+          case InfluxDBWriteMessage(_: Point, _, _, _) => {
+            true
+          }
+          case InfluxDBWriteMessage(_: AnyRef, _, _, _) => {
+            failStage(new RuntimeException(s"unexpected type Point required"))
```
ennru:
```suggestion
            failStage(new RuntimeException(s"unexpected type ${v.getClass}, Point required"))
```
But this could be stopped via a type constraint for `T`, I believe.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
```diff
@@ -0,0 +1,131 @@
...
+      messages
+        .groupBy(groupByDbRp)
+        .map(convertToBatchPoints)
+        .foreach(influxDB.write)
+
+    def groupByDbRp(im: InfluxDBWriteMessage[T, C]) =
+      (
+        im.databaseName match {
+          case Some(databaseName) => Some(databaseName)
+          case None => Some(mapperHelper.databaseName(im.point))
```
ennru: Is there anything we can do to enable caching even in this usage? Or will it cache with the first lookup anyway?

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
```diff
@@ -0,0 +1,131 @@
...
+    override protected def write(messages: immutable.Seq[InfluxDBWriteMessage[T, C]]): Unit =
+      messages
+        .groupBy(groupByDbRp)
+        .map(convertToBatchPoints)
+        .foreach(influxDB.write)
+
+    def groupByDbRp(im: InfluxDBWriteMessage[T, C]) =
+      (
+        im.databaseName match {
+          case Some(databaseName) => Some(databaseName)
```
ennru:
```suggestion
          case dbn: Some(databaseName) => dbn
```

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
```diff
@@ -0,0 +1,131 @@
...
+  override protected def initialAttributes: Attributes =
+    super.initialAttributes and Attributes(ActorAttributes.IODispatcher)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    clazz match {
+      case Some(c) => new InfluxDBMapperRecordLogic
+      case None => new InfluxDBRecordLogic
+    }
+
+  sealed abstract class InfluxDBLogic extends GraphStageLogic(shape) with InHandler with OutHandler {
```
ennru: I find it often easier to evolve the code when the logic is a non-inner class, as you do with `InfluxDBSourceLogic`.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBSourceStage.scala
```diff
@@ -0,0 +1,123 @@
...
+  setHandler(outlet, this)
+
+  var dataRetrieved: Option[QueryResult] = None
+  var resultMapperHelper: InfluxDBResultMapperHelper = _
+
+  override def preStart(): Unit = {
+    resultMapperHelper = new InfluxDBResultMapperHelper
+    resultMapperHelper.cacheClassFields(clazz)
+
+    val queryResult = influxDB.query(query)
```
ennru: This will issue the query before there is any demand.

I assume this should run on the `IODispatcher`, as well.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBSourceStage.scala
```diff
@@ -0,0 +1,123 @@
...
+    if (!queryResult.hasError) {
```
ennru: What happens with the error?

influxdb/src/main/scala/akka/stream/alpakka/influxdb/InfluxDBSettings.scala
```diff
@@ -0,0 +1,33 @@
...
+ * Copyright (C) 2016-2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.influxdb
+
+import java.util.concurrent.TimeUnit
+
+import scala.concurrent.duration.TimeUnit
+
+object InfluxDBSettings {
```
ennru: We often see the need for specific settings for flows vs sources. Might be better to start off with both `InfluxDbSourceSettings` and `InfluxDbWriteSetting`.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/InfluxDBWriteMessage.scala
gkatzioura: This functionality is pretty close to using a schema prefix when using sql. The connection string for the connectivity is not bound to the database (although you can set a default database on the driver) thus the user has the ability to specify the database per measurement.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
mingraham: Taking the suggested change one step further:
```
retentionPolicy.foreach(builder.retentionPolicy)
```

influxdb/src/main/scala/akka/stream/alpakka/influxdb/InfluxDBWriteMessage.scala
gkatzioura: I was using it for pattern matching. Will proceed on the change as advised.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBSourceStage.scala
gkatzioura: Due to the http api being capable of executing many queries there might be cases of a partially erroneous payload. 

For example

```bash
SELECT*FROM valid_measure; SELECT invalid_action FROM invalid_measure;
```

This will lead to a payload which contains data for the first query and error for the next.
My recent commit throws an error (failStage) when it comes to all queries from a request failing. In cases of partial errors the user shall be able to handle what went wrong and what went bad.
When it comes to the typed queries
which are bound to one type of data, thus one type of result, an error will be thrown anyway.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
gkatzioura: Create a separate FlowStage for points

influxdb/src/main/scala/akka/stream/alpakka/influxdb/InfluxDBSettings.scala
gkatzioura: Refactored to InfluxDbReadSettings and InfluxDbWriteSettings

influxdb/src/main/scala/akka/stream/alpakka/influxdb/InfluxDbReadSettings.scala
```diff
@@ -0,0 +1,26 @@
...
+final class InfluxDbReadSettings private (val precision: TimeUnit) {
+
+  def withPrecision(value: TimeUnit): InfluxDbReadSettings = copy(precision = value)
+
+  private def copy(
+      precision: TimeUnit = precision
+  ): InfluxDbReadSettings = new InfluxDbReadSettings(
+    precision = precision
+  )
+
```
ennru: Add a `toString`, please.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/InfluxDbWriteMessage.scala
```diff
@@ -0,0 +1,57 @@
...
+  // Java-api - without passThrough
+  def create[T](point: T): InfluxDbWriteMessage[T, NotUsed] =
+    new InfluxDbWriteMessage(point, NotUsed)
+
+  // Java-api - with passThrough
+  def create[T, C](point: T, passThrough: C) =
+    new InfluxDbWriteMessage(point, passThrough)
+}
+
+final class InfluxDbWriteMessage[T, C](val point: T,
```
ennru: `private` constructor

influxdb/src/main/scala/akka/stream/alpakka/influxdb/InfluxDbWriteSettings.scala
```diff
@@ -0,0 +1,26 @@
...
+) {
+
+  def withBatchSize(value: Int): InfluxDbWriteSettings = copy(batchSize = value)
+
+  private def copy(
+      batchSize: Int = batchSize,
+  ): InfluxDbWriteSettings = new InfluxDbWriteSettings(
+    batchSize = batchSize
+  )
+
```
ennru: Add a `toString`, please.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/scaladsl/InfluxDbSource.scala
```diff
@@ -0,0 +1,38 @@
...
+import org.influxdb.InfluxDB
+import org.influxdb.dto.{Query, QueryResult}
+
+/**
+ * Scala API.
+ */
+object InfluxDbSource {
+
+  /**
+   * Java API: creates an [[akka.stream.alpakka.influxdb.impl.InfluxDbRawSourceStage]] from a given statement.
```
ennru: Copy-paste error, this is Scala API.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/scaladsl/InfluxDbSink.scala
```diff
@@ -0,0 +1,30 @@
...
+import akka.stream.alpakka.influxdb.{InfluxDbWriteMessage, InfluxDbWriteSettings}
+import akka.stream.scaladsl.{Keep, Sink}
+import org.influxdb.InfluxDB
+import org.influxdb.dto.Point
+
+import scala.concurrent.Future
+
+object InfluxDbSink {
+
+  def apply(
```
ennru: To align with `Flow`, you might use `create` here.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/scaladsl/InfluxDbFlow.scala
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Scala API to create InfluxDB flows.
+ */
+object InfluxDbFlow {
+
+  def create(settings: InfluxDbWriteSettings)(
+      implicit influxDB: InfluxDB
+  ): Flow[InfluxDbWriteMessage[Point, NotUsed], InfluxDbWriteResult[Point, NotUsed], NotUsed] =
+    Flow[InfluxDbWriteMessage[Point, NotUsed]]
+      .batch(settings.batchSize, immutable.Seq(_))(_ :+ _)
```
ennru: I think we should consider leaving the grouping to the user instead of always using `batch`. `batch` will only group if there is back-pressure, but there might be benefits of grouping even when writing isn't slow eg. with [`groupedWeightedWithin`](https://doc.akka.io/docs/akka/current/stream/operators/Source-or-Flow/groupedWeightedWithin.html). I admit that the API becomes a bit cruder, showing batching in the examples should balance that, though.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/scaladsl/InfluxDbFlow.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  def typed[T](clazz: Class[T], settings: InfluxDbWriteSettings)(
+      implicit influxDB: InfluxDB
+  ): Flow[InfluxDbWriteMessage[T, NotUsed], InfluxDbWriteResult[T, NotUsed], NotUsed] =
+    Flow[InfluxDbWriteMessage[T, NotUsed]]
+      .batch(settings.batchSize, immutable.Seq(_))(_ :+ _)
+      .via(new impl.InfluxDbMapperFlowStage[T, NotUsed](clazz, influxDB))
+      .mapConcat(identity)
+
+  def createWithPassThrough[C](settings: InfluxDbWriteSettings)(
```
ennru: Alternatively, just offer the pass-through versions and leave it to the user to set the `NotUsed` type. It shows anyway in the elements.

ennru:
The Scala 2.11 build fails, I'm OK with excluding 2.11. Add `crossScalaVersions -= Dependencies.Scala211` to the settings to switch it off.

ennru:
Can you please direct the logging during tests to a file, as in all other modules?
See https://travis-ci.com/akka/alpakka/jobs/202982406#L622

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDBFlowStage.scala
gkatzioura: @ennru 
This in memory grouping certainly has a cost. The goal is to use the batching capabilities of the driver. The requests are batched per database and retention policy.
To make this more efficient I propose to add the optional fields databaseName and retentionPolicy on the InfluxDbWriteSettings. Those fields will override the database/retention policy information per message. If those optional fields are set then all the measurements/points shall be written (and thus batched) to the same database and retention policy as in the settings. Therefore there won't be the need for the grouping above.
Eventually the user will have the one database/retention policy option, or the various database/retention policy combinations.

gkatzioura:
@ennru 
Greeting below is a list of issues on which I would like to consult you on  you on how I should proceed

**InfluxDBResultMapperHelper**

https://github.com/akka/alpakka/pull/1680#discussion_r282461301

The reason why It currently resides at the `org.influxdb.impl` package is due to some package private methods. I can extract the methods being used and thus put the InfluxDBResultMapperHelper.java at the `akka.stream.alpakka.influxdb` package. If this is more preferable than its current state then I will proceed so. 

**Database name per  element**

https://github.com/akka/alpakka/pull/1680#discussion_r285118562

If you reckon that having a database per element It is not a functionality of great value, I can proceed on removing it, thus the users shall specify the database through the settings or the driver configuration.

**Grouping and batch**

https://github.com/akka/alpakka/pull/1680#discussion_r285129316
https://github.com/akka/alpakka/pull/1680#discussion_r286060171

Based on the batch retrieved from the settings when creating a flow a batch operation is applied on the driver side. Based on my comment I can provide methods which are not based on any batching in the first place and methods which shall do batching only for one database (thus no in memory actions with the underlying collection).

**Tuples instead of pass through field in the messages**

https://github.com/akka/alpakka/pull/1680#discussion_r282460938

I failed to find an implementation which offers the passthrough using tuples. Correct me If I am wrong, what is proposed is an implementation where the passthrough field does not exist on the InfluxDbWriteMessage, instead any passthrough related information exists through a tuple (InfluxDbWriteMessage,T)? If the above is correct I shall proceed on that.

ennru:
> InfluxDBResultMapperHelper
> The reason why It currently resides at the org.influxdb.impl package is due to some package private methods. I can extract the methods being used and thus put the InfluxDBResultMapperHelper.java at the akka.stream.alpakka.influxdb package. If this is more preferable than its current state then I will proceed so.

If that is technically possible, please do so. It not intended as user-facing API, so `akka.stream.alpakka.influxdb.impl` would be the best package.

> Database name per element
> If you reckon that having a database per element It is not a functionality of great value, I can proceed on removing it, thus the users shall specify the database through the settings or the driver configuration.

I haven't used InfluxDB in any real set-up. Maybe @adkafka has some advice?

> Grouping and batch
> Based on the batch retrieved from the settings when creating a flow a batch operation is applied on the driver side. Based on my comment I can provide methods which are not based on any batching in the first place and methods which shall do batching only for one database (thus no in memory actions with the underlying collection).
> Therefore shall I provide more options, remove any batched write functionality or provide batching just per database?

No, please leverage the drivers possibilities for batching. Looking closer at  https://github.com/akka/alpakka/pull/1680/files/eaf6060c1c3e09be148209112cdb0d3ad285b994#diff-1baa4b8bbbb8f810c355306b9966993bR90 you should remove the wrapping in `Some` for both tuple fields.
About the batching, I think you should remove the `batch` operator and instead make the flow accept `immutable.Seq[InfluxDbWriteMessage[Point, NotUsed]]` so that the user may choose an appropriate way to create batches to Influx. 

> Tuples instead of pass through field in the messages
> I failed to find an implementation which offers the passthrough using tuples.  Correct me If I am wrong, what is proposed is an implementation where the passthrough field does not exist on the InfluxDbWriteMessage, instead any passthrough related information exists through a tuple (InfluxDbWriteMessage,T)? If the above is correct I shall proceed on that.

Remove the `withContext` for now, `passThrough` in the message is easier and more flexible.

adkafka:
@ennru @gkatzioura 

> Database name per element
If you reckon that having a database per element It is not a functionality of great value, I can proceed on removing it, thus the users shall specify the database through the settings or the driver configuration.

It does seem a bit awkward to me to **need** to specify the database for every point. A write to Influx is just an HTTP POST, where the database name is included as a query parameter, and the data points are included in the body of the POST. See [their docs here](https://docs.influxdata.com/influxdb/v1.7/guides/writing_data/#write-data-using-the-http-api).

I think the majority of the time, an application will be writing to a single database. There definitely will be cases where there will need to be partitioning of incoming elements to two flows, so it is worth thinking this through. Would it be possible to include a `BatchInfluxDbWriteMessage` (feel free to change name) that looks the same as `InfluxDbWriteMessage` but instead of `point: T`, it had `points: Seq[T]`? I think I would find that interface cleaner to use. Seems like something we could also choose to add later too if we want to merge this in sooner.

To provide some contrast / another point of view, we can look at how the "Paul Goldbaum" InfluxDB client is set up. By no means is that library perfect, and I think what we have in this PR will provide a better interface, but it does shine some light.
* The [`Point`](https://github.com/paulgoldbaum/scala-influxdb-client/blob/master/src/main/scala/com/paulgoldbaum/influxdbclient/Point.scala#L3) class has no notion of a database.
* The user can [choose to perform a `write` (one point) or a `bulkWrite` (`Seq` of points)](https://github.com/paulgoldbaum/scala-influxdb-client/blob/master/src/main/scala/com/paulgoldbaum/influxdbclient/Database.scala#L15-L26) on a particular database.
* The user selects which database to use by creating an instance of [`Database`](https://github.com/paulgoldbaum/scala-influxdb-client/blob/master/src/main/scala/com/paulgoldbaum/influxdbclient/InfluxDB.scala#L31) through the main class in the API.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDbFlowStage.scala
```diff
@@ -0,0 +1,153 @@
...
+  override def onPull(): Unit = if (!isClosed(in) && !hasBeenPulled(in)) pull(in)
+
+  override def onPush(): Unit = {
+    val messages = grab(in)
+    if (messages.nonEmpty) {
+
+      influxDB.enableBatch(BatchOptions.DEFAULTS)
+      write(messages)
+      val writtenMessages = messages.map(m => new InfluxDbWriteResult(m, None))
+      influxDB.close()
```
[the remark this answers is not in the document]
gkatzioura: @ennru 
Exactly. Once close() is called the batch is disabled and the messages are send to the database. 
However this is no longer needed and It shall removed.
I extracted the functionality, that the driver provided with batching enabled, and is currently executed by the connector itself. 
This means that on a onPush call, the messages are written to the database in one http request.
Therefore the user has control over the batch size and the delay through the connector without the need to be aware of the driver internals.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/InfluxDbFlowStage.scala
```diff
@@ -0,0 +1,153 @@
...
+
+  protected def write(messages: immutable.Seq[InfluxDbWriteMessage[T, C]]): Unit
+
+  override def onPull(): Unit = if (!isClosed(in) && !hasBeenPulled(in)) pull(in)
+
+  override def onPush(): Unit = {
+    val messages = grab(in)
+    if (messages.nonEmpty) {
+
+      influxDB.enableBatch(BatchOptions.DEFAULTS)
```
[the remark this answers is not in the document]
gkatzioura: This shall be removed as explained [here](https://github.com/akka/alpakka/pull/1680#discussion_r293718043)

ennru:
@gkatzioura This looks good now, I propose we'll merge it and mark it as "API may change" so we keep the door open to apply more of @adkafka's suggestions.
We require documentation, though.
It would be perfect to get this in before we release Alpakka 1.0.3 on 2019-07-03.

gkatzioura:
@ennru 
Just added the documentation and also added the @ApiMayChange annotation.
The api will indeed change apart from extra recommendations due to a major release on InfluxDB 2.0. Is there any faq on api changes?

ennru:
Thank's for the heads-up about the new version. I'll put a not in the documentation about both "API may change" and the upcoming InfluxDB version.

ennru:
Thank you so much for this effort! We've seen the interest for InfluxDB support for a while, but didn't get to look into it. Highly appreciated.

gkatzioura:
@ennru 
I am keeping a close eye on the changes that occur. The version 2.0 is going to be huge, with extra functionality so I will be back with some contributions.

PR #1685, opened 2019-05-09
Add alpakka-csv benchmark and optimize.

htmldoug:
## Purpose

Improves parsing speed of alpakka-csv from 1.5 MiB/s to 78 MiB/s for the most extreme case.

## References

I searched, but didn't find any other issues or discussions.

## Changes

* New jmh benchmark: `CsvBench`
* `CsvParser` is faster.

## Background Context

While evaluating alpakka-csv as a non-blocking replacement for our currently blocking usage of OpenCSV, I benched it and found that it parses slower than I'd expect, even over benchmarks that play to alpakka-csv's strengths. After poking around, I found some slow spots and optimized them.

#### Before
```
[info] Benchmark       (bsSize)   Mode  Cnt    Score   Error  Units
[info] CsvBench.parse        32  thrpt   10    1.551 ± 0.028  ops/s
[info] CsvBench.parse      1024  thrpt   10   60.279 ± 0.720  ops/s
[info] CsvBench.parse      8192  thrpt   10  117.125 ± 1.138  ops/s
[info] CsvBench.parse     65536  thrpt   10  137.510 ± 3.048  ops/s
```

#### After
```
[info] Benchmark        (bsSize)   Mode  Cnt    Score   Error  Units
[info] CsvBench.parse        32  thrpt   10   78.424 ± 1.351  ops/s
[info] CsvBench.parse      1024  thrpt   10  158.948 ± 3.187  ops/s
[info] CsvBench.parse      8192  thrpt   10  167.617 ± 2.759  ops/s
[info] CsvBench.parse     65536  thrpt   10  170.670 ± 2.462  ops/s
```

### Slow spot 1: Duplicate parsing
On every new input `ByteString`, `parseLine: Option[List[ByteString]]` attempts to parse an entire row. If it doesn't have enough data to complete the row, it rewinds all progress on that row and waits for the next chunk then tries again. This causes duplicate work on every new chunk that continues an existing line. It gets particularly bad when the chunk size is significantly smaller than the row size.

There was already some state in `CsvParser` fields, so I pulled up the method state to class fields as well. It's fully incremental now and doesn't repeat work.

### Slow spot 2: `ByteStrings`/`MultiByteArrayIterator`
Accessing each byte and doing slices is hot inner-loop code and the composites (`ByteStrings`/`MultiByteArrayIterator`) add significant overhead to that.

We can't avoid `ByteStrings.slice(...)` when a field spans multiple chunks, but we can mitigate subsequent slices by calling `dropReadBuffer()` as soon as we complete a field, rather than waiting until we complete the whole line. It should allow `buffer: ByteString` to return to a fast `ByteString1` for all subsequent slices on the row.

Indexing into `ByteStrings.apply(pos)` for each byte is also costly. Even the `MultiByteArrayIterator` produced by `ByteStrings.iterator()` has to do non-free normalization on each `.next(): Byte` call.

An optimization is possible from the observation that `CsvParser` parses each `ByteString` chunk before receiving the next chunk. We can traverse only the single chunk in `parseLine` while keeping the `buffer: ByteStrings` around for slicing which may also need to contain the previous chunk.

Downside is we have to track a bit more state which adds a bit more complexity to the implementation. Upside is at least a 25% improvement across all scenarios, which makes it competitive with OpenCsv.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -103,6 +103,11 @@ lazy val couchbase =
 
 lazy val csv = alpakkaProject("csv", "csv", Dependencies.Csv, whitesourceGroup := Whitesource.Group.Supported)
 
+lazy val csvBench = alpakkaProject("csv-bench", "csvBench", publish / skip := true)
+  .dependsOn(csv)
+  .enablePlugins(JmhPlugin)
+  .disablePlugins(BintrayPlugin, MimaPlugin)
+
 lazy val dynamodb = alpakkaProject("dynamodb", "aws.dynamodb", Dependencies.DynamoDB)
 
 lazy val elasticsearch = alpakkaProject(
--- a/csv-bench/src/main/scala/akka/stream/alpakka/csv/scaladsl/CsvBench.scala
+++ b/csv-bench/src/main/scala/akka/stream/alpakka/csv/scaladsl/CsvBench.scala
@@ -1,0 +1,139 @@
+/*
+ * Copyright (C) 2016-2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.csv.scaladsl
+
+import java.util.concurrent.TimeUnit
+
+import akka.NotUsed
+import akka.actor.ActorSystem
+import akka.stream.ActorMaterializer
+import akka.stream.scaladsl.Source
+import akka.util.ByteString
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra.Blackhole
+
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+
+/**
+ * Measures the time to parse a 1 MB CSV file, consuming the first field of each row.
+ *
+ * ==Using Oracle Flight Recorder==
+ * To record a Flight Recorder file from a JMH run, run it using the jmh.extras.JFR profiler:
+ * > csv-bench/jmh:run -prof jmh.extras.JFR -t1 -f1 -wi 5 -i 10 .*CsvBench
+ *
+ * This will result in flight recording file which you can open and analyze offline using JMC.
+ * Start with "jmc" from a terminal.
+ *
+ * ==Sample benchmark results==
+ * Your results may differ.
+ * Rerun these on YOUR OWN MACHINE before/after making changes.
+ *
+ * {{{
+ * > csv-bench/jmh:run -t1 -f1 -wi 10 -i 10 .*CsvBench
+ * [info] Benchmark        (bsSize)   Mode  Cnt    Score   Error  Units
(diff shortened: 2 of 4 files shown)


ennru:
Thank you!

Great OSS attitude to dive into it when evaluating!

I was aware of slow spot 1 and tried to solve it earlier, but didn't have the time.

build.sbt
```diff
@@ -102,6 +103,14 @@ lazy val couchbase =
                  whitesourceGroup := Whitesource.Group.Supported)
 
 lazy val csv = alpakkaProject("csv", "csv", Dependencies.Csv, whitesourceGroup := Whitesource.Group.Supported)
+  .settings(
+    mimaBinaryIssueFilters += ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.csv.impl.CsvParser#FieldBuilder.this")
```
htmldoug: Seems safe to suppress the removal of a constructor on a private class? Would be great if someone could confirm.

https://github.com/akka/alpakka/blob/0baab3fba160f9832b28eb430a4711b86b2373c6/csv/src/main/scala/akka/stream/alpakka/csv/impl/CsvParser.scala#L81

ennru: These excludes go into files instead. 

Add `csv/src/main/mima-filters/1.0.x.backwards.excludes` with content

```
# Allow changes to impl
ProblemFilters.exclude[Problem]("akka.stream.alpakka.csv.impl.*")
```

instead.

htmldoug:
I tried `@inline` in a few places, but didn't see any improvement beyond the margin of error. In some cases, javap showed that scalac didn't actually inline the methods at all.

Top suggestion from JITWatch was:
```
The call at bytecode 1289 to
Class: akka.stream.alpakka.csv.impl.CsvParser
Member: dropReadBuffer()
was not inlined for reason: 'too big'
The callee method is not 'hot' but is too big to be inlined into the caller method.
Invocations: 8924
Size of callee bytecode: 39
```

One `dropReadBuffer()` stack frame for every csv field doesn't seem terrible. At least it's less frequent than every byte. Here's as much of the compile chain from JITWatch as I could capture on a single monitor: https://user-images.githubusercontent.com/663139/57494232-4c2a1d00-7296-11e9-82c9-47a7b758757f.png

I also tried extracting parts of `churn()` out into smaller methods to give the JIT more freedom to inline selectively, but perf actually got worse.

ennru:
Thank you so much, this is great.

htmldoug:
Thank you, @ennru! The fast review and helpful feedback makes contributing patches fun.

ennru:
You can expect a release with the fastest Alpakka CSV ever next week.

PR #1686, opened 2019-05-09
AWS S3: use a single date/time for signing

ennru:
## Purpose

During AWS request signing the request timestamp and the date in the signed authorization credential should denote the same date, so they should be constructed from a single timestamp.

## References

This is a follow-up to #1678 

## Changes

* Add `ZonedDateTime` to `SigningKey`
* Dirive date in `CredentialScope` from the request date in `SigningKey`
* unrelated: add missing (but impossible) match case in `processCheckIfExistsResponse`

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
@@ -4,8 +4,8 @@
 
 package akka.stream.alpakka.s3.impl
 
-import java.time.{Instant, LocalDate, ZoneOffset}
 import java.net.InetSocketAddress
+import java.time.{Instant, ZoneOffset, ZonedDateTime}
 
 import akka.actor.ActorSystem
 import akka.annotation.InternalApi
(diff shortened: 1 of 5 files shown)


s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -98,10 +98,14 @@ import akka.util.ByteString
...
-  def signingKey(implicit settings: S3Settings) = SigningKey(
-    settings.credentialsProvider,
-    CredentialScope(LocalDate.now(ZoneOffset.UTC), settings.s3RegionProvider.getRegion, "s3")
-  )
+  def signingKey(implicit settings: S3Settings) = {
+    val requestDate = ZonedDateTime.now(ZoneOffset.UTC)
+    SigningKey(
+      requestDate,
+      settings.credentialsProvider,
+      CredentialScope(requestDate.toLocalDate, settings.s3RegionProvider.getRegion, "s3")
```
thadeusb: We are still seeing akka.stream.alpakka.s3.S3Exception: The authorization header is malformed; Invalid credential date. Date is not the same as X-Amz-Date. with alpakka 1.0.1.

I'm wondering, does going back to localDate in the CredentialScope screw things up? Because now requestDate and CredentialScope.date are going to have different dates, causing S3 to reject the request still.

ennru: I do not see how the dates can be different.
From `toLocalDate` Javadoc: "This returns a LocalDate with the same year, month and day as this date-time."

aalosb: Okay I see I misunderstood and thought it would convert it to a date in the "local" timezone. It is working as expected. Thank you for clarifying!

Issue #1649, opened 2019-04-16
Failed: UnixDomainSocketSpec should not be able to connect to a non-existent file


2m:
```
[info] - should not be able to connect to a non-existent file *** FAILED *** (14 milliseconds)
[info]   scala.MatchError: java.util.NoSuchElementException: head of empty stream (of class java.util.NoSuchElementException)
[info]   at docs.scaladsl.UnixDomainSocketSpec$$anonfun$1$$anonfun$apply$mcV$sp$7$$anonfun$apply$15.apply(UnixDomainSocketSpec.scala:177)
[info]   at docs.scaladsl.UnixDomainSocketSpec$$anonfun$1$$anonfun$apply$mcV$sp$7$$anonfun$apply$15.apply(UnixDomainSocketSpec.scala:177)
[info]   at scala.util.Success$$anonfun$map$1.apply(Try.scala:237)
[info]   at scala.util.Try$.apply(Try.scala:192)
[info]   at scala.util.Success.map(Try.scala:237)
[info]   at scala.concurrent.Future$$anonfun$map$1.apply(Future.scala:237)
[info]   at scala.concurrent.Future$$anonfun$map$1.apply(Future.scala:237)
[info]   at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:36)
[info]   at org.scalatest.concurrent.SerialExecutionContext.recRunNow(SerialExecutionContext.scala:120)
```

https://travis-ci.com/akka/alpakka/jobs/192800945#L869

2m:
https://travis-ci.com/akka/alpakka/jobs/197199565

2m:
https://travis-ci.com/akka/alpakka/jobs/198867107

2m:
https://travis-ci.com/akka/alpakka/jobs/199906903

2m:
https://travis-ci.com/akka/alpakka/jobs/201036061

2m:
https://travis-ci.com/akka/alpakka/jobs/200518681

ennru:
https://travis-ci.com/akka/alpakka/jobs/205945075#L614

2m:
https://travis-ci.com/akka/alpakka/jobs/208998431

2m:
https://travis-ci.com/akka/alpakka/jobs/212966921

2m:
https://travis-ci.com/akka/alpakka/jobs/226263604

seglo:
[...]

PR #1688, opened 2019-05-09
Sync grpc version to the one that comes from akka-grpc

2m:
Otherwise coursier fails to resolve dependencies.


2m:

Issue #1665, opened 2019-04-25
SQS stream stops consuming messages


gipeshka:
### Versions used 

Alpakka version: 1.0.0
Akka version: 2.5.17

### Expected Behavior

Processing should go on

### Actual Behavior

What we see after update from M1 to 1.0.0 is that queue consumption stops randomly with no error reported to a supervisor.

To track errors we use a supervision policy in materializer:
```scala
  implicit val materializer: Materializer = ActorMaterializer(
    ActorMaterializerSettings(as).withSupervisionStrategy { error =>
      logger.error("Stream processing has failed", error)

      error match {
        case NonFatal(_) => Supervision.Resume
        case _ => Supervision.Stop
      }
    }
  )
```

The stream consists of a source spawned via `SqsSource.apply`, a grouping ack-flow created via `SqsAckFlow.grouped` and the business logic in between. We use the default `SqsAsyncClient` as spawned via `SqsAsyncClient.create()`. The resulting sink is simply `Sink.ignore`

What we see in thread pool monitoring is the following drop at the moment when consumption stops:

![Screen Shot 2019-04-25 at 16 15 06](https://user-images.githubusercontent.com/1704933/56742595-57992680-6775-11e9-9446-385a96260995.png)

Unfortunately I'm unable to reproduce it locally, failure only happens under high load on production servers, but it usually happens every hour or so.

ennru:
Thank you for reporting. 
I guess we need to revisit the changes between M1 and 1.0.0.

ranchoiver:
I've also had this problem (my comment from another issue: https://github.com/akka/alpakka/issues/1588#issuecomment-475584379)

ennru:
For 1.0-M3 the move to AWS SDK 2.0 in #1450 replaced the source stage with the much simpler implementation we have today: https://github.com/akka/alpakka/blob/docs/v1.0-M2/sqs/src/main/scala/akka/stream/alpakka/sqs/impl/SqsSourceStage.scala

@gipeshka Just to be sure, you DON'T use `closeOnEmptyReceive`, right?

gipeshka:
@ennru yes, we don't `closeOnEmptyReceive`.

Also according to SQS monitoring dashboard we barely have any empty receives at the time when the consumptions stops.

Vict0r7:
I'm experiencing the exact same issue since 1.0.0 too 😕

avdv:
Hi.

[...]

PR #1693, opened 2019-05-10
SQS/SNS: put aws-spi-akka-http dependency in test scope

ennru:
## Purpose

For some reason the `aws-spi-akka-http` dependency was added on Compile scope, but used only in tests. This gave the impression Alpakka SQS and SNS were built on top of it (https://github.com/akka/alpakka/issues/1665#issuecomment-489404667) which they are not by default. But it is possible to use it as shown in https://github.com/akka/alpakka/blob/v1.0.0/sqs/src/test/java/docs/javadsl/SqsSourceTest.java#L94-L104

## References

Detected while reading up on https://github.com/akka/alpakka/issues/1665

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -341,15 +341,15 @@ object Dependencies {
 
   val Sns = Seq(
     libraryDependencies ++= Seq(
-      "com.github.matsluni" %% "aws-spi-akka-http" % AwsSpiAkkaHttpVersion excludeAll ExclusionRule(
-        organization = "com.typesafe.akka"
-      ), // ApacheV2
       "software.amazon.awssdk" % "sns" % AwsSdk2Version excludeAll (ExclusionRule(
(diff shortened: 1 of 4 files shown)


gabfssilva:
I added this lib as compile time in order to use Akka HTTP as the HTTP client implementation for the AWS SDK 2. There's no need to explicitly configure it in AWS SDK that we are using this lib since it uses reflection to load the class, if that's what you are wondering. We could always switch for the default one, which is Netty, but, most people will prefer to use Akka HTTP since they already are using Alpakka and Akka Streams.

ennru:
Thank you for intervening, I wasn't looking close enough.

So `AkkaHttpClient` is configured via the `services` file in that jar. I only found the reference in the Java test.

ennru:
I'll open a new PR to improve the docs to explain which client is used.

Issue #1588, opened 2019-03-18
OutOfMemoryError with AWS SQS


avdv:
### Versions used 
Alpakka version: 1.0-M3

Akka version: 2.5.20

Amazon SQS library: 2.5.8

### Expected Behavior

When starting a SqsSource I would expect that the flow handles a slow consumer, ie. it only pulls new messages when needed.

```scala
val settings = SqsWatcherSettings(
      queue-url,
      SqsSourceSettings()
        .withMaxBatchSize(cfg.get[Int]("max-number-of-messages"))
        .withWaitTimeSeconds(cfg.get[Int]("wait-time-seconds"))
        .withMessageAttributes(cfg.get[Seq[String]]("message-attributes").map(MessageAttributeName.apply).toList)
    )

implicit val sqsClient: SqsAsyncClient = SqsAsyncClient
    .builder()
    .httpClient(new AkkaHttpAsyncHttpService().createAsyncHttpClientFactory().build())
    .build()

val source = SqsSource(queueUrl, settings)
        .withAttributes(supervisionStrategy(resumingDecider))

val sink = SqsAckSink(queueUrl)

source
  .map(...)
  .throttle(100, 1.minute)
  .mapAsync(parallelism = 1)(...)
  .map {
      case (msg, Right(_)) ⇒
        MessageAction.Delete(msg)
      case (msg, _) ⇒
        MessageAction.Ignore(msg)
    }
  .(sink)(Keep.both)
  .run()
```

### Actual Behavior

When starting up the flow, it seems to eagerly pull down all messages in the queue (probably multiple times, when the visibility timeout for messages in flight is reached).

![image](https://user-images.githubusercontent.com/3471749/54533405-54e32e80-498a-11e9-8211-b9584a893945.png)

Note, we were using https://github.com/s12v/akka-stream-sqs before, but tried to switch to alpakka-sqs.

### Relevant logs

![flo](https://user-images.githubusercontent.com/3471749/54532939-58c28100-4989-11e9-9206-0dcec94433da.png)

### Reproducible Test Case

n/a

ennru:
Thank you for trying it out, that is definitely not how it's meant to be.

avdv:
Hi, 

I tried to pinpoint the problem with localstack and the throttling seems to be the culprit...

I have created a queue with 500 messages.

```scala
implicit val system: ActorSystem = ActorSystem()
implicit val mat: Materializer = ActorMaterializer()
[...]

PR #1696, opened 2019-05-13
SQS source: make parallelism configurable

ennru:
## Purpose

Make the number of parallel requests issued by `SqsSource` configurable. Use 1 as default. 

## References

References #1605 

## Changes

* Set the parallelism for AWS client fetching via `SqsSourceSettings` (which fetches up to `maxBatchSize` messages)
* cleanup: limit all `CompletionStage.get` calls with a finite duration

## Background Context

As discussed in https://github.com/akka/alpakka/issues/1665#issuecomment-491521609 the higher parallelism could help to fill the buffer when starting up. (See even https://github.com/akka/alpakka/issues/1588#issuecomment-476105607 for the original discussion).

The `buffer` operator in `SqsSource` makes sure more data is fetched eagerly, even if the stream back-pressures, but does not buffer data as long there is down-stream demand ([buffer operator](https://doc.akka.io/docs/akka/current/stream/operators/Source-or-Flow/buffer.html)).

With slow down-stream processing, this may play badly together with the visibility timeout if messages are buffered too long and show up for other consumers as well.

### DIFF
--- a/docs/src/main/paradox/sqs.md
+++ b/docs/src/main/paradox/sqs.md
@@ -80,7 +80,7 @@ Java
 
 Options:
 
- - `maxBatchSize` - the maximum number of messages to return (see `MaxNumberOfMessages` in AWS docs). Default: 10
+ - `maxBatchSize` - the maximum number of messages to return per request (allowed values 1-10, see `MaxNumberOfMessages` in AWS docs). Default: 10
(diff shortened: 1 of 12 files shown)


ennru:
I made the parallelism configurable instead as there might be some benefit in very high-load scenarios.

Issue #1701, opened 2019-05-14
LogRotatorSink does not terminate when consuming an empty Source


samidalouche:
### Versions used 

Latest master . commit : a0725277321c471242feadabccec2207a941bf47
Same problem with alpakka 1.0.0 / akka 2.5.22

### Expected Behavior

Please describe the expected behavior of the issue, starting from the first action.

```
  "LogRotatorSink" must {

    "work with empty sources" in assertAllStagesStopped {
      val triggerCreator: () => ByteString => Option[Path] = () => { element: ByteString =>
        ??? // never called, doesn't matter
      }

      val rotatorSink: Sink[ByteString, Future[Done]] =
        LogRotatorSink(triggerCreator)

      val fileSizeCompletion = Source.empty[ByteString].runWith(rotatorSink)
      fileSizeCompletion.futureValue shouldBe Done
    }
 ...
}
```

### Actual Behavior

Please provide a description of what actually happens, working from the same starting point.

```
[info] - must work with empty sources *** FAILED *** (5 seconds, 198 milliseconds)
[info]   A timeout occurred waiting for a future to complete. Queried 50 times, sleeping 100000000 nanoseconds between each query. (LogRotatorSinkSpec.scala:95)
[info]   org.scalatest.concurrent.Futures$FutureConcept$$anon$1:
[info]   at org.scalatest.concurrent.Futures$FutureConcept.tryTryAgain$1(Futures.scala:538)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValueImpl(Futures.scala:550)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValueImpl$(Futures.scala:479)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:275)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue(Futures.scala:476)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue$(Futures.scala:475)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:275)
[info]   at docs.scaladsl.LogRotatorSinkSpec.$anonfun$new$3(LogRotatorSinkSpec.scala:95)
[info]   at akka.stream.testkit.scaladsl.StreamTestKit$.assertAllStagesStopped(StreamTestKit.scala:32)
[...]

PR #1703, opened 2019-05-15
Fix: LogRotatorSink does not complete when consuming an empty source

samidalouche:
## Purpose

Fix bug: LogRotatorSink does not terminate when consuming an empty Source

## References

[Related bug report](https://github.com/akka/alpakka/issues/1701)

## Changes

- Complete the stage's promise when`onUpstreamFinish` is called

## Legal info

I signed the CLA


ennru:
Great, thank you!

PR #1708, opened 2019-05-17
Migrate to publish rsync plugin

2m:
## Purpose

PublishRsync sbt auto plugin has been extracted to a separate artifact, so it could be reused across projects.


2m:

Issue #1682, opened 2019-05-06
DynamoDB: InternalServerErrorException: Internal server Error


AyushTiwary:
I am using Alpakka in one of my projects for DynamoDB, and we are facing a problem i.e. any random request is failing with Internal Server Error. The occurrence of facing this exception is once or twice a month.

I am also using DynamoDB in my other projects too where we are not using Alpakka there everything works fine but this is not happening in the case of Alpakka.

One of the example of such exception is as follow:

com.amazonaws.services.dynamodbv2.model.InternalServerErrorException: Internal server error (Service: dynamodb; Status Code: 500; Error Code: InternalServerError; Request ID: randonStringForRequestID)

I have a few questions regarding the Alpakka implementation for DynamoDB.
Are you using one of AWS SDKs?
If Not, then Is Alpakka support self retry mechanism that is by default available if we directly use the AWS SDK for DynamoDB?
As this problem is very common in case of DynamoDB. That for any millisecond, InternalServer errors are expected every once in a while, SDK clients will do automatic retries until the request goes through, to your application requests will just look successful.

PR #1712, opened 2019-05-21
DynamoDB: Issue/1682: Adds retries on DynamoDB failed requests as per the DynamoDB documentation

mavericksid:
# Pull Request Checklist

## Purpose

To add retry mechanism for failed requests in DynamoDB as per the following document:
[Error Handling](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Programming.Errors.html
)

## References

References #1682 

## Changes

- Retry settings to be configured via retry.maximum-attempts, retry.initial-timeout and retry.strategy
- Failures are recovered using a similar pattern as followed in Flow.recoverWithRetries but with the addition of TimerGraphStageLogic. Not really sure if this is even a sane solution since it's pretty new to me, will be looking forward to any suggestions :)

## Background Context

It seemed to be a clean solution without making a lot of changes to the existing codebase.


mavericksid:
Hi @ennru 

Please review the PR once you can spare some time :)

mavericksid:
Hi @2m 

Build is failing on some Cassandra step which does not concern this PR, how to rerun the build?

2m:
Thanks for the PR. We will take a look.

It is not a problem with unrelated test failures. We will still merge it, if the failure is of an unrelated test.

mavericksid:
Hi @2m 

The build failure was fixed.

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/impl/AwsClient.scala
```diff
@@ -94,6 +97,31 @@ private[dynamodb] trait AwsClient[S <: AwsClientSettings] {
...
+
+    opFlow
+      .retryWithBackoff(
+        settings.retrySettings.maximumAttempts,
+        settings.retrySettings.initialRetryTimeout,
+        settings.retrySettings.backoffStrategy, {
+          case _ @(_: InternalServerErrorException | _: ItemCollectionSizeLimitExceededException |
+              _: LimitExceededException | _: ProvisionedThroughputExceededException |
+              _: RequestLimitExceededException) =>
+            Source.single(op).via(opFlow)
```
2m: I think we need to improve the retry logic here. As it currently stands, if the stage fails, the retry will happen with the `op` that was passed into the `flowOp` factory at the moment o the flow creation. So it could happen that quite a few operations went through that flow, and then one failed. The retry will happen with the operation that was initially given during the creation of the flow instead of the one, that failed.

I think we should adapt DynamoDb connector flows to be composable with a retry flow in akka-stream-contrib: https://github.com/akka/akka-stream-contrib/blob/master/src/main/scala/akka/stream/contrib/Retry.scala#L42 instead.

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/impl/AwsClient.scala
```diff
@@ -94,6 +97,31 @@ private[dynamodb] trait AwsClient[S <: AwsClientSettings] {
...
+        }
+
+    opFlow
+      .retryWithBackoff(
+        settings.retrySettings.maximumAttempts,
+        settings.retrySettings.initialRetryTimeout,
+        settings.retrySettings.backoffStrategy, {
+          case _ @(_: InternalServerErrorException | _: ItemCollectionSizeLimitExceededException |
+              _: LimitExceededException | _: ProvisionedThroughputExceededException |
+              _: RequestLimitExceededException) =>
```
2m: It is possible to set supervision strategy on a stream operator as a stream attribute. We should use that instead so users can customize supervision strategies: https://doc.akka.io/docs/akka/current/stream/stream-error.html#supervision-strategies

2m:
Thank you for the initial PR. This has sparked a discussion which has materialized a general RetryFlow stage.

This PR has been superseded by #1896 where DynamoDb connector is changed to compose with a general RetryFlow.

Issue #1662, opened 2019-04-23
Setting tls=true in DynamoSettings overwrites the configured port to 443


Ophirr33:
### Versions used 

Alpakka version: 1.0.0

Akka version: 2.5.22

### Expected Behavior
Using the alpakka client with:
```
    val settings = DynamoSettings(region, host)
      .withPort(8080)
      .withParallelism(5)
      .withMaxOpenRequests(Some(4))
      .withTls(true)
    val client  DynamoClients(settings)(someActorSystem, someMat)
    DynamoDb.source(someOp).withAttributes(DynamoAttributes.client(client)).runWith(Sink.head)
```
will attempt to connect to the specified region/host on port 8080

### Actual Behavior

The client attempts to connect on port 443 instead of port 8080

The issue seems to originate here: https://github.com/akka/alpakka/blob/master/dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/impl/DynamoClientImpl.scala#L41

In the `tls = true` branch, no port is passed to `cachedHostConnectionPoolHttps`, whereas in the `tls = false` branch, the port _is_ passed to `cachedHostConnectionPool`.

ennru:
Thanks for reporting. Would you be able to propose a fix in a PR?

mavericksid:
Hi @ennru,

PR has been raised, please have a look:
https://github.com/akka/alpakka/pull/1722

ennru:
The PR is merged and https://github.com/akka/alpakka/pull/1724 as a follow-up, as well.

PR #1722, opened 2019-05-23
DynamoDb: set port from settings even when tls=true

mavericksid:
## Purpose

Fixes the issue wherein setting tls=true in DynamoSettings overwrites the configured port to 443

## References

References #1662 

## Changes

Passed port to Http().cachedHostConnectionPoolHttps


2m:
Thanks for the fix!

mavericksid:
Glad it helped!

Issue #372, opened 2017-07-01
Update to AWS SDK 2.0


tlockney:
This would only make sense as a POC at this point, since the [new AWS SDK](https://aws.amazon.com/blogs/developer/aws-sdk-for-java-2-0-developer-preview/) is still only in preview, but the new implementation includes a few key features that should make integration with with Akka world more effective:

- non-blocking IO; specifically leveraging Java's CompletableFuture
- AsyncRequestProvider, which is a reactive streams Publisher, for supplying inputs in a way that supports backpressure
- a pluggable HTTP layer, which allows for usage of Akka HTTP for client operations reducing concerns around resource management and providing more seamless interop in Akka-based systems

areyouspiffy:
> AsyncRequestProvider, which is a reactive streams Publisher, for supplying inputs in a way that supports backpressure

How cool is that? I'd be happy to try it out on the `Kinesis` source--it has a pretty minimal surface area so it wouldn't be much trouble.

joearasin:
This looks like a fantastic idea -- I don't think the approach I took on the S3 uploader is sustainable, and being able to offload as much of the AWS integration to Amazon would be great for keeping things maintainable.

joost-de-vries:
@raboof Arnout, this is also relevant for the label `p:dynamodb`...

francisdb:
Don't think we have to wait until this is out of preview?

joost-de-vries:
Relevant issues: [reactive streams api for dynamodb streams](https://github.com/aws/aws-sdk-java-v2/issues/93), [use TCK to test RS compliance](https://github.com/aws/aws-sdk-java-v2/issues/406)

ktoso:
I would very much welcome help to finish that PR as I'm overloaded with other "urgent" things :(
Will try to get it done of course, but help would be tremendously useful

joost-de-vries:
@ktoso [here](https://github.com/joost-de-vries/aws-sdk-java-v2/tree/wip-show-how-to-use-TCK-test-a-few-types) you go.

ktoso:
Awesome! Let me have a look there :)

jtjeferreira:
AWS SDK 2.0 is now GA https://aws.amazon.com/blogs/developer/aws-sdk-for-java-2-x-released/

gabfssilva:
[...]

PR #1725, opened 2019-05-25
DynamoDb: Update to AWS SDK 2.0

RustedBones:
# Pull Request Checklist

## Purpose

Use the AWS java SDK v2 for dynamodb connector

## References

References #372 #1345

## Changes

- `source` is only made available for paginated request (user should use `single` for other requests instead)

## Background Context

- Change `AwsOps` to implicit objects instead of classes. Make use of type parameters to pass them as implicits when building streams.

## Comment

Documentation is not yet updated. Waiting for code approval before doing it.


ennru:
Great that you started to work on this migration. It can't be done without breaking the API, of course, so we'll need to keep this until Alpakka 1.1 is approaching.

ennru:
@RustedBones 	Now is a great time to move to AWS SDK 2. It would be great if you could re-vive this PR to do it.

RustedBones:
Looking again on the PR, I'm wondering if it is OK that the java API for operation is now with capital letter. eg:
```scala
AwsOp.CreateTable()
```
To avoid the boilerplate
```scala
/** Java API */
def createTable = CreateTable
```

I was thinking of renaming the implicit vals in the `AwsOp` to lowercase even if it does not follow the scala naming convention

RustedBones:
For the MIMA, should I create a `dynamodb/src/main/mima-filters/2.0.0.backwards.excludes` with
```
# Allow breaking changes
ProblemFilters.exclude[Problem]("akka.stream.alpakka.dynamodb.*")
```

ennru:
It should be in `1.1.x.backwards.excludes`.
A bit more narrow would be nicer.

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/AwsOp.scala
```diff
@@ -4,241 +4,64 @@
...
-
-  /** Java Api */
-  def create(r: UpdateTimeToLiveRequest): UpdateTimeToLive = r
+/**
+ * Representation on an AWS dynamodb sdk operation
+ * @param sdkExecute function to be executed on the AWS client
+ * @tparam In dynamodb request type
+ * @tparam Out dynamodb response type
+ */
+sealed class AwsOp[In <: DynamoDbRequest, Out <: DynamoDbResponse](
```
2m: Since we are changing the structure of the connector quite a bit, it is good time to rename this as well. I think renaming this to `DynamoDbOp` is a better match. WDYT?

dynamodb/src/test/scala/akka/stream/alpakka/dynamodb/ItemSpec.scala
```diff
@@ -20,59 +26,67 @@ class ItemSpec extends TestKit(ActorSystem("ItemSpec")) with AsyncWordSpecLike w
...
+    super.afterAll()
   }
 
   "DynamoDB" should {
 
     import ItemSpecOps._
 
     "1) list zero tables" in assertAllStagesStopped {
-      DynamoDb.single(listTablesRequest).map(_.getTableNames.asScala.count(_ == tableName) shouldBe 0)
+      DynamoDb.single(listTablesRequest).map(_.tableNames().asScala shouldBe empty)
```
2m: Here and below `()` can be omitted in Scala. In fact usually they are omitted for methods without parameters, if the method does not side-effect.

dynamodb/src/test/scala/docs/scaladsl/ExampleSpec.scala
```diff
@@ -31,83 +35,68 @@ class ExampleSpec
...
-      val listTablesResult: Future[ListTablesResult] =
-        DynamoDb.single(new ListTablesRequest())
+      val listTablesResult: Future[ListTablesResponse] =
+        DynamoDb.single(ListTablesRequest.builder().build())
       //##simple-request
 
       listTablesResult.futureValue
     }
 
     "allow multiple requests - explicit types" in assertAllStagesStopped {
```
2m: The ` - explicit types` is not relavant in the test name anymore, as that is not required in this improved API.

dynamodb/src/test/scala/docs/scaladsl/ExampleSpec.scala
```diff
@@ -31,83 +35,68 @@ class ExampleSpec
...
-        .via(DynamoDb.flow)
-        .map(_.getTable.getItemCount)
-      val streamCompletion = source.runWith(Sink.seq)
-      streamCompletion.failed.futureValue shouldBe a[AmazonDynamoDBException]
+      Source
+        .single(CreateTableRequest.builder().tableName("testTable").build())
+        .via(DynamoDb.flow(1))
+        .map(result => DescribeTableRequest.builder().tableName(result.tableDescription().tableName()).build())
+        .via(DynamoDb.flow(1))
+        .map(_.table().itemCount())
```
2m: I really like the lack of the typo annotations in the new API.

dynamodb/src/test/java/docs/javadsl/ExampleTest.java
```diff
@@ -59,41 +66,26 @@ public void checkForStageLeaks() {
...
   public void allowMultipleRequests() throws Exception {
     // #flow
     Source<String, NotUsed> tableArnSource =
-        Source.single(AwsOp.create(new CreateTableRequest().withTableName("testTable")))
-            .via(DynamoDb.flow())
-            .map(r -> (CreateTableResult) r)
-            .map(result -> result.getTableDescription().getTableArn());
+        Source.single(CreateTableRequest.builder().tableName("testTable").build())
+            .via(DynamoDb.flow(client, AwsOp.createTable(), 1))
+            .map(result -> result.tableDescription().tableArn());
```
2m: This should be expanded to actually demonstrated multiple requests, according to the test name and as in Scala spec:

```
  .map(result -> DescribeTableRequest.builder().tableName(result.tableDescription().tableName()).build())
  .via(DynamoDb.flow(client, AwsOp.describeTable(), 1))
```

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/AwsOp.scala
RustedBones: I also prefer `DynamoDbOp` considering the bounds on the type parameters

dynamodb/src/test/scala/akka/stream/alpakka/dynamodb/ItemSpec.scala
RustedBones: Sure. My IDE auto completes with parenthesis on java methods. Will clean this

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/AwsOp.scala
RustedBones: What about defining the ops in a trait and mixing it in both `scaladsl.DynamoDb` and `javadsl.DynamoDb` so `DynamoDb` is the only thing to be imported by the users?

2m: Interesting idea.

But then the `{java,scala}dsl.DynamoDb` namespace would get two groups of quite different members: Akka Stream factories and DynamoDb operations themselves. That might get distracting when using autocomplete in `DynamoDb` where the `single`, `source` and `flow` methods, that are API entry methods, will be scattered accross all DynamoDb operation values.

2m:
The latest CI validation failure is because we have enabled fatal warnings on master. Rebasing this PR on the top of current master should fix that.

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/AwsOp.scala
RustedBones: Legit. I'll leave it in `DynamoDbOp`

2m:
Unrelated failures were #1417 and #1897

PR #1730, opened 2019-05-28
Finish move of CSV example; use Committer in pass-through examples

ennru:
## Purpose

* The self-contained example "Fetch CSV from Internet and publish the data as JSON to Kafka" is now part of https://github.com/akka/alpakka-samples This PR drops the code from Alpakka
* The pass-through examples with Kafka showed the old API, this PR makes it use the Committer


2m:
Failure was unrelated #1735

Issue #838, opened 2018-03-14
S3 Connector - Using `+` (plus) symbol causes mismatched request signature


SeanCheatham:
I noticed a couple other closed issues indicating errors when uploading files with either an `@` symbol or a `:` symbol, so I apologize if I should've posted in one of those instead of creating a new one.  There appears to be similar behavior when uploading a file with a `+` (plus) symbol in its name.

This code was run using Alpakka S3 0.17.

Sample Code:
```
val client: S3Client = ???
Source.single(ByteString("Test"))
  .runWith(
    client.multipartUpload("test-bucket", "Special+Character.txt")
  )
```

Produces the following error:
```
The request signature we calculated does not match the signature you provided. Check your key and signing method.
akka.stream.alpakka.s3.S3Exception: The request signature we calculated does not match the signature you provided. Check your key and signing method. (Code: SignatureDoesNotMatch, RequestID: , HostID: )
	at 
akka.stream.alpakka.s3.impl.S3Stream.$anonfun$initiateMultipartUpload$4(S3Stream.scala:216)
```

SeanCheatham:
For reference, [this document](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-keys) lists the characters that are generally safe, characters that require special encoding, and characters that should generally be avoided.

tuleism:
This seems to be the problem that S3 interpretes plus sign as space (see https://forums.aws.amazon.com/thread.jspa?messageID=722673). S3 interpretes `Special+Character.txt` as `Special Character.txt`, while the generated Canonical Request is for `Special+Character.txt`, which leads to mismatched signature error.

I see the way to fix this is to use `Raw-Request-URI` header and manually replace the plus sign by `%2B` (as mentioned in https://github.com/akka/akka-http/issues/451). Any thoughts?

SeanCheatham:
Ah that's interesting indeed!  And a bit disappointing that the issue has existed in S3 for nearly a decade now.

It sounds to me like modifying https://github.com/akka/alpakka/blob/master/s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala#L103 would be a good place to start.

Could possibly modify that method to declare a base HttpRequest:
1. Check to see if the Request Uri contains a `+` symbol
2. If so, stringify it and replace instances of `+` with `%2B`
3. Attach a Raw-Request-URI header to the base Request

[...]

PR #1738, opened 2019-05-31
AWS S3: Fix plus char usage in s3 key

gipeshka:
## Purpose

Fixes issue with using `+` char in a S3 object key

## References

https://github.com/akka/alpakka/issues/838 S3 Connector - Using `+` (plus) symbol causes mismatched request signature

## Changes

- HttpRequests adds a synthetic akka http header `Raw-Request-URI` to the built request. The header contains stringified uri of a request with all `+` characters replaced with the url-encoded representation `%2B`.
- Signer now ignores a synthetic akka http header `Raw-Request-URI`.

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/auth/Signer.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/auth/Signer.scala
@@ -10,7 +10,7 @@ import java.time.ZonedDateTime
 
 import akka.NotUsed
 import akka.annotation.InternalApi
-import akka.http.scaladsl.model.headers.{RawHeader, `Raw-Request-URI`}
+import akka.http.scaladsl.model.headers.{`Raw-Request-URI`, RawHeader}
 import akka.http.scaladsl.model.{HttpHeader, HttpRequest}
 import akka.stream.scaladsl.Source
 
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/auth/Signer.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/auth/Signer.scala
@@ -10,7 +10,7 @@ import java.time.ZonedDateTime
 
 import akka.NotUsed
 import akka.annotation.InternalApi
-import akka.http.scaladsl.model.headers.{`Raw-Request-URI`, RawHeader}
+import akka.http.scaladsl.model.headers.RawHeader
 import akka.http.scaladsl.model.{HttpHeader, HttpRequest}
 import akka.stream.scaladsl.Source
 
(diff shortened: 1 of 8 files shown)


s3/src/main/scala/akka/stream/alpakka/s3/impl/auth/Signer.scala
```diff
@@ -25,7 +25,9 @@ import akka.stream.scaladsl.Source
         val headersToAdd = Vector(RawHeader("x-amz-date", key.requestDate.format(dateFormatter)),
                                   RawHeader("x-amz-content-sha256", hb)) ++ sessionHeader(key)
         val reqWithHeaders = request.withHeaders(request.headers ++ headersToAdd)
-        val cr = CanonicalRequest.from(reqWithHeaders)
+        val cr = CanonicalRequest.from(
+          reqWithHeaders.withHeaders(reqWithHeaders.headers.filterNot(_.isInstanceOf[`Raw-Request-URI`]))
```
ennru: It would be more expected to do the filtering directly on the headers in line 27.

gipeshka: that's the trick here, the request to build a signature differs from the original request, so we can skip the raw-request-uri synthetic header.

gipeshka: should this go to `CanonicalRequest` itself? `CanonicalRequest.from` could filter out known synthetic headers before calling `*HeaderString` methods.

aparra: I think you can filter out un-rendered headers inside of `CanonicalRequest`:

```scala
  def from(request: HttpRequest): CanonicalRequest = {
    val hashedBody =
      request.headers
        .collectFirst { case header if header.is("x-amz-content-sha256") => header.value }
        .getOrElse("")

    val renderedHeaders = request.headers.filterNot(_.is(`Raw-Request-URI`.lowercaseName))

    CanonicalRequest(
      request.method.value,
      pathEncode(request.uri.path),
      canonicalQueryString(request.uri.query()),
      canonicalHeaderString(renderedHeaders),
      signedHeadersString(renderedHeaders),
      hashedBody
    )
  }
```
Then you could have a spec like this one:

```scala
  it should "correctly build a canonicalString when `Raw-Request-URI` header is present" in {
    val req = HttpRequest(
      HttpMethods.GET,
      Uri("https://s3-eu-central-1.amazonaws.com/my.test.bucket/file+name.txt")
    ).withHeaders(
      RawHeader("x-amz-content-sha256", "testhash"),
      `Content-Type`(ContentTypes.`application/json`),
      `Raw-Request-URI`("/my.test.bucket/file%2Bname.txt")
    )
    val canonical = CanonicalRequest.from(req)
    canonical.canonicalString should equal(
      """GET
        |/my.test.bucket/file%2Bname.txt
        |
        |content-type:application/json
        |x-amz-content-sha256:testhash
        |
        |content-type;x-amz-content-sha256
        |testhash""".stripMargin
    )
  }
```

ennru:
Thank you for fixing it!

Issue #1740, opened 2019-06-05
SQS connector produces plenty of empty receives


gipeshka:
### Versions used 

1.0.1

Akka version: 

2.5.17

### Expected Behavior

As of version 1.0.0-M1 the behaviour of Alpakka sqs connector was to reduce the amount of requests sent to SQS in case there is an empty receive.

### Actual Behavior

Current implementation doesn't scale down the consumption on empty receives. Users switching to the new version might experience huge increase in AWS costs.

ennru:
Thank you for reporting, that seems to have been lost in the SQS make-over.

RustedBones:
In `1.0.0-M1`, the concurrency was decreased to 1 in case of empty receive. See [here](https://github.com/akka/alpakka/blob/960f49a1a654b80b38ee83f2a9677785a8df17e4/sqs/src/main/scala/akka/stream/alpakka/sqs/impl/SqsSourceStage.scala#L82).

Now, the [`mapConcat`](https://github.com/akka/alpakka/blob/c137fcf1c174e0685546d14699f9bdca876413b4/sqs/src/main/scala/akka/stream/alpakka/sqs/scaladsl/SqsSource.scala#L48) from the source will request messages from upstream in a greedy fashion. Instead of having 1 request per `waitTimeSeconds`, we will have `parallelRequests` in the same duration.

Adding a `throtle` stage above the `mapConcat`  that delays in case of empty response could maybe solve the issue.

RustedBones:
@gipeshka did you set a `parallelRequests` > 1 ?

gipeshka:
@RustedBones yes, it's set to about 30

RustedBones:
@gabfssilva was partially right in his [comment](https://github.com/akka/alpakka/issues/1588#issuecomment-476070078).

> And now I see that there's kind of a problem with the approach I took by using pre-defined stages since `mapAsync` is executed eagerly and more messages are getting fetched than the buffer predefined. (even with your fix, @RustedBones, tho it fixes the non stopping polling)
> 
> Also, since the `mapAsync` is stateless, it cannot know how many messages it can poll from SQS based on the size of the current buffer, it always fetches based on `maxBatchSize` setting. So, even if the buffer is set to 20 and the stream has 19 elements, it'll try to fetch more 10 messages, which is not what should happen.
> 
[...]

PR #1743, opened 2019-06-06
AWS SQS source: reduce consumption on empty receives

gipeshka:
## Purpose

This PR is supposed to provide functionality of quick feedback to a throttling stage in SQSSource to reduce amount of requests to SQS when there is nothing to consume.

## References

References #1740

## Changes

Added a `BalancingMapAsyncUnordered ` stage, that wraps the call to SQS. When received empty SQS response parallelism scales down to 1 until there is a non-empty response.
Added specs for some basic cases and the new one.

## Background Context

Inspired by https://github.com/akka/alpakka/blob/960f49a1a654b80b38ee83f2a9677785a8df17e4/sqs/src/main/scala/akka/stream/alpakka/sqs/impl/SqsSourceStage.scala


sqs/src/main/scala/akka/stream/alpakka/sqs/ControlledThrottling.scala
```diff
@@ -0,0 +1,56 @@
...
+
+import java.time.Clock
+
+import akka.stream.Attributes
+import akka.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
+import akka.stream.stage.{GraphStageLogic, InHandler, OutHandler, TimerGraphStageLogic}
+
+import scala.concurrent.duration.FiniteDuration
+
+class ControlledThrottling[T](clock: Clock, delay: FiniteDuration) extends SimpleLinearGraphStage[T] {
```
ennru: This helper stage should move to the `impl` package and be marked as `InternalApi`.

sqs/src/main/scala/akka/stream/alpakka/sqs/ControlledThrottling.scala
```diff
@@ -0,0 +1,56 @@
...
+  private var isThrottling: Boolean = false
```
ennru: This is mutable state which would be shared by all streams using this stage. You are not allowed to have any mutable state here. We need to find a different implementation to achieve this feedback loop. I guess reaching out to the graph DSL is appropriate.

sqs/src/main/scala/akka/stream/alpakka/sqs/scaladsl/SqsSource.scala
```diff
@@ -25,7 +28,10 @@ object SqsSource {
   def apply(
       queueUrl: String,
       settings: SqsSourceSettings = SqsSourceSettings.Defaults
-  )(implicit sqsClient: SqsAsyncClient): Source[Message, NotUsed] =
+  )(implicit sqsClient: SqsAsyncClient): Source[Message, NotUsed] = {
+    val controlledThrottling =
+      new ControlledThrottling[ReceiveMessageRequest](Clock.systemDefaultZone(), settings.waitTimeSeconds.seconds)
```
ennru: Would there be a reason to introduce another setting, to configure it independently of `waitTimeSeconds`?

sqs/src/main/scala/akka/stream/alpakka/sqs/impl/AutoBalancingSqsReceive.scala
```diff
@@ -0,0 +1,32 @@
...
+
+import akka.stream.FlowShape
+import akka.stream.scaladsl.{Broadcast, Flow, GraphDSL, Merge, Partition}
+import software.amazon.awssdk.services.sqs.SqsAsyncClient
+import software.amazon.awssdk.services.sqs.model.{Message, ReceiveMessageRequest, ReceiveMessageResponse}
+
+import scala.collection.JavaConverters._
+import scala.compat.java8.FutureConverters._
+
+class AutoBalancingSqsReceive(parallelism: Int)(implicit sqsClient: SqsAsyncClient) {
```
gipeshka: hey @ennru 
I've put another way to solve the same problem.

This one uses routing to a reduced parallelism async stage instead of throttling, you might find this one cleaner.

sqs/src/main/scala/akka/stream/alpakka/sqs/ControlledThrottling.scala
gipeshka: I am not sure I understand why this is going to be reused, the idea I had was that anyone who uses this class spawns a new instance of `ControlledThrottling` and therefore the state is not shared

sqs/src/main/scala/akka/stream/alpakka/sqs/scaladsl/SqsSource.scala
gipeshka: Fair point. I think that in case of using throttling it might make more sense to have a configuration to switch "autothrottling" off, so that clients that are interested in low latencies could still use the library in the same way it works now.

sqs/src/main/scala/akka/stream/alpakka/sqs/ControlledThrottling.scala
ennru: This problem is the difference between creating the class and materialising the stream. You may create the source and even a whole stream and materialise it multiple times.

gipeshka: You're right, I've created a custom stage (which is basically autobalancing `MapAsyncUnordered`).

sqs/src/main/scala/akka/stream/alpakka/sqs/impl/BalancingMapAsyncUnordered.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) with InHandler with OutHandler {
+      override def toString = s"BalancingMapAsyncUnordered.Logic(inFlight=$inFlight, buffer=$buffer)"
+
+      val decider = inheritedAttributes.mandatoryAttribute[SupervisionStrategy].decider
+
+      private var inFlight = 0
+      private var buffer: BufferImpl[Out] = _
+      private var parallelism = initialParallelism
```
2m: Great addition to the `mapAsyncUnordered`. I think we should drop `initialParallelism` from the arguments of this stage, as whatever it is, it is reset after first Future completes. So we could have it hardcoded to 1 and still have the same bahaviour.

sqs/src/main/scala/akka/stream/alpakka/sqs/scaladsl/SqsSource.scala
```diff
@@ -42,7 +43,13 @@ object SqsSource {
...
         }
       }
-      .mapAsync(settings.parallelRequests)(sqsClient.receiveMessage(_).toScala)
+      .via(
+        BalancingMapAsyncUnordered[ReceiveMessageRequest, ReceiveMessageResponse](
+          settings.parallelRequests,
+          sqsClient.receiveMessage(_).toScala,
+          response => if (response.messages().isEmpty) 1 else settings.parallelRequests
+        )
+      )
```
2m: This changes the semantics from `mapAsync` to `mapAsyncUnordered` which means that messages received with an later `sqsClient.receiveMessage` call might overtake messages received with an earlier `sqsClient.receiveMessage` call in the stream.

sqs/src/main/scala/akka/stream/alpakka/sqs/impl/BalancingMapAsyncUnordered.scala
gipeshka: yeah, slow start might make sense.

After some thought I also want to change the signature of balancing function to (Out, Int) => Int, where the second argument is the current value of parallelism, wdyt? So one can increase parallelism gradually, which might be useful for something like a circuit-breaker case.

gipeshka: however I still need maxParallelism parameter to initialise the buffer inside the stage

sqs/src/main/scala/akka/stream/alpakka/sqs/scaladsl/SqsSource.scala
```diff
@@ -43,15 +44,24 @@ object SqsSource {
...
       .takeWhile(messages => !settings.closeOnEmptyReceive || messages.nonEmpty)
       .mapConcat(identity)
       .buffer(settings.maxBufferSize, OverflowStrategy.backpressure)
+
+  private def resolveHandler(parallelism: Int)(implicit sqsClient: SqsAsyncClient) = {
+    if (parallelism == 1) {
+      MapAsyncUnordered[ReceiveMessageRequest, ReceiveMessageResponse](
+        parallelism,
+        sqsClient.receiveMessage(_).toScala
+      )
```
2m: You can replace this with `Flow[ReceiveMessageRequest].mapAsyncUnordered(parallelism)(sqsClient.receiveMessage(_).toScala)` and then you will not need to import `MapAsyncUnordered` from the `impl` package.

sqs/src/main/scala/akka/stream/alpakka/sqs/impl/BalancingMapAsyncUnordered.scala
2m: Balancing function that takes previous parallelism sounds good!

> however I still need maxParallelism parameter to initialise the buffer inside the stage

Ahh, thats right. What is more, the balancing function should not return larger parallelism than `maxParallelism`. But since this is not user facing API, it can be left like this for now.

gipeshka: > balancing function should not return larger parallelism than `maxParallelism`
Yeah, I also thought about it

I'll create a PR to akka-streams later this week and will address this point there.

2m: Sounds good. Create an issue first, with your proposal, so any interested parties could have a discussion before the code gets written. :)

PR #1746, opened 2019-06-07
Couchbase: Akka Discovery support

ennru:
## Purpose

Add support for use of Akka Discovery to configure Couchbase nodes.

## Changes

* Add `enrichAsync` to `CouchbaseSessionSettings`
* Apply `enrichAsync` before using settings
* Test case with `DiscoverySupport` usage
* Documentation

### DIFF
--- a/couchbase/src/main/java/akka/stream/alpakka/couchbase/javadsl/DiscoverySupport.java
+++ b/couchbase/src/main/java/akka/stream/alpakka/couchbase/javadsl/DiscoverySupport.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (C) 2016-2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.couchbase.javadsl;
+
+import akka.actor.ActorSystem;
+import akka.stream.alpakka.couchbase.CouchbaseSessionSettings;
+import com.typesafe.config.Config;
+
+import java.util.concurrent.CompletionStage;
+
+/**
+ * Utility to delegate Couchbase node address lookup to [[https://doc.akka.io/docs/akka/current/discovery/index.html Akka Discovery]].
+ */
+public final class DiscoverySupport {
+
+  private static final akka.stream.alpakka.couchbase.scaladsl.DiscoverySupport SUPPORT =
+      akka.stream.alpakka.couchbase.scaladsl.DiscoverySupport.INSTANCE();
+
+  /**
+   * Expects a `service` section in the given Config and reads the given service name's address to
+   * be used as Couchbase `nodes`.
+   */
+  public static final java.util.function.Function<
+          CouchbaseSessionSettings, CompletionStage<CouchbaseSessionSettings>>
+      getNodes(Config config, ActorSystem system) {
+    return SUPPORT.getNodes(config, system);
+  }
+
+  /**
+   * Expects a `service` section in the given Config and reads the given service name's address to
+   * be used as Couchbase `nodes`.
+   */
+  public static final java.util.function.Function<
+          CouchbaseSessionSettings, CompletionStage<CouchbaseSessionSettings>>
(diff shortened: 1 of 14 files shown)


couchbase/src/main/scala/akka/stream/alpakka/couchbase/model.scala
```diff
@@ -172,14 +180,38 @@ final class CouchbaseSessionSettings private (val username: String,
...
+  /** Scala API:
+   * Allows to provide an asynchronous method to update the settings.
+   */
+  def withEnrichAsync(value: CouchbaseSessionSettings => Future[CouchbaseSessionSettings]): CouchbaseSessionSettings =
+    copy(enrichAsync = value)
+
+  /** Java API:
+   * Allows to provide an asynchronous method to update the settings.
+   */
+  def withEnrichCompletionStage(
```
2m: Maybe `withEnrichAsyncCs`

2m:
Needs a rebase.

couchbase/src/main/scala/akka/stream/alpakka/couchbase/scaladsl/DiscoverySupport.scala
```diff
@@ -0,0 +1,92 @@
...
+   * Use Akka Discovery to read the addresses for `serviceName` within `lookupTimeout`.
+   */
+  private def readNodes(
+      serviceName: String,
+      lookupTimeout: FiniteDuration
+  )(implicit system: ActorSystem): Future[immutable.Seq[String]] = {
+    import system.dispatcher
+    val discovery = Discovery(system).discovery
+    discovery.lookup(serviceName, lookupTimeout).map { resolved =>
+      resolved.addresses.map(_.host)
```
ignasi35: This is dropping the port and favoring host before `InetAddress`. I'm guessing the `port` could still be necessary?

Put another way, I think the typing of the return type could be stronger. OTOH, if there's a reason to drop all and just keep `host` then I think a comment with that reason would be great.

couchbase/src/main/java/akka/stream/alpakka/couchbase/javadsl/DiscoverySupport.java
```diff
@@ -0,0 +1,44 @@
...
+    return SUPPORT.getNodes(config, system);
+  }
+
+  /**
+   * Expects a `service` section in the given Config and reads the given service name's address to
+   * be used as Couchbase `nodes`.
+   */
+  public static final java.util.function.Function<
+          CouchbaseSessionSettings, CompletionStage<CouchbaseSessionSettings>>
+      getNodes(ActorSystem system) {
```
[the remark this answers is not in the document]
ennru: That's automated.

couchbase/src/main/scala/akka/stream/alpakka/couchbase/model.scala
```diff
@@ -126,7 +129,7 @@ object CouchbaseSessionSettings {
    * Scala API:
    */
   def apply(username: String, password: String): CouchbaseSessionSettings =
-    new CouchbaseSessionSettings(username, password, Nil, None)
+    new CouchbaseSessionSettings(username, password, Nil, None, EnrichAsync)
```
johanandren: Why not just `None, Future.successful)`?

couchbase/src/main/scala/akka/stream/alpakka/couchbase/scaladsl/DiscoverySupport.scala
johanandren: I think the port isn't really used with Couchbase because it opens up a whole set of ports on each cluster node.

couchbase/src/main/scala/akka/stream/alpakka/couchbase/model.scala
ennru: That's better, thank you.

PR #1752, opened 2019-06-15
Add `withProperties` to all jms message types

kubukoz:


kubukoz:
force pushing to trigger CI, apparently some test was flaky

PR #1760, opened 2019-06-19
HDFS: Unit-test target file overwrite

markarasev:
## Purpose

Adds a unit test for HDFS target file overwrite.

## References

References #1761


2m:
This is superseded by #1834 where the same testcase and a fix is implemented.

Issue #793, opened 2018-02-19
Jackson Databind vulnerability


johanandren:
There are a couple of quite serious vulnerabilities (potentially remote code execution) in jackson-databind:

* https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-15095 - fixed in 2.8.10, 2.9.1
* https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-17485 - fixed in 2.8.10, 2.9.3
* https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-7525 - fixed in 2.6.7.1, 2.7.9.1 and 2.8.9

Among the Alpakka modules older versions are pulled in (directly or transitively) by:
* kinesis - 2.6.7.1
* elastic search - 2.9.1
* geode - 2.9.2
* orientdb - 2.6.0
* s3 - 2.6.7.1
* sns - 2.6.7.1
* sqs - 2.6.7.1
* awslambda - 2.6.7.1
* dynamodb - 2.6.7.1

I think all these needs to be updated

takezoe:
I investigated the status of Jackson dependency in each connector. The summary is as follows:

## elasticsearch

Upgraded in #804

## kinesis, s3, sns, sqs, awslambda, dynamodb

Jackson dependency in these connectors are comming from aws-sdk-java. I sent a pull request to aws-sdk-java to upgrade Jackson: https://github.com/aws/aws-sdk-java/pull/1480

However aws-sdk-java have to support Java 6 and some of fixes for Jackson vulnerabilities have not been backported to Jackson 2.6.x which is for Java 6. Hence, we might have to overwrite Jackson dependency in Alpakka side.

## azure-storage-queue

Same as aws-sdk-java in the point of Java 6 support. I created an issue instead of sending a pull request: https://github.com/Azure/azure-storage-java/issues/264

## geode

Jackson was upgraded in develop branch recently: https://github.com/apache/geode/commit/c076bef6471f43e227531c4829dd90b32931dd23#diff-1fa89afcc2bb350e76df4a548d82c912R46
We can wait for the next geode release.

## orientdb

I sent a pull request to orientdb repository: https://github.com/orientechnologies/orientdb/pull/8118

Note: orientdb has dependency to blueprints and blueprints has a dependency to old version of Jackson. But I guess it will be evicted by the latest version of Jackson if above pull request is merged.

ennru:
[...]

PR #1777, opened 2019-06-25
Override Jackson Databind dependency to 2.9.9

ennru:
## Purpose

Make sure all Alpakka connectors use the Jackson Databind version 2.9.9 to have fixes for the known security issues.

## References

See #793


project/Dependencies.scala
```diff
@@ -169,6 +169,8 @@ object Dependencies {
           .map("org.apache.geode" % _ % GeodeVersion exclude ("org.slf4j", "slf4j-log4j12")) ++
         Seq(
           "com.chuusai" %% "shapeless" % "2.3.3",
+          // overriding version from geode to avoid some security vulnerabilities
+          "com.fasterxml.jackson.core" % "jackson-databind" % "2.9.9",
```
2m: Should we extract the version to a val?

ennru:

PR #1780, opened 2019-06-26
Fix link in CSV release notes

ennru:
Remove extra char breaking the link address.


2m:
Merging this one without validation, as Travis seems to not be in the best of a relantionship with Maven Central.

Issue #1786, opened 2019-06-27
Google Cloud PubSub PubSubMessage.messageId should be optional


LGLO:
As per REST API reference: https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage
`messageId` property should not be set when publishing.

I understand it's a breaking change.

I think I can provide fix that will make `PubSubMessage.messageId` optional or introduce separate case class for publishing that doesn't have `messageId` and `publishTime` fields.

2m:
In the generated code of the Alpakka Google Cloud Pub/Sub gRPC connector the logic is to check if the `messageId` string is not empty:

```
val __v = messageId
if (__v != "") {
  _output__.writeString(3, __v)
}
```

We could go with the same approach, where JsonWriter would not write out `messageId` if it is empty.

And then for nicer user API, we could add `PubSubMessage.apply/create` methods that do not take `messageId` or `publishTime`.

LGLO:
According to API specs, "messageId" should never be present when publishing. PubSub could start rejecting messages, although it would go against robustness principle.

Maybe we can make this two step? In the first step I'll add functions to create a `PubSubMessage` without `messageId` and `publishTime` and remove these fields from JSON when publishing.

Then we could change `messageId` to `Option[String]` when some breaking changes are going in anyway? WDYT?

2m:
Two step process sounds good. Then we can release non-breaking changes sooner, and API breaking changes will have to wait next major release.

LGLO:
I've opened https://github.com/akka/alpakka/pull/1795 could you take a look?
On Monday I'll re-learn MiMa and check if it breaks binary compatibility or not.

2m:
Change looks good to me and MiMa.

When you will be tackling step 2, it could be done together with https://github.com/akka/alpakka/issues/1797 since it will break binary API as well.

LGLO:
Thank you! I've updated `@deprecated` with `1.1.0`.

ennru:
Fixed with #1795

PR #1795, opened 2019-06-29
Issue 1786 - 'messageId' and 'publishTime' should not be published

LGLO:
Nor should be 'messageId' be required when creating PubSubMessage for publishing.

## Purpose
Improve Scala and Java API - don't require redundant parameter
Improve compliance with Google Cloud PubSub HTTP API - don't send fields that shouldn't be sent.

## References

References #1786

## Changes

* Added Scala and Java api that doesn't require 'messageId'
* Do not encode 'messageId' and 'publishTime' when sending JSON body

## Background Context
I wanted to create separate case class for publishing PubSub message but I went in a direction to not break code.

If for 1.1 we can break source and binary compatibility, I'd like to contribute other change.

### DIFF
--- a/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/model.scala
+++ b/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/model.scala
@@ -80,11 +80,12 @@ object PubSubConfig {
 /**
  *
  * @param data the base64 encoded data
- * @param messageId the message id
+ * @param messageId the message id given by server. It must not be populated when publishing.
  * @param attributes optional extra attributes for this message.
  * @param publishTime the time the message was published. It must not be populated when publishing.
  */
 final case class PubSubMessage(data: String,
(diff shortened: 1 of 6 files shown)


google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/model.scala
```diff
@@ -99,9 +100,18 @@ final case class PubSubMessage(data: String,
...
+  def apply(data: String, attributes: immutable.Map[String, String]): PubSubMessage =
+    PubSubMessage(data, "", Some(attributes), None)
+
   /**
    * Java API: create [[PubSubMessage]]
    */
+  def create(data: String) =
+    PubSubMessage(data)
+
+  @deprecated("Setting messageId when creating message for publishing is futile.", "1.0.3")
```
2m: The version should be 1.1.0, as this is the version we are aiming to release in the middle of this week.

ennru:
Thank you for your first contribution! Keep them coming!

Issue #1356, opened 2018-12-05
SSE: support gzip encoding


ennru:
The Alpakka SSE client doesn't support gzipped replies out of the box.
This can easily be added by amending the `send` function as shown in the test added in https://github.com/akka/alpakka/pull/1355/commits/82b7ac8118c77b47518bc51225ab77acdb8dec64.

jefftt:
Hi! Hopefully this is still wanted. I put up a PR that addresses this.

PR #1796, opened 2019-06-30
SSE: Support gzip encoding

jefftt:
## Purpose

Support gzip responses out of the box

## References

References #1356 

## Changes

SSE Source

## Background Context

Based on previous test


2m:
Fixes #1356

PR #1799, opened 2019-07-01
Nightly with Akka 2.6

2m:
## Purpose

When in nightly, build with Akka 2.6.

## Changes

When in nightly:
* switches Akka to the latest mileston
* drops 2.11 cross Scala version


2m:
Unrelated failures were #1801 and #1802

PR #1807, opened 2019-07-03
Complete promise from the locally submitted ack

huntc:
Prior to this PR any acks to a PUBLISH could fail and have their failure swallowed (hidden). I now think that it is best to signal completions so that the calling code is able to act on them; as per the existing SUBACK completion mechanism. To illustrate:

1. Publisher publishes PUBLISH with QoS 1
2. Consumer receives the PUBLISH
3. Machine is really slow (like my ancient single core Atom processor!)
4. Consumer throws a ConsumeFailed given that 30 seconds (the former local ack timeout) has passed and it hasn't received a local PUBACK
5. A PUBACK is then provided locally

Step 5 will result in there being no route to the Consumer as it previously stopped at step 4. The local calling code has no knowledge of this and assumes that life is normal.

The test changes illustrate how this mechanism works.

I've also increased the local ack timeout to reduce the chances of above timeout. Testing on my slow Atom based machine shows that the changes work.


huntc:
@longshorej WDYT? It appears to work for us.

huntc:
Further to https://github.com/akka/alpakka/pull/1807#issuecomment-507981237, @longshorej and I chatted and concluded that a notification mechanism would be better than failing the stream. I've now updated the PR to use the existing SUBACK completion mechanism. I'm pretty happy with it, but will have another chat with Jason to get his feedback.

huntc:
I'm pretty happy with this PR now. Feel free to review/merge etc. There's no change to the existing behaviour other than the increased timeout. The PR mostly introduces a new feature whereby all types of locally submitted ack can now receive an additional `completed` promise parameter.

huntc:
@ennru @2m This PR is ready for review. :-)

huntc:
@2m Any chance we can get this merged? :-)

2m:
Sorry for the delay. Will review today.

huntc:
Thanks @2m !

PR #1812, opened 2019-07-04
Add imports to Ftp settings sample, actually run listFiles

raboof:


raboof:
oops

ftp/src/test/java/docs/javadsl/FtpTraversingExample.java
```diff
@@ -6,16 +6,18 @@
...
 import akka.stream.javadsl.Source;
 
 public class FtpTraversingExample {
 
-  public Source<FtpFile, NotUsed> listFiles(String basePath, FtpSettings settings)
+  public void listFiles(String basePath, FtpSettings settings, Materializer materializer)
       throws Exception {
-    return Ftp.ls(basePath, settings);
+    Ftp.ls(basePath, settings)
+        .runForeach(ftpFile -> System.out.println(ftpFile.toString()), materializer);
```
raboof: Not quite sure about this one: on the one hand it is nice to show in the example that this produces a source of `FtpFile`, on the other hand I think it is good to show actually running the stream, not just creating it, especially in 'early' examples.

ennru: I agree. The FTP examples are from the early days of Alpakka...

PR #1818, opened 2019-07-05
Update to the latest Akka 2.6-M4

2m:
## Purpose

Keeps nightlies in sync with the latest Akka release.


mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/QueueOfferState.scala
```diff
@@ -43,5 +43,5 @@ private[mqtt] object QueueOfferState {
         case (_, other) =>
           waitForQueueOfferCompleted(behavior, stash = stash :+ other)
       }
```
patriknw: not related to this change, but I wonder if that unstashing by sending to `self` is correct? it would be ok if the order of `QueueOfferCompleted` messages doesn't matter, otherwise another `QueueOfferCompleted` could already be in the mailbox and be processed by the behavior before the unstashed messages. Solution would be to use `StashBuffer`.

PR #1822, opened 2019-07-05
Solr: deleteByQuery with routing field

### LINKED ISSUE #1821: [Solr] deleteByQuery ignore routing field value when composeId router is used
The consequence is that the request is broadcast on all nodes of the solrCloud cluster.
That lead to dramatic performances.
I will sent a PR to fix this issue.

newca12:


2m:
Fixes #1821

2m:
Thanks for the fix!

PR #1830, opened 2019-07-17
Added withHeader(s) withProperty(-ies) and withoutDestination methods…

WellingR:
This is an addition to the functionality added to #1752 the withProperties method is now available on the `JmsMessage` trait, as well as withHeaders and withDestination.

All these methods are implemented such that the subtypes return their own subtype. E.g. `JmsTextMessage.withProperties` previously returned `JmsTextMessagePassThrough[NotUsed]` instead of `JmsTextMessage`

### DIFF
--- a/jms/src/main/mima-filters/1.1.0.backwards.excludes
+++ b/jms/src/main/mima-filters/1.1.0.backwards.excludes
@@ -1,0 +1,6 @@
+# PR #1830
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.stream.alpakka.jms.JmsMessage.withProperty")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.stream.alpakka.jms.JmsMessage.withHeaders")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.stream.alpakka.jms.JmsMessage.withProperties")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.stream.alpakka.jms.JmsMessage.withHeader")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.stream.alpakka.jms.JmsMessage.withoutDestination")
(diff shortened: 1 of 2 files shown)


WellingR:
I added some MiMa problem filters because of the newly added methods, did I add these in the right place?

jms/src/main/mima-filters/1.0.2.backwards.excludes
```diff
@@ -8,3 +8,10 @@ ProblemFilters.exclude[MissingTypesProblem]("akka.stream.alpakka.jms.DurableTopi
 
 # add Java API
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.stream.alpakka.jms.JmsEnvelope.*")
+
+# PR #1830
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.stream.alpakka.jms.JmsMessage.withProperty")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.stream.alpakka.jms.JmsMessage.withHeaders")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.stream.alpakka.jms.JmsMessage.withProperties")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.stream.alpakka.jms.JmsMessage.withHeader")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.stream.alpakka.jms.JmsMessage.withoutDestination")
```
2m: These have to go to a separate file named by the latest released, so `1.1.0.backwards.excludes`

WellingR:
One question, this is now targeted for 2.0, however this PR is actually backwards binary compatible, but not forwards binary compatible. Or do I get this wrong?

The only difference with #1752, is that this PR also adds methods to the `sealed trait JmsMessage`. But as the trait is sealed, can this incompatibility really be a problem?

ennru:
We now start merging things for Alpakka 2.0 after the 1.1.1 release.
Thank you for adding this!

ennru:
Backported for 1.1.2 with https://github.com/akka/alpakka/pull/1974

Issue #1761, opened 2019-06-19
HDFS: Overwrite target file


markarasev:
### Versions used 

Akka version: 2.5.22
Alpakka version: 1.0.2

### Expected Behavior

Given an existent file in HDFS, overwriting it with a different content should change the file's contents.

### Actual Behavior

The stream execution succeeds but the target file's contents are unchanged.

### Reproducible Test Case

Run `docs.scaladsl.HdfsWriterSpec` from https://github.com/akka/alpakka/pull/1760

2m:
Thanks for the reproducer. Are you willing to take a stab at fixing this?

markarasev:
I could try but I first need to be sure that I'm not misusing the API in the reproducer. Can someone confirm this is a bug and not an error from my side?

PR #1834, opened 2019-07-22
HDFS: fix target file overwrite

markarasev:
Fixes https://github.com/akka/alpakka/issues/1761

## Purpose

Overwrite HDFS target file when the `overwrite` setting is set to true.

## Changes

* propagated overwrite information to HdfsWriter base trait
* deleted existing file if overwrite is enabled in HdfsWriter#moveToTarget function

## Background Context

This is the most simple fix I could come up with but there are still some issues with the code imho:
* First the result of the `moveToTarget` operation is ignored at https://github.com/akka/alpakka/blob/master/hdfs/src/main/scala/akka/stream/alpakka/hdfs/impl/HdfsFlowStage.scala#L144 . This results in silent errors which lead to the original issue. What would be the best approach to handle errors here?
* Secondly, `org.apache.hadoop.fs.FileSystem` does not provide a `rename` function that can take an overwrite flag or option. This is why I had to break the atomicity of the `rename` operation with a `delete`. `org.apache.hadoop.fs.FileContext` does provide such a `rename` method https://hadoop.apache.org/docs/r3.1.1/api/org/apache/hadoop/fs/FileContext.html#rename-org.apache.hadoop.fs.Path-org.apache.hadoop.fs.Path-org.apache.hadoop.fs.Options.Rename...- but using it would require to use `FileContext` instead of `FileSystem` all the way around, and ultimately, break the public API.

### DIFF
--- a/hdfs/src/main/mima-filters/1.1.x.backwards.excludes
+++ b/hdfs/src/main/mima-filters/1.1.x.backwards.excludes
@@ -1,0 +1,1 @@
+ProblemFilters.exclude[Problem]("akka.stream.alpakka.hdfs.impl.*")
--- a/hdfs/src/main/scala/akka/stream/alpakka/hdfs/impl/writer/CompressedDataWriter.scala
+++ b/hdfs/src/main/scala/akka/stream/alpakka/hdfs/impl/writer/CompressedDataWriter.scala
@@ -17,36 +17,36 @@ import org.apache.hadoop.io.compress.{CodecPool, CompressionCodec, CompressionOu
  */
 @InternalApi
 private[writer] final case class CompressedDataWriter(
-    fs: FileSystem,
+    override val fs: FileSystem,
     compressionCodec: CompressionCodec,
-    pathGenerator: FilePathGenerator,
+    override val pathGenerator: FilePathGenerator,
     maybeTargetPath: Option[Path],
-    overwrite: Boolean
+    override val overwrite: Boolean
 ) extends HdfsWriter[FSDataOutputStream, ByteString] {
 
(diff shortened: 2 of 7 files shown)


lightbend-cla-validator:

markarasev:
Oops, I forgot to link the email I used in the commits to my GitHub account but I did sign the CLA. I now have linked this email address, how can I re-trigger the CLA-validator?

markarasev:
I ignored binary compatibility issues since it affects only internal APIs.

2m:
Thanks for the PR. It looks good to me, and I really appreciate you looking into the broken atomicity and suggesting a fix for that as well.

We are planning to release Alpakka 2.0 in a month or so. Some connectors have accumulated API breaking changes and thus the time for a major release. If you would be willing to upgrade this PR to use `FileContext` then we could target this PR to Alpakka 2.0.

markarasev:
Should the migration to `FileContext` be in this PR or in another one with an associated issue? I can't guarantee I'll find time to work on this in the next month and succeed.

Also, according to https://hadoop.apache.org/docs/r3.1.1/api/org/apache/hadoop/fs/FileContext.html#rename-org.apache.hadoop.fs.Path-org.apache.hadoop.fs.Path-org.apache.hadoop.fs.Options.Rename...-

> atomicity of rename is dependent on the file system implementation

So even if we migrate to `FileContext` the atomicity will not be guaranteed (but it is very likely to happen).

2m:
A new ticket for migration to `FileContext` would be good. Maybe that will attract the attention of other HDFS connector users and spark a discussion.

A new PR that does the migration then will close that new ticket and #1761 as well.

ennru:
Thank you for fixing this! Keep the PRs coming.

PR #1851, opened 2019-08-01
Build all connectors when building in nightly

### LINKED ISSUE #1850: Build changed script should always build a module during nightly build
Currently nightly builds do not run connector tests, because `buildChanged` script does not detect any changes. It should run tests of all modules.

2m:
## Purpose

`buildChanged.sh` should build modules even with no changes when the Travis job is a nightly.

## References

Fixes #1850


2m:
Merging as it can only be tested in a nightly run.

PR #1853, opened 2019-08-04
Automatically unregister packet ids and other resources

huntc:
Prior to this PR, actors associated with packet identifiers were responsible for not only registering with a packet router but also unregistering. This commit changes simplifies the behaviour to avoid errors so that actors don't need to explicitly unregister with a packet router to free resources. Instead, by virtue of terminating, an actor's associated packet identifier and other resources can be cleaned up via being watched by its router. These actors previously unregistered on termination so there is no real change in when unregistration occurs.

There was also the potential for a memory leak with the previous implementation given that a message could be routed and end up in the deadletter queue, yet its associated promise would never be completed.

Also, this PR reverses back https://github.com/akka/alpakka/pull/1852 behaviour where `ConsumeFailed` should cause the shutdown of a connection. In hindsight, the original view of letting `ConsumeFailed` drop on the floor is the right approach. What has changed since then is that, if a `Publish` is acknowledged locally but cannot be routed then its completion promise will fail. This wasn't happening before and is all that we require.

Lastly, I've re-instated a timeout that shouldn't have been changed as it was related to an application concern.


huntc:
Failing with MiMa given that I've now declared a couple of `impl` classes as private - which should be fine. Should we not exclude `impl` by default?

ennru:
Yes, it is generally OK to exclude `impl` with 
```
# Allow changes to impl
ProblemFilters.exclude[Problem]("akka.stream.alpakka.mqtt.streaming.impl.*")
```
in `1.1.x.backwards.excludes`.

ennru:
You'll be glad to hear that I added a general solution to Mima excludes for `impl` in https://github.com/akka/alpakka/pull/1860
I rebased this PR on master.

ennru:
My bad, it needs to be rebased on #1861 as well...

ennru:
This PR causes two tests in mqtt-streaming to fail.

huntc:
It is late at night and I've just forced pushed some changes and then noticed that others have also made some changes. I'll yield to @ennru - I apologise for my forced push.  I'll let you take it on from here.

ennru:
Sleep well.

huntc:
Looks as though we have a few flaky tests. I don't think it is due to this PR. I can look further if required, but I won't have time for a while.

ennru:
We decided to not merge it for 1.1.1 as the failures are new to us.

huntc:
> We decided to not merge it for 1.1.1 as the failures are new to us.

Quite right. It turned out to be an inconsistency in our code with the new logic highlighting a problem. Fixed here: https://github.com/akka/alpakka/pull/1853/commits/749709bfaeb469c73615cd182e6f50def793d8a8

huntc:
@longshorej Could you please review this also? Perhaps by commit...

huntc:
The build failed for an unrelated reason

huntc:
I’m getting some downstream test failures which I’ll have to look into. There could well be something strange going on with the changes in this  PR.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/QueueOfferState.scala
```diff
@@ -23,15 +26,64 @@ private[mqtt] object QueueOfferState {
...
       .receive[T] {
         case (context, completed: QueueOfferCompleted) =>
           completed.result match {
             case Right(QueueOfferResult.Enqueued) =>
-              stash.foreach(context.self.tell)
+              var b = Behavior.start(behavior, context)
+
+              stash.foreach {
+                case Right(msg) =>
+                  val nextBehavior = Behavior.interpretMessage(b, context, msg)
```
huntc: @longshorej Seeing an error in production at this line:

```
java.lang.IllegalArgumentException: cannot execute with [Unhandled] as behavior
	at akka.actor.typed.Behavior$.interpret(Behavior.scala:420)
	at akka.actor.typed.Behavior$.interpretMessage(Behavior.scala:393)
	at akka.stream.alpakka.mqtt.streaming.impl.QueueOfferState$.$anonfun$behaviorImpl$2(QueueOfferState.scala:71)
	at akka.stream.alpakka.mqtt.streaming.impl.QueueOfferState$.$anonfun$behaviorImpl$2$adapted(QueueOfferState.scala:69)
	at akka.actor.typed.internal.StashBufferImpl$Node.apply(StashBufferImpl.scala:27)
	at akka.actor.typed.internal.StashBufferImpl.foreach(StashBufferImpl.scala:92)
	at akka.stream.alpakka.mqtt.streaming.impl.QueueOfferState$.$anonfun$behaviorImpl$1(QueueOfferState.scala:69)
	at akka.actor.typed.internal.BehaviorImpl$ReceiveBehavior.receive(BehaviorImpl.scala:37)
	at akka.actor.typed.Behavior$.interpret(Behavior.scala:437)
	at akka.actor.typed.Behavior$.interpretMessage(Behavior.scala:393)
	at akka.actor.typed.internal.adapter.ActorAdapter.handleMessage(ActorAdapter.scala:118)
	at akka.actor.typed.internal.adapter.ActorAdapter.aroundReceive(ActorAdapter.scala:102)
	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:612)
	at akka.actor.ActorCell.invoke(ActorCell.scala:581)
	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:268)
	at akka.dispatch.Mailbox.run(Mailbox.scala:229)
	at akka.dispatch.Mailbox.exec(Mailbox.scala:241)
	at akka.dispatch.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
	at akka.dispatch.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
	at akka.dispatch.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
	at akka.dispatch.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
```

huntc:
@ennru @2m We've now tested this PR in production and are pretty happy with it. It is read for a final review/merge.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala
```diff
@@ -50,7 +50,7 @@ import scala.util.{Either, Failure, Success}
             settings: MqttSessionSettings)(implicit mat: Materializer): Behavior[Event] =
     disconnected(
       Disconnected(
-        Vector.empty,
+        StashBuffer(Int.MaxValue),
```
2m: In Akka 2.6 milestones, the StashBuffer has been made `akka` private and is instead exposed as a behaviour with stash capabilities:

https://github.com/akka/akka/pull/27200

Can the stash functionality change be reverted here, as there is currently no way to cross-build StashBuffer usage across Akka 2.5 and 2.6.

huntc: Clearly, StashBuffer is useful. I believe it was also recommended to us. We’ve put a ton of effort into this PR already, and tested it in production. Can it not be made public once more in Akka 2.6?

cc @patriknw

huntc: @2m @patriknw I just had a chat with @longshorej about this. His response:

> I think there's more to it to be completely correct, StashBuffer provides methods to have behaviors receive all of the stashed events before processing the mailbox. The replacement would need to do that, similar to the code in QueueOfferState right now where the stashed messages are passed to interpretMessage/interpretSignal.

I think that the simplest path forward is to make `StashBuffer` public once more.

2m: The visibility of `StashBuffer` is not the main problem here. Alpakka modules live in the `akka` package namespace. Therefore we can access the internals of Akka.

The problem here is not having a single factory method for `StashBuffer` across Akka 2.5 and 2.6. This prevents this module to be easily cross-compilable across different Akka versions.

I see that in Akka 2.6 public stash buffer API has been refactored from a standalone data structure to only being accessible via `Behaviors.withStash`.

And the internal API has been modified to take ActorContext which it did not in Akka 2.5.

@patriknw Can we get a constructor of `StashBuffer` that takes `ActorContext` in Akka 2.5 as well?

2m:
Validation failure was a test failure from this connector, but which have been seen before (#1549). So it might not be related to the changes in this PR.

2m:
Pushed a Akka 2.6 API usage fix.

huntc:
Lgtm

ennru:
The build seems to consistently fail on a test
https://travis-ci.com/akka/alpakka/jobs/229592564#L777

huntc:
That’s so weird as it does get through that test many times. I can also see that it gets through that test in the same run given the typed/untyped execution of it.

ennru:
Retriggering doesn't solve it. I guess it requires a slow machine to surface...

ennru:
Now it went through. I hope this won't keep us busy.

huntc:
> Now it went through. I hope this won't keep us busy.

I'm happy to look into it, but I'm starved of time at the moment. If anything, I'd suspect the test itself. Things are working pretty well in production. That said, if it becomes a problem then please let me know.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/BehaviorRunner.scala
```diff
@@ -0,0 +1,43 @@
...
+      case (b, StoredSignal(signal)) =>
+        val nextBehavior = Behavior.interpretSignal(b, context, signal)
+
+        if ((nextBehavior ne Behaviors.same) && (nextBehavior ne Behaviors.unhandled)) {
+          nextBehavior
+        } else {
+          b
+        }
+    }
+}
```
johanandren: Can't you use `StashBuffer` and `unstashAll` which comes out of the box instead?

huntc: We used to use StashBuffer: https://github.com/akka/alpakka/pull/1853#discussion_r317562543

johanandren: Ah I see, thanks I missed that conversation. That's ok I guess. Could be a future improvement then perhaps. 

The main shortcoming with the home made stash, except for being used/tested by less people, will be that if you use LB telemetry you won't get any metrics for it.

huntc:
> Now it went through. I hope this won't keep us busy.

@ennru I took a look at the test, but don't see any obvious reason for it to fail - and the failure log has disappeared from Travis. I will certainly take a look if/when it occurs again though.

ennru:
Thank you for pushing your improvements over to the world of Open Source!

ennru:
Backported to 1.1.x with #1961

Issue #1801, opened 2019-07-01
Failed: MqttFlowSpec should send an ack after sent confirmation


2m:
```
[info] - should send an ack after sent confirmation *** FAILED *** (15 seconds, 211 milliseconds)
[info]   The code passed to eventually never returned normally. Attempted 136 times over 15.099454129000002 seconds. Last failure message: false was not true. (MqttFlowSpec.scala:117)
[info]   org.scalatest.exceptions.TestFailedDueToTimeoutException:
[info]   at org.scalatest.concurrent.Eventually.tryTryAgain$1(Eventually.scala:432)
[info]   at org.scalatest.concurrent.Eventually.eventually(Eventually.scala:439)
[info]   at org.scalatest.concurrent.Eventually.eventually$(Eventually.scala:391)
[info]   at docs.scaladsl.MqttFlowSpec.eventually(MqttFlowSpec.scala:23)
[info]   at org.scalatest.concurrent.Eventually.eventually(Eventually.scala:337)
[info]   at org.scalatest.concurrent.Eventually.eventually$(Eventually.scala:336)
[info]   at docs.scaladsl.MqttFlowSpec.eventually(MqttFlowSpec.scala:23)
```

https://travis-ci.com/akka/alpakka/jobs/212184447

ennru:
https://travis-ci.com/akka/alpakka/jobs/222858261#L928

2m:
https://travis-ci.com/akka/alpakka/jobs/223291797

ennru:
https://travis-ci.com/akka/alpakka/jobs/226752099#L690

seglo:
https://travis-ci.com/akka/alpakka/jobs/242975205#L581

seglo:
Failure from PR #1981 build

https://travis-ci.com/akka/alpakka/jobs/244115463#L593

seglo:
https://travis-ci.com/akka/alpakka/jobs/254015130#L652

ennru:
https://travis-ci.com/akka/alpakka/jobs/260963211#L674 in #1347

ennru:
https://travis-ci.com/akka/alpakka/jobs/266709851#L526 in #2006

seglo:
https://travis-ci.com/akka/alpakka/jobs/275777735#L420

raboof:
again https://travis-ci.com/github/akka/alpakka/jobs/298526633#L705

raboof:
[...]

PR #1861, opened 2019-08-06
Mima: don't check reference connector, as it is not published

ennru:
## Purpose

Use the `interalProject` setup for the reference connector to not check Mima for it.

## References

#1860

References #xxxx

## Background Context

#1860 enabled Mima for the reference connector which isn't published, so it doesn't work.


ennru:
Failure was https://github.com/akka/alpakka/issues/1801

PR #1864, opened 2019-08-07
Scala 2.12.9

sullis:
## References

References #xxxx


ennru:
Thank you for trying this. 
Alpakka FTP uses a pattern which doesn't work with Scala 2.12.8, 2.12.9 and 2.13.0.
So we need to stay on Scala 2.12.7 for now.

Issue #1735, opened 2019-05-28
Failed: EventSourceSpec should read gzipped replies


2m:
```
[info] - should read gzipped replies *** FAILED *** (8 milliseconds)
[info]   java.net.BindException: Cannot assign requested address
[info]   at sun.nio.ch.Net.bind0(Native Method)
[info]   at sun.nio.ch.Net.bind(Net.java:433)
[info]   at sun.nio.ch.Net.bind(Net.java:425)
[info]   at sun.nio.ch.ServerSocketChannelImpl.bind(ServerSocketChannelImpl.java:223)
[info]   at sun.nio.ch.ServerSocketAdaptor.bind(ServerSocketAdaptor.java:74)
[info]   at sun.nio.ch.ServerSocketAdaptor.bind(ServerSocketAdaptor.java:67)
```

https://travis-ci.com/akka/alpakka/jobs/203473870

2m:
https://travis-ci.com/akka/alpakka/jobs/223291808

ennru:
This hasn't surfaced since over a year.
Closing.

Issue #1716, opened 2019-05-22
Failed JmsConnectorsSpec should producer disconnect exceptional completion


2m:
```
[info] - should producer disconnect exceptional completion *** FAILED *** (2 minutes)
[info]   A timeout occurred waiting for a future to complete. Queried 1996 times, sleeping 60 milliseconds between each query. (JmsConnectorsSpec.scala:584)
[info]   org.scalatest.concurrent.Futures$FutureConcept$$anon$1:
[info]   at org.scalatest.concurrent.Futures$FutureConcept.tryTryAgain$1(Futures.scala:538)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValueImpl(Futures.scala:550)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValueImpl$(Futures.scala:479)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:275)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue(Futures.scala:476)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue$(Futures.scala:475)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:275)
[info]   at docs.scaladsl.JmsConnectorsSpec.$anonfun$new$60(JmsConnectorsSpec.scala:584)
[info]   at docs.scaladsl.JmsConnectorsSpec.$anonfun$new$60$adapted(JmsConnectorsSpec.scala:561)
[info]   at akka.stream.alpakka.jms.JmsSpec.withServer(JmsSpec.scala:50)
[info]   at docs.scaladsl.JmsConnectorsSpec.$anonfun$new$59(JmsConnectorsSpec.scala:561)
[info]   at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
```

[...]

PR #1869, opened 2019-08-07
Scaladoc: exclude jwt-core 3.0.1 during unidoc

ennru:
## Purpose

Exclude a conflicting version during the unidoc task to avoid compilation errors due to classpath order.


2m:
Failures were unrelated: #1716 #1801 #1549 #1735

Issue #1306, opened 2018-11-06
Failure: JmsConnector connection status should reflect connection status on connection retries


2m:
```
[info] - should reflect connection status on connection retries *** FAILED *** (16 seconds, 792 milliseconds)
[info]   The code passed to eventually never returned normally. Attempted 16 times over 10.034311338 seconds. Last failure message: Published connection state Disconnected was not Connected. (JmsConnectionStatusSpec.scala:349)
[info]   org.scalatest.exceptions.TestFailedDueToTimeoutException:
[info]   at org.scalatest.concurrent.Eventually.tryTryAgain$1(Eventually.scala:432)
[info]   at org.scalatest.concurrent.Eventually.eventually(Eventually.scala:439)
[info]   at org.scalatest.concurrent.Eventually.eventually$(Eventually.scala:391)
[info]   at akka.stream.alpakka.jms.JmsSpec.eventually(JmsSpec.scala:18)
[info]   at akka.stream.alpakka.jms.JmsConnectionStatusSpec.$anonfun$new$43(JmsConnectionStatusSpec.scala:349)
[info]   at scala.collection.immutable.Range.foreach$mVc$sp(Range.scala:155)
[info]   at akka.stream.alpakka.jms.JmsConnectionStatusSpec.$anonfun$new$42(JmsConnectionStatusSpec.scala:348)
[info]   at akka.stream.alpakka.jms.JmsConnectionStatusSpec.$anonfun$new$42$adapted(JmsConnectionStatusSpec.scala:313)
[info]   at akka.stream.alpakka.jms.JmsSpec.withServer(JmsSpec.scala:54)
[info]   at akka.stream.alpakka.jms.JmsConnectionStatusSpec.$anonfun$new$41(JmsConnectionStatusSpec.scala:313)
```

https://travis-ci.org/akka/alpakka/jobs/451313850

ennru:
A different variant in the same test:

```
[info] JmsConnectionStatusSpec:
[info] JmsConnector connection status
[info] - should reflect connection status on connection retries *** FAILED *** (3 seconds, 916 milliseconds)
[info]   java.io.IOException: VMTransportServer already bound at: vm://brokerName-a6671ba7-3b6c-4aca-a52e-b1dc55936cc9?create=false
```

https://travis-ci.org/akka/alpakka/jobs/492957863#L706

2m:
```
[...]

PR #1874, opened 2019-08-09
JMS & Google pubsub grpc: Update Mima excludes

ennru:
* Rename the JMS excludes file
* Add exclude file for generated Google Cloud Pub-Sub gRPC classes


google-cloud-pub-sub-grpc/src/main/mima-filters/2.0.0.backwards.excludes
```diff
@@ -0,0 +1,6 @@
+# PR #1781 Updated dependencies
```
2m: The file should be `1.1.x.backwards.excludes`.

ennru:
Thank, I always get that wrong.

ennru:
Failure was https://github.com/akka/alpakka/issues/1306

PR #1876, opened 2019-08-09
Add DynamoDb tryFlow to be used with RetryFlow

### LINKED ISSUE #1682: DynamoDB: InternalServerErrorException: Internal server Error

2m:
## Purpose

Adds a `DynamoDb.tryFlow` of the shape `Req => Try[Resp]`. This allows composition with a RetryFlow (https://github.com/akka/akka/pull/27461).

To let the tests pass, RetryFlow is now included verbatim in this PR.

## References

Supersedes https://github.com/akka/alpakka/pull/1712
Fixes #1682


dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/impl/scaladsl/RetryFlow.scala
```diff
@@ -0,0 +1,291 @@
+/*
+ * Copyright (C) 2016-2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.scaladsl
```
ennru: The packages must be internal in Alpakka.

2m:
Yes, there are two API considerations regarding retrying failed requests. One is for the connector to provide a retry-friendly API that can be composed with the RetryFlow. This is what this PR does.

Then it is up for a connector to decide whether a factory that already incorporates RetryFlow should be provided. For example the new test-case that is added in this PR is quite use-case specific (retrying failed parts of BatchGetItems). But I could see how there could be a factory in the DynamoDb connector that retries requests when it gets QuotaExceeded exceptions.

2m:
Will have to be rebased and adapted once #1725 is merged. Closing this and will create a new PR for that.

Issue #1827, opened 2019-07-11
Upgrade to the latest Solr library version before next major Alpakka release


giena:
Hello,

I use alpakka solr in a sbt project with dependency 7.7.1. An exception occurs:
java.lang.NoSuchMethodError: org.apache.solr.client.solrj.request.UpdateRequest.setCommitWithin(I)V
        at akka.stream.alpakka.solr.impl.SolrFlowLogic.$anonfun$deleteEachByQuery$1(SolrFlowStage.scala:151)

If i compile my own alpakka version with dependency 7.7.1, there's not any exception.

Could you please upgrade to a 7.7 version? 7.4 starts to be old now.

2m:
Thanks for the ticket. We will make sure to upgrade to the latest before Alpakka Solr 2.0.0 comes out. Do you think upgrading to the latest 7.x is fine, or should we use the chance and upgrade all the way to the latest of 8.x?

giena:
I think the version 8 is too recent. The new version 7.7.2 could be a good choice.

ennru:
We're now open for dependency upgrades for Alpakka 2.0. Would you have the time to prepare a PR?

giena:
See #1880

giena:
Thx

PR #1880, opened 2019-08-10
Solr: Upgrade to 7_7_2

giena:
See #1827


ennru:

PR #1883, opened 2019-08-12
Enable fatal warnings

### LINKED ISSUE #1875: Fail CI build for compiler warnings
### Short description

To improve code quality compiler warnings should become fatal. 

### Details

To start with, this can be required per module so that nothing degrades. For now, this should apply to the Scala 2.12 build, only.

2m:
## Purpose

Enables fatal warnings by default with selective opt-out for the modules that still have compilation warnings.

## References

Fixes #1875 

## Notes

Quite a few fixed warnings where unused parameter warnings in `copy` methods that only had a single parameter. Having a default parameter on the single argument copy method is not reasonable, as copy is always called with at least one parameter.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,10 +22,12 @@ jobs:
     - script: sbt scalafmtSbtCheck || { echo "[error] Unformatted sbt code found. Please run 'scalafmtSbt' and commit the reformatted code."; false; }
       name: "Build code style check (fixed with `sbt scalafmtSbt`)"
     - env: CMD="++2.11.12 Test/compile"
-      name: "Compile all tests (with Scala 2.11)"
+      name: "Compile all code with Scala 2.11"
       if: type != cron
+    - env: CMD="++2.12.7 Test/compile"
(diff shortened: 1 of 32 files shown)


2m:
I am not sure where to add it so it is useful. We are only enabling fatal warnings on Scala 2.12, and there is no separate compilation step that is done for all the code with 2.12. Maybe mention that in the sbt welcome message for the build?

ennru:
Funnily enough, scalafmt failed.

The fatal warnings for 2.12 shows first in the MiMa step of the build -- as Travis shows for your PR: https://travis-ci.com/akka/alpakka/jobs/224381883#L873

Maybe put a note on that step name? "Fails for compiler warnings and Mima"?

ennru:
But the Mima step doesn't compile test code.

2m:
Yea. Lets add another job in the "check" stage that compiles all the code with Scala 2.12?

2m:
Failure was #1801

PR #1889, opened 2019-08-14
Expect correct exception in Akka 2.6

2m:
## Purpose

Akka 2.6 changes what exception is being thrown in case of FileIO failure. This PR adapts the testcase to expect a correct exception.

### DIFF
--- a/file/src/main/java/akka/stream/alpakka/file/impl/FileTailSource.java
+++ b/file/src/main/java/akka/stream/alpakka/file/impl/FileTailSource.java
@@ -36,7 +36,8 @@ import java.nio.file.StandardOpenOption;
  *
  * <p>Aborting the stage can be done by combining with a [[akka.stream.KillSwitch]]
  *
- * <p>To use the stage from Scala see the factory methods in {@link akka.stream.alpakka.file.scaladsl.FileTailSource}
+ * <p>To use the stage from Scala see the factory methods in {@link
+ * akka.stream.alpakka.file.scaladsl.FileTailSource}
  */
 @InternalApi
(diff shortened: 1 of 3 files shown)


file/src/main/java/akka/stream/alpakka/file/impl/FileTailSource.java
```diff
@@ -36,7 +36,8 @@
  *
  * <p>Aborting the stage can be done by combining with a [[akka.stream.KillSwitch]]
  *
- * <p>To use the stage from Scala see the factory methods in {@link akka.stream.alpakka.file.scaladsl.FileTailSource}
+ * <p>To use the stage from Scala see the factory methods in {@link
+ * akka.stream.alpakka.file.scaladsl.FileTailSource}
```
ennru: It is not legal JavaDoc to break the `@link` tag. I recently change these lines to make Scaladoc more happy. Did the autoformat do this?

2m: Yes, it was auto-formatted. I have seen this flipping back and forth recently.

2m: And even if I manually add a linebreak before `{@link`, google java formatter reformats it like in this PR.

ennru: Cool blend of Scaladoc markdown and Javadoc tag notation BTW. Guess what HTML it spits out...
https://doc.akka.io/api/alpakka/current/akka/stream/alpakka/file/impl/FileTailSource.html

PR #1896, opened 2019-08-19
DynamoDb: AWS SDK2; Add tryFlow to be used with RetryFlow

### LINKED ISSUE #1682: DynamoDB: InternalServerErrorException: Internal server Error

2m:
*This is a continuation of #1876. ~Built on top of #1725 and therefore Draft until it gets merged~*

## Purpose

Adds a `DynamoDb.tryFlow` of the shape `Req => Try[Resp]`. This allows composition with a RetryFlow (https://github.com/akka/akka/pull/27461).

To let the tests pass, RetryFlow is now included verbatim in this PR.

## References

Supersedes https://github.com/akka/alpakka/pull/1712
Fixes #1682

### DIFF
--- a/dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/javadsl/DynamoDb.scala
+++ b/dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/javadsl/DynamoDb.scala
@@ -14,6 +14,8 @@ import software.amazon.awssdk.core.async.SdkPublisher
 import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient
 import software.amazon.awssdk.services.dynamodb.model.{DynamoDbRequest, DynamoDbResponse}
 
+import scala.util.Try
+
 /**
  * Factory of DynamoDb Akka Stream operators.
  */
@@ -26,6 +28,22 @@ object DynamoDb {
                                                            operation: DynamoDbOp[In, Out],
                                                            parallelism: Int): Flow[In, Out, NotUsed] =
     scaladsl.DynamoDb.flow(parallelism)(client, operation).asJava
+
+  /**
+   * Create a Flow that emits a response for every request to DynamoDB.
+   * A successful response is wrapped in [scala.util.success] and a failed
+   * response is wrapped in [scala.util.Failure].
+   */
+  def tryFlow[In <: DynamoDbRequest, Out <: DynamoDbResponse, State](
+      client: DynamoDbAsyncClient,
+      operation: DynamoDbOp[In, Out],
+      parallelism: Int
+  ): Flow[akka.japi.Pair[In, State], akka.japi.Pair[Try[Out], State], NotUsed] =
+    Flow
+      .create[akka.japi.Pair[In, State]]()
+      .map(func(p => (p.first, p.second)))
+      .via(scaladsl.DynamoDb.tryFlow(parallelism)(client, operation))
+      .map(func(t => akka.japi.Pair.create(t._1, t._2)))
 
   /**
    * Create a Source that will emit potentially multiple responses for a given request.
@@ -46,4 +64,8 @@ object DynamoDb {
                                                              request: In,
                                                              mat: Materializer): CompletionStage[Out] =
     Source.single(request).via(flow(client, operation, 1)).runWith(Sink.head(), mat)
+
+  private def func[T, R](f: T => R) = new akka.japi.function.Function[T, R] {
+    override def apply(param: T): R = f(param)
+  }
 }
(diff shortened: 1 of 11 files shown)


dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/javadsl/DynamoDb.scala
```diff
@@ -27,6 +29,22 @@ object DynamoDb {
                                                            parallelism: Int): Flow[In, Out, NotUsed] =
     scaladsl.DynamoDb.flow(parallelism)(client, operation).asJava
 
+  /**
+   * Create a Flow that emits a response for every request to DynamoDB.
+   * A successful response is wrapped in [scala.util.success] and a failed
```
ennru:
```suggestion
   * A successful response is wrapped in [scala.util.Success] and a failed
```

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/javadsl/DynamoDb.scala
```diff
@@ -27,6 +29,22 @@ object DynamoDb {
                                                            parallelism: Int): Flow[In, Out, NotUsed] =
     scaladsl.DynamoDb.flow(parallelism)(client, operation).asJava
 
+  /**
+   * Create a Flow that emits a response for every request to DynamoDB.
```
ennru: Add something along the lines "Intended to be used with Akka retry flows which expect the Scala `Try` type"

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/scaladsl/DynamoDb.scala
```diff
@@ -27,6 +28,38 @@ object DynamoDb {
   )(implicit client: DynamoDbAsyncClient, operation: DynamoDbOp[In, Out]): Flow[In, Out, NotUsed] =
     Flow[In].mapAsync(parallelism)(operation.execute(_))
 
+  /**
+   * Create a Flow that emits a response for every request to DynamoDB.
+   * A successful response is wrapped in [scala.util.success] and a failed
```
ennru:
```suggestion
   * A successful response is wrapped in [scala.util.Success] and a failed
```

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/scaladsl/DynamoDb.scala
```diff
@@ -27,6 +28,38 @@ object DynamoDb {
...
+   * response is wrapped in [scala.util.Failure].
+   */
```
ennru: Add `@param` documentation with an explanation of `parallelism`.
Add `@tparam` for `State`.

dynamodb/src/test/java/docs/javadsl/RetryTest.java
```diff
@@ -0,0 +1,203 @@
...
+        .toCompletableFuture()
+        .get(5, TimeUnit.SECONDS);
+    DynamoDb.single(
+            client, DynamoDbOp.batchWriteItem(), batchWriteLargeItemRequest(26, 50), materializer)
+        .toCompletableFuture()
+        .get(5, TimeUnit.SECONDS);
+
+    final JavaPartialFunction<
+            Pair<Try<BatchGetItemResponse>, NotUsed>,
+            akka.japi.Option<Collection<Pair<BatchGetItemRequest, NotUsed>>>>
```
ennru: The Java code improves a tiny bit if you `import akka.japi.Option` and `import static akka.japi.Option.*` and `NotUsed.notUsed` for less need to qualify things.

dynamodb/src/test/resources/application.conf
```diff
@@ -3,3 +3,5 @@ akka {
   logging-filter = "akka.event.slf4j.Slf4jLoggingFilter"
   loglevel = "DEBUG"
 }
+
+akka.http.client.parsing.max-content-length = 20m
```
ennru: Should it be recommended in the docs to increase this from the Akka HTTP default (8 MByte)?

2m: That is only needed for tests. The RetryFlow test sends a request larger than DynamoDb limit of 16mb. And then it partially fails. RetryFlow is then used to retry the failed part.

PR #1898, opened 2019-08-20
Do not enable fatal warnings on nightly

2m:
## Purpose

Disables fatal warnings when CI is running a nightly build. Fatal warnings are still enabled in PR builds and regular master builds.

Alpakka nightly builds switch to the latest milestone of Akka 2.6. Some of the APIs in the 2.6 milestones are deprecated which prevents running nightly builds with fatal warnings.


ennru:
LGTM.

PR #1903, opened 2019-08-27
Scaladoc: link to source and canonical

ennru:
* Use new Scaladoc flags from Scala 2.12.9.
* Remove hack to get correct links to Java sources.

### DIFF
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -66,12 +66,6 @@ object Common extends AutoPlugin {
           version.value,
           "-sourcepath",
           (baseDirectory in ThisBuild).value.toString,
-          "-doc-source-url", {
-            val branch = if (isSnapshot.value) "master" else s"v${version.value}"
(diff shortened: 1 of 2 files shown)


project/Common.scala
```diff
@@ -66,17 +66,31 @@ object Common extends AutoPlugin {
...
           "com.google.longrunning:com.google.protobuf:com.google.rpc:com.google.type"
         ),
+      Compile / doc / scalacOptions ++= (scalaVersion.value match {
+          case Dependencies.Scala211 =>
+            Seq(
+              "-doc-source-url", {
+                val branch = if (isSnapshot.value) "master" else s"v${version.value}"
+                s"https://github.com/akka/alpakka/tree/${branch}€{FILE_PATH}.scala#L1"
+              }
+            )
```
2m: That's right. Even if we generate the unified scaladoc only with Scala 2.12, the scaladoc in the javadoc jars is generated with whatever Scala version is used to compile the code.

2m:
Unrelated failure was #1649

Issue #600, opened 2017-11-20
sFTP proxy server support


tonymurphy:
When working with sFTP, the SSHJ session does not support the use of a proxy server.

ennru:
Implemented with #1907

Issue #1723, opened 2019-05-23
[Failure] MQTT streaming: publish twice with a QoS of 1 so that the second is queued


ennru:
"MQTT client connector should publish twice with a QoS of 1 so that the second is queued."

https://travis-ci.com/akka/alpakka/jobs/202157323#L890

Stages didn't stop in time

```
activeShells (actor: akka://mqtt-spec/system/StreamSupervisor-3/flow-105-0-ignoreSink):
  GraphInterpreterShell(
  logics: [
    akka.stream.impl.fusing.GraphStages$IgnoreSink$@6db721cf attrs: [Name(ignoreSink), InputBuffer(4,16), SupervisionStrategy(<function1>), Dispatcher(akka.actor.default-dispatcher)],
    LazyFlow attrs: [Name(lazyFlow), Name(lazyFlow), InputBuffer(4,16), SupervisionStrategy(<function1>), Dispatcher(akka.actor.default-dispatcher)],
    MapAsync(10,akka.stream.alpakka.mqtt.streaming.scaladsl.ActorMqttClientSession$$Lambda$10779/1372930626@6f1ddca5) attrs: [Name(mapAsync), LogLevels(LogLevel(4),LogLevel(4),LogLevel(4)), LogLevels(LogLevel(4),LogLevel(4),LogLevel(4)), InputBuffer(4,16), SupervisionStrategy(<function1>), Dispatcher(akka.actor.default-dispatcher)],
    Log attrs: [LogLevels(LogLevel(4),LogLevel(4),LogLevel(4)), LogLevels(LogLevel(4),LogLevel(4),LogLevel(4)), InputBuffer(4,16), SupervisionStrategy(<function1>), Dispatcher(akka.actor.default-dispatcher)],
    Map(akka.stream.alpakka.mqtt.streaming.scaladsl.ActorMqttClientSession$$Lambda$10778/686177776@6ecec637) attrs: [Name(map), LogLevels(LogLevel(4),LogLevel(4),LogLevel(4)), LogLevels(LogLevel(4),LogLevel(4),LogLevel(4)), InputBuffer(4,16), SupervisionStrategy(<function1>), Dispatcher(akka.actor.default-dispatcher)],
[...]

PR #1907, opened 2019-08-28
FTP - Add proxy support to FTP, SFTP, and FTPS connectors

mjp426:
## Purpose

Adds proxy support to FTP, SFTP, and FTPS connectors

## References

References #600

## Changes

 * Add `withProxy` method on all three FTPSettings builders that takes a `java.net.Proxy` object
 * Pass the proxy object to the underlying client before the connection is made
 * Added tests for the three supported protocols via a proxy
 * Added squid proxy and configuration to docker-compose so this can be tested. 

## Background Context

The approach is relatively straight forward. The extra option fits neatly into to existing builder pattern and passing the proxy to the client at the start of the connect operation is the natural place for it to go. 

I've chosen to use a squid proxy in the testing as this is a widely used proxy.

### DIFF
--- a/docs/src/main/paradox/ftp.md
+++ b/docs/src/main/paradox/ftp.md
@@ -37,7 +37,7 @@ The example demonstrates optional use of `configureConnection` option available 
 
 For non-anonymous connection, please provide an instance of @scaladoc[NonAnonFtpCredentials](akka.stream.alpakka.ftp.FtpCredentials$$NonAnonFtpCredentials) instead.
 
-For connection via a proxy, please provide an instance of `java.net.Proxy`.
+For connection via a proxy, please provide an instance of `java.net.Proxy` by using the `withProxy` method.
 
 For connection using a private key, please provide an instance of @scaladoc[SftpIdentity](akka.stream.alpakka.ftp.SftpIdentity) to @scaladoc[SftpSettings](akka.stream.alpakka.ftp.SftpSettings).
 
--- a/docs/src/main/paradox/ftp.md
+++ b/docs/src/main/paradox/ftp.md
@@ -37,6 +37,8 @@ The example demonstrates optional use of `configureConnection` option available 
 
 For non-anonymous connection, please provide an instance of @scaladoc[NonAnonFtpCredentials](akka.stream.alpakka.ftp.FtpCredentials$$NonAnonFtpCredentials) instead.
 
+For connection via a proxy, please provide an instance of `java.net.Proxy`.
+
 For connection using a private key, please provide an instance of @scaladoc[SftpIdentity](akka.stream.alpakka.ftp.SftpIdentity) to @scaladoc[SftpSettings](akka.stream.alpakka.ftp.SftpSettings).
 
 In order to use a custom SSH client for SFTP please provide an instance of [SSHClient](https://static.javadoc.io/com.hierynomus/sshj/0.26.0/net/schmizz/sshj/SSHClient.html).
--- a/ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
+++ b/ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
@@ -112,7 +112,14 @@ final class FtpSettings private (
   )
 
   override def toString =
-    s"""FtpSettings(host=$host,port=$port,credentials=$credentials,binary=$binary,passiveMode=$passiveMode,configureConnection=$configureConnection),proxy=$proxy"""
+    "FtpSettings(" +
+      s"host=$host," +
+      s"port=$port," +
+      s"credentials=$credentials," +
+      s"binary=$binary," +
+      s"passiveMode=$passiveMode," +
+      s"configureConnection=$configureConnection," +
+      s"proxy=$proxy)"
 }
 
 /**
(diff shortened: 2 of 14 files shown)


mjp426:
The build is now failing on the MiMa check. Could someone from the core team confirm I'm ok to add an exception for this change?

mjp426:
Also, the unidoc task was failing because is was picking up an old version of commons-net from the hbase dependency. I've fixed it by upgrading that dependency, but perhaps there's a better way to do this?

ennru:
Hi @mjp426 	- Welcome to Alpakka!

Thank you for working on this, as you've seen this feature has been requested for a long while.

The Mima exclusions for the settings classes are OK to add.
The Unidoc problem should be fixed in here instead:
https://github.com/akka/alpakka/blob/e48dd81f59fa00bd5beffabcc60720c83332c3c4/build.sbt#L69-L77

mjp426:
Hi @ennru, thanks for the pointers. I've made the changes you suggested. 

My FTP tests are passing, but there are a few failing modules. I can't reproduce the failures locally, and they seem unrelated to my changes. Do you have any tips to get the tests green?

mjp426:
Also, I think the dependency-change flag can be removed now.

ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
```diff
@@ -95,18 +99,20 @@ final class FtpSettings private (
...
     binary = binary,
     passiveMode = passiveMode,
-    configureConnection = configureConnection
+    configureConnection = configureConnection,
+    proxy = proxy
   )
 
   override def toString =
-    s"""FtpSettings(host=$host,port=$port,credentials=$credentials,binary=$binary,passiveMode=$passiveMode,configureConnection=$configureConnection)"""
+    s"""FtpSettings(host=$host,port=$port,credentials=$credentials,binary=$binary,passiveMode=$passiveMode,configureConnection=$configureConnection),proxy=$proxy"""
```
ennru: The `)` should move to the end. You may make this a single-" string and split it with one field per line, so this doesn't happen again.

ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
```diff
@@ -182,18 +192,20 @@ final class FtpsSettings private (
       credentials: FtpCredentials = credentials,
       binary: Boolean = binary,
       passiveMode: Boolean = passiveMode,
-      configureConnection: FTPSClient => Unit = configureConnection
+      configureConnection: FTPSClient => Unit = configureConnection,
+      proxy: Option[Proxy] = None
```
ennru:
```suggestion
      proxy: Option[Proxy] = proxy
```

ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
```diff
@@ -182,18 +192,20 @@ final class FtpsSettings private (
...
     binary = binary,
     passiveMode = passiveMode,
-    configureConnection = configureConnection
+    configureConnection = configureConnection,
+    proxy = proxy
   )
 
   override def toString =
-    s"""FtpsSettings(host=$host,port=$port,credentials=$credentials,binary=$binary,passiveMode=$passiveMode,configureConnection=$configureConnection)"""
+    s"""FtpsSettings(host=$host,port=$port,credentials=$credentials,binary=$binary,passiveMode=$passiveMode,configureConnection=$configureConnection),proxy=$proxy"""
```
ennru: Same thing with the `)`.

ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
```diff
@@ -250,25 +265,28 @@ final class SftpSettings private (
...
 
   private def copy(
       host: java.net.InetAddress = host,
       port: Int = port,
       credentials: FtpCredentials = credentials,
       strictHostKeyChecking: Boolean = strictHostKeyChecking,
       knownHosts: Option[String] = knownHosts,
-      sftpIdentity: Option[SftpIdentity] = sftpIdentity
+      sftpIdentity: Option[SftpIdentity] = sftpIdentity,
+      proxy: Option[Proxy] = None
```
ennru:
```suggestion
      proxy: Option[Proxy] = proxy
```

ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
```diff
@@ -95,18 +99,20 @@ final class FtpSettings private (
       credentials: FtpCredentials = credentials,
       binary: Boolean = binary,
       passiveMode: Boolean = passiveMode,
-      configureConnection: FTPClient => Unit = configureConnection
+      configureConnection: FTPClient => Unit = configureConnection,
+      proxy: Option[Proxy] = None
```
ennru:
```suggestion
      proxy: Option[Proxy] = proxy
```

docs/src/main/paradox/ftp.md
```diff
@@ -37,6 +37,8 @@ The example demonstrates optional use of `configureConnection` option available
 
 For non-anonymous connection, please provide an instance of @scaladoc[NonAnonFtpCredentials](akka.stream.alpakka.ftp.FtpCredentials$$NonAnonFtpCredentials) instead.
 
+For connection via a proxy, please provide an instance of `java.net.Proxy`.
```
ennru: Ah, there is this. Maybe refer to the `withProxy` method.

mjp426:
Good catch on the copy method. I've committed your suggestions.

ennru:
Failures were 
https://github.com/akka/alpakka/issues/1909
https://github.com/akka/alpakka/issues/1723

ennru:
Thank you! Great to see this long-standing feature request fixed.

PR #1915, opened 2019-09-04
Switch to torkild credentials

2m:
## Purpose

Switches to torkild credentials for publishing artifacts.


2m:
Merging since as credentials are not available in PR validation. Thus the only way to test it is to build master or release.

PR #1921, opened 2019-09-09
AMQP: flow stages

kstrek:
## Purpose

Following the lack of progress in #1330 I prepared my own proposal that covers changes discussed in that PR - addition of Flow stages for AMQP connector.

## References

References #1132 #1330

## Changes

- Added 4 new Flow variants: without confirms, blocking confirms, async confirms, async unordered confirms.
- Modified `WriteMessage` to include `passThrough` element. According to MiMa it breaks binary compatibility. Error was reported for private constructor of this class - I'm not sure if this is an issue or how else should the addition of pass-through be handled so I have to ask you for advise on this 😉.

## Background Context

Context for most changes can be found in #1330.


ennru:
Thank you for pushing this. The new argument on the private constructor is acceptable and may be added to the MiMa exclusions.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
```diff
@@ -0,0 +1,338 @@
...
+        } else {
+          log.debug("Received upstream finish signal - stage will be closed when all buffered messages are processed")
+        }
+
+      private def publish(message: WriteMessage[T]): DeliveryTag = {
+        val tag: DeliveryTag = channel.getNextPublishSeqNo
+
+        log.debug("Publishing message {} with deliveryTag {}.", message, tag)
+
+        Try(
```
fancywriter: what's the reason using `Try` class and extract it right here, why not just `try` clause?

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
```diff
@@ -0,0 +1,338 @@
...
+            exchange,
+            message.routingKey.getOrElse(routingKey),
+            message.mandatory,
+            message.immediate,
+            message.properties.orNull,
+            message.bytes.toArray
+          )
+        ) match {
+          case Failure(exception) => failStage(exception)
+          case _ => ()
```
fancywriter: this line won't be needed for `try`:
```
try channel.basicPublish(...) catch {
  case NonFatal(e) => failStage(e)
}
```

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
kstrek: Well, you are right, I'm not sure why I have done it this way (it was some time ago), but anyway it looks like throwing exception from `onPush` has the same effect as failing the stage manually, so there is no need to handle this exception here at all.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
kstrek: Removed the try completely - please see my answer to your previous comment.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
```diff
@@ -0,0 +1,332 @@
...
+import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.{Future, Promise}
+
+/**
+ * Internal API.
+ */
+@InternalApi private final case class AwaitingMessage[T](
+    tag: DeliveryTag,
+    passThrough: T,
+    var ready: Boolean = false
```
fancywriter: does it have to be var? I see that in `setReady` the object is copied, not modified.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
```diff
@@ -0,0 +1,332 @@
...
+
+    dequeued.foreach(m => cancelTimer(m.tag))
+
+    pushOrEnqueueResults(
+      dequeued.map(m => WriteResult.rejected(m.passThrough))
+    )
+  }
+
+  private def pushOrEnqueueResults(results: Iterable[WriteResult[T]]): Unit = {
+    results.foreach(result => {
```
fancywriter: I would prefer `results.foreach { result => }` syntax than `results.foreach(result => {})`, less parentheses, easier to read.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
```diff
@@ -0,0 +1,332 @@
...
+      if (exitQueue.nonEmpty || !isAvailable(out)) {
```
fancywriter: I might be wrong, but for me will be eaiser to read logic "if output is available and queue is empty then push, otherwise enqueue for future".

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
```diff
@@ -0,0 +1,332 @@
...
+        if (isFinished)
+          closeStage()
+        else
+          log.debug("Received upstream failure signal - stage will be failed when all buffered messages are processed")
+
+      }
+
+      override def onUpstreamFinish(): Unit =
+        if (noAwaitingMessages && exitQueue.isEmpty) {
+          promise.success(Done)
```
fancywriter: `tryFailure`, but `success`, why not `trySuccess`? From reading code it is not entirely clear how often each method is called... could it be possible to be called more than once?

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
```diff
@@ -0,0 +1,332 @@
...
+        log.debug("Received timeout for deliveryTag {}.", tag)
+        onRejection(tag, multiple = false)
+      }
+      case _ => ()
+    }
+
+  private def closeStage(): Unit =
+    upstreamException match {
+      case Some(throwable) =>
+        promise.failure(throwable)
```
fancywriter: and here is failure/complete, not tryFailure/tryComplete.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
```diff
@@ -0,0 +1,332 @@
...
+      override def enqueueMessage(tag: DeliveryTag, passThrough: T): Unit =
+        buffer += (tag -> AwaitingMessage(tag, passThrough))
+
+      override def dequeueAwaitingMessages(tag: DeliveryTag, multiple: Boolean): Iterable[AwaitingMessage[T]] = {
+        setReady(tag)
+
+        if (multiple) {
+          dequeueWhile((t, _) => t <= tag)
+        } else {
+          if (isAtHead(tag)) {
```
fancywriter: could be just plain if-else-if-else, not nested.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
```diff
@@ -0,0 +1,332 @@
...
+            dequeueWhile((_, message) => message.ready)
+          } else {
+            Seq.empty
+          }
+        }
+      }
```
fancywriter: or even better:
```
val predicate: ??? = if (multiple) (t, _) => t <= tag else if (isAtHead(tag)) (_, message) => message.ready else (_,_) => false
dequeueWhile(predicate)
```

Maybe could be written more elegantly...

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
```diff
@@ -0,0 +1,332 @@
...
+
+      override def enqueueMessage(tag: DeliveryTag, passThrough: T): Unit =
+        buffer += AwaitingMessage(tag, passThrough)
+
+      override def dequeueAwaitingMessages(tag: DeliveryTag, multiple: Boolean): Iterable[AwaitingMessage[T]] =
+        if (multiple)
+          buffer.dequeueAll(_.tag <= tag)
+        else
+          buffer
+            .dequeueFirst(_.tag == tag)
```
fancywriter: `Option#fold`?

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpFlow.scala
```diff
@@ -0,0 +1,185 @@
...
+
+      override def publish(message: WriteMessage[T]): Unit = {
+        val publicationResult: Try[Boolean] = for {
+          _ <- publishWithBlockingConfirm(message)
+          result <- waitForConfirmation()
+        } yield result
+
+        log.debug("Publication result: {}", publicationResult)
+
+        publicationResult
```
fancywriter: looks too complex and long, could it be simplified?

I think everything could be done in one `Future#transform` or `Future#transformWith`...

amqp/src/main/scala/akka/stream/alpakka/amqp/javadsl/AmqpFlow.scala
```diff
@@ -0,0 +1,109 @@
...
+      bufferSize: Int,
+      confirmationTimeout: java.time.Duration
+  ): akka.stream.javadsl.Flow[WriteMessage[T], WriteResult[T], CompletionStage[Done]] =
+    akka.stream.alpakka.amqp.scaladsl.AmqpFlow
+      .withAsyncUnorderedConfirm(
+        settings = settings,
+        bufferSize = bufferSize,
+        confirmationTimeout = JavaDurationConverters.asFiniteDuration(confirmationTimeout)
+      )
+      .mapMaterializedValue(f => f.toJava)
```
fancywriter: `.mapMaterializedValue(_.toJava)` ?

amqp/src/main/scala/akka/stream/alpakka/amqp/model.scala
```diff
@@ -28,57 +31,110 @@ object ReadResult {
...
     WriteMessage(bytes, immediate, mandatory)
 }
+
+final class WriteResult[T] private (val confirmed: Boolean, val passThrough: T) {
+  def rejected: Boolean = !confirmed
+
+  override def toString: String =
+    s"WriteResult(confirmed=$confirmed, passThrough=$passThrough)"
+
+  override def equals(other: Any): Boolean = other match {
```
fancywriter: `PartialFunction.cond`

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
kstrek: `onUpstreamFinish` is called only once, thats why it should be safe to call just `success`. `tryFailure` is used in `postStop` and `onFailure`. `postStop` is executed even if you close stage manually, so it's more like a fallback if promise is not resolved in any correct way. `onFailure` is called in a connection shutdown handler (see `AmqpConnectorLogic`) so I'm not sure if it would be safe to call `failure`.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
kstrek: `try...` is used just in two cases described in my answert to your comment above. Here I don's see any reason why it wouldn't be safe to call failure/complete.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
kstrek: `setReady(tag)` should be called before the second `if`, it doesn't have to be called for `multiple` branch - I misplaced it during refactoring. So it can't be simplified as you suggested.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
kstrek: It's certainly not easier to read for me. And with `setReady` moved it can't be changed to predicate like this.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpFlow.scala
kstrek: I moved timeout handling code inside `waitForConfirmation` as it seems to better fit there, and used fold instead of transform - code looks way better now. If you have further suggestions i'm open 😉

amqp/src/main/scala/akka/stream/alpakka/amqp/model.scala
kstrek: It doesn't make it much simpler, and every single equals method in whole alpakka seems to be implemented using `match`, so I'll leave that as is.

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpFlow.scala
```diff
@@ -0,0 +1,96 @@
...
+import akka.stream.alpakka.amqp._
+import akka.stream.scaladsl.Flow
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+object AmqpFlow {
+
+  /**
+   * Creates an [[AmqpFlow]] that accepts [[WriteMessage]] elements and emits [[WriteResult]].
```
ennru: Scaladoc requires fully-qualified class names for everything that is not package local.
You can remove the links and just put the classes in backticks -- the types are referenced in the type signature anyway.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
```diff
@@ -0,0 +1,330 @@
...
+ *
+ * AMQP flow that uses asynchronous confirmations but preserves the order in which messages were pulled.
+ * Internally messages awaiting confirmation are stored in a ordered buffer. Initially messages have `ready`
+ * flag set to `false`. On confirmation flag is changed to `true` and messages from the beginning of the queue
+ * are dequeued in sequence until first non-ready message - this means that if confirmed message is not a
+ * first element of the buffer then nothing will be dequeued. Flag `multiple` on a confirmation means that
+ * broker confirms all messages up to a given delivery tag, which means that so all messages up to (and including)
+ * this delivery tag can be safely dequeued.
+ */
+@InternalApi private[amqp] final class AmqpAsyncFlow[T](
```
ennru: Please append `Stage` to the stage class names. We can do without `sealed`.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpFlow.scala
```diff
@@ -0,0 +1,179 @@
...
+/**
+ * Internal API.
+ *
+ * Simplest AMQP flow. Published messages without any delivery guarantees. Result is emitted as soon as message is
+ * published. Because there it's not verified whether message is confirmed or rejected, results are emitted always
+ * as confirmed. Since result is always confirmed, it would also make sense to emit just plain `passThrough` object
+ * instead of complete [[WriteResult]] (possibly it would be less confusing for users), but [[WriteResult]] is used
+ * for consistency with other variants and to make the flow ready for any possible future [[WriteResult]] extensions.
+ */
+@InternalApi private[amqp] final class AmqpSimpleFlow[T](writeSettings: AmqpWriteSettings)
```
ennru: Please create different files for the implementations.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpFlow.scala
```diff
@@ -0,0 +1,179 @@
...
+import scala.util.{Failure, Success, Try}
+
+/**
+ * Internal API.
+ *
+ * Base stage for AMQP flows that don't use asynchronous confirmations.
+ */
+@InternalApi private sealed abstract class AbstractAmqpFlowStage[T](
+    override val settings: AmqpWriteSettings,
+    promise: Promise[Done],
```
ennru: I know it wasn't you, but wouldn't `streamCompletion` be a better name?

kstrek: Yup, good catch. Changed.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpAsyncFlow.scala
kstrek: I changed it, although I'm not sure now what is the convenion. In reference connector there are no `Stage` suffixes 😛 .

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpFlow.scala
kstrek: Done.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AmqpFlow.scala
kstrek: Done.

kstrek:
I like the idea of `FlowWithContext` but I would like to make sure how this should look like before implementing:
Do I understand it correctly that `passThrough` should be removed from `WriteMessage` completely and signature of flow stages changed from `Flow[WriteMessage[T], WriteResult[T], ...]` to `Flow[(WriteMessage, T), (WriteResult, T), ...]`? I like that approach but it's a bit different from what I see in `ElasticsearchFlow`.
Also, what should be exposed in api? Both `FlowWithContext` and tupled versions? Or just `FlowWithContext` variants? I would like to be consistent with something but I see this used only in Elasticseatch connector 😛 . It would be nice to add something like this to reference flow as it seems to be pretty common case.

ennru:
Yes, Elasticsearch got the pass-through support before we introduced `FlowWithContext`. Passing tuples is the more generic solution and the recommended way going forward.
A benefit of using tuples here is, that the `WriteMessage` doesn't change so that the existing API stays source-compatible.

kstrek:
Sorry it took so long, I flooded my computer 😩 

I've added `FlowWithContext` variants. It's implemented as a separate factories (`AmqpFlowWithContext`) to avoid strange names like `AmqpFlow.withAsyncConfirmAndContext`. Let me know what you think about that.

I had to implement java api using stages from `impl` package directly, instead of mapping over connectors from scala api, because for some reason there is no `.mapMaterializedValue` on a `FlowWithContext`. Maybe you will have an idea on how to work around this in a cleaner way.

seglo:
> I had to implement java api using stages from `impl` package directly, instead of mapping over connectors from scala api, because for some reason there is no `.mapMaterializedValue` on a `FlowWithContext`. Maybe you will have an idea on how to work around this in a cleaner way.

I asked and there's no particular reason `FlowWithContext` does not include `mapMaterializedValue`.  `FlowWithContext` is a relatively recent addition, so there may be some operations that could still be added.  There are operations that we can't add without further design discussion, such as batching or merging operations, because we don't have a good way of handling the context generically in these scenarios, but maybe we can do something about this in the future.  In the meantime you could either drop down to a normal flow (with `FlowWithContext.asFlow`) and back again, or just rewrap the custom stage as you've currently done.

ennru:
If you rebase this PR on master, you'll get the latest Akka 2.5 release which has `mapMaterializedValue` for `FlowWithContext`.

seglo:
Enno beat me to the punch! Apparently `mapMaterializedValue` has already been added (and released)  as part of Akka 2.5.24.  Alpakka master now references 2.5.26 (upgraded from 2.5.23 recently).

amqp/src/test/java/docs/javadsl/AmqpDocsTest.java
```diff
@@ -26,6 +27,8 @@
 import akka.stream.testkit.javadsl.TestSink;
 import akka.testkit.javadsl.TestKit;
 import akka.util.ByteString;
+import scala.collection.JavaConverters;
```
ennru: This looks wrong here (and unused AFAICS).

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpSink.scala
```diff
@@ -6,46 +6,40 @@ package akka.stream.alpakka.amqp.scaladsl
...
+import akka.stream.scaladsl.{Keep, Sink}
 import akka.util.ByteString
 
 import scala.concurrent.Future
 
 object AmqpSink {
 
   /**
-   * Scala API: Creates an [[AmqpSink]] that accepts ByteString elements.
+   * Creates an [[AmqpSink]] that accepts [[WriteMessage]] elements.
```
ennru: Scaladoc requires fully-qualified class names even if the class is imported.

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpFlow.scala
```diff
@@ -0,0 +1,102 @@
...
+  def withAsyncUnorderedConfirm[T](
+      settings: AmqpWriteSettings,
+      bufferSize: Int,
+      confirmationTimeout: FiniteDuration
+  ): Flow[WriteMessage, WriteResult, Future[Done]] =
+    asFlowWithoutContext(
+      Flow.fromGraph(new impl.AmqpAsyncUnorderedFlowStage(settings, bufferSize, confirmationTimeout))
+    )
+
+  private def asFlowWithoutContext(flow: Flow[(WriteMessage, NotUsed), (WriteResult, NotUsed), Future[Done]]) =
```
ennru: I think you are the first to turn this around so that `withContext` is the default thing to do. Great!

amqp/src/main/scala/akka/stream/alpakka/amqp/javadsl/AmqpSink.scala
```diff
@@ -6,44 +6,43 @@ package akka.stream.alpakka.amqp.javadsl
...
 import akka.stream.alpakka.amqp._
 import akka.util.ByteString
 
+import scala.compat.java8.FutureConverters._
+
 object AmqpSink {
 
   /**
-   * Java API: Creates an [[AmqpSink]] that accepts [[akka.stream.alpakka.amqp.WriteMessage WriteMessage]] elements.
+   * Creates an [[AmqpSink]] that accepts [[WriteMessage]] elements.
```
ennru: Scaladoc requires fully-qualified class names (even though IDE's are happy without).

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpFlowWithContext.scala
```diff
@@ -0,0 +1,68 @@
...
+   * @see [[AmqpFlow.withAsyncUnorderedConfirm]]
+   */
+  def withAsyncUnorderedConfirm[T](
+      settings: AmqpWriteSettings,
+      bufferSize: Int,
+      confirmationTimeout: FiniteDuration
+  ): FlowWithContext[WriteMessage, T, WriteResult, T, Future[Done]] =
+    FlowWithContext.fromTuples(
+      Flow.fromGraph(new impl.AmqpAsyncUnorderedFlowStage(settings, bufferSize, confirmationTimeout))
+    )
```
ennru: Hm, `FlowWithContext` is supposed to always keep message order. So I guess the unordered flow stage should not support context passing.
A motivation for this criteria are Kafka offsets which should not be reordered.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AbstractAmqpAsyncFlowStageLogic.scala
```diff
@@ -0,0 +1,213 @@
...
+@InternalApi private object AbstractAmqpAsyncFlowStageLogic {
+  type DeliveryTag = Long
+}
+
+/**
+ * Internal API.
+ *
+ * Base stage for AMQP flows with asynchronous confirmations.
+ */
+@InternalApi private abstract class AbstractAmqpAsyncFlowStageLogic[T](
```
ennru: Quite an impressive machinery in here! Great work.

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpFlowWithContext.scala
kstrek: Well, I never used Kafka connector, but here are my thoughts on that: it seems strange to enforce such ordering constraints since `FlowWithContext` can be used there to carry arbitrary objects, not only offsets (if I understand correctly). Offsets are a fundamental concept in Kafka so I would expect users to be familiar with it and having "unordered" somewhere in a name would be enough warning sign to check if that can be used with offsets (warning could be also included in docs ofc).

Nevertheless, regardless of Kafka, I don't see the reason to enforce such convention across all alpakka connectors. If I understand correctly `FlowWithContext` in general is not meant to preserve order (nothing about that in akka streams docs anyway) so unordered variants should not be suprising for users. Unordered flows are a quick way to boost performance and it would be sad to not have them with context 😉. 

Ofc I do not insist on that and I can remove it if you want so.

amqp/src/main/scala/akka/stream/alpakka/amqp/javadsl/AmqpSink.scala
kstrek: I changed it to backticks, as you recommended for `AmqpFlow` in one of the previous comments.

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpSink.scala
kstrek: I changed it to backticks, as you recommended for AmqpFlow in one of the previous comments.

amqp/src/test/java/docs/javadsl/AmqpDocsTest.java
kstrek: Right, removed.

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpFlowWithContext.scala
ennru: You're right, the Akka documentation around `FlowWithContext` does hardly exist. It's in our plans to improve on it. Even for a normal Akka Stream keeping order is expected unless you go unordered explicitly (as you do here).

Taking a step back: In which scenarios does AMQP reply out of order to the published messages? Are you sure it provides significantly better performance so it is worth to have the unordered variants?
If it does, you may expose the `withAsyncUnorderedConfirm` as a tuple flow (which can easily be wrapped by the user).

Similarly, is the distinction between `withConfirm` and `withAsyncConfirm` really important? 
Wouldn't a buffer size of 1 effectively give `withConfirm` behaviour? The buffer size could become a value in the `AmqpWriteSettings`.
The `confirmationTimeout` could be moved to the settings, as well.

kstrek: Looking at the RabbitMQ docs right now to answer your questions and I just realized that confirmation mechanism I used is not a part of AMQP protocol but a RabbitMQ-specific extension to it. I guess that disqualifies all variants except for basic one that don't use it. Am I right?


Anyway, my thoughts on your concerns:

When exactly messages are confirmed by broker depends on few things. It's described shortly in [docs](https://www.rabbitmq.com/confirms.html#when-publishes-are-confirmed), but most notably for persistent messages acks are sent only after message is persisted to disk (which is done in batches, so the latency here may vary) and for mirrored queues messages are ack'ed if all mirrors accept the message, so the network unreliability is involved here. I can imagine scenario where I would expect unordered flow would be significantly more performant, but I admit I haven't done any tests to prove that.
I am not pleased by inconsistency of having unordered variant as a tuple flow, but if there is no other way then I'm ok with that 😉.

Indeed, `withAsyncConfirm` with buffer size of 1 should give something similar as `withConfirm`. I haven't thought about that. It can be removed.

I didn't want to add buffer size and timeout to `AmqpWriteSettings` because now it contains settings common to all flows and sinks, while those two are more specific. I saw the same approach used in `AmqpRpcFlow` as well.

ennru: Sorry for my late reply.

> I guess that disqualifies all variants except for basic one that don't use it. Am I right?

The RabbitMQ specific thing is a problem. But I wonder if anything but RabbitMQ is used somewhere... I assume it is the most common AMQP broker so we might mark those flows as specific to RabbitMQ. But we need to make sure the existing ones are not broken.

> I am not pleased by inconsistency of having unordered variant as a tuple flow, but if there is no other way then I'm ok with that 😉.

I understand that, but for now, it's the best solution I see.

> I didn't want to add buffer size and timeout to AmqpWriteSettings because now it contains settings common to all flows and sinks

That is always a balance, but settings are the solution to add more later without breaking the API. It is better to have settings that are not used in certain factories.

kstrek: I removed the `withConfirm` variants and adjusted the namings. Unordered flow with context is removed and replaced by flow of pairs, as discussed. I moved the `bufferSize` and `confirmationTimeout` to `WriteSettings` but I didn't apply this to rpc flow - I don't know what are the plans to release this so I didn't want to introduce breaking change there.

Please check out the changes.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AbstractAmqpAsyncFlowStageLogic.scala
```diff
@@ -30,6 +30,9 @@ import scala.concurrent.duration.FiniteDuration
  */
 @InternalApi private object AbstractAmqpAsyncFlowStageLogic {
   type DeliveryTag = Long
+  val DefaultBufferSize: Int = 10
+  val DefaultConfirmationTimeout: FiniteDuration = 100.millis
```
ennru: The defaults should go in `AmqpConnectorSettings` so they are always defined instead.

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpFlow.scala
```diff
@@ -61,14 +43,14 @@ object AmqpFlow {
    *
    * This stage materializes to a `Future` of `Done`, which can be used to know when the Flow completes,
    * either normally or because of an amqp failure.
+   *
+   * @param settings default value is 20 for `bufferSize` and 100ms for `confirmationTimeout`
```
ennru: Risky to have the values in the docs, AFAICS you changed the buffer size to 10 already.

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
```diff
@@ -208,18 +212,34 @@ final class AmqpWriteSettings private (
...
+                   confirmationTimeout: Option[FiniteDuration] = confirmationTimeout) =
+    new AmqpWriteSettings(connectionProvider, exchange, routingKey, declarations, bufferSize, confirmationTimeout)
 
   override def toString: String =
     "AmqpSinkSettings(" +
     s"connectionProvider=$connectionProvider, " +
     s"exchange=$exchange, " +
     s"routingKey=$routingKey, " +
     s"declarations=$declarations" +
+    s"bufferSize=$bufferSize" +
```
ennru: Add commas to separate the values.

docs/src/main/paradox/amqp.md
```diff
@@ -54,6 +56,33 @@ Scala
 Java
 : @@snip [snip](/amqp/src/test/java/docs/javadsl/AmqpDocsTest.java) { #create-sink }
 
+### With flow
```
ennru: Put the flow variant first in the docs. Sinks are not as important when flows are available.

amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AbstractAmqpAsyncFlowStageLogic.scala
kstrek: I wanted to make the values in `AmqpWriteSettings` optional to indicate they are sometimes ignored, but indeed, when I moved them to settings it looks cleaner.

amqp/src/main/scala/akka/stream/alpakka/amqp/scaladsl/AmqpFlow.scala
kstrek: Right, good catch 😉

amqp/src/main/scala/akka/stream/alpakka/amqp/AmqpConnectorSettings.scala
kstrek: Done.

docs/src/main/paradox/amqp.md
kstrek: Done.

amqp/src/main/mima-filters/1.1.x.backwards.excludes
```diff
@@ -0,0 +1,3 @@
+# PR #1921
+# new argument to private constructor
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.amqp.AmqpWriteSettings.this")
```
ennru: As we've now released 2.0.0-M1, you need to rename this file to `2.0.0-M1.backwards.excludes`.

amqp/src/main/scala/akka/stream/alpakka/amqp/javadsl/AmqpFlow.scala
```diff
@@ -0,0 +1,108 @@
...
+   * confirmation) before propagating failure/finish downstream.
+   *
+   * This stage materializes to a `CompletionStage` of `Done`, which can be used to know when the Flow completes,
+   * either normally or because of an amqp failure.
+   *
+   * NOTE: This connector uses RabbitMQ's extension to AMQP protocol
+   * ([[https://www.rabbitmq.com/confirms.html#publisher-confirms Publisher Confirms]]), therefore it is not
+   * supposed to be used with another AMQP brokers.
+   */
+  def createWithUnorderedConfirm(
```
ennru: I'd prefer `createWithConfirmUnordered` as the `unordered` is more a postfix marker to the method.

amqp/src/main/mima-filters/1.1.x.backwards.excludes
kstrek: Done.

amqp/src/main/scala/akka/stream/alpakka/amqp/javadsl/AmqpFlow.scala
kstrek: Right, and it seems more in line with naming of `map` operators in akka-streams.

ennru:
Thank you @kstrek! This is quite a bunch of great work with enormous improvements to the Alpakka AMQP usefulness.

Issue #1919, opened 2019-09-08
S3 1.1.1 doesn't render custom endpoint correctly for getObjectMetadata


relgames:
I'm using a custom s3 implementation and trying to use getObjectMetadata().
It fails on alpakka s3 1.1.1 and 1.1.0 but works on 1.0.2. 

### Versions used 

Akka version: 2.5.23
Akka http version: 10.1.8
Alpakka s3 version: 1.1.1 or 1.1.0

### Expected Behavior

getObjectMetadata() succeeds for custom endpointUrl

### Actual Behavior

getObjectMetadata() fails for custom endpointUrl

### Relevant logs

With trace logging enabled, it renders HEAD request including a custom endpoint:
```
vax.net.ssl|DEBUG|1F|default-akka.actor.default-dispatcher-17|2019-09-08 17:33:23.454 CEST|SSLCipher.java:1727|Plaintext before ENCRYPTION (
  0000: 48 45 41 44 20 68 74 74   70 73 3A 2F 2F 73 33 2E  HEAD https://s3.
  0010: 6E 6C 2D 61 6D 73 2E 73   63 77 2E 63 6C 6F 75 64  nl-ams.scw.cloud
  0020: 2F 66 73 2D 63 6F 70 79   2E 65 61 74 65 6C 6C 69  /fs-copy.eatelli
  0030: 61 2E 63 6F 6D 2F 31 2F   74 68 75 6D 62 5F 36 30  a.com/1/thumb_60
  0040: 30 2E 6A 70 67 20 48 54   54 50 2F 31 2E 31 0D 0A  0.jpg HTTP/1.1..
```  

So it renders full endpoint as part of HEAD request. While it should only render relative path, not the full path. It works fine in v1.0.2

ennru:
Thank you for this report with a proper investigation of the problem.
Would you be able to suggest a fix in a Pull request?

relgames:
I can try. Can you suggest where it is happening - rendering of endpoint into HEAD request? 
Are there unit tests so I can reproduce the issue with a test first?..

jrudolph:
Likely culprit: https://github.com/akka/alpakka/commit/3e98c1c3a5fb615c4830a6ec8c471a54d5cfedd4

ennru:
With @jrudolph's help, I have a fix incoming. Amazon S3's non-standard HTTP continues to annoy.
I guess we'll need to cut an Alpakka 1.1.2 release with the fix.

[...]

PR #1923, opened 2019-09-10
AWS S3: make raw request URI contain path and query 

ennru:
## Purpose

Just pass the URI path and query as `Raw-Request-URI` so that Amazon S3 understands Alpakka's HTTP requests.

## References

Reported in #1919
https://github.com/akka/akka-http/issues/451
#1738 
#838 

## Changes

* Commit 1 solves the actual problem
* Commit 2 makes sure the tests shut down the ActorSystems they create

## Background Context

As Amazon S3's use of HTTP is non-standard the fix in #1738 introduced a new problem. Originally discussed in https://github.com/akka/akka-http/issues/451 the Raw Request URI needs to be without request scheme and authority.


ennru:
Backported to 1.1.x with #1963

Issue #1764, opened 2019-06-19
HDFS: Build for Scala 2.13


ennru:
It requires an upgrade of Cats from 1.6 to 2.0 which is in milestone state and needs to be investigated for compatibility rules.
It is stated to be binary-compatible https://github.com/typelevel/cats/releases
Alternatively the use of Cats' `State` could be replaced to lower the dependency burden.

ennru:
Fixed with #1925

PR #1925, opened 2019-09-10
HDFS: upgrade Cats to 2.0.0 and enable Scala 2.13

ennru:
## Purpose

Make Alpakka HDFS build with Scala 2.13 by upgrading cats-core to the binary compatible 2.0.0 release.

## References

#1764


ennru:
Failures are
* #1417
* #1927 
* #1549

Issue #1902, opened 2019-08-26
PubSub http connector exposes no observability into http pull errors


jacobg:
### Versions used 

Akka version: 2.5.23

### Expected Behavior

When creating a stream with source from `GooglePubSub.subscribe(subscriptionId, config)`, if the subscription does not exist, either the stream should fail or there should be some logging or other observability into the error.

### Actual Behavior

`PubSubApi.pull` returns a `Success(PullResponse(None))`, as if there were simply no messages, and which swallows the actual 404 http error response.

The actual http response is:
```
HttpResponse(404 Not Found,List(),HttpEntity.Strict(application/json,{"error":{"code":404,"message":"Subscription does not exist","status":"NOT_FOUND"}}),HttpProtocol(HTTP/1.1))
```

The consequence is that a misconfigured application will simply not receive any events, and unless there is some metrics alert on message throughput through which the problem can be inferred, then there is no observability of the problem.

### Tagging people
@stephennancekivell 
@2m 
@ennru

ennru:
Thank you for reporting this. I guess it took you a while before you understood what's going on, sorry for that.
Would you be in a position to suggest an improvement in a Pull Request?

LGLO:
Can I try to provide some solution?

ennru:
Sure, go ahead.

PR #1928, opened 2019-09-10
Google Pub/Sub: Fail source when pull status code is not 200 OK

LGLO:
## Purpose
Changes behavior for pull subscribe stream.
Before: error responses were successfully Unmarshalled to empty PullResponse, giving no signal that errors occurred.
That was because Google PubSub error response bodies are valid JSON messages and PullResponse contains only one optional field.
After: when status code for pull request is not 200 OK, then source will fail with exception that contain no OK response status.

## References
References #1902

## Background Context
I've used technique similar to current handling of failed publish requests.

### DIFF
--- a/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApi.scala
+++ b/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApi.scala
@@ -141,8 +141,13 @@ private[pubsub] trait PubSubApi {
       request <- Marshal((HttpMethods.POST, uri, request)).to[HttpRequest]
       response <- doRequest(request, maybeAccessToken)
       pullResponse <- response.status match {
-        case StatusCodes.OK => Unmarshal(response).to[PullResponse]
-        case status => Future.failed(new RuntimeException(s"Unexpected publish response. Code: [${status}]"))
+        case StatusCodes.Success(_) => Unmarshal(response).to[PullResponse]
+        case status =>
+          Unmarshal(response)
+            .to[String]
+            .map { entity =>
+              throw new RuntimeException(s"Unexpected pull response. Code: [$status]. Entity: [$entity]")
+            }
       }
     } yield pullResponse
   }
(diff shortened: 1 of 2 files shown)


google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApi.scala
```diff
@@ -140,7 +140,10 @@ private[pubsub] trait PubSubApi {
     for {
       request <- Marshal((HttpMethods.POST, uri, request)).to[HttpRequest]
       response <- doRequest(request, maybeAccessToken)
-      pullResponse <- Unmarshal(response).to[PullResponse]
+      pullResponse <- response.status match {
+        case StatusCodes.OK => Unmarshal(response).to[PullResponse]
+        case status => Future.failed(new RuntimeException(s"Unexpected publish response. Code: [${status}]"))
```
LGLO: Google PubSub sometimes returns quite nice explanations in body of response.
I was wondering about `Unmarshal.to[String]` for non OK responses and adding this information to exception message. WDYT?

ennru: Yes, more error context is always a plus.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApi.scala
```diff
@@ -140,7 +140,15 @@ private[pubsub] trait PubSubApi {
...
-      pullResponse <- Unmarshal(response).to[PullResponse]
+      pullResponse <- response.status match {
+        case StatusCodes.Success(_) => Unmarshal(response).to[PullResponse]
+        case status =>
+          Unmarshal(response)
+            .to[String]
+            .map { entity =>
+              throw new RuntimeException(s"Unexpected pull response. Code: [$status]. Entity: [$entity]")
+            }
+      }
```
LGLO: I'm using same pattern in pull, acknowledge and publish.
I'm matching for every success status code. If it's not successful, then trying to read error as String to add it to an exception. It works for response with no entity, returning empty string (checked in added spec for acknowledge).

ennru:
Thank you!

PR #1932, opened 2019-09-13
GCS add versions support for listing, add generation support fo…

### LINKED ISSUE #1922: GCS listBucket should have an option to list all versions
### Short description

When listing a versioned bucket in GCS, an option to list all the versions would be useful as an aid to programmatically selecting a particular version

### Details

As per the [google docs](https://cloud.google.com/storage/docs/using-object-versioning#list) it looks like this can be achieved by adding `versions=true` as a query parameter.

We would be willing to work on a PR for this if it is considered a desirable feature.

RomanLebid:
## Purpose

This PR adds support for listing both live and archived versions of objects in a bucket if versioning is enabled. It also adds support to select a generation of object to get, download or delete.

Fixes #1922 

## References

Using Object Versioning
https://cloud.google.com/storage/docs/using-object-versioning

## Changes

Added `Boolean` parameter `versions` with default value of false to `listBucket` method.
Added `Long` optional parameter `generation` to `getObject`, `deleteObject` and `download` methods.

The same applies to Java API, but using overloaded methods instead of default parameters.

## Background Context

It is currently unavailable to list both live and archived version of GCS bucket if versioning is enabled. And if you list archived versions as well, there is no way to interact with those.

Instead of wrapping google Java API to do that, using Alpakka for that purpose would be useful.

### DIFF
--- a/google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/impl/GCStorageStream.scala
+++ b/google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/impl/GCStorageStream.scala
@@ -87,7 +87,7 @@ import scala.util.control.NonFatal
   def deleteBucket(bucketName: String)(implicit mat: Materializer, attr: Attributes): Future[Done] =
     deleteBucketSource(bucketName).withAttributes(attr).runWith(Sink.head)
 
-  def listBucket(bucket: String, prefix: Option[String], versions: Boolean): Source[StorageObject, NotUsed] = {
+  def listBucket(bucket: String, prefix: Option[String], versions: Boolean = false): Source[StorageObject, NotUsed] = {
     sealed trait ListBucketState
     case object Starting extends ListBucketState
     case class Running(nextPageToken: String) extends ListBucketState
@@ -101,7 +101,7 @@ import scala.util.control.NonFatal
         Map(
           pageToken.map(token => "pageToken" -> token).toList :::
           prefix.map(pref => "prefix" -> pref).toList :::
-          List("versions" -> s"$versions").filter(_ => versions): _*
+          (if (versions) List("versions" -> true.toString) else Nil): _*
         )
 
       makeRequestSource(
@@ -136,7 +136,9 @@ import scala.util.control.NonFatal
       .mapMaterializedValue(_ => NotUsed)
   }
 
-  def getObject(bucket: String, objectName: String, generation: Option[Long]): Source[Option[StorageObject], NotUsed] =
+  def getObject(bucket: String,
+                objectName: String,
+                generation: Option[Long] = None): Source[Option[StorageObject], NotUsed] =
     Setup
       .source { implicit mat => implicit attr =>
         makeRequestSource(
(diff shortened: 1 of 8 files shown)


google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/impl/GCStorageStream.scala
```diff
@@ -106,8 +99,9 @@ import scala.util.control.NonFatal
       import mat.executionContext
       val queryParams =
         Map(
-          pageToken.map(token => "pageToken" -> token).toList ++
-          prefix.map(pref => "prefix" -> pref).toList: _*
+          pageToken.map(token => "pageToken" -> token).toList :::
+          prefix.map(pref => "prefix" -> pref).toList :::
+          List("versions" -> s"$versions").filter(_ => versions): _*
```
ennru: What about `(if (versions) List("versions" -> true.toString) else Nil)`?

google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/impl/GCStorageStream.scala
```diff
@@ -142,36 +136,41 @@ import scala.util.control.NonFatal
...
         ).mapAsync(parallelism)(entityForSuccessOption)
           .map(option => option.isDefined)
       }
       .mapMaterializedValue(_ => NotUsed)
 
   def deleteObjectsByPrefixSource(bucket: String, prefix: Option[String]): Source[Boolean, NotUsed] =
-    listBucket(bucket, prefix)
-      .flatMapConcat(listBucketResult => deleteObjectSource(bucket, listBucketResult.name))
+    listBucket(bucket, prefix, versions = false)
+      .flatMapConcat(listBucketResult => deleteObjectSource(bucket, listBucketResult.name, None))
```
ennru: Might be good to use `generation = None` for these.

RomanLebid: I also thought of that, but wouldn't it break the binary compatibility?
or we can just add this to exclusion rules, since those are parts of `InternalApi`?

google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/impl/GCStorageStream.scala
RomanLebid: yeah, I agree, it's a bit more readable this way

google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/impl/GCStorageStream.scala
ennru: I mean to add the parameter name at use-side, that won't change the API. But yes, `impl` is automatically excluded from Mima checks as it is always considered internal API.

ennru:
Thank you, great to have you here!

PR #1933, opened 2019-09-14
sbt 1.3.0

sullis:
## References

References #xxxx


ennru:
When going to sbt 1.3, we need to make sure Travis caches are used in a good way.
We have a minor issue with the Paradox dependency listing which doesn't show the licenses when used with sbt 1.3.

sullis:
obsolete

Issue #1934, opened 2019-09-16
Cassandra: not consuming the entire resultset in all cases


anzecesar:
### Versions used 

Alpakka 1.1.1

Akka version: 2.5.25

### Expected Behavior

When using `setFetchSize`, the source should always load all pages

### Actual Behavior

Under certain conditions, which I find hard to reproduce, it seems like the source completes just after loading a few pages of results. Looks like that happens when the page size is small and downstream consumer is fast.

In my tests I needed to have a small page size due to larger response size.

I'll open a PR with a fix

ennru:
Fixed with #1935

PR #1935, opened 2019-09-16
Cassandra: check `isFullyFetched` before completion

anzecesar:
## Purpose

Under some conditions, the source completed before all results were loaded from the result set.

Now it only completes when rs is fully fetched and all elements were emitted. Before it was completing in case where the current fetched buffer was empty.

## References

References https://github.com/akka/alpakka/issues/1934

## Changes

* only completes the stream when `rs.isFullyFetched`
* if just the currently loaded set is empty, it fetches more (an edge case)

## Background Context

I observed this issue where fetchSize was set relatively low and the
downstream consumer was fast.

### DIFF
--- a/cassandra/src/main/scala/akka/stream/alpakka/cassandra/impl/CassandraSourceStage.scala
+++ b/cassandra/src/main/scala/akka/stream/alpakka/cassandra/impl/CassandraSourceStage.scala
@@ -36,6 +36,14 @@ import scala.util.{Failure, Success, Try}
       }(exec)
     }
 
+    private[this] def fetch(rs: ResultSet): Unit = {
+      isFetching = true
+      // fetch next page
+      val gFut = rs.fetchMoreResults()
+      val exec = materializer.executionContext
+      GuavaFutures.invokeTryCallback(gFut, exec)(futFetchedCallback)
+    }
+
     setHandler(
       out,
       new OutHandler {
@@ -43,11 +51,7 @@ import scala.util.{Failure, Success, Try}
           val currentlyAvailableRows = rs.getAvailableWithoutFetching
 
           if (!isFetching && currentlyAvailableRows < minimumPreFetchSize) {
-            isFetching = true
-            // fetch next page
-            val gFut = rs.fetchMoreResults()
-            val exec = materializer.executionContext
-            GuavaFutures.invokeTryCallback(gFut, exec)(futFetchedCallback)
+            fetch(rs)
           }
 
           if (currentlyAvailableRows > 0)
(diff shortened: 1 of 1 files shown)


anzecesar:
@ennru I can try, but this was already hard to reproduce. Are there any existing tests with cassandra already?

ennru:
Yes, they use Cassandra via Docker compose.

anzecesar:
@ennru Where can I find them? There's nothing in https://github.com/akka/alpakka/tree/master/cassandra/src/test

ennru:
https://github.com/akka/alpakka/blob/master/cassandra/src/test/scala/docs/scaladsl/CassandraSourceSpec.scala

anzecesar:
@ennru Tnx. For some reason I assumed the purpose of that was more or less just for the docs and haven't even considered it a working spec 🤷‍♂️. Well, we all know where assumption leads 😂.

I'll try to add a test case for this although I suspect it will be at best very flaky.

anzecesar:
I finally got some time to work on this today and I was not able to reproduce this locally with a spec. The conditions under which I originally encountered the issue were reading ~10k rows containing quite large jsons (around 150 mb for the whole dataset). I'm not exactly sure what exactly led to the buffer being exhausted prematurely.

There is however an exiting test, that generally ensures this doesn't break any functionality: https://github.com/akka/alpakka/blob/aafad95ffa49a08c99342bd64752a2e4f1cae399/cassandra/src/test/scala/docs/scaladsl/CassandraSourceSpec.scala#L107

cassandra/src/main/scala/akka/stream/alpakka/cassandra/impl/CassandraSourceStage.scala
```diff
@@ -36,18 +36,22 @@ import scala.util.{Failure, Success, Try}
       }(exec)
     }
 
+    private[this] def fetch(rs: ResultSet): Unit = {
+      isFetching = true
+      // fetch next page
+      val gFut = rs.fetchMoreResults()
+      val exec = materializer.executionContext
+      GuavaFutures.invokeTryCallback(gFut, exec)(futFetchedCallback)
```
chbatey: should this be using an async call back?

ennru: It looks OK to me, it is only called within synchronous code. 
`tryPushAfterFetch` is called via the async handler `futFetchedCallback` when the future returns.

anzecesar: Yes, I just refactored it out into a method to not duplicate code.

ennru:
Thank you for this fix.

Issue #1909, opened 2019-08-30
[Failed] File: docs.javadsl.ArchiveTest.flowShouldCreateZIPArchive


ennru:
```
[error] Test docs.javadsl.ArchiveTest.flowShouldCreateZIPArchive failed: java.lang.AssertionError: expected:<ByteString(80, 75, 3, 4, 20, 0, 8, 8, 8, 0, 8, -100, 29, 79, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 19, 0, 0, 0, 97, 107, 107, 97, 95, 102, 117, 108, 108, 95, 99, 111, 108, 111, 114, 46, 115, 118, 103, -19, 85, 73, 110, 27, 49, 16, -4, 74, 99, 114, 110, -118, -5, 18, 88, 2, -30, 32, 71, 61, -62, -98, 40, -110, -31, -79, 108, 88, -54, -56, -48, -21, 83, 69, -59, -122, 125, 10, -112, 91, -128, 0, -94, -72, -11, 82, -84, 46, 114, -82)... and [1078] more> but was:<ByteString(80, 75, 3, 4, 20, 0, 8, 8, 8, 0, 9, -100, 29, 79, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 19, 0, 0, 0, 97, 107, 107, 97, 95, 102, 117, 108, 108, 95, 99, 111, 108, 111, 114, 46, 115, 118, 103, -19, 85, 73, 110, 27, 49, 16, -4, 74, 99, 114, 110, -118, -5, 18, 88, 2, -30, 32, 71, 61, -62, -98, 40, -110, -31, -79, 108, 88, -54, -56, -48, -21, 83, 69, -59, -122, 125, 10, -112, 91, -128, 0, -94, -72, -11, 82, -84, 46, 114, -82)... and [1078] more>, took 0.371 sec

[error]     at docs.javadsl.ArchiveTest.flowShouldCreateZIPArchive(ArchiveTest.java:97)

[error]     ...
```

https://travis-ci.com/akka/alpakka/jobs/229397111#L657

michalbogacz:
It seems that in all error cases, the difference between ZIP files is in metadata of ZIP. I'm not sure why it happens. 
But I can change this test to create ZIP file, then unzip it, and compare the source files with the unzipped.

ennru:
That would be great.

ennru:
[...]

PR #1939, opened 2019-09-20
Scala 2.12.10; Akka 2.6-M8; Akka HTTP 10.1.10

ennru:
* Use Scala 2.12.10
* Use Akka 2.6.0-M8
* Use Akka HTTP 10.1.10 to get an Akka 2.6 compatible version

(Scala 2.13.1 lacks the Silencer plugin at the moment.)


ennru:
Failures were
* https://github.com/akka/alpakka/issues/1942
* https://github.com/akka/alpakka/issues/1909
* #1549 
* #1649

ennru:
For the cron builds it is important to get the Akka HTTP upgrade in.

project/Common.scala
```diff
@@ -39,7 +39,7 @@ object Common extends AutoPlugin {
       whitesourceGroup := Whitesource.Group.Community,
       crossVersion := CrossVersion.binary,
       crossScalaVersions := Dependencies.ScalaVersions,
-      scalaVersion := Dependencies.Scala212,
+      scalaVersion := Dependencies.ScalaVersions.head,
```
seglo: This will obfuscate the actual Scala version used. The way `ScalaVersions` is built took me a little bit to figure out, but it seems in both cases when this is or isn't a cron build, the `Scala212` version will be used.

ennru: Right, I reverted this.

PR #1940, opened 2019-09-21
InfluxDB docs: fix artifact name for sbt, Maven and Gradle

laszlovandenhoek:
The documentation at https://doc.akka.io/docs/alpakka/current/influxdb.html renders the artifact name as `akka-stream-alpakka-influxdb2.12`. This should of course be `akka-stream-alpakka-influxdb_2.12` for Maven and Gradle, and just `akka-stream-alpakka-influxdb` for sbt.

I checked other Alpakka projects that do get it right, and the only difference appears to be the missing underscore.


ennru:
Thank you!

PR #1941, opened 2019-09-22
Elasticsearch: refactor to make retrying more explicit

ennru:
## Purpose

This PR introduces a simple flow stage implementation without retrying.

## References

See RetryFlow in https://github.com/akka/akka/pull/27742
DynamoDB retrying https://github.com/akka/alpakka/pull/1712

## Changes

* Introduce `ElasticsearchJsonBase`
* Add `ElasticsearchSimpleFlowStage`
* Rework the `javadsl` factory methods to use `scaladsl` implementations
* Rework the factory methods to choose stage depending on the retry settings

## Background Context

The Elasticsearch flow stage implementation is more involved than needed, as it contains retrying logic. I work on implementing generic retrying flows so that these concerns can be abstracted over in https://github.com/akka/akka/pull/27742

### DIFF
--- a/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchJsonBase.scala
+++ b/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchJsonBase.scala
@@ -1,0 +1,124 @@
+/*
+ * Copyright (C) 2016-2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.elasticsearch.impl
+
+import akka.annotation.InternalApi
+import akka.stream.alpakka.elasticsearch.Operation.{Create, Delete, Index, Update, Upsert}
+import akka.stream.alpakka.elasticsearch.{MessageWriter, WriteMessage, WriteResult}
+import akka.stream.stage.StageLogging
+import spray.json._
+
+import scala.collection.immutable
+
+/**
+ * Internal API.
+ */
+@InternalApi
+trait ElasticsearchJsonBase[T, C] {
+
+  self: StageLogging =>
+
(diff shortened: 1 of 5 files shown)


elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/scaladsl/ElasticsearchFlow.scala
```diff
@@ -141,12 +126,55 @@ object ElasticsearchFlow {
     require(settings.retryLogic == RetryNever,
             "`withContext` may not be used with retrying enabled, as it disturbs element order")
     Flow[WriteMessage[T, C]]
-      .via(createWithPassThrough(indexName, typeName, settings, writer))
+      .batch(settings.bufferSize, immutable.Seq(_)) { case (seq, wm) => seq :+ wm }
+      .via(
```
seglo: IUC you can still `.via(stageFlow)` because it has a condition handling `RetryNever`?

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchJsonBase.scala
```diff
@@ -0,0 +1,124 @@
...
+  def typeName: String
+  def versionType: Option[String]
+  def messageWriter: MessageWriter[T]
+
+  private lazy val typeNameTuple = "_type" -> JsString(typeName)
+  private lazy val versionTypeTuple: Option[(String, JsString)] = versionType.map { versionType =>
+    "version_type" -> JsString(versionType)
+  }
+
+  protected def updateJson(messages: immutable.Seq[WriteMessage[T, C]]): String =
```
seglo: [minor] Maybe `toJson`?

Issue #1936, opened 2019-09-16
XML: add ability to provide a specific XMLOutputFactory


jphelp32:
### Short description

Currently, the Xml Writer can only take the default XMLOutputFactory as configured for the JVM instance. It would be helpful if one could provide an alternate XMLOutputFactory to be used for the Xml Writer, rather than always having to take the default.

### Details

Should be fairly simple to add XMLOutputFactory as an arg to the `StreamingXmlWriter` constructor, and then add another method to `XmlWriting` as follows:
```
def writer(charset: Charset, outputFactory: XMLOutputFactory): Flow[ParseEvent, ByteString, NotUsed] =
    Flow.fromGraph(new impl.StreamingXmlWriter(charset, outputFactory))
```

ennru:
Hi @jphelp32 	
Thank you for this useful suggestion. Would you be in a position to suggest the necessary changes in a Pull Request?

jphelp32:
@ennru I'm inquiring w/my employer re: the CLA..  Not sure how difficult it will be to get that part of it done.
In the meantime I've forked alpakka and pulled it down, but having some trouble importing the project into IDEA and being able to build/test/etc. I'm not too familiar with sbt projects unfortunately.

Side note, there's a similar issue here on the parsing side. I know that it relies upon the aalto parser, but currently there is no way to set any specific configuration on that factory. And by default, the factory configuration supports DTD and external entities, which is not secure. Should I log a separate issue for this or could that be added into this one?

ennru:
If you have sbt installed it might help to run `sbt Test/update` outside of IntelliJ before importing. You could comment out the other modules in `build.sbt` before importing.

Please create a separate issue for the XML parser.

ennru:
Fixed with #1945

PR #1945, opened 2019-09-24
XML: add ability to provide a specific XMLOutputFactory

jphelp32:
## Purpose

add ability to provide a specific XMLOutputFactory when instantiating an Xml Writer

## References

References #1936


jphelp32:
@ennru I've added a scala and a java unit test. Tried to mock/spy the XMLOutputFactory w/mockito to verify that it's actually being used, but was unable to get that to work. I think b/c the implementation class is final and therefore cannot be mocked by mockito. I hope this suffices.

ennru:
Great, thank you for proposing this.

PR #1946, opened 2019-09-24
Files: used in ArchiveTest.java are now read only once

michalbogacz:
## Purpose
Make stable "ArchiveTest.java" test.

## References
References #1909

## Changes
Instead of reading files multiple times in test. Now files are kept in memory. 

## Background Context
After checking the test I decided to give a try different solution than proposed in the issue.
I suspect, something was accessing files, while the tests were running and file metadata has changed before creating reference zip. After this change, files are read from disk only once and reused for both methods of creation ZIP.


ennru:
Thank you, great to fix these spurious test failures.

Issue #1938, opened 2019-09-18
S3: Enable X-Ray tracing


vigoo:
### Short description

I'm currently working on enabling AWS X-Ray tracing of our services using Akka-HTTP and Alpakka. With the current Alpakka S3 API this is not possible as it requires injecting a new header into all requests, and also capturing at least the request ID from the response headers.

### Details
To trace S3 calls the following steps would be necessary:
- Needs some operation-specific metadata before sending the request, such as bucket name and key. This is not a problem currently as the existing API can be wrapped to capture this information.
- An `X-Amzn-Trace-Id` header has to be injected to _all_ requests to S3 (its value is generated by the X-Ray implementation outside the scope of the Alpakka S3 lib). This is currently only possible in `multipartUploadWithHeaders` not for the other operations.
- Needs to capture some information from the response such as the request ID, content length (coming as a header), and status code.

To enable injecting the header I would add an `S3Headers` parameter to _all_ public operations on the `S3` object, defaulting to an empty set of headers. 
I'm not so sure about extracting things from the response though as we should not break the current quite clean API with these unnecessary metadata. Maybe all methods could have an "advanced" version annotating the return value with these elements?

I would be happy to implement this and send in a PR if we can agree on how it should work.

jypma:
Other solutions like [kamon's jaeger tracing](https://github.com/kamon-io/kamon-jaeger) use AOP techniques for this ([kanela](https://github.com/kamon-io/kamon-akka-http/blob/master/kamon-akka-http/src/main/scala-2.12/kamon/instrumentation/akka/http/AkkaHttpServerInstrumentation.scala#L29) in their case, AspectJ would do as well).

Using AOP has the advantage that you shield the "unnecessary metadata", as you rightly say, from ordinary API callers that have no interest in the tracing. It's a cross-cutting concern, after all. Additionally, using AOP would allow you to add it to _all_ `akka-http` calls, if so desired.

The downside of AOP of course is, that it's AOP :)

ennru:
Thank you for this suggestion.
I agree with @jypma that tracing should be as little intrusive to the normal use case as possible. If we can come up with a solution that doesn't affect other users, that would good. Duplicating the API is not an option.

vigoo:
[...]

PR #1947, opened 2019-09-24
AWS S3: Custom header injection

vigoo:
## Purpose

This PR adds the possibility to add custom headers to any of the AWS S3 operations.

## References

The motivation was discussed in the following issue: https://github.com/akka/alpakka/issues/1938

## Changes

- Adds a new factory class beside the existing `S3` called `S3WithHeaders` that contains the same set of operations but modified with the capability to pass an `S3Headers` value to any of them.

## Background Context

I assumed (based on the contributor guidelines) that binary compatibility must be maintained, so I created a new API `S3WithHeaders` beside the existing one. If this would not be true, the contents of this new object could be the new `S3` instead.


lightbend-cla-validator:
Hi @vigoo,

vigoo:
CLA signed

ennru:
There are some unused imports in your code
https://travis-ci.com/akka/alpakka/jobs/238787362#L973

and this needs extra parens for Scala 2.13
https://travis-ci.com/akka/alpakka/jobs/238787363#L1249

ennru:
Please add the new methods to the existing `S3`. Binary-compatiblity is provided even when adding new methods, as long as the old signature is available and the class is not intended for extension.

vigoo:
So I should rather add a `xxxWithHeaders` or something variant beside each existing method in `S3`?

vigoo:
Ok so I realized that they can be just overloaded methods, I'm pushing the change soon

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -140,7 +142,8 @@ import akka.util.ByteString
       import mat.executionContext
       implicit val conf = resolveSettings(attr, mat.system)
 
-      signAndGetAs[ListBucketResult](HttpRequests.listBucket(bucket, prefix, token))
+      // TODO: pass the headers
```
ennru: This comment is obsolete now, right?

vigoo:
I removed the obsolete comment and added an `empty` `S3Headers` value.

s3/src/main/scala/akka/stream/alpakka/s3/S3Headers.scala
```diff
@@ -118,6 +120,12 @@ final class S3Headers private (val cannedAcl: Option[CannedAcl] = None,
  * Convenience apply methods for creation of canned acl, meta, storage class or custom headers.
  */
 object S3Headers {
+
+  /**
+   * Empty set of headers
+   */
+  val empty: S3Headers = S3Headers()
+
   def apply() = new S3Headers()
```
ennru: I'd suggest you do it like: 
```
val empty = new S3Headers()
def apply() = empty
def create() = empty
```

vigoo: Would you prefer keeping the `empty` on the API too as I did or just hide it and use in `apply` and `create`?

ennru:
Thank you, great work! We hope to see you around more.

vigoo:
Thanks for merging!

Issue #1944, opened 2019-09-23
XML: uptake aalto-xml version 1.2.2 for thread-safety fix


jphelp32:
### Short description

Uptake aalto-xml version 1.2.2.
https://github.com/FasterXML/aalto-xml/blob/master/release-notes/VERSION

### Details

aalto-xml 1.2.2 fixes a thread-safety issue where its internal buffers may be corrupted when there are multiple concurrently processing materializations of XmlParser/XmlWriter w/in the same akka stream.

ennru:
Yes, now that we are approaching Alpakka 2.0 would be a good time for this upgrade. Can you create a PR for it?

jphelp32:
pull request #1948 created

ennru:
Fixed with #1948

PR #1948, opened 2019-09-25
XML: update aalto-xml to version 1.2.2

jphelp32:
## Purpose

Updates aalto-xml to version 1.2.2
## References

References #1944 

## Background Context

aalto-xml version 1.2.2 has a thread-safety fix that is especially important when used in akka streams


ennru:
Thank you.

Issue #1942, opened 2019-09-23
[Failed] DynamoDB: ItemSpec: DynamoDB should list zero tables


ennru:
```
[info] ItemSpec:
[info] DynamoDB
[info] - should 1) list zero tables *** FAILED *** (124 milliseconds)
[info]   Buffer("RetryTest") was not empty (ItemSpec.scala:47)
```

https://travis-ci.com/akka/alpakka/jobs/237572223#L827

See #1927

ennru:
We haven't logged this failure since https://github.com/akka/alpakka/pull/1949 switched off parallel execution.
Closing.

Issue #1927, opened 2019-09-10
Failed: DyanmoDB: ItemSpec "Cannot do operations on a non-existent table"


ennru:
This test failure seems new since the upgrade to AWS SDK 2 in #1896 

```
[info] DynamoDB

[info] - should 4) put an item and read it back *** FAILED *** (164 milliseconds)
[info]   java.util.concurrent.CompletionException: software.amazon.awssdk.services.dynamodb.model.ResourceNotFoundException: Cannot do operations on a non-existent table (Service: DynamoDb, Status Code: 400, Request ID: 3c98fcfc-f213-4268-924f-ccaedefd0ac6)
[info]   at software.amazon.awssdk.utils.CompletableFutureUtils.errorAsCompletionException(CompletableFutureUtils.java:61)
```

https://travis-ci.com/akka/alpakka/jobs/232712155#L742

Most of the tests after this fail, as well.

ennru:
We haven't logged this failure since https://github.com/akka/alpakka/pull/1949 switched off parallel execution.
Closing.

PR #1949, opened 2019-09-26
DynamoDB: make tryFlow a flowWithContext

ennru:
## Purpose

Implements the `tryFlow` can be implemented much, much simpler and makes it a `FlowWithContext`.

## References

The `tryFlow` was added as part of #1896 

## Changes

* simpler implementation without Graph DSL
* rename `tryFlow` to `flowWithContext`
* change tests to run sequentially (see #1927, #1942)

## Background Context

I'm not happy with the current state of the `RetryFlow` and will update in Alpakka DynamoDb before this is released.
While working on that I found this nice simplification.

I'm not really happy with exposing Scala's `Try` in the Java API.

### DIFF
--- a/dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/javadsl/DynamoDb.scala
+++ b/dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/javadsl/DynamoDb.scala
@@ -9,7 +9,7 @@ import java.util.concurrent.CompletionStage
 import akka.NotUsed
 import akka.stream.Materializer
 import akka.stream.alpakka.dynamodb.{scaladsl, DynamoDbOp, DynamoDbPaginatedOp}
-import akka.stream.javadsl.{Flow, Sink, Source}
+import akka.stream.javadsl.{Flow, FlowWithContext, Sink, Source}
 import software.amazon.awssdk.core.async.SdkPublisher
 import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient
 import software.amazon.awssdk.services.dynamodb.model.{DynamoDbRequest, DynamoDbResponse}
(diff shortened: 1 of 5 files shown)


dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/javadsl/DynamoDb.scala
```diff
@@ -32,32 +32,21 @@ object DynamoDb {
...
       client: DynamoDbAsyncClient,
       operation: DynamoDbOp[In, Out],
       parallelism: Int
-  ): Flow[akka.japi.Pair[In, State], akka.japi.Pair[Try[Out], State], NotUsed] =
-    Flow
-      .create[akka.japi.Pair[In, State]]()
-      .map(func(p => (p.first, p.second)))
-      .via(scaladsl.DynamoDb.tryFlow(parallelism)(client, operation))
-      .map(func(t => akka.japi.Pair.create(t._1, t._2)))
+  ): FlowWithContext[In, Ctx, Try[Out], Ctx, NotUsed] =
```
seglo: I saw your note about using `Try` in the Java DSL. How come we even expose a `Try[Out]`? We could leave it up to the user to recover the stream.  Another option is to use `Either`, and then add a simple Alpakka implementation (and propose it upstream in Akka's `japi` package).  The only other Java alternatives I can think of are `Optional`, which loses information, or `japi.Pair` which seems like the wrong fit.

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/scaladsl/DynamoDb.scala
```diff
@@ -31,48 +32,29 @@ object DynamoDb {
...
+   * The context is merely passed through to the emitted element.
    *
    * @param parallelism maximum number of in-flight requests at any given time
-   *
-   * @tparam State the type of the pass-through value that is taken together with
-   *               requests and is emitted together with responses. Can be used,
-   *               for example:
-   *                 * to correlate a request with a response
-   *                 * to track number of retries
-   *                 * to store request to be issued in the case of retry
```
seglo: I like the explanation here.  It may help motivate people to try `withContext`.  Any reason why you truncated it?

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/javadsl/DynamoDb.scala
ennru: It is this new API that exposes the `Try`, so we might remove it before releasing. Offering a Java counterpart might be a solution.

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/scaladsl/DynamoDb.scala
ennru: These examples are not relevant with the simpler RetryFlow I'm working on.

seglo: Aren't the examples are applicable for users who use `withContext`, regardless of the `RetryFlow`?

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/javadsl/DynamoDb.scala
seglo: The old API still had an output type of `akka.japi.Pair[Try[Out], State]`, which makes sense I guess since it was called the `tryFlow`.

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/scaladsl/DynamoDb.scala
ennru: Just the first example is, and that should be documented in the `withContext` docs.

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/javadsl/DynamoDb.scala
ennru: Ah, that "old API" was recently added by @2m, it wasn't part of an Alpakka release.

seglo: Oh, I see.  Ok.

dynamodb/src/main/scala/akka/stream/alpakka/dynamodb/scaladsl/DynamoDb.scala
seglo: Ok.  Makes sense.

PR #1953, opened 2019-09-30
FileTailSource examples when file deleted or idle

seglo:
## Purpose

When a file is deleted that is being tailed with `FileTailSource` the source will do nothing.  The underlying API used to stream files from the filesystem uses [`AsynchronousFileChannel`](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/AsynchronousFileChannel.html).  The `FileTailSource` will issue an asynchronous read request to the channel which wil callback to the `CompletionHandler` successfully even if the file is deleted.  The handler will signal that no bytes were read with the `-1` special case that we use to trigger a new poll.  ~~Instead of repolling indefinitely, this PR checks if the file exists first.  If it does not exist, then we successfully complete the stage.~~

This PR adds examples to the docs to demonstrate how to shutdown the stream if the file is deleted or if it's been idle for an interval of time.

## References

* [`AsynchronousFileChannel.read`](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/AsynchronousFileChannel.html#read-java.nio.ByteBuffer-long-)

## Changes

* Docs example to shutdown `FileTailSource` when file is deleted
* Docs example to shutdown `FileTailSource` when file has been idle for a period of time.
* New tests

### DIFF
--- a/file/src/main/java/akka/stream/alpakka/file/impl/FileTailSource.java
+++ b/file/src/main/java/akka/stream/alpakka/file/impl/FileTailSource.java
@@ -111,11 +111,15 @@ public final class FileTailSource extends GraphStage<SourceShape<ByteString>> {
                 (tryInteger) -> {
                   if (tryInteger.isSuccess()) {
                     int readBytes = tryInteger.get();
+                    // when the number of bytes read is -1 it signals that no new data is available
                     if (readBytes > 0) {
                       buffer.flip();
                       push(out, ByteString.fromByteBuffer(buffer));
                       position += readBytes;
                       buffer.clear();
+                    } else if (!Files.exists(path)) {
+                      // when the file no longer exists complete the stream
+                      this.completeStage();
                     } else {
                       // hit end, try again in a while
                       scheduleOnce("poll", pollingInterval);
--- a/file/src/main/java/akka/stream/alpakka/file/impl/FileTailSource.java
+++ b/file/src/main/java/akka/stream/alpakka/file/impl/FileTailSource.java
@@ -4,10 +4,13 @@
 
 package akka.stream.alpakka.file.impl;
 
+import akka.NotUsed;
 import akka.annotation.InternalApi;
 import akka.stream.Attributes;
 import akka.stream.Outlet;
 import akka.stream.SourceShape;
+import akka.stream.javadsl.Framing;
+import akka.stream.javadsl.Source;
 import akka.stream.stage.*;
 import akka.util.ByteString;
 import scala.concurrent.duration.FiniteDuration;
@@ -19,6 +22,8 @@ import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.channels.AsynchronousFileChannel;
 import java.nio.channels.CompletionHandler;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
@@ -41,8 +46,6 @@ public final class FileTailSource extends GraphStage<SourceShape<ByteString>> {
   private final int maxChunkSize;
   private final long startingPosition;
   private final FiniteDuration pollingInterval;
-  private final String pollTimer = "poll";
-  private final String fileExistsCheckTimer = "fileCheck";
   private final Outlet<ByteString> out = Outlet.create("FileTailSource.out");
   private final SourceShape<ByteString> shape = SourceShape.of(out);
 
@@ -108,20 +111,14 @@ public final class FileTailSource extends GraphStage<SourceShape<ByteString>> {
                 (tryInteger) -> {
                   if (tryInteger.isSuccess()) {
                     int readBytes = tryInteger.get();
-                    // when the number of bytes read is -1 it signals that no new data is available
                     if (readBytes > 0) {
-                      if (isTimerActive(fileExistsCheckTimer)) {
-                        cancelTimer(fileExistsCheckTimer);
-                      }
                       buffer.flip();
                       push(out, ByteString.fromByteBuffer(buffer));
                       position += readBytes;
                       buffer.clear();
                     } else {
                       // hit end, try again in a while
-                      scheduleOnce(pollTimer, pollingInterval);
-                      // check to see if the file still exists
-                      scheduleOnce(fileExistsCheckTimer, pollingInterval);
+                      scheduleOnce("poll", pollingInterval);
                     }
 
                   } else {
@@ -132,21 +129,11 @@ public final class FileTailSource extends GraphStage<SourceShape<ByteString>> {
 
       @Override
       public void onTimer(Object timerKey) {
-        if (timerKey == pollTimer) {
-          doPull();
-        } else if (timerKey == fileExistsCheckTimer) {
-          fileExistsCheck();
-        }
+        doPull();
       }
 
       private void doPull() {
         channel.read(buffer, position, chunkCallback, completionHandler);
-      }
-
-      private void fileExistsCheck() {
-        if (!Files.exists(path)) {
-          this.completeStage();
-        }
       }
 
       @Override
--- a/file/src/main/java/akka/stream/alpakka/file/impl/FileTailSource.java
+++ b/file/src/main/java/akka/stream/alpakka/file/impl/FileTailSource.java
@@ -4,13 +4,10 @@
 
 package akka.stream.alpakka.file.impl;
 
-import akka.NotUsed;
 import akka.annotation.InternalApi;
 import akka.stream.Attributes;
 import akka.stream.Outlet;
 import akka.stream.SourceShape;
-import akka.stream.javadsl.Framing;
-import akka.stream.javadsl.Source;
 import akka.stream.stage.*;
 import akka.util.ByteString;
 import scala.concurrent.duration.FiniteDuration;
@@ -22,8 +19,6 @@ import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.channels.AsynchronousFileChannel;
 import java.nio.channels.CompletionHandler;
-import java.nio.charset.Charset;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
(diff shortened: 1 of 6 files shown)


file/src/main/java/akka/stream/alpakka/file/impl/FileTailSource.java
```diff
@@ -111,11 +111,15 @@ public void preStart() {
                 (tryInteger) -> {
                   if (tryInteger.isSuccess()) {
                     int readBytes = tryInteger.get();
+                    // when the number of bytes read is -1 it signals that no new data is available
                     if (readBytes > 0) {
                       buffer.flip();
                       push(out, ByteString.fromByteBuffer(buffer));
                       position += readBytes;
                       buffer.clear();
+                    } else if (!Files.exists(path)) {
```
ennru: This is a blocking call, I believe it should be wrapped in a future and use an `asyncCallback` to trigger completion.
It should possibly not try this for every poll but have some counter?

seglo: @ennru Good point.  It will be executed for every poll where no new data was read from the channel, which could be frequent if the log is inactive for some time, but there is still lots of demand from downstream.  I can add the `asyncCallback`, but if we only check every `X` poll then we could have a situation where the stream stays open indefinitely when there's no demand.  That currently happens now as well, but with a counter it would happen more frequently.

This Graph Stage already uses a `TimerGraphStageLogic`.  I could add a 2nd timer which ticks every second to see if the file exists.

seglo: Another option would be to use `DirectorySourceChange` internally to trigger the shutdown on file deletes, but maybe that's overkill.

ennru: Schedule a second timer to check if the file is still there when no bytes are read, cancel it when bytes are read.

seglo: I'll have to add a new parameter to define a new interval for the file check, since reusing the poll interval is effectively check if the file exists as frequently as the current implementation (though, in a non-blocking manner).

seglo: Maybe that's not so bad though.. I'll reuse the poll duration for now.

file/src/main/java/akka/stream/alpakka/file/impl/FileTailSource.java
```diff
@@ -111,14 +108,20 @@ public void preStart() {
                 (tryInteger) -> {
                   if (tryInteger.isSuccess()) {
                     int readBytes = tryInteger.get();
+                    // when the number of bytes read is -1 it signals that no new data is available
                     if (readBytes > 0) {
+                      if (isTimerActive(fileExistsCheckTimer)) {
```
ennru: The test is not required, it is OK to call `cancelTimer` directly.

file/src/main/java/akka/stream/alpakka/file/impl/FileTailSource.java
```diff
@@ -129,13 +132,23 @@ public void preStart() {
 
       @Override
       public void onTimer(Object timerKey) {
-        doPull();
+        if (timerKey == pollTimer) {
```
ennru: Use a `switch` with Strings?

seglo:
I updated the implementation to avoid making changes to `impl.FileTailSource`, and just use the operator DSL with the `recoverWith` `Source.empty` trick that @johanandren mentioned the other day to complete the stream successfully.

file/src/main/java/akka/stream/alpakka/file/impl/FileTailSource.java
seglo: Trying a different approach with operator DSL.

seglo:
I'll update the scaladsl tomorrow.

file/src/main/java/akka/stream/alpakka/file/javadsl/FileTailSource.java
```diff
@@ -43,12 +52,36 @@
...
+        .mapConcat(tick -> {
+          if (Files.exists(path)) {
+            return Collections.<ByteString>emptyList();
+          }
+          throw new FileNotFoundException();
+        })
+        .recoverWith(new PFBuilder<Throwable, Source<ByteString, NotUsed>>()
+            .match(FileNotFoundException.class, t -> Source.empty())
+            .build()
+        );
```
johanandren: If you are sure you want this as a separate stream + a public API I think `.takeWhile(e -> fileExists)` would be a less roundabout way of doing it.

However spontaneously it sounds like a concern that should be covered by the graph stage rather than through composition, now you will end up with a race condition where the last bytes the FileTailSource may or may not arrive before the ´fileCheckSource´ completes if the file is deleted.

johanandren:
BTW isn't the semantics on unixy filesystems that the file could be deleted from the filesystem but two processes reading and writing it will keep it around, and that is fine?

file/src/main/java/akka/stream/alpakka/file/javadsl/FileTailSource.java
seglo: @johanandren Thanks for the review.  That's a good point about the race condition, I hadn't considered that.

The original implementation was in the `FileTailSource` implementation.  One of the issues we found was that checking if the file exists is a synchronous operation, and even if it were made asynchronous with a timer, it could still block the stage from processing other log content.  I had considered creating another actor within the stage to do the checking and message back if the file was detected as deleted, but a separate stream seemed a little more elegant at the time.

seglo:
> BTW isn't the semantics on unixy filesystems that the file could be deleted from the filesystem but two processes reading and writing it will keep it around, and that is fine?

@johanandren I think you're right. That seems reinforced by the fact that `AsynchronousFileChannel` seems unpeturbed by deleted files.  The use case that inspired this PR was from the [Alpakka samples project](https://github.com/akka/alpakka-samples/pull/38) I'm working on, which creates `FileTailSources` on file creation events, which made me wonder what happened when files were removed.  In that case the sub-streams remain active indefinitely.

file/src/main/java/akka/stream/alpakka/file/javadsl/FileTailSource.java
seglo: Moved this implementation to a documented use case with the race condition caveat.

seglo:
Repurposed PR to add docs and sample code for dealing with file deletes and idle files.

file/src/test/java/docs/javadsl/FileTailSourceTest.java
```diff
@@ -112,6 +118,116 @@ public void willReadNewLinesAppendedAfterReadingTheInitialContents() throws Exce
...
+
+  /**
+   * Periodically checks if the file at `path` exists every `pollingInterval`. Shutdown the stream
+   * when the file is not found using an empty Source.
+   * @param pollingInterval Interval to check if file exists.
+   * @param path a file to check
+   */
+  private static Source<String, Cancellable> fileCheckSource(java.time.Duration pollingInterval, Path path) {
+    return Source
+            .tick(pollingInterval, pollingInterval, new Tick())
```
ennru: You do not want to create a new instance for every tick. Create a static field with the instance.

file/src/test/scala/docs/scaladsl/FileTailSourceExtrasSpec.scala
```diff
@@ -0,0 +1,116 @@
...
+       * @param pollingInterval Interval to check if file exists.
+       * @param path a file to check
+       */
+      def fileCheckSource(pollingInterval: FiniteDuration, path: Path): Source[String, Cancellable] =
+        Source
+          .tick(pollingInterval, pollingInterval, Tick)
+          .mapConcat { _ =>
+            if (Files.exists(path))
+              Collections.emptyList[String]
+            throw new FileNotFoundException
```
ennru: Isn't this messing an `else`?

file/src/test/java/docs/javadsl/FileTailSourceTest.java
```diff
@@ -112,6 +118,116 @@ public void willReadNewLinesAppendedAfterReadingTheInitialContents() throws Exce
...
+            .mapConcat(tick -> {
+              if (Files.exists(path)) {
+                return Collections.<String>emptyList();
+              }
+              throw new FileNotFoundException();
```
ennru: I would add an `else` to make it more obvious.

file/src/test/scala/docs/scaladsl/FileTailSourceExtrasSpec.scala
```diff
@@ -0,0 +1,116 @@
...
+      Files.write(path, "a\n".getBytes(UTF_8))
+
+      // just for docs
+      // #shutdown-on-idle-timeout
+
+      val idleTimeout = 30.seconds
+
+      val stream = FileTailSource
+        .lines(path = path, maxLineSize = 8192, pollingInterval = 250.millis)
+        .idleTimeout(idleTimeout)
```
ennru: Just put these values in, it reads better.

file/src/test/java/docs/javadsl/FileTailSourceTest.java
```diff
@@ -112,22 +118,56 @@ public void willReadNewLinesAppendedAfterReadingTheInitialContents() throws Exce
...
+            })
+            .recoverWith(new PFBuilder<Throwable, Source<String, NotUsed>>()
+                    .match(FileNotFoundException.class, t -> Source.empty())
+                    .build()
+            );
+  }
```
johanandren: One thing that struck me: don't we have a FS-change-event-source-connector as well, couldn't one just filter-map that down to only deletes and this file?

seglo: Yes, that would be a valid use case as well to manage the stream from the outside. The original purpose of this example was to shutdown the stream from within when a file was deleted. You could create a graph that uses the `DirectoryChangesSource` to do the same thing within, but it would be scanning for filesystem changes on a whole directory instead of individual files.

seglo: Good point.  I'll take a look at that soon.

seglo: I updated the example to use `DirectoryChangesSource`.  I noticed that it's a bit slower to detect the event then eagerly checking if the file exists, no matter what I set the interval frequenty to.  It required me to increase the test probe timeout to `akka.test.single-expect-default = 10s` so the test would pass.

PR #1959, opened 2019-10-04
Idle timeout experiment

seglo:
## Purpose

An experiment using the `idleTimeout` operator.  This is a WIP.  I think it's a good idea to add an option for the idle timeout, but I don't like how it requires us to add parameters across all the DSLs.  I think we should consider using configuration types as parameters to Alpakka DSLs to make it easier to add options in the future.

Or maybe I misunderstood what you wanted to do here.  Was this a feature you wanted to include, or just document?


ennru:
We want to promote composition by the user over more built-in functionality. So just add it in the docs by including new snippets from tests would be great.

seglo:
Documented in #1953 .

PR #1973, opened 2019-10-08
Fix typo.

takayahilton:
## Purpose

fix typo.


ennru:
Thank you for engaging with Alpakka!

"Iff" stands for "If and only if".

PR #1978, opened 2019-10-09
Release notes for 1.1.2

ennru:


docs/src/main/paradox/release-notes/1.1.x/amqp.md
```diff
@@ -1,5 +1,11 @@
 # AMQP
 
+## 1.1.2
+
+No changes.
+
+[*closed in 1.1.2*](https://github.com/akka/alpakka/issues?q=is%3Aclosed+milestone%3A1.1.2+label%3Ap%3Aamqp)
```
johanandren: What are these, "no changes" but then still a link to the milestone list, which will be empty if there are no changes?

ennru: Sometimes there are minor things which were not relevant enough to write out.
(Or we miss things in this round...)

Issue #1943, opened 2019-09-23
XML: add ability to provide configuration properties on the XMLInputFactory used by the Xml Parser


jphelp32:
### Short description

Add the ability to provide configuration properties on the XMLInputFactory used by the Xml Parser.

### Details

The biggest reason that this is necessary is b/c the default configuration of the aalto-xml parser does not disable external DTDs or DTD entity expansion, which is considered to be unsecure. At the very least, it should be possible to set the configuration properties needed to achieve secure XML parsing.

ennru:
This would be good, indeed. We should even consider to disable external DTDs and DTD entity expansion by default in Alpakka.

jphelp32:
Disabling those by default in alpakka would suffice for my interests.

I was a little unsure about how to approach this one, since the parser requires a specific factory (ie. aalto) rather than any generic XMLInputFactory. So a change similar to the one I made on the output side would not be appropriate.

jphelp32:
@ennru would you like a separate issue and/or pull request just for the disabling of external DTDs and DTD entity expansion (short of full configurability of the input factory)?

ennru:
An alternative could be to let the user pass an `XMLInputFactory => Unit` function. Or even `AsyncXMLInputFactory => Unit`.
And have some default implementation of it which disables the bad ones.

jphelp32:
Something like the following:

```
@InternalApi private[xml] class StreamingXmlParser(ignoreInvalidChars: Boolean, configureXmlInputFactory: AsyncXMLInputFactory => Unit = _ => Unit) {
      ...
      private val factory: AsyncXMLInputFactory = new InputFactoryImpl()
      factory.setProperty(XMLInputFactory.SUPPORT_DTD, false)
      factory.setProperty(XMLInputFactory.IS_SUPPORTING_EXTERNAL_ENTITIES, false)
      factory.setProperty(XMLConstants.FEATURE_SECURE_PROCESSING, true)   // I'm not sure about this one
      configureXmlInputFactory(factory)
      private val parser: AsyncXMLStreamReader[AsyncByteArrayFeeder] = factory.createAsyncFor(Array.empty)
      if (ignoreInvalidChars) {
        parser.getConfig.setIllegalCharHandler(new ReplacingIllegalCharHandler(0))
      }
      ...
}

**Comment by ennru** [comment]
I'd like to suggest to create a function in `XmlParsing` which does the settings we'd recommend as default.
```
object XmlParsing {
[...]

PR #1982, opened 2019-10-10
XML: add support for supplying XMLInputFactory configuration

jphelp32:
## Purpose

Adds support to pass in a configure function for configuring the XMLInputFactory used by XmlParsing

## References

References #1943

### DIFF
--- a/xml/src/main/scala/akka/stream/alpakka/xml/impl/StreamingXmlParser.scala
+++ b/xml/src/main/scala/akka/stream/alpakka/xml/impl/StreamingXmlParser.scala
@@ -17,7 +17,8 @@ import scala.annotation.tailrec
 /**
  * INTERNAL API
  */
-@InternalApi private[xml] class StreamingXmlParser(ignoreInvalidChars: Boolean)
+@InternalApi private[xml] class StreamingXmlParser(ignoreInvalidChars: Boolean,
+                                                   configureFactory: AsyncXMLInputFactory => Unit)
     extends GraphStage[FlowShape[ByteString, ParseEvent]] {
   val in: Inlet[ByteString] = Inlet("XMLParser.in")
   val out: Outlet[ParseEvent] = Outlet("XMLParser.out")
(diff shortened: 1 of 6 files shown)


xml/src/main/scala/akka/stream/alpakka/xml/scaladsl/XmlParsing.scala
```diff
@@ -9,22 +9,38 @@ import akka.stream.alpakka.xml.ParseEvent
...
 
+import javax.xml.stream.XMLInputFactory
+import javax.xml.XMLConstants
+
 import scala.collection.immutable
 
 object XmlParsing {
 
+  /**
+   * The default factory configuration is to enable secure processing.
```
ennru: It should disable it.

xml/src/main/scala/akka/stream/alpakka/xml/impl/StreamingXmlParser.scala
```diff
@@ -28,8 +29,9 @@ import scala.annotation.tailrec
...
       import javax.xml.stream.XMLStreamConstants
 
-      private val feeder: AsyncXMLInputFactory = new InputFactoryImpl()
-      private val parser: AsyncXMLStreamReader[AsyncByteArrayFeeder] = feeder.createAsyncFor(Array.empty)
+      private val factory: AsyncXMLInputFactory = new InputFactoryImpl()
+      configureFactory(factory)
+      private val parser: AsyncXMLStreamReader[AsyncByteArrayFeeder] = factory.createAsyncFor(Array.empty)
       if (ignoreInvalidChars) {
         parser.getConfig.setIllegalCharHandler(new ReplacingIllegalCharHandler(0))
       }
```
ennru: While you're at it, would it make sense to allow for parser configuration with the same pattern?

jphelp32: not sure there's any other configuration settings for the parser itself that couldn't be set in the factory. I'll look into that though...

jphelp32: ok, so I can't spend a ton of time on this, but from the list of public methods available on the ReaderConfig instance (returned from parser.getConfig) that mutate the configuration, they all fall into one of these categories:
1) it's a configuration that could be set on the factory prior to creation of the parser
2) it's something that, from what I can tell, should perhaps not be public at all
3) it's the setIllegalCharHandler() config

1 is now implemented with what we have so far in this pull request
2, it's possible that one or two of the methods I put into this category ARE in fact useful, but since that's not clear, and since there's no current ask from anyone, I would say adding support for these into alpakka is not necessary at this time.
3, this is only partially supported in alpakka right now. Alpakka creates a specific IllegalCharHandler now, whereas the underlying config would allow the use of any custom handler. It might be useful to address this, but it might require a more significant impact on the signature of the parser constructors, in order to work nicely with the existing ignoreInvalidChars parameter. So I'm inclined to not take up that effort with this pull-request.

xml/src/main/scala/akka/stream/alpakka/xml/scaladsl/XmlParsing.scala
jphelp32: done

ennru:
Thank you for hanging in here!

Issue #1958, opened 2019-10-03
google-cloud-pub-sub should use own connection pool


btomala:
### Short description

Currently, google-cloud-pub-sub is using `Http().singleRequest` to make HTTP calls. Those requests are run on a shared connection pool. User unintentionally can affect PubSub requests when running other requests using a shared pool.

### Details

> The request-level API is implemented on top of a connection pool that is shared inside the actor system. A consequence of using a pool is that long-running requests block a connection while running and starve other requests. Make sure not to use the request-level API for long-running requests like long-polling GET requests. Use the Connection-Level Client-Side API or an extra pool just for the long-running connection instead.

Above quotes comes from [request level client-side api documentation](https://doc.akka.io/docs/akka-http/current/client-side/request-level.html#request-level-client-side-api)

To avoid issues with that google-cloud-pub-sub should use `Http().cachedHostConnectionPool` for `pubsub.googleapis.com` host.

Other advantages of this solution:
* separate connection pool should improve performance
* stream API provides a context which can be used to track publishing messages to PubSub in an asynchronous manner

ennru:
It would make sense to allow for a specific connection pool for pub/sub. Would you be able to suggest the necessary changes in a pull request?

btomala:
Yes, I'll prepare a pull request.

Kreinoee:
Sorry, got by mistake referenced this from the slick pull request.

PR #1983, opened 2019-10-12
Google-cloud-pub-sub: dedicated connection pool; `withContext`

btomala:
## Purpose

This PR introduces cached host connection pool for google cloud pub-sub HTTP integration.

## References

References  #1958

## Changes

* google-cloud-pub-sub start using cashed host connection pool instead of shared one
* replace `GooglePubSubSource` with `Source.tick`
* subscribe source from pub-sub is Cancellable (you can cancel it during graceful shutdown process)
* binary compatibility is broken
* all public methods takes `HttpsConnectionContext` as a parameter, I need to do this for testing purposes, not sure if this is a correct approach.

## Background Context

Previously the google-cloud-pub-sub user single request to fetch/push data to pub-sub. Single request use shared connections pool, for performance reasons makes sense to use separate connection pool for pub-sub only. 
Using streaming akka-http API we have Cancellation implemented in the tick source. It is easier to understand than `GooglePubSubSource`. However, if someone writes it maybe there is something special what I missed.
We can also add context passing to correlate publishing requests with responses.
I decided to pass `HttpsConnectionContex` as one of the params because I don't know how to skip validation SSL config in the test. Looks like `akka.ssl-config.loose.acceptAnyCertificate = true` flag doesn't work in this case.

If you like this change let me know, then I'll update documentation.


google-cloud-pub-sub/src/test/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApiSpec.scala
```diff
@@ -30,17 +47,28 @@ class PubSubApiSpec extends FlatSpec with BeforeAndAfterAll with ScalaFutures wi
   implicit val defaultPatience =
     PatienceConfig(timeout = 5.seconds, interval = 100.millis)
 
+  val sslContext: SSLContext = SSLContext.getInstance("TLS")
+  sslContext.init(null, Array(new NoopTrustManager()), null)
+
+  val hcc = Some(ConnectionContext.https(sslContext))
```
ennru: You replaced the trust manager, that switches off much more than the hostname verification.
Instead of passing the context around, set it on the HTTP extension via `Http().setDefaultClientHttpsContext(ConnectionContext.https(sslContext))`.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApi.scala
```diff
@@ -121,87 +128,138 @@ private[pubsub] trait PubSubApi {
...
+  def pull[T](project: String,
+              subscription: String,
+              returnImmediately: Boolean,
+              maxMessages: Int,
+              parallelism: Int,
+              hcc: Option[HttpsConnectionContext] = None)(
       implicit as: ActorSystem,
       materializer: Materializer
-  ): Future[PullResponse] = {
+  ): Flow[(Done, Option[String], T), (Future[PullResponse], T), NotUsed] = {
```
ennru: The code doesn't use the context in `T` right now. The `pull` implementation is only used with `subscribe`, so I think we do not need it here.
If there is a use case for passing context, the `mapAsyncUnordered` should become a `mapAsync` as it could reorder elements which `FlowWithContext` shouldn't allow.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/scaladsl/GooglePubSub.scala
```diff
@@ -27,61 +28,46 @@ protected[pubsub] trait GooglePubSub {
...
-    if (httpApi.isEmulated) {
-      Flow[PublishRequest].mapAsyncUnordered(parallelism) { request =>
-        httpApi.publish(config.projectId, topic, maybeAccessToken = None, request)
-      }
-    } else {
-      Flow[PublishRequest].mapAsyncUnordered(parallelism) { request =>
-        config.session.getToken().flatMap { accessToken =>
-          httpApi.publish(config.projectId, topic, Some(accessToken), request)
-        }
+  ): Flow[PublishRequest, immutable.Seq[String], NotUsed] =
```
ennru: `publish` would benefit from a `FlowWithContext` variant.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/scaladsl/GooglePubSub.scala
```diff
@@ -27,61 +28,46 @@ protected[pubsub] trait GooglePubSub {
...
+    httpApi
+      .accessToken[PublishRequest](config, parallelism)
+      .via(httpApi.publish(config.projectId, topic, parallelism))
+      .mapAsyncUnordered(parallelism) {
```
ennru: This should be a `mapAsync(1)` to be sure the response order is kept.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/scaladsl/GooglePubSub.scala
```diff
@@ -27,61 +28,46 @@ protected[pubsub] trait GooglePubSub {
...
       )
-    )
-
-  def acknowledge(
-      subscription: String,
-      config: PubSubConfig,
-      parallelism: Int = 1
-  )(implicit actorSystem: ActorSystem, materializer: Materializer): Sink[AcknowledgeRequest, Future[Done]] = {
-    import materializer.executionContext
+      .mapAsyncUnordered(parallelism) {
```
ennru: This should be a `mapAsync(1)` to be sure the response order is kept.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApi.scala
```diff
@@ -121,87 +128,138 @@ private[pubsub] trait PubSubApi {
       AcknowledgeRequest(json.asJsObject.fields("ackIds").convertTo[immutable.Seq[String]]: _*)
     def write(ar: AcknowledgeRequest): JsValue = JsObject("ackIds" -> ar.ackIds.toJson)
   }
-  private implicit val pullRequestFormat = DefaultJsonProtocol.jsonFormat2(PubSubApi.PullRequest)
+  private implicit val pullRequestFormat = DefaultJsonProtocol.jsonFormat2(PullRequest)
+
+  private def pool[T](hcc: Option[HttpsConnectionContext])(
+      implicit as: ActorSystem
+  ): Flow[(HttpRequest, T), (Try[HttpResponse], T), Http.HostConnectionPool] =
+    if (isEmulated) {
```
ennru: Very nice to have the emulation pushed to a central place instead.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/scaladsl/GooglePubSub.scala
```diff
@@ -27,61 +28,46 @@ protected[pubsub] trait GooglePubSub {
...
-        subscription = subscription,
-        returnImmediately = config.pullReturnImmediately,
-        maxMessages = config.pullMaxMessagesPerInternalBatch,
-        httpApi = httpApi
+  /**
+   * Creates a source pulling messages from subscription
+   */
+  def subscribe(subscription: String, config: PubSubConfig, parallelism: Int = 1)(
+      implicit actorSystem: ActorSystem,
+      materializer: Materializer
```
ennru: While changing APIs... 
This could make use of [Source.setup](https://doc.akka.io/docs/akka/current/stream/operators/Source-or-Flow/setup.html) to not require the Actor System and Materializer.
But it might be better to keep that in a different PR.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/scaladsl/GooglePubSub.scala
```diff
@@ -27,61 +28,46 @@ protected[pubsub] trait GooglePubSub {
...
+  ): Source[ReceivedMessage, Cancellable] =
+    Source
+      .tick(0.seconds, 1.second, Done)
```
ennru: This is nice. It might happen that the requests behave a bit differently depending on the downstream demand than with the graph stage, but I don't see any major difference.

btomala:
Hi, sorry for delay, I was on holiday. I'll review your comments soon.

google-cloud-pub-sub/src/test/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApiSpec.scala
btomala: Switching off the host verification wasn't enough. That was the reason why I replace the trust manager.
I think I try it already and doesn't work. However, I'll give a try again.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApi.scala
btomala: Yes, it is not used yet. I have a plan to use it. However, I want to limit number of changes in the first step. I'll expose it to the public API in next commit. I'll change `mapAsync` to unordered.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/scaladsl/GooglePubSub.scala
btomala: I'll add context than this should stay as it is.

google-cloud-pub-sub/src/test/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApiSpec.scala
btomala: That's works! I'll remove HttpsConnectionContext from API

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/scaladsl/GooglePubSub.scala
btomala: Yes, I want to add context here.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/scaladsl/GooglePubSub.scala
btomala: I make leave it only for `publish` and change to `mapAsync(1)` for others

btomala:
Do you have milestone 2.0 roadmap? When do you have plan to release those changes? Do you have a milestone release?

ennru:
We'll start releasing 2.0 milestones soon, snapshots are published after successfull builds of master https://doc.akka.io/docs/alpakka/current/other-docs/snapshots.html (which is sadly not often)

ennru:
We'll cut a first 2.0.0 milestone this week https://github.com/akka/alpakka/issues/1998

btomala:
Ok, thanks! Please check the public API after my last changes. If this is ok, I'll upgrade the documentation.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/scaladsl/GooglePubSub.scala
ennru: Created https://github.com/akka/alpakka/issues/2017

ennru:
The errors MiMa detected require you to add exclusion files. I just suggested a more comprehensive instruction for that in https://github.com/akka/alpakka/pull/2015/files

btomala:
I rebased branch to master and add a file with excludes. Unfortunately, looks like it is not taken into consideration when filters are run. Don't know why.

ennru:
The exclusions file should be `.../mima-filters/2.0.0-M1.backwards.excludes/pr-1983-use-dedicated-connection-pool.excludes`

btomala:
Ok, thanks I didn't know that milestones are considered as binary compatible. I test it locally and it passed now.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApi.scala
```diff
@@ -121,87 +129,138 @@ private[pubsub] trait PubSubApi {
...
+    val url: Uri = s"/v1/projects/$project/subscriptions/$subscription:pull"
+
+    Flow[(Done, Option[String])]
+      .mapAsync(1) {
+        case (_, maybeAccessToken) =>
+          Marshal((HttpMethods.POST, url, PullRequest(returnImmediately, maxMessages)))
+            .to[HttpRequest]
+            .map(request => (authorize(maybeAccessToken)(request), NotUsed))
+      }
+      .via(pool[NotUsed]())
```
ennru: Please make add a `pool` variant without context so that the using methods don't get to create strange tuples, such as 
```
  private def poolSimple()(
    implicit as: ActorSystem
  ): Flow[HttpRequest, Try[HttpResponse], Http.HostConnectionPool] =
    Flow[HttpRequest]
      .map(_ -> NotUsed)
      .viaMat(pool())(Keep.right)
      .map(_._1)
```

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApi.scala
```diff
@@ -16,39 +16,47 @@ import akka.http.scaladsl.model.headers.OAuth2BearerToken
 import akka.http.scaladsl.unmarshalling.Unmarshal
 import akka.stream.Materializer
 import akka.stream.alpakka.googlecloud.pubsub._
-import spray.json._
+import akka.stream.scaladsl.FlowWithContext
+import akka.stream.scaladsl.Flow
+import akka.{Done, NotUsed}
 import spray.json.DefaultJsonProtocol._
-import spray.json.{deserializationError, DefaultJsonProtocol, JsObject, JsString, JsValue, RootJsonFormat}
+import spray.json.{DefaultJsonProtocol, JsObject, JsString, JsValue, RootJsonFormat, deserializationError, _}
```
ennru: `spray.json._` will do

google-cloud-pub-sub/src/test/java/docs/javadsl/ExampleUsageJava.java
```diff
@@ -67,21 +67,38 @@ private static void example() throws NoSuchAlgorithmException, InvalidKeySpecExc
...
+
     // #publish-fast
     Source<PublishMessage, NotUsed> messageSource = Source.single(publishMessage);
     messageSource
         .groupedWithin(1000, Duration.ofMinutes(1))
-        .map(messages -> PublishRequest.create(messages))
-        .via(publishFlow)
+        .map(messages -> Pair.apply(PublishRequest.create(messages), context))
+        .via(publishFlowWithContext)
         .runWith(Sink.ignore(), materializer);
```
ennru: This example would be simpler without the context.

google-cloud-pub-sub/src/test/scala/docs/scaladsl/ExampleUsage.scala
```diff
@@ -59,13 +59,34 @@ class ExampleUsage {
...
+  //#publish-single-with-context
+
   //#publish-fast
   val messageSource: Source[PublishMessage, NotUsed] = Source(List(publishMessage, publishMessage))
-  messageSource.groupedWithin(1000, 1.minute).map(PublishRequest.apply).via(publishFlow).to(Sink.seq)
+  messageSource
+    .groupedWithin(1000, 1.minute)
+    .map(grouped => PublishRequest(grouped) -> resultPromise)
+    .via(publishFlowWithContext)
+    .to(Sink.seq)
```
ennru: Same here, the context doesn't seem to provide any value.

btomala:
I made them simpler.

ennru:
Thank you! Sorry if it grew bigger than you originally intended...
Great to have `withContext` support, it proves useful in many situations.

btomala:
Thank you for your help in review!

Issue #767, opened 2018-02-08
Backpressure on ProvisionedThroughputExceededException in DynamoDB?


an-tex:
Every DynamoDB table has a provisioned throughput. Once exceeded, the client returns a ProvisionedThroughputExceededException . Wouldn't this be an indication to back pressure?

joost-de-vries:
I seem to remember that the AWS client itself does already do a retry/backoff mechanism. So I'm not sure doing another backoff mechanism on top of that makes sense.

an-tex:
I'm not sure about the implementation details of the backoff / retry but it certainly doesn't expose any back pressure to the upstream, so it's either buffering or dropping items. Either way it's not the akka streams way of handling things ;)
Sure it's better than nothing, but I guess especially as the aws client is already signalling an overloading / throttling of the next stage, why don't make use of it and make it visible to the whole stream.

codynhat:
I have run into many issues of `ProvisionedThroughputExceededException` being hit and retries by the AWS client just making the problem worse. The retries do not necessarily stop new requests from coming in flight. Especially when the upstream is much faster. Signaling back pressure when the AWS client sees this would help.

It's also very akka stream-y and intuitively seems like the expected behavior.

2m:
I would be in favor of adding a new `flow` to the DynamoDB connector that would return responses wrapped in `Try[...]`. Then upon failure a [`Retry`](https://developer.lightbend.com/docs/api/akka-stream-contrib/current/akka/stream/contrib/Retry$.html) could be used to retry the failed request after some time, which would be backpressuring.

julianhowarth:
Just been trying to use the connector and have immediately run into this issue - this makes this lib a complete non-starter for me. My use case is that I need to scan a 100k row DynamoDB table when bootstrapping my service instances - I need to get the data as fast as possible but after 5k rows the flow just stops.

Just for additional information I'm use OpenJDK 11.0.4, scala 2.12.7, akka-streams 2.5.25, and 1.1.1 of Alpakka Dynamo - even with a logging supervision strategy the flow just stops silently. The onlmy clue something is wrong is a JDK warning:
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.fasterxml.jackson.databind.util.ClassUtil (file:/home/julian/.m2/repository/com/fasterxml/jackson/core/jackson-databind/2.8.6/jackson-databind-2.8.6.jar) to field java.lang.Throwable.cause
[...]

Issue #1924, opened 2019-09-10
DynamoDb: update RetryFlow to latest version from Akka


ennru:
### Short description

The internal `RetryFlow` in Alpakka DynamoDb should be updated with the latest version from Akka before releasing Alpakka.

### Details

Alpakka DynamoDb has the first usage of the suggested Akka Stream `RetryFlow` and implemented with a work-in-progress version of https://github.com/akka/akka/pull/27461

ennru:
By using the built-in retrying in the AWS SDK 2 for DynamoDB we do not need the `RetryFlow`.

PR #1986, opened 2019-10-14
DynamoDB: document retry configuration

ennru:
## Purpose

Instead of wrapping the DynamoDB calls with retrying in Akka Streams, the AWS SDK 2 built-in "retry with backoff" logic should be used.

## References

This seems to be the answer to #767 (which was suggested when Alpakka used AWS SDK 1).
The [AWS documentation](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Programming.Errors.html#Programming.Errors.RetryAndBackoff) is not up-to-date right now but indicates that this is a recommended approach.

## Changes

* Drop the Alpakka-internal copies of `RetryFlow` added in https://github.com/akka/alpakka/pull/1896
* Add an example client configuration that shows `overrideConfiguration`


ennru:
This will make #1924 obsolete.

Issue #1972, opened 2019-10-04
Make cross-compilation Akka 2.5.x/2.6.0-M8 work


ennru:
The code needs to be adapted are a few places.

https://travis-ci.com/akka/alpakka/jobs/242016134#L1028

https://travis-ci.com/akka/alpakka/jobs/242016134#L1117

https://travis-ci.com/akka/alpakka/jobs/242016134#L1211

https://travis-ci.com/akka/alpakka/jobs/242016134#L1348

These changes need to be backported to 1.1.x, as well.

ennru:
As it turns out, cross-compilation is not possible for the time being. Some API changes need to be made available in another release of Akka 2.5.

ennru:
All modules work with Akka 2.6.0 now.

PR #1988, opened 2019-10-18
Akka 2.6/2.5 cross-compilation

ennru:
* Akka 2.6.0-RC1
* Akka 2.5.26

* Adapt some method names
* Copy in old Buffer factory to SQS


ennru:
Fixes #1972

PR #1991, opened 2019-10-21
Google Cloud Pub/Sub gRPC: Akka Discovery

ennru:
## Purpose

Directly depend on Akka Discovery so it always uses the correct Akka version.

## References

#1988
https://travis-ci.com/akka/alpakka/jobs/247554088#L528

## Changes

* Add Akka Discovery dependency
* Recommend direct Akka Discovery dependency in the docs
* Commit formatting updates for Java files

### DIFF
--- a/docs/src/main/paradox/google-cloud-pub-sub-grpc.md
+++ b/docs/src/main/paradox/google-cloud-pub-sub-grpc.md
@@ -14,10 +14,15 @@ look at the alternative @ref[Alpakka Google Cloud Pub/Sub](google-cloud-pub-sub.
 
 ## Artifacts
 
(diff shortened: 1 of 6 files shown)


docs/src/main/paradox/google-cloud-pub-sub-grpc.md
```diff
@@ -14,10 +14,15 @@ look at the alternative @ref[Alpakka Google Cloud Pub/Sub](google-cloud-pub-sub.
 
 ## Artifacts
 
+As Akka gRPC uses Akka Discovery internally, make sure to add Akka Discovery of the same Akka version as the application uses.
```
seglo:
```suggestion
Akka gRPC uses Akka Discovery internally. Make sure to add Akka Discovery with the same Akka version that the application uses.
```

Issue #1679, opened 2019-05-05
S3 path-style access is going away


jrudolph:
https://forums.aws.amazon.com/ann.jspa?annID=6776

> Amazon S3 currently supports two request URI styles in all regions: path-style (also known as V1) that includes bucket name in the path of the URI (example: //s3.amazonaws.com/<bucketname>/key), and virtual-hosted style (also known as V2) which uses the bucket name as part of the domain name (example: //<bucketname>.s3.amazonaws.com/key). In our effort to continuously improve customer experience, the path-style naming convention is being retired in favor of virtual-hosted style request format. Customers should update their applications to use the virtual-hosted style request format when making S3 API requests before September 30th, 2020 to avoid any service disruptions. Customers using the AWS SDK can upgrade to the most recent version of the SDK to ensure their applications are using the virtual-hosted style request format.

> Virtual-hosted style requests are supported for all S3 endpoints in all AWS regions. S3 will stop accepting requests made using the path-style request format in all regions starting September 30th, 2020. Any requests using the path-style request format made after this time will fail.

> If there is any reason why your application is not able to utilize the virtual-hosted style request format, or if you have any questions or concerns, please reach out to AWS Support. 

That means that some solution will have to found for the HTTPS issue for buckets with dots in the name by then. Probably nothing should be done for now because AWS will have to figure that out by themselves so that customers are left with the invalid certificates when they are finally forced to use the domain style access.

2m:
Most of the Alpakka S3 tests use capital letters in the bucket name, which also does not work with path-style-access. So that is another thing to avoid in bucket names.

ennru:
With #2099 Alpakka is back to using virtual-host-style as the default bucket access, this restricts the bucket names much more than path-style access.
Warnings are issued in the logs if path-style access is configured.

PR #1993, opened 2019-10-29
Set path style access to true by default

laszlovandenhoek:
According to `reference.conf`, `path-style-access` is `true` by default:

https://github.com/akka/alpakka/blob/c143ee821feb650bc1c7b9902db3c535dad45f59/s3/src/main/resources/reference.conf#L80

(this has been in effect since https://github.com/akka/alpakka/pull/1308)

However, the non-deprecated version of `S3Settings.apply(...)` sets it to `false`. Following the motivation quoted in `reference.conf` above, I believe this should be `true` instead.

It appears this discrepancy was introduced as part of the polish that @2m applied to https://github.com/akka/alpakka/pull/1639 in https://github.com/akka/alpakka/pull/1639/commits/88217f57856f2e3ccc89fce7d1dfeba0b17525e2. I mention this only because he might have had a good reason for this that I'm unaware of.

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/settings.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/settings.scala
@@ -431,7 +431,7 @@ object S3Settings {
     bufferType,
     credentialsProvider,
     s3RegionProvider,
-    false,
+    true,
     None,
     listBucketApiVersion,
     None


laszlovandenhoek:
Hmm, only now I'm reading https://github.com/akka/alpakka/issues/1679. I guess `reference.conf` is the one that should be changed, not `S3Settings`.

Issue #1509, opened 2019-02-15
Slick/JDBC create SlickSession from Database and Profile.


Kreinoee:
### Short description

It should be possible to create a SlickSession, from a Database object and a Profile object.

### Details

A SlickSession is basically just a wrapper arround a Database and a Profile, but you ironically, you cannot create it, if these two objects are your starting point. Currently its only possiby to create a SlickSession from a Config or a DatabaseConfig.

I have a situation, where I use another Library, that supplies a Database object to me, for using in unit tests. So my problem is, that I have the Database object, and now what Profile to use with it, and therefore have the two things that makes up a SlickSession, but I have no easy way of actually creating the needed SlickSession instance.

I would suggest creating an extra forDatabaseAndProfile factory method, to SlickSessions companion object. I have created a Local fork to test it out, and it works fine,  so I can provide it as a pull request on the other side of the weekend.

ennru:
Yes, this sounds useful (even the `forConfig` hasn't been around too long). It would be great to get in your PR for this.

Kreinoee:
Sorry for the delay, but gut stuck on other projects.

I may need some help here. I have made a fix, that works for me now. However for being able to build alpakka at all, I had to remove a lot of stuff from the build files. As you contributions policies require, I would like add some documentation for this new way of creating a SlickSession. However when i try to run docs/previewSite sbt task, I get a lot of errors like:
```
error] Failed to resolve [akka.stream.alpakka.sqs.SqsAckResultEntry] because property [scaladoc.akka.stream.alpakka.base_url] contains an invalid URL [//] at /home/anders/cloudSolution/projects/alpakka/docs/src/main/paradox/sqs.md:292
[error] `SqsAckFlow.grouped` batches actions on their type and forwards a @scaladoc[`SqsAckResultEntry`](akka.stream.alpakka.sqs.SqsAckResultEntry) 
```
This could very well be caused by all the stuff that I have deleted in the build file, but if I revert the changes in the build files, or just try on a fresh clone of alpakka, I cannot even start sbt, due to this error:

```
/tmp/alpakka/ sbt
OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=2G; support was removed in 8.0
[info] Loading global plugins from /home/anders/.sbt/1.0/plugins
[info] Loading settings for project alpakka-build from plugins.sbt ...
[info] Loading project definition from /tmp/alpakka/project
[...]

PR #1996, opened 2019-10-31
Slick: factory method SlickSession.forDbAndProfile

Kreinoee:
## Purpose

Introduces an extra factory method for creating a SlickSession out of Slick Database and Profile instances.

## References

References #1509

## Changes

* Created method forDbAndProfile(db: JdbcBackend#Database, profile: JdbcProfile): SlickSession in the scaladsl version.
* Before this changes, the scala version of SlickSession object was the same as the javadsl version. Now these two are different, as the above method, only makes sense from scala, due to Slicks missing Java API.
* To avoid duplicat code, all the shared logic of the two SlickSession objects, has been moved to a SlickSessionFactory abstract class, that they both extend.

## Background Context

It makes it possible to create a SlickSession, if you starting point is a Slick Database instance. This can be the case, when working with other libraries, that are based on Slick, but now nothing about Alpakka. In my case, it was a test library, that provides each test, with its own test database, provided as a Slick Database instance.

### DIFF
--- a/docs/src/main/paradox/slick.md
+++ b/docs/src/main/paradox/slick.md
@@ -84,6 +84,13 @@ SQL Server
 
 Of course these are just examples. Please visit the @extref[Slick documentation for `DatabaseConfig.fromConfig`](slick:api/index.html#slick.jdbc.JdbcBackend$DatabaseFactoryDef@forConfig(String,Config,Driver,ClassLoader%29:Database) for the full list of things to configure.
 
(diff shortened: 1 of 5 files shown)


ennru:
You need to add MiMa exclusions in a file `slick/src/main/mima-filters/1.1.x.backwards.excludes` 
See https://travis-ci.com/akka/alpakka/jobs/251451264#L679

Kreinoee:
Done, thanks for the hint.

ennru:
Thank you!

PR #2000, opened 2019-11-06
Upgrade plugins and common dependencies

ennru:
## Purpose

Get more versions in shape for the 2.0.0-M1 milestone.

## Changes

* Current plugin versions (commit 1)
* Common and test library updates (commit 2)
	* Scala 2.13.1
    * Akka 2.6.0 for scheduled build
* Reformatted Java code (commit 3)
* Show JDK 11 in project info (commit 4)
* Depend directly on Jackson Core when Databind is used (commit 6)


ennru:
S3 tests fail, most likely connected to the upgrade of WireMock.

PR #2001, opened 2019-11-07
AWS S3: discard entity on failed sign request

ennru:
## Purpose

Ensure the HTTP response entity data is discarded when a signing response fails.

## Changes

* discard entity data in `signAndRequest`
* change Javadsl `listBucket` to use `Optional` instead of `scala.Option` (old methods deprecated)
* add shutdown of Actor System and WireMock in `S3Test`
* upgrade WireMock to 2.25.1 and adapt empty request mocking


seglo:
Some failures from other projects

https://github.com/akka/alpakka/issues/2003 (first observation)
https://github.com/akka/alpakka/issues/1649

PR #2009, opened 2019-11-13
Add more metadata about storage objects to google-cloud-storage api

DanieleSassoli:
## Purpose

The intent of this PR is to be able to retrieve more metadata about storage objects via the alpakka api. 
The fields that I have added are fairly arbitrary and I'm happy to remove some if people think they aren't needed or add others if people think they'd be more valuable.

Mainly I'm interested in the creation and update times of the objects.

## Background Context

I'm writing an interface that should enable me to treat Sftp, Ftps, S3 and GCS files in the same way, I noticed that I could get information around when the document was last modified for Sftp and Ftps(still haven't got round to S3) but not for GCS, so thought I'd add it in.


DanieleSassoli:
I see the binary compatibilities checks are failing, any suggestions on how/if to fix them? Would I need to provide default values?

seglo:
Hi @DanieleSassoli.  Thanks for the PR.  The MiMa error can be ignored in this case because the constructor that was modified is private, but you added the appropriate builder methods (`withSomething`) for each property, so it's OK.  Add the violations as exceptions as documented in the [`CONTRIBUTING`](https://github.com/akka/alpakka/blob/master/CONTRIBUTING.md#binary-compatibility-mima) doc to suppress the errors.

DanieleSassoli:
Hi @seglo, thanks for the advice, I hope I've done the correct change.

seglo:
Hi @DanieleSassoli.  When adding a MiMa exception you can use the problem definition mentioned in the violation message.  In this case your exception file should include:

```
# Add more metadata about storage objects to google-cloud-storage api #2009
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.googlecloud.storage.StorageObject.this")
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.googlecloud.storage.StorageObject.apply")
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.googlecloud.storage.StorageObject.create")

``` 

You'll also need to use the filename convention `2.0.0-M1.backwards.excludes` (`2.0.0-M1` was released last week)

DanieleSassoli:
@seglo, thanks for explaining that, will amend.

google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/StorageObject.scala
```diff
@@ -140,7 +190,13 @@ object StorageObject {
       etag: String,
       md5Hash: String,
       crc32c: String,
-      mediaLink: String
+      mediaLink: String,
```
seglo: Modifying `StorageObject.apply` and `StorageObject.create` will cause a binary incompatibility with this API.  Ideally we would like to avoid this scenario because a user may have code that already exists to create `StorageObjects` to transmit to the google-cloud-storage connector's `Sink` or `Flow` APIs.  If there's a good reason to modify these methods (and the `with` builder methods are not good enough), then we should consider adding constructor overloads to accomodate the extra parameters.  Do you need to modify these methods?

DanieleSassoli: Hi @seglo, good point! I believe the `with` builder methods should be good enough, I'll amend my PR.

seglo: Thanks.  You'll also be able to revert these two MiMa exceptions, but keep the one on `this`.

```
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.googlecloud.storage.StorageObject.apply")
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.googlecloud.storage.StorageObject.create")
```

DanieleSassoli: hi @seglo, just before I go off and do this, the only way I can see to implement this is by making the parameters I'm adding optional and giving them a default value of none. 
Is that ok?

seglo: Hi @DanieleSassoli.  Thanks for checking in, it made me take a closer look at this.  I looked at the [Google Cloud Storage API spec](https://cloud.google.com/storage/docs/json_api/v1/objects#resource-representations) and noticed the additional fields you would like to add.  I assume that when this Alpakka connector was first created, that these fields didn't exist and that they were added later on as non-breaking changes by Google (non-breaking because it's just a JSON API).

I also noticed that nowhere in the Alpakka DSL do we actually expose a way for users to create their own `StorageObject` for the purpose of writing data to google cloud.  We only let the user provide a storage object name identifier and a `ByteString`, and `StorageObjects` are only ever deserialized from google API responses.  In fact, I don't see any internal use of our `with` methods either, so I assume these were added at some point for reasons that are unclear to me (maybe for internal testing?)

Therefore I don't see any reason for not modifying the constructor with additional fields as your PR originally intended.  
Since you are more familiar with this API I would like your advice on how to proceed.  Is my understanding of the usage of the `StorageObject` correct?  Is there ever a reason (using this Alpakka connector) to mutate fields of the `StorageObject`?  I notice that the spec suggests some fields are "writable", but I don't think that's relevant to this Alpakka connector?

DanieleSassoli: Thanks for looking into this and getting back to me. 
Your analysis makes sense, the only reason we have to generate a `StorageObject` and mutate its fields is for testing purposes. 
I've been able to generate an `Arbitrary[StorageObject]` via the `StorageObject.apply` method, this allows me to generate random instances of `StorageObject` for my tests, I then use the `with` method to modify those 2 or 3 fields that I need to have control over(like `name` and `bucket`).

You're also correct when you say that within the API they're only ever deserialised from responses. 

Does this answer your questions?

DanieleSassoli: No, not all the fields are there at the moment, I think a few are still missing as I didn't know if this was of any interest so to begin with I only took the ones I considered more relevant. 
I've got a busy day today and tomorrow but will try to get this sorted before the weekend. 
Thanks again for your help.

DanieleSassoli:
@seglo fixing the current issues ended up taking less than expected, build should be passing now 🤞 
We're still missing a fair amount of fields from the official GCS documentation that you linked above:
```
"metageneration": long,
"timeDeleted": datetime,
"temporaryHold": boolean,
"eventBasedHold": boolean,
"retentionExpirationTime": datetime,
"timeStorageClassUpdated": datetime,
"contentLanguage": string,
"cacheControl": string,
"metadata": {
  (key): string
},
"acl": [
  objectAccessControls Resource
],
"owner": {
  "entity": string,
  "entityId": string
},
"componentCount": integer,
"customerEncryption": {
  "encryptionAlgorithm": string,
  "keySha256": string
},
"kmsKeyName": string
```
Unless you have comments on this, I'll go ahead and add this when I get a moment...

seglo:
> Unless you have comments on this, I'll go ahead and add this when I get a moment...

Thanks.  I think it would be great to add them while we're at it.

DanieleSassoli:
So had a go at implementing this and I'm afraid I've encountered a bit of a hard limit, spray.json.DefaulJsonProtocol(used in Formats.scala) only accepts up to 22 parameters (there a `jsonFormat22` function but nothing north of that), how do you suggest we resolve this?

seglo:
That's a pain.  You'll have to implement your own `JsonFormat` for the type to workaround this limitation.  There's an example in the [`spray-json`](https://github.com/spray/spray-json#providing-jsonformats-for-other-types) docs.

seglo:
@DanieleSassoli it would be great to get this into Alpakka 2.0.0 since it's technically a breaking change.  If you want to commit what you have I could implement the `JsonFormat`, but I'd like to rely on your expertise with the actual usage of GCS to assure that the implementation works.

DanieleSassoli:
hey @seglo, sorry, haven't had time to look into this... I don't have anything compiling at the moment, I'll try to sort something out this evening and close it off. Sorry about this, would have like to finish it sooner but didn't manage...

DanieleSassoli:
@seglo, I didn't manage to get this to work just yet, not being able to get alpakka to compile in intelliJ doesn't help either.
There's still 1 or 2 fields missing that I was going to add once I had finished this part. It would be helpful if you could implement the correct `JsonFormat` for this so that I can then finish off with the other fields.

seglo:
@DanieleSassoli I created a custom `JsonFormat` for `StorageObjectJson`.  To project is compiling, but there are a bunch of tests failing.  I wasn't able to push the commit to your fork for some reason, and I rebased on master without thinking.  Anyway, the relevant changes are in this commit:

https://github.com/DanieleSassoli/alpakka/pull/2/commits/ff32eb621ac4b9905bd06c2a6215ee712d114969

I hope that helps.  I'll be on vaca until next Weds.

DanieleSassoli:
hey @seglo I've now added all of the fields to the storage object. Please double check that having the `acl` field as an `Option[List[ObjectAccessControls]]` makes sense to you as well.
I'll try and sort out the compilation errors in `Formats.scala` next, but please if you have time and I haven't finished yet feel free to complete that side of things.

seglo:
Thanks @DanieleSassoli .  I may have some cycles next week. Can you give me access to push changes directly to your branch?  That will make collaboration easier.  I wasn't able to do this earlier, so I think you need to do something to allow it.

https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

DanieleSassoli:
Thanks for that, I believe you should now be able to push changes to this branch, let me know if not.
I noticed that compilation failed as it can't find the `com.google.api.services.storage.model.ObjectAccessControls`, not sure what's the best way forward here, I'm sure alpakka would import the google api libraries so I expected this to be available. I can create a new class like I did with Owner and CustomerEncryption but not sure that's the best way forward?

DanieleSassoli:
Hey @seglo, I went ahead and did my own implementation of ObjectAccessControls for what it was needed in our case, I hope this makes sense. I'm not 100% sure how to fix the current compilation issues in `Formats.scala`, I'll keep having a look but any help is welcome.

DanieleSassoli:
right so it looks like I might have figured out a way of getting it to work, please let me know if this is not the intended way.

seglo:
Hi @DanieleSassoli.  Thanks for continuing this effort.  The MiMa check failed because we released Alpakka 2.0.0-M2 earlier this week and the excludes file needed to be updated to reflect that these are breaking changes _since_ that version.  I can't seem to push to your master branch, but you can find the commit on my branch here https://github.com/seglo/alpakka/commit/5e79e66d3319701f018196d93e171f6d2e55bc83

The jsonformat's you added look good :+1: 

There are still a number of test failures.  If you're having trouble running them locally in IntelliJ, you can run them from sbt with `sbt google-cloud-storage/test`.  It would be faster turn around to go into an sbt session, and then run the tests as you make updates.

I think part of the problem I'm having contributing to your branch is that you are making changes in your local master.  I suggest creating a new branch and squashing all your commits.  You'll probably have to recreate this PR, but it will make it easier for me to collaborate with you, and to rebase it on master as it's falling behind.

DanieleSassoli:
hey @seglo, thanks for that, good to know that's alright. I'll have a look at the tests and see if I can fix them between today and tomorrow. 
What you say about me working off my master makes sense, if I can't fix the tests I'll create a branch on my fork and create a PR from that branch. Sorry for the confusion.

lightbend-cla-validator:

DanieleSassoli:
Hey @seglo I'm gonna close this as we now have #2058 that has been created from a branch so should hopefully allow you to collaborate more easily. 
Thanks so much for the help on this PR, much appreciated and sorry if I've caused a bit of a mess.

Issue #1802, opened 2019-07-01
Failed: HttpRequestsSpec should support custom endpoint configured by `endpointUrl`


2m:
```
[info] - should support custom endpoint configured by `endpointUrl` *** FAILED *** (85 milliseconds)
[info]   java.net.BindException: Cannot assign requested address
[info]   at sun.nio.ch.Net.bind0(Native Method)
[info]   at sun.nio.ch.Net.bind(Net.java:433)
[info]   at sun.nio.ch.Net.bind(Net.java:425)
[info]   at sun.nio.ch.ServerSocketChannelImpl.bind(ServerSocketChannelImpl.java:223)
[info]   at sun.nio.ch.ServerSocketAdaptor.bind(ServerSocketAdaptor.java:74)
[info]   at sun.nio.ch.ServerSocketAdaptor.bind(ServerSocketAdaptor.java:67)
[info]   at akka.testkit.SocketUtil$.$anonfun$temporaryServerAddresses$1(SocketUtil.scala:97)
[info]   at scala.collection.generic.GenTraversableFactory.fill(GenTraversableFactory.scala:89)
[info]   at akka.testkit.SocketUtil$.temporaryServerAddresses(SocketUtil.scala:81)
[info]   at akka.testkit.SocketUtil$.temporaryServerAddress(SocketUtil.scala:74)
```

https://travis-ci.com/akka/alpakka/jobs/212184452

seglo:
Same test, different error (timeout)

https://travis-ci.com/akka/alpakka/jobs/260163601#L911

seglo:
https://travis-ci.com/akka/alpakka/jobs/260670123#L723

ennru:
https://travis-ci.com/akka/alpakka/jobs/259855779#L779 in #2006

seglo:
https://travis-ci.com/akka/alpakka/jobs/268926798#L767

seglo:
https://travis-ci.com/akka/alpakka/jobs/272322371#L1470

seglo:
https://travis-ci.com/akka/alpakka/jobs/275777739#L433

PR #2011, opened 2019-11-15
Use MiMa 0.6.0

ennru:


seglo:
https://github.com/akka/alpakka/issues/1802

PR #2012, opened 2019-11-18
Unix domain socket: API changes

huntc:
Taking advantage of the 2.0 release to change the use of File to Path (more correct) and the exposure of JNR in the public interface. The latter then paves the way for re-implementing JNR should we wish to e.g. as per https://github.com/akka/alpakka/pull/1659.


huntc:
@ennru This is failing the binary compatibility test - which is to be expected. Are tests for binary compatibility really necessary on the master branch given that it is for 2.0? Thanks.

ennru:
Yes, these changes are important so I think it is OK to break the API.

unix-domain-socket/src/main/scala/akka/stream/alpakka/unixdomainsocket/impl/UnixDomainSocketImpl.scala
```diff
@@ -397,7 +405,7 @@ private[unixdomainsocket] abstract class UnixDomainSocketImpl(system: ExtendedAc
                     .andThen {
                       case _ =>
                         try {
-                          file.delete()
+                          path.toFile.delete()
```
seglo: We could use `Files.delete(path)` to stay consistent with the usage of NIO types.  According to `File.delete` javadocs, `Files.delete` will also throw an `IOException` when applicable, but since we're squelching the exception anyway I guess it's not that important.

unix-domain-socket/src/main/scala/akka/stream/alpakka/unixdomainsocket/impl/UnixDomainSocketImpl.scala
```diff
@@ -256,12 +258,18 @@ private[unixdomainsocket] object UnixDomainSocketImpl {
...
       } catch { case _: IOException => }
       incomingConnectionQueue.offer(
-        IncomingConnection(localAddress, acceptingChannel.getRemoteSocketAddress, connectionFlow)
+        IncomingConnection(
+          UnixSocketAddress(Paths.get(localAddress.path())),
+          UnixSocketAddress(
+            Paths.get(Option(acceptingChannel.getRemoteSocketAddress).getOrElse(new JnrUnixSocketAddress("")).path())
+          ),
+          connectionFlow
+        )
```
seglo: The instantiation of `IncomingConnection` may benefit from some named parameters in this expanded form.

huntc:
> If I understand correctly, your comment about removing the exposure of JNR in the public interface is so that the underlying JNR implementation may be more easily swapped out in the future by creating the new `UnixSocketAddress` type.

Correct. JNR is implementation and shouldn't appear in the public API.

> I left a few other minor comments to consider, but otherwise it LGTM.

Thanks!

unix-domain-socket/src/main/scala/akka/stream/alpakka/unixdomainsocket/impl/UnixDomainSocketImpl.scala
huntc: Will do.

unix-domain-socket/src/main/scala/akka/stream/alpakka/unixdomainsocket/impl/UnixDomainSocketImpl.scala
huntc: Will do.

huntc:
> Since we're ok with the breaking changes for 2.0.0 release and that there are quite some number of MiMa violations reported I suggest we create several [package level rules](https://github.com/lightbend/mima#filtering-binary-incompatibilities) for `javadsl` and `scaladsl` packages instead of adding each violation entry.

Is there anything I need to do re this?

seglo:
> Is there anything I need to do re this?

Yes, sorry I should have mentioned that.  You'll need to add a MiMa excludes file to the unix domain socket project that suppresses the violations raised by the `mimaReportBinaryIssues` sbt task.  There's a [section in the contributing](https://github.com/akka/alpakka/blob/master/CONTRIBUTING.md#binary-compatibility-mima) doc that explains what you need to do (LMK if you have questions).  FYI the previous release version is the milestone release: 2.0.0-M1.

huntc:
> You'll need to add a MiMa excludes file to the unix domain socket project...

MiMa excludes are now in.

unix-domain-socket/src/main/mima-filters/2.0.0-M1.backwards.excludes
```diff
@@ -0,0 +1,35 @@
+# PR #2012
```
seglo: We normally put MiMa excludes in a separate file per PR for better organization.

```
<module>/src/main/mima-filters/<last-version>.backwards.excludes/<pr-or-issue>-<issue-number>-<description>.excludes
```

seglo: Well, not _normally_, but it's something we've started doing since MiMa now supports it :)

huntc: I've just gone through all of the sub projects and couldn't find an example. May I recommend that you separately create another PR reorganising all of the MiMa files to your new convention? I was expecting this to be a quick PR. ;-)

seglo: Yes, that's a good point.  I don't think we've done this in Alpakka yet since adopting the new convention.  

You can find an example in Alpakka Kafka here: https://github.com/akka/alpakka-kafka/tree/master/core/src/main/mima-filters/1.1.0.backwards.excludes

All that needs to be done is to create `./unix-domain-socket/src/main/mima-filters/2.0.0-M1.backwards.excludes` as a directory and move your MiMa file into it and give it a filename like: `PR2012-unix-domain-socket-api-changes.excludes`.

huntc: Unfortunately, I've no more time for this PR. I will adopt the convention going forward.

seglo: Alright.  I pushed an update to your branch.  I'll merge it after the build is complete.

huntc: Thanks mate.

seglo:
Thanks for your contribution!

PR #2015, opened 2019-11-18
Docs: more MiMa docs

ennru:
Explain the use of MiMa a bit more (copied in from https://github.com/akka/akka/blob/master/CONTRIBUTING.md#binary-compatibility) now that we intend to use the directory-style.

### DIFF
--- a/contributor-advice.md
+++ b/contributor-advice.md
@@ -96,7 +96,30 @@ See [Binary Compatibilty Rules](https://doc.akka.io/docs/akka/current/common/bin
 
 See [Binary Compatibility for library authors](https://docs.scala-lang.org/overviews/core/binary-compatibility-for-library-authors.html)
 
-Use [MigrationManager (MiMa)](https://github.com/lightbend/migration-manager) to validate, if versions are binary compatible. MiMa is part of the Alpakka build and its checks can be triggered by `mimaReportBinaryIssues`.
+### Using MiMa
+
+Alpakka uses [MiMa](https://github.com/lightbend/mima) to validate binary compatibility of incoming pull requests. If your PR fails due to binary compatibility issues, you may see an error like this:
+
+```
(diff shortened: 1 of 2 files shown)


contributor-advice.md
```diff
@@ -96,7 +96,30 @@ See [Binary Compatibilty Rules](https://doc.akka.io/docs/akka/current/common/bin
...
+
+Alpakka uses [MiMa](https://github.com/lightbend/mima) to validate binary compatibility of incoming pull requests. If your PR fails due to binary compatibility issues, you may see an error like this:
+
+```
+[info] akka-stream: found 1 potential binary incompatibilities while checking against com.typesafe.akka:akka-stream_2.12:2.4.2  (filtered 222)
+[error]  * method foldAsync(java.lang.Object,scala.Function2)akka.stream.scaladsl.FlowOps in trait akka.stream.scaladsl.FlowOps is present only in current version
+[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.stream.scaladsl.FlowOps.foldAsync")
+```
+
+In such situations it's good to consult with a core team member whether the violation can be safely ignored or if it would indeed
```
seglo:
```suggestion
In such situations, it's good to consult with a core team member about whether the violation can be safely ignored or if it would indeed
```

contributor-advice.md
```diff
@@ -96,7 +96,30 @@ See [Binary Compatibilty Rules](https://doc.akka.io/docs/akka/current/common/bin
...
+break binary compatibility. If the violation can be ignored add exclude statements from the mima output to
```
seglo:
```suggestion
break binary compatibility. If the violation can be ignored add exclude statements from the MiMa output to
```

ennru:
You're right, I should update `CONTRIBUTING.MD` instead.

PR #2019, opened 2019-11-19
mkdir flow

ennru:
## Purpose

Add new flow factories which will create a directory for every element sent through.

## Background Context

When refreshing the [Alpakka Samples](https://akka.io/alpakka-samples/) I noticed the "Download all file from FTP" example is lacking the creation of local directories.

### DIFF
--- a/docs/src/main/paradox/file.md
+++ b/docs/src/main/paradox/file.md
@@ -76,6 +76,19 @@ Scala
 Java
 : @@snip [snip](/file/src/test/java/docs/javadsl/FileTailSourceTest.java) { #shutdown-on-idle-timeout }
 
+
+## Creating directories
+
+`Directory.mkdirs()` and `Directory.mkdirsWithContext()` create directories for @javadoc[Path](java.nio.file.Path) elements in the stream. The `withContext`-variant allows to pass through other information with every path.
+
+Scala
+: @@snip [snip](/file/src/test/scala/docs/scaladsl/DirectorySpec.scala) { #mkdirs }
+
+Java
+: @@snip [snip](/file/src/test/java/docs/javadsl/DirectoryTest.java) { #mkdirs }
+
+
+
 ## Listing directory contents
 
 `Directory.ls(path)` lists all files and directories
--- a/docs/src/main/paradox/file.md
+++ b/docs/src/main/paradox/file.md
@@ -79,7 +79,7 @@ Java
 
 ## Creating directories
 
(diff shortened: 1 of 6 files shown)


file/src/main/scala/akka/stream/alpakka/file/scaladsl/Directory.scala
```diff
@@ -42,4 +43,26 @@ object Directory {
     StreamConverters.fromJavaStream(factory)
   }
 
+  /**
+   * Create local directories, including any parent directories.
+   */
+  def mkdirs(): Flow[Path, Path, NotUsed] =
+    Flow[Path]
+      .map(Files.createDirectories(_))
```
seglo: I know filesystem ops are synchronous, but I wonder if this could be a bottleneck when we want to create lots of directories.  Would it make sense to offer an async version or leave it as a task to the user?

docs/src/main/paradox/file.md
```diff
@@ -76,6 +76,19 @@ Scala
 Java
 : @@snip [snip](/file/src/test/java/docs/javadsl/FileTailSourceTest.java) { #shutdown-on-idle-timeout }
 
+
+## Creating directories
+
+`Directory.mkdirs()` and `Directory.mkdirsWithContext()` create directories for @javadoc[Path](java.nio.file.Path) elements in the stream. The `withContext`-variant allows to pass through other information with every path.
```
seglo:
```suggestion
`Directory.mkdirs()` and `Directory.mkdirsWithContext()` create directories for @javadoc[Path](java.nio.file.Path) elements in the stream. The `withContext`-variant allows the user to pass through additional information with every path.
```

file/src/main/scala/akka/stream/alpakka/file/scaladsl/Directory.scala
ennru: This will run on Akka's IO Dispatcher so it is async. What kind of async do you have in mind?

seglo: I mean should we use `mapAsync` with some default parallelism when creating directories?

ennru: I don't think that's useful. It would require wrapping it in a `Future` which would need an execution context.
In the end, the file system is a giant shared resource with a big semaphore somewhere. Creating directories in parallel might even show strange behaviour.

PR #2020, opened 2019-11-19
some review comments/questions on mqtt-streaming

patriknw:
@huntc you asked for feedback so I looked at the mqtt-streaming code...

The most important question is what is the reason for using Promise for the replies, instead of using messages and ActorRef?

### DIFF
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ServerState.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ServerState.scala
@@ -126,8 +126,7 @@ import scala.util.{Failure, Success}
           data.unpublisherPacketRouter ! RemotePacketRouter.UnregisterConnection(connectionId)
 
           QueueOfferState.waitForQueueOfferCompleted(
-            data.terminations
-              .offer(ClientSessionTerminated(clientId)),
+            data.terminations.offer(ClientSessionTerminated(clientId)),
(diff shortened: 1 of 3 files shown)


huntc:
> The most important question is what is the reason for using Promise for the replies, instead of using messages and ActorRef?

No particular reason for when requester is an actor.

Also, I was thinking that you’d comment on the use of partial handlers. That’s one thing I was thinking of eliminating.

patriknw:
This has served it's purpose - to give feedback

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ServerState.scala
```diff
@@ -953,6 +924,7 @@ import scala.util.{Failure, Success}
   // State event handling
 
   def prepareServerUnpublisher(data: Start): Behavior[Event] = Behaviors.setup { context =>
+    // FIXME using Promise for replies seems strange. I would prefer ActorRef and messages.
     val reply = Promise[RemotePacketRouter.Registered.type]
     data.packetRouter ! RemotePacketRouter.Register(context.self.unsafeUpcast, data.clientId, data.packetId, reply)
```
huntc: @patriknw I’m having a further think on the use of promises for ask patterns within types actors. The convention is to wrap actor reply messages, but is this really better than passing in a promise for a reply?

The only downside to the promise that I can see is that one has to attempt to fail it after a timeout given no success (so that resources are cleared up).

The upside is the avoidance of having to create a wrapped message type.

Issue #2021, opened 2019-11-19
AWS S3 List only one hierarchy level of objects by using the delimiter parameter


an-tex:
### Short description

By using the [delimiter parameter](https://docs.aws.amazon.com/AmazonS3/latest/dev/ListingKeysHierarchy.html) it is possible to list the common prefixes of only one hierarchical level.

### Details

In the current implementation only the prefix can be set. To allow browsing level by level it should be possible to set the delimiter parameter as well. See: [S3 API Documentation](https://docs.amazonaws.cn/en_us/AmazonS3/latest/API/API_ListObjectsV2.html#API_ListObjectsV2_RequestSyntax)

I have already started working on a PullRequest and will add it shortly.

seglo:
Good idea.  Looking forward to the PR!

seglo:
Implemented in #2023

PR #2023, opened 2019-11-20
AWS S3 added listObjects endpoint including common prefixes for a delimiter

an-tex:
## Purpose

The listObjects endpoint allows to specify a delimiter, which is used to return only one hierarchy level of objects. This allows for a directory style browsing (like in the AWS S3 console) without having to retrieve all objects.

## References

References https://github.com/akka/alpakka/issues/2021

## Changes

* added listObjects function taking an optional delimiter parameter

## Background Context

I've choosen to create a new function instead of changing the existing listBucket one. The listBucket function returns a `ListBucketResultContents`, which can't be a common prefix. Compared to the actual listObjects endpoint, which can return both, `ListBucketResultContents` and `ListBucketResultCommonPrefixes` ([API](https://docs.amazonaws.cn/en_us/AmazonS3/latest/API/API_ListObjectsV2.html#API_ListObjectsV2_ResponseSyntax), [Guide](https://docs.aws.amazon.com/AmazonS3/latest/dev/ListingKeysHierarchy.html)). 

To reflect the two kinds of stream events, the new listObjects Source emits a `ListBucketResultBase` trait, which can either be a `ListBucketResultContents` or `ListBucketResultCommonPrefixes`. 

This change would have broken existing implementations, that's why I opted to create a new function. Does this make sense? 

I postponed updating the documentation until this approach is validated.


lightbend-cla-validator:

an-tex:
> At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

I've added the missing email address and signed the CLA

an-tex:
I'm not sure what the story with the failing build is. The same test is failing for me in Intellij, but even before my commits. When I run it in SBT directly it's fine for both.

seglo:
> I'm not sure what the story with the failing build is. The same test is failing for me in Intellij, but even before my commits. When I run it in SBT directly it's fine for both.

That's alright.  We've previously identified this as a flakey test.  It needs further investigation, but it may just need a longer timeout since travis VMs can be rather underpowered.

https://github.com/akka/alpakka/issues/1802

Thanks for the PR!  I'll do a review later today.

s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3.scala
```diff
@@ -483,6 +483,27 @@ object S3 {
       .listBucket(bucket, prefix.asScala, s3Headers)
       .asJava
 
+  /**
+   * Will return a source of object metadata and common prefixes for a given bucket with optional prefix and delimiter using version 2 of the List Bucket API.
+   * This will automatically page through all keys with the given parameters.
+   *
+   * The `alpakka.s3.list-bucket-api-version` can be set to 1 to use the older API version 1
+   *
+   * @see https://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html  (version 1 API)
```
seglo: I assume this is actually the v2 API.  Can you update this reference and the one in `listBucket` javadoc?

s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3.scala
```diff
@@ -483,6 +483,27 @@ object S3 {
...
+   * @see https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html (version 1 API)
+   * @see https://docs.aws.amazon.com/AmazonS3/latest/dev/ListingKeysHierarchy.html (prefix and delimiter documentation)
+   * @param bucket    Which bucket that you list object metadata for
+   * @param prefix    Prefix of the keys you want to list under passed bucket
+   * @param delimiter Delimiter to use for listing only one level of hierarchy
+   * @param s3Headers any headers you want to add
+   * @return [[akka.stream.scaladsl.Source Source]] of [[ListBucketResultBase]]
+   */
+  def listObjects(bucket: String,
```
seglo: It seems that this should be another `listBucket` overload.  Is there a reason you gave it a different name?

seglo: Or better yet, `listBucketWithPrefixes` to accommodate the different return type.

an-tex:
Thanks for the feedback. Sorry about the DRY, it seemed some discussion would follow so I wanted to keep it WET and wait with premature optimisations until we've decided the final interface. Probably should have mentioned that ;) Once it's in place I'll follow your guide.

> For the object with prefix results, you could return a tuple of (ListBucketResultContents, ListBucketResultCommonPrefixes) instead of trying to combine them together with the new base trait you added.

One streaming event can be EITHER a Contents OR a CommonPrefixes item. So I'm not sure how returning a tuple would work here. The only way to keep it separate would be to return a tuple of Sources with those types.

s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3.scala
an-tex: listBucketWithPrefixes could work. Then we could make the delimiter parameter even non-optional. 

I've taken the listObjects name from the actual S3 API Naming as I was shortly confused myself which alpakka S3 function is linked to which API call. But I guess this only makes sense if in the long run the listBucket function gets deprecated and completely replaced by listObjects*. Until then it would probably just confuse more. So yea lets make it listBucketWithPrefixes

an-tex: Ahh it seems this "list contents/prefixes" endpoint used to be called "GET Bucket (List Objects)" in the documentation but has been renamed to "ListObjectsV2". Now I get where the original name is coming from

s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3.scala
an-tex: Those links seem to be outdated and redirect to https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html (and https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html accordingly). I'll update the links and fix the version numbers

ennru: Please do, see eg. https://github.com/akka/alpakka/pull/2024

an-tex: done. i've updated the other api links too

seglo:
> Thanks for the feedback. Sorry about the DRY, it seemed some discussion would follow so I wanted to keep it WET and wait with premature optimisations until we've decided the final interface. Probably should have mentioned that ;) Once it's in place I'll follow your guide.

Sounds good!

> One streaming event can be EITHER a Contents OR a CommonPrefixes item. So I'm not sure how returning a tuple would work here. The only way to keep it separate would be to return a tuple of Sources with those types.

I see.  Thanks for the clarification.  I understand now why you took this approach.  Since we're exploding the contents results there's no real appropriate place to include the prefixes result too, so you've added them to the stream as elements.  This leaves it to the end user to match on the correct concrete type.  I suppose one alternative would be to add a reference to the prefixes in every `ListBucketResultContents` emitted for that response.  That would keep the return type the same.  WDYT?

an-tex:
> I suppose one alternative would be to add a reference to the prefixes in every `ListBucketResultContents` emitted for that response. That would keep the return type the same. WDYT?

You mean every `ListBucketResultContents` element contains an additional field of type `Source[ListBucketResultCommonPrefixes]` as the reference?

seglo:
> You mean every `ListBucketResultContents` element contains an additional field of type `Source[ListBucketResultCommonPrefixes]` as the reference?

I mean iterate over `ListBucketResult.contents` and emit downstream a new type like `ListBucketResultContents`, but that includes a `Seq[ListBucketResultCommonPrefixes]`.

Though your reply made me think of another approach.  Instead of merging results and common prefixes into a single stream you could return a `Source` for each, which the user could then handle separately.

an-tex:
> I mean iterate over `ListBucketResult.contents` and emit downstream a new type like `ListBucketResultContents`, but that includes a `Seq[ListBucketResultCommonPrefixes]`.
> 

I think the CommonPrefixes should be Elements of a Source as well as there might still be a lot of them.

> Though your reply made me think of another approach. Instead of merging results and common prefixes into a single stream you could return a `Source` for each, which the user could then handle separately.

That's what I meant earlier by a tuple of sources. I prolly should use more code as examples instead of confusing descriptions ;)

How about this interface:

```scala
  def listBucketWithCommonPrefixes(
                  bucket: String,
                  delimiter: String,
                  prefix: Option[String] = None,
                  s3Headers: S3Headers = S3Headers.empty
): Source[(ListBucketResultCommonPrefixes, ListBucketResultContents), NotUsed]
```

This way the user could consume both independently. Even though I wonder if we should internally keep the common base trait. Then the user could still opt in to concat both sources and do pattern matching on a single stream (similar to the original approach).

seglo:
> I think the CommonPrefixes should be Elements of a Source as well as there might still be a lot of them.

True, but they've already been deserialized to a `Seq` from that page/response, so referencing that `Seq` multiple times shouldn't be a concern, though it could be confusing if the common prefixes are paged as well.

> `Source[(ListBucketResultCommonPrefixes, ListBucketResultContents), NotUsed]`

Do you mean this:

```scala
Source[(Source[ListBucketResultCommonPrefixes, NotUsed], Source[ListBucketResultContents, NotUsed]), NotUsed]
```

an-tex:
> True, but they've already been deserialized to a `Seq` from that page/response, so referencing that `Seq` multiple times shouldn't be a concern, though it could be confusing if the common prefixes are paged as well.

Exactly, the `CommonPrefixes` should be treated just like the `Contents` especially due to the paging.
 
> > `Source[(ListBucketResultCommonPrefixes, ListBucketResultContents), NotUsed]`
> 
> Do you mean this:
> 
> ```scala
> Source[(Source[ListBucketResultCommonPrefixes, NotUsed], Source[ListBucketResultContents, NotUsed]), NotUsed]
> ```

Ah sorry, I actually meant:

```scala
  def listBucketWithCommonPrefixes(
                                    bucket: String,
                                    delimiter: String,
                                    prefix: Option[String] = None,
                                    s3Headers: S3Headers = S3Headers.empty
                                  ):
  (
    Source[ListBucketResultCommonPrefixes, NotUsed],
    Source[ListBucketResultContents, NotUsed]
  ) = ???
```

Here are a few usage use cases I'd see with this interface:

```scala
// I don't care about the contents, only commonPrefixes
val (commonPrefixes, _) = listBucketWithCommonPrefixes("bucket","/")
commonPrefixes.runForeach(???)

// I care about both separately
val (commonPrefixes, contents) = listBucketWithCommonPrefixes("bucket","/")
commonPrefixes.runForeach(???)
contents.runForeach(???)

// That's not really a use case as you can just use the existing listBucket function if you're only interested in the contents
val (_, contents) = listBucketWithCommonPrefixes("bucket","/")

// if we keep the ListContentsResultBase trait one could even merge them to one single stream and use pattern matching later
val (commonPrefixes, contents) = listBucketWithCommonPrefixes("bucket","/")
val merged: Source[ListBucketResultBase, NotUsed] = commonPrefixes.merge(contents)
  merged.runForeach {
    case contents: ListBucketResultContents => ???
    case prefixes: ListBucketResultCommonPrefixes => ???
  }
```

seglo:
I see.  That could work.  I would have thought this would be the most common use case:

```scala
// That's not really a use case as you can just use the existing listBucket function if you're only interested in the contents
val (_, contents) = listBucketWithCommonPrefixes("bucket","/")
```

Because you're using the delimiter to essentially filter one level of results, but you don't care about the common prefixes.

Do you want to experiment with this API?

seglo:
> ```scala
> // if we keep the ListContentsResultBase trait one could even merge them to one single stream and use pattern matching later
> val (commonPrefixes, contents) = listBucketWithCommonPrefixes("bucket","/")
> val merged: Source[ListBucketResultBase, NotUsed] = commonPrefixes.merge(contents)
>   merged.runForeach {
>     case contents: ListBucketResultContents => ???
>     case prefixes: ListBucketResultCommonPrefixes => ???
>   }
> ```

I don't think keeping `ListBucketResultBase` is that useful.  It seems odd to me to stream these two different things in one stream.  If the user really wants to do that they can do it themselves.

an-tex:
> Because you're using the delimiter to essentially filter one level of results, but you don't care about the common prefixes.

True! 
 
> Do you want to experiment with this API?

Yep I'll update the PR

an-tex:
> I don't think keeping `ListBucketResultBase` is that useful. It seems odd to me to have these two different things in one stream. If the user really wants to do that they can do it themselves.

My current use case is similar to the S3 console inside AWS: A "directory" based browser, listing both, files and commonPrefixes. Pretty much like an `ls` in the shell. So I need exactly that one stream ;)

Doing it yourself without a common base trait would involve mapping to an Either and then pattern match later again. It's not too bad but seems rather unnecessarily clunky:

```scala
  val (commonPrefixes, contents) = listBucketWithCommonPrefixes("bucket","/")
  
  val commonPrefixesEither = commonPrefixes.map(Right(_))
  val contentsEither = contents.map(Left(_))
  
  commonPrefixesEither.merge(contentsEither).runForeach {
    case Left(content) => ???
    case Right(commonPrefix) => ???
  }
```

Or am I missing some easier way?

I'd have thought leaving a common base trait wouldn't affect the API directly anyway so it does not hurt but makes it easier for the end user in case he needs it. In any case, that's not a show stopper for me I'll continue doing the other changes.

an-tex:
Well, the implementation of the Tuple Source seems to be tricky. I'm now down to having a
```scala
Source[(Seq[ListBucketResultCommonPrefixes], Seq[ListBucketResultContents]), NotUsed]
```
(every element is a page from the result)
but I don't know how to turn that into the desired 
```scala
(
    Source[ListBucketResultCommonPrefixes, NotUsed],
    Source[ListBucketResultContents, NotUsed]
  )
```
without forcing the client to use custom graph logic like e.g. described in https://stackoverflow.com/questions/38438983/split-akka-stream-source-into-two

Do you have any suggestions @seglo ? 

Otherwise returning to the original single source using a base trait might be the better solution after all? You can still fairly easy consume e.g. only the contents:
```scala
 def listBucketWithCommonPrefixes(
                                    bucket: String,
                                    delimiter: String,
                                    prefix: Option[String] = None,
                                    s3Headers: S3Headers = S3Headers.empty
                                  ): Source[ListBucketResultBase, NotUsed] = ???

  val contentsOnly: Source[ListBucketResultContents, NotUsed] =
    listBucketWithCommonPrefixes("bucket", "/").collect {
      case contents: ListBucketResultContents => contents
    }
```

seglo:
> My current use case is similar to the S3 console inside AWS: A "directory" based browser, listing both, files and commonPrefixes. Pretty much like an `ls` in the shell. So I need exactly that one stream ;)

I see.  I'm unfamiliar with this AWS S3 utility.  When you do a `ls` does it mix common prefixes with results?  Could you paste a snippet?

> Well, the implementation of the Tuple Source seems to be tricky. I'm now down to having a
> 
> ```scala
> Source[(Seq[ListBucketResultCommonPrefixes], Seq[ListBucketResultContents]), NotUsed]
> ```
> 
> (every element is a page from the result)
> but I don't know how to turn that into the desired

I think the return type you mention here is fine.  I tried playing around with this as well and discovered that it doesn't really make sense to return the nested sources since they would be emitted for every result anyway.  

I realize this makes your use case a little more tricky, but I think it's still the best option over mixing the elements together with a common base trait.  We could also forego returning a tuple/pair and just return the concrete upstream result object (or something like it).

I think you already have this code, but in case it's useful I pushed my experimental code as a demonstration.

https://github.com/an-tex/alpakka/pull/1/files

an-tex:
> > My current use case is similar to the S3 console inside AWS: A "directory" based browser, listing both, files and commonPrefixes. Pretty much like an `ls` in the shell. So I need exactly that one stream ;)
> 
> I see. I'm unfamiliar with this AWS S3 utility. When you do a `ls` does it mix common prefixes with results? Could you paste a snippet?

Exactly, it's just like an `ls` in your terminal/bash/console I meant ;)


> > Well, the implementation of the Tuple Source seems to be tricky. I'm now down to having a
> > ```scala
> > Source[(Seq[ListBucketResultCommonPrefixes], Seq[ListBucketResultContents]), NotUsed]
> > ```
> > 
> > 
> > (every element is a page from the result)
> > but I don't know how to turn that into the desired
> 
> I think the return type you mention here is fine. I tried playing around with this as well and discovered that it doesn't really make sense to return the nested sources since they would be emitted for every result anyway.

Sounds good let's just stick with that.

> I realize this makes your use case a little more tricky, but I think it's still the best option over mixing the elements together with a common base trait. We could also forego returning a tuple/pair and just return the concrete upstream result object (or something like it).

That's fine, it's just another streaming stage on my side. Let's just keep is like this and move forward.

> I think you already have this code, but in case it's useful I pushed my experimental code as a demonstration.
> 
> https://github.com/an-tex/alpakka/pull/1/files

Thanks! Yep it's mostly what I have on my side too. I'll finish it up and push the changes.

an-tex:
Alright, I've pushed it all. The only main change I made to your code was making the `delimiter` parameter mandatory. As otherwise the `commonPrefixes` would always be empty and the function would behave exactly like the original `listBucket`.

Otherwise:
- [x] paradox documentation
- [x] DRY
- [x] scala&java specs
- [x] merge upstream
- [x] build successful

s3/src/test/scala/docs/scaladsl/S3SourceSpec.scala
```diff
@@ -243,6 +243,44 @@ class S3SourceSpec extends S3WireMockBase with S3ClientIntegrationSpec {
...
+    val commonPrefixes = result.futureValue._2
+
+    contents.head.key shouldBe listKey
+    commonPrefixes.head.prefix shouldBe listCommonPrefix
+  }
+
+  it should "list keys and common prefixes for a given bucket with a prefix and delimiter using the version 1 api" in {
+    mockListBucketAndCommonPrefixesVersion1()
+
+    //#list-bucket-and-common-prefixes-attributes
```
seglo: I don't see a reference to this code snippet in docs.  These types of comments anchor code snippets that get extracted into paradox.

s3/src/main/scala/akka/stream/alpakka/s3/scaladsl/S3.scala
```diff
@@ -209,6 +209,29 @@ object S3 {
...
+   * The `alpakka.s3.list-bucket-api-version` can be set to 1 to use the older API version 1
+   *
+   * @see https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html  (version 2 API)
+   * @see https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html (version 1 API)
+   * @see https://docs.aws.amazon.com/AmazonS3/latest/dev/ListingKeysHierarchy.html (prefix and delimiter documentation)
+   * @param bucket    Which bucket that you list object metadata for
+   * @param delimiter Delimiter to use for listing only one level of hierarchy
+   * @param prefix    Prefix of the keys you want to list under passed bucket
+   * @param s3Headers any headers you want to add
+   * @return [[akka.stream.scaladsl.Source Source]] of [[ListBucketResultBase]]
```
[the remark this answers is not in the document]
an-tex: ok i'll check

s3/src/main/scala/akka/stream/alpakka/s3/model.scala
```diff
@@ -132,6 +132,11 @@ object FailedUpload {
   def create(reasons: Seq[Throwable]) = FailedUpload(reasons)
 }
 
+/**
+ * Base class for listObjects endpoint
+ */
+sealed trait ListBucketResultBase
```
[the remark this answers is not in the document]
an-tex: ah right that sneaky trait. i'll get rid of it

s3/src/test/scala/docs/scaladsl/S3SourceSpec.scala
an-tex: true, that was c&p from the listBucket which is used for the S3 Attributes example. I guess it's still ok to test the passing of the attributes but no need for the paradox comments. I'll get rid of them

an-tex:
> Something else I was considering was adding a new `listBucket` overload that only returns `ListBucketResultContents` for the one hierarchy level since that's going to be a common use case. It could have the same return type as the other `listBuckets` because it's just flattening out the one sequence and skipping common prefixes. We can do this in another PR though.

I've added it. Have a look if that's what you're after https://github.com/akka/alpakka/pull/2023/commits/3ea4954f2f9538df9a93d69fa8ae3d70a9f116f7 .

seglo:
> I've added it. Have a look if that's what you're after [3ea4954](https://github.com/akka/alpakka/commit/3ea4954f2f9538df9a93d69fa8ae3d70a9f116f7) .

Exactly what I was looking for.  Thanks!

seglo:
Some stage 1 check failures.  If you look at the right-most field it gives you the command you can run locally.

https://github.com/akka/alpakka/pull/2023/checks?check_run_id=321556931

an-tex:
> Some stage 1 check failures. If you look at the right-most field it gives you the command you can run locally.

Thanks for the tip. Was some back and forth due to cross compilation, looking good now!

seglo:
Thanks for the contribution @an-tex !  I appreciated your patience during the discussion :)

an-tex:
> Thanks for the contribution @an-tex ! I appreciated your patience during the discussion :)

Glad I could contribute! Thanks for the guidance @seglo 👍

PR #2024, opened 2019-11-20
Fixes typo in ScalaDoc

odd:


ennru:
Thank you, the same error applies to the other doc sections.
I guess we should remove the link to the API 1.x versions altogehter, as well.

seglo:
This has been addressed in #2023 .  Thanks for bringing it to our attention!

PR #2026, opened 2019-11-21
Google Cloud Storage: Add exponential backoff on server errors

### LINKED ISSUE #1931: Add backoff mechanism for google-cloud-storage
It appears that there is currently no backoff mechanism built into the Google Cloud Storage adapter in the case of HTTP 500 errors, but intended use of GCS according to the SLA stipulates that exponential backoff be used, and that HTTP 500 errors are somewhat expected.

The [GCS SLA](https://cloud.google.com/storage/sla) specifies uptime of >= 99.9% for single-region storage, with uptime being defined as the average percentage of requests that fail with HTTP 500 per 5 minutes over a one-month period.

My understanding of this SLA is that it is considered acceptable for an average of one request per thousand to fail with HTTP 500.

The SLA also specifies that exponential backoff is to be used in the case of such an error condition.

Seeing as such a relatively high error rate is expected and handling of it is specified within the SLA, it looks like exponential backoff should be built into the GCS adapter (and not be expected to be performed by downstream users of Alpakka).

gabrielreid:
## Purpose

Brings the GCS implementation in line with the documented
expected behavior for dealing with HTTP 500 responses.

## References

This PR resolves #1931 

Ticket #1931 pointed to akka/akka#27461 as another potential resolution, but that ticket appears to have been discarded in the meantime.

## Changes

* Exponential backoff is now used for waits of up to 32 seconds on 500 responses returned from GCS, as specified in the GCS SLA.

### DIFF
--- a/google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/impl/GCStorageStream.scala
+++ b/google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/impl/GCStorageStream.scala
@@ -4,6 +4,8 @@
 
 package akka.stream.alpakka.googlecloud.storage.impl
 
+import java.util.concurrent.atomic.AtomicInteger
+
 import akka.{Done, NotUsed}
 import akka.actor.ActorSystem
 import akka.annotation.InternalApi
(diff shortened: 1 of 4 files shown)


ennru:
Great! 
As this provides `Source`s, the use of a `RestartSource` makes total sense.
Just for the record, the `RetryFlow` ended up in Akka 2.6.0 in its simpler form in https://github.com/akka/akka/pull/27742

I wonder if this calls for introducing settings so that the back-off can be configured by the user.

gabrielreid:
> I wonder if this calls for introducing settings so that the back-off can be configured by the user.

I was thinking about having these things in settings as well -- I decided against it (for now) because these specific timings are specified in the GCS SLA, so (in my view) they are part of the specification of the interface. I'm definitely open to another point of view on this though of course.

ennru:
Oh, some trivial warnings make the build fail
https://travis-ci.com/akka/alpakka/jobs/258961188#L775

gabrielreid:
> Oh, some trivial warnings make the build fail
> https://travis-ci.com/akka/alpakka/jobs/258961188#L775

Oops, I guess the Travis build runs with some flags that I didn't use locally, sorry about that.

I've fixed the warnings (I believe) and pushed a new version.

ennru:
Thank you! A great improvement.

PR #2029, opened 2019-11-24
AWS Kinesis: Use AWS SDK v2

### LINKED ISSUE #372: Update to AWS SDK 2.0

### LINKED ISSUE #1347: Kinesis: Use AWS SDK 2.0

regiskuckaertz:
## Purpose

Updates the Kinesis library to use v2 of the AWS SDK.

## References

This PR resolves #1347 and resolves #372 as this is the last AWS-related project to depend on v1.


regiskuckaertz:
Not sure how to get around the MIMA check 🤔

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSource.scala
```diff
@@ -9,24 +9,24 @@ import akka.stream.alpakka.kinesis.KinesisErrors.NoShardsError
...
-    Source.fromGraph(new KinesisSourceStage(shardSettings, amazonKinesisAsync))
+  def basic(shardSettings: ShardSettings, KinesisAsyncClient: KinesisAsyncClient): Source[Record, NotUsed] =
+    Source.fromGraph(new KinesisSourceStage(shardSettings, KinesisAsyncClient))
 
   /**
    * Read from multiple shards into a single stream.
    */
   def basicMerge(shardSettings: List[ShardSettings],
-                 amazonKinesisAsync: AmazonKinesisAsync): Source[Record, NotUsed] = {
+                 KinesisAsyncClient: KinesisAsyncClient): Source[Record, NotUsed] = {
```
ennru:
```suggestion
                 kinesisAsyncClient: KinesisAsyncClient): Source[Record, NotUsed] = {
```

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSource.scala
```diff
@@ -9,24 +9,24 @@ import akka.stream.alpakka.kinesis.KinesisErrors.NoShardsError
...
+import software.amazon.awssdk.services.kinesis.model.Record
 
 object KinesisSource {
 
   /**
    * Read from one shard into a stream.
    */
-  def basic(shardSettings: ShardSettings, amazonKinesisAsync: AmazonKinesisAsync): Source[Record, NotUsed] =
-    Source.fromGraph(new KinesisSourceStage(shardSettings, amazonKinesisAsync))
+  def basic(shardSettings: ShardSettings, KinesisAsyncClient: KinesisAsyncClient): Source[Record, NotUsed] =
```
ennru:
```suggestion
  def basic(shardSettings: ShardSettings, kinesisAsyncClient: KinesisAsyncClient): Source[Record, NotUsed] =
```

kinesis/src/test/scala/akka/stream/alpakka/kinesis/KinesisSourceSpec.scala
```diff
@@ -151,70 +152,59 @@ class KinesisSourceSpec extends WordSpecLike with Matchers with KinesisMock {
...
-    when(describeStreamResult.getStreamDescription).thenReturn(streamDescription)
-    when(streamDescription.getShards).thenReturn(shards)
-    when(streamDescription.getHasMoreShards).thenReturn(false)
+    when(amazonKinesisAsync.describeStream(any[DescribeStreamRequest]))
+      .thenReturn(CompletableFuture.completedFuture(describeStreamResult))
 
-    val getShardIteratorRequest = new GetShardIteratorRequest
-    val getShardIteratorResult = new GetShardIteratorResult().withShardIterator(shardIterator)
-    val getRecordsRequest = new GetRecordsRequest
+    val getShardIteratorRequest = GetShardIteratorRequest.builder().build()
```
ennru: This looks unused.

kinesis/src/test/scala/akka/stream/alpakka/kinesis/KinesisSourceSpec.scala
```diff
@@ -151,70 +152,59 @@ class KinesisSourceSpec extends WordSpecLike with Matchers with KinesisMock {
...
+    val getShardIteratorResult = GetShardIteratorResponse.builder().shardIterator(shardIterator).build()
+    val getRecordsRequest = GetRecordsRequest.builder().build()
```
ennru: So does this.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSource.scala
regiskuckaertz: 🤦‍♂ thanks!

regiskuckaertz:
I'm happy to remove the retry logic ✌️

regiskuckaertz:
I've done so for `KinesisFlowStage`. If this looks correct to you, I'll apply the same to the firehose stage.

regiskuckaertz:
I think I can replace `TimerGraphStageLogic` with a simple `GraphStageLogic`.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisFlowStage.scala
```diff
@@ -74,41 +63,32 @@ private[kinesis] final class KinesisFlowStage[T](
...
+        val request = PutRecordsRequest
+          .builder()
+          .streamName(streamName)
+          .records(job.records.map(_._1).asJavaCollection)
+          .build()
+
+        def handleResult(result: PutRecordsResponse): immutable.Seq[(PutRecordsResultEntry, T)] = {
+          val (correlatedRequestResultSuccess, correlatedRequestResultFailure) =
+            result.records.asScala.zip(job.records).toList.partition { case (res, _) => res.errorCode eq null }
+          if (result.failedRecordCount > 0) {
```
ennru: Next question is if we should leave it to the user to decide to fail the stream on error. Just emitting the result even if it contains errors?
That's what Alpakka Elasticsearch does https://doc.akka.io/docs/alpakka/current/elasticsearch.html#passing-data-through-elasticsearchflow

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisFlowStage.scala
```diff
@@ -49,21 +44,15 @@ private[kinesis] final class KinesisFlowStage[T](
...
-    new TimerGraphStageLogic(shape) with StageLogging with InHandler with OutHandler {
+    new GraphStageLogic(shape) with StageLogging with InHandler with OutHandler {
 
       type Token = Int
-      type RetryCount = Int
-
       private var completionState: Option[Try[Unit]] = None
 
       private val pendingRequests: mutable.Queue[Job[T]] = mutable.Queue.empty
       private var inFlight: Int = 0
```
ennru: I believe these can go, as well.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisFlowStage.scala
regiskuckaertz: I think it would give more control to users and reduce complexity in the library, so it's a win-win imo. I'll give that a go and remove the partial error tests.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSourceStage.scala
```diff
@@ -126,38 +129,44 @@ private[kinesis] class KinesisSourceStage(shardSettings: ShardSettings, amazonKi
         case 'GET_RECORDS => requestRecords()
```
ennru: While at it, could you please replace the symbol with a `case object`.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisFlow.scala
```diff
@@ -7,29 +7,38 @@ package akka.stream.alpakka.kinesis.scaladsl
...
   ): Flow[PutRecordsRequestEntry, PutRecordsResultEntry, NotUsed] =
     Flow[PutRecordsRequestEntry]
       .map((_, ()))
       .via(withUserContext(streamName, settings))
       .map(_._1)
 
   def withUserContext[T](streamName: String, settings: KinesisFlowSettings = KinesisFlowSettings.Defaults)(
-      implicit kinesisClient: AmazonKinesisAsync
+      implicit kinesisClient: KinesisAsyncClient
   ): Flow[(PutRecordsRequestEntry, T), (PutRecordsResultEntry, T), NotUsed] =
```
ennru: Akka has been extended to have native support for this pattern now.
Please make it return `FlowWithContext[PutRecordsRequestEntry, T, PutRecordsResultEntry, T, NotUsed]`. The simplest solution is to use `FlowWithContext.fromTuples(...)`.
The methods should be renamed to `withContext`/`createWithContext` to align nicely.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisFlow.scala
```diff
@@ -39,41 +48,48 @@ object KinesisFlow {
...
+            .toScala
+            .transform(handlePutRecordsSuccess(entries), FailurePublishingRecords(_))(sameThreadExecutionContext)
       )
-      .mapAsync(settings.parallelism)(identity)
       .mapConcat(identity)
 
+  private def handlePutRecordsSuccess[T](
+      entries: Iterable[(PutRecordsRequestEntry, T)]
+  )(result: PutRecordsResponse): List[(PutRecordsResultEntry, T)] =
+    result.records.asScala.toList.zip(entries).map { case (res, (_, t)) => (res, t) }
```
ennru: Hm, this unrolls the `PutRecordsResponse` so the user can't actually react on a non-zero `failedRecordCount`. I guess the records with errors will be missing in the response, that's why the old implementation had some correlation logic.
Ideas?

regiskuckaertz: The `PutRecordsResultEntry` is enough to determine whether an error occurred at the record level (via `errorCode`), isn't it?

ennru: Ok, I didn't discover that API.
The examples should show something along the lines of
```
  .map { response =>
    if (response.errorCode != 0) throw new RuntimeException(...)
    response
  }
```
so that users understand it is their responsibility to check for response errors.

kinesis/src/main/scala/akka/stream/alpakka/kinesisfirehose/scaladsl/KinesisFirehoseFlow.scala
```diff
@@ -5,37 +5,44 @@
...
+                .build()
+            )
+            .toScala
+            .transform(identity, FailurePublishingRecords(_))(sameThreadExecutionContext)
       )
-      .mapAsync(settings.parallelism)(identity)
-      .mapConcat(_.getRequestResponses.asScala.toIndexedSeq)
-      .filter(_.getErrorCode == null)
+      .mapConcat(_.requestResponses.asScala.toIndexedSeq)
+      .filter(_.errorCode == null)
```
ennru: This is interesting, are elements with errors silently discarded?

regiskuckaertz: Ha, I failed to remove that line, I think now that it's all handled by the AWS client, it is expected users will either intentionally set up a retry strategy or simply be oblivious to it. In both cases, errors should not be discarded. Thanks for the catch 👍

regiskuckaertz:
Another question: the firehose classes are now in the `firehose` package in the SDK. Should the `kinesisfirehose` package be renamed accordingly?

ennru:
We can keep `kinesisfirehose`.

regiskuckaertz:
Looks like there were some flaky tests in other projects, otherwise this is ready for review.

ennru:
Failures were
#1306 
#1801

ennru:
Thank you, this was the last connector on the old AWS SDK! 💯

PR #2032, opened 2019-11-25
SQS: allow sub-types in the Java API; improve tests

ennru:
## Purpose

Change the type in the Java API of the batch operations to allow for sub-types of `Iterable` so it composes more nicely with eg. `grouped`.

## Changes

* Improved closing the SQS client in the tests
* Use Alpakka to publish to the topic in `SqsSourceTest` to avoid too many parallel HTTP connections when tests run on Travis

### DIFF
--- a/sqs/src/test/java/akka/stream/alpakka/sqs/javadsl/BaseSqsTest.java
+++ b/sqs/src/test/java/akka/stream/alpakka/sqs/javadsl/BaseSqsTest.java
@@ -29,6 +29,7 @@ public abstract class BaseSqsTest {
 
   protected static ActorSystem system;
   protected static Materializer materializer;
+  private static SqsAsyncClient sqsClientForClose;
 
   private boolean initialized = false;
   protected String sqsEndpoint = "http://localhost:9324";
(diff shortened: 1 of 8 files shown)


sqs/src/test/java/akka/stream/alpakka/sqs/javadsl/BaseSqsTest.java
```diff
@@ -44,13 +45,17 @@ public static void setup() {
 
   @AfterClass
   public static void teardown() {
+    if (sqsClientForClose != null) {
```
seglo: I don't understand the context of why this is necessary.  How come we can't just always call `sysClient.close`?  What is the purpose of `sqsClientForClose`?

ennru: This is a "static vs. instance" dance.

seglo: Ah, I didn't notice that it was static.

Issue #2017, opened 2019-11-18
Google Cloud Pub/Sub: remove `ActorSystem` and `Materializer` from the factory methods


ennru:
### Short description

The `GooglePubSub` flow factory methods could be changed to not require an `ActorSystem` and a `Materializer` by using the `Source.setup` operator internally.

### Details

As https://github.com/akka/alpakka/pull/1983 breaks some factory methods' signatures anyway, it would be great to get this in for Alpakka 2.0.

ennru:
Fixed with #2047

PR #2047, opened 2019-12-13
Google Cloud Pub/Sub: remove ActorSystem and Materializer

ennru:
## Purpose

Use `Flow.setup` to get access to the `ActorSystem` and `Materializer` on materialization so that those don't need to be passed as (implicit) parameters.

## References

#2017 

## Background Context

The APIs were changed by #1983 which introduced the need to access the materializer.

### DIFF
--- a/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/scaladsl/GooglePubSub.scala
+++ b/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/scaladsl/GooglePubSub.scala
@@ -23,12 +23,11 @@ protected[pubsub] trait GooglePubSub {
   private[pubsub] def httpApi: PubSubApi
 
   /**
-   * Creates a flow to that publish messages to a topic and emits the message ids
+   * Creates a flow to that publishes messages to a topic and emits the message ids.
    */
-  def publish(topic: String, config: PubSubConfig, parallelism: Int = 1)(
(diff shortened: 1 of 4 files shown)


google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/scaladsl/GooglePubSub.scala
```diff
@@ -23,65 +23,95 @@ protected[pubsub] trait GooglePubSub {
...
+  def publishWithContext[C](
+      topic: String,
+      config: PubSubConfig,
+      parallelism: Int = 1
+  ): FlowWithContext[PublishRequest, C, immutable.Seq[String], C, NotUsed] = {
+    // some wrapping back and forth as FlowWithContext doesn't offer `setup`
+    // https://github.com/akka/akka/issues/27883
+    FlowWithContext.fromTuples {
+      Flow
+        .setup { (mat, _) =>
```
seglo: `Flow.setup` is deprecated in Akka 2.6, but the recommended alternative doesn't exist in 2.5 (`Flow.fromMaterializer`).  Maybe we can add a comment to change it when 2.5 is dropped.

ennru: We'll revisit the Akka deprecations when we drop Akka 2.5 (in a year or so).

Issue #2039, opened 2019-11-29
Alpakka: JMS session isn't closed after ack timeout when stream is failed


matwojcik:
### Versions used 

Alpakka version: 1.1.2

### Behaviour

When consumer is defined to have ack timeout and fails the stream on ack timeout, then after timeout the underlying JMS session is not closed. It results in session being still open and brokers (like ActiveMQ) assign messages to this session, making other connection unable to read them.

### Reproducible Test Case

```scala
val jmsSource =
  JmsConsumer.txSource(
    JmsConsumerSettings(actorSystem.settings.config.getConfig(JmsConsumerSettings.configPath), connectionFactory)
      .withSessionCount(1)
      .withAckTimeout(1.second)
      .withQueue("test-queue")
      .withFailStreamOnAckTimeout(true)
  )

jmsSource
  .map { txEnvelope =>
    Thread.sleep(5000L)
    txEnvelope.commit()
    txEnvelope.message
  }
  .runWith(Sink.seq)
```

ennru:
Thank you for reporting.
Would you be in a position to suggest the necessary changes in a Pull Request?

matwojcik:
@ennru unfortunately I'm not sure where the bug origins from :/

ennru:
Fixed with #2048

kubukoz:
Thanks!

PR #2048, opened 2019-12-13
JMS: close sessions in postStop

ennru:
## Purpose

Close JMS sessions on stage stop.

## Changes

(See commits)
* Isolate the access to `sessions` within `JmsConnector`.
* Switch some tests to use a shared JMS broker with by using random queue names
* Close sessions in `finishStop` which is called from `postStop`

## Background Context

While exploring possible solutions to #2039 I noticed the JMS sessions are not properly closed in more scenarios.


ennru:
@andreas-schroeder Would be great if you can help me to come up with a good solution to #2039.

Issue #2025, opened 2019-11-20
ElasticSearch: Support for v7 ? 


peternitro:
### Short description

It looks like the alpakka ES project is compiled against ES client v6.3.1. ES 7.x has been out for a while and is supported by AWS.  

### Details

Using a 7.x client is not possible (binary interface changes results in a runtime exception) and using a 6.x client with a 7.x server results in the following warnings:

```
WARN  o.e.c.RestClient - request [POST http://localhost:9200/_bulk] returned 1 warnings: [299 Elasticsearch-7.1.1-7a013de "[types removal] Specifying types in bulk requests is deprecated."]
```

ennru:
Alpakka Elasticsearch uses the JSON interface to be able to talk to different versions of ES. As the ES JSON changes over time we'd need to introduce some version support as I suggested in https://github.com/akka/alpakka/issues/1232#issuecomment-534723623

ennru:
https://www.elastic.co/guide/en/elasticsearch/reference/7.x/removal-of-types.html

ennru:
We just merged #2052 so that we can implement different REST messages according to the ES version.
Would be great if ES users can help and implement support for ES 6 and ES 7.

ennru:
Implemented with #2277

Issue #1843, opened 2019-07-27
Allow elastic search _bulk operations to set the index name in the URL 


Tommmster:
### Short description

The elasticsearch Sink will fail sending messages when `rest.action.multi.allow_explicit_index` is set to `true` in the `elasticsearch.yml`. Moreover, this sink provides no functionality to set the index in the URL.

### Details
`_bulk` operations do not include the index name in the URL, but add the `_index` field to [each document](https://github.com/akka/alpakka/blob/master/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchFlowStage.scala#L143). However, the elasticsearch cluster may not allow this. (see https://www.elastic.co/guide/en/elasticsearch/reference/current/url-access-control.html). If this is the case, the cluster will reject the bulk operation.

The [`ElasticsearchFlowStage` constructor](https://github.com/akka/alpakka/blob/master/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchFlowStage.scala#L27) has all the information needed to handle this situation. By adding a field to the `ElasticsearchWriteSettings` to indicate whether explicit index names are allowed, the `sendBulkUpdateRequest` method can change it's behaviour accordingly (that is, using the /indexname/_bulk endpoint and *not* setting the `_index` field in the payload)

#### Proposal
* Add a `multiAllowExplicitIndex` field to the `ElasticsearchWriteSettings` case class, with a default value of `true` (keeps current behaviour)

* Create two `WriteMessage[T,C] => Seq[(String, JsString)]` functions. One will provide the `("_index" -> JsString(indexname))` tuple, the other one will not. Keep the appropiate one in a field.

* Modify this signature
```
private def sendBulkUpdateRequest(messages: immutable.Seq[WriteMessage[T,C]]): Unit
```
to this one
```
private def sendBulkUpdateRequest(messages: immutable.Seq[WriteMessage[T,C]], 
                                                            bulkSharedFields: WriteMessage[T,C] => Seq[(String, JsString)]): Unit
```

and provide the appropiate function 

* Add a field indicating the endpoint to pass to the `RestClient`. Possible values are `/_bulk` (current) and `/$indexName/_bulk` 

* In the `ElasticsearchFlowStage[T,C]` constructor, read this property and determine

2m:
Thanks for the detailed analysis. Approach sounds good. Reviewing the PR.

ennru:
Implemented with #2326

Issue #1232, opened 2018-09-24
Elasticsearch: Use of _version generates deprecation warnings for ES>=6


semafor:
_version and other related fields have been deprecated and will be removed in elasticsearch v7.

Migrating to a higher level client library would "permanently" fix this, but maybe there's rationale for keeping the low level client around.

Reproduce:
Index/Update docs in ES>=6

What do I expect:
No deprecation warnings/the ability to bump a higher level ES client to fix these errors

What happened instead:
_version is hard coded throughout the library, so there's no real way to upgrade to ES7 and keep using this library.

ennru:
Thank you for reporting this. Elasticsearch is quite fast-paced on their JSON API, we should add a warning to the docs.
I believe the choice for the client implementation using raw JSON was made to achieve non-blocking API calls, does the ES client library offer that nowadays?

takezoe:
> I believe the choice for the client implementation using raw JSON was made to achieve non-blocking API calls, does the ES client library offer that nowadays?

Official Elasticsearch client for REST API required raw JSON handling at that moment, but now a high-level client which can handle request and response as typed models is available. It may be possible to replace the current Alpakka implementation with the high-level client.

Fs02:
I'm interested in replacing the client library with high level client, as it also open the possibility to remove `spray-json` from dependency (which i need because i'm using circe).

I've experimented some change using @takezoe original repository [here](https://github.com/Fs02/akka-stream-elasticsearch), I can try to modify and update current alpakka implementation if it allowed

DJLemkes:
@Fs02 your work looks promising. Did you ever make a PR out of this?

I'm using Circe as well so I'd love to bring that into my flow instead of Spray. 

Besides, I'm also getting the `Specifying types in bulk requests is deprecated` warnings when using ES version > 7. This is due to the [`_type`](https://github.com/akka/alpakka/blob/44feb05e666dbfc2baa203822aff6ba28220b428/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchFlowStage.scala#L144) field being sent by the current implementation.

If I'm not mistaken, this would be fixed when the high-level REST client is used instead of the current one. Only problem could be that Elastic recommends to keep cluster version and high-level client version in [sync with each other](https://www.elastic.co/guide/en/elasticsearch/client/java-rest/7.2/java-rest-high-compatibility.html).

ennru:
Hi Alpakka Elasticsearch users,

[...]

Issue #945, opened 2018-05-09
ES connector throwing NoSuchElementException on Elasticsearch 2.4.6


brunoballekens:
I use the ElasicsearchSink to insert documents in an index in bulk.
It works perfectly on elastic search version `5.6.8`, but on `2.4.6` it fails with the following exception:
```
java.util.NoSuchElementException: key not found: index
	at scala.collection.immutable.Map$Map1.apply(Map.scala:111)
	at akka.stream.alpakka.elasticsearch.ElasticsearchFlowStage$$anon$1.$anonfun$handleResponse$1(ElasticsearchFlowStage.scala:153)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:234)
	at scala.collection.Iterator.foreach(Iterator.scala:944)
	at scala.collection.Iterator.foreach$(Iterator.scala:944)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1432)
	at scala.collection.IterableLike.foreach(IterableLike.scala:71)
	at scala.collection.IterableLike.foreach$(IterableLike.scala:70)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:54)
	at scala.collection.TraversableLike.map(TraversableLike.scala:234)
	at scala.collection.TraversableLike.map$(TraversableLike.scala:227)
	at scala.collection.AbstractTraversable.map(Traversable.scala:104)
	at akka.stream.alpakka.elasticsearch.ElasticsearchFlowStage$$anon$1.handleResponse(ElasticsearchFlowStage.scala:151)
	at akka.stream.alpakka.elasticsearch.ElasticsearchFlowStage$$anon$1.$anonfun$responseHandler$1(ElasticsearchFlowStage.scala:107)
	at akka.stream.alpakka.elasticsearch.ElasticsearchFlowStage$$anon$1.$anonfun$responseHandler$1$adapted(ElasticsearchFlowStage.scala:107)
	at akka.stream.impl.fusing.GraphInterpreter.runAsyncInput(GraphInterpreter.scala:454)
	at akka.stream.impl.fusing.GraphInterpreterShell$AsyncInput.execute(ActorGraphInterpreter.scala:468)
	at akka.stream.impl.fusing.GraphInterpreterShell.processEvent(ActorGraphInterpreter.scala:563)
[...]

PR #2052, opened 2019-12-18
Elasticsearch: support multiple REST versions

ennru:
## Purpose

Introduce machinery to support different versions of the Elasticsearch REST protocol.

## References

https://github.com/akka/alpakka/issues/1232
https://github.com/akka/alpakka/issues/2025
https://github.com/akka/alpakka/issues/1843
https://github.com/akka/alpakka/issues/945

## Changes

* Add `ApiVersion` to `WriteSettings`
* Make in the JSON handling a class instead of mixing it in

### DIFF
--- a/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchFlowStage.scala
+++ b/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchFlowStage.scala
@@ -36,14 +36,12 @@ private[elasticsearch] final class ElasticsearchFlowStage[T, C](
   private val out = Outlet[immutable.Seq[WriteResult[T, C]]]("result")
   override val shape = FlowShape(in, out)
 
+  private val restApi: RestApi[T, C] =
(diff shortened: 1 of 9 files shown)


elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchFlowStage.scala
```diff
@@ -92,7 +87,8 @@ private[elasticsearch] final class ElasticsearchFlowStage[T, C](
     private def handleResponse(args: (immutable.Seq[WriteMessage[T, C]], Response)): Unit = {
       val (messages, response) = args
       val jsonString = EntityUtils.toString(response.getEntity)
-      val messageResults = toWriteResults(messages, jsonString)
+      if (log.isDebugEnabled) log.debug("response {}", jsonString)
```
seglo: This log line no longer prettyprints the JSON.  i.e.

```scala
      import spray.json._
      if (log.isDebugEnabled) log.debug("response {}", jsonString.parseJson.prettyPrint)
```

PR #2053, opened 2019-12-19
AWS S3: disallow sub-dir selection with bucket names

ennru:
## Purpose

Validate the bucket name even when using path-style access so that using `..` doesn't give access to other S3 buckets levels.

## References

#1679

## Changes

* validate bucket names for path-style access
* shut down WireMock in tests
* avoid nesting `Source.setup` in `request`

## Background Context

When using path-style access, the S3 bucket name becomes a part of the URLs and opens up to select sub-directories on S3.
Path-style access is being discontinued by AWS from September 2020 see #1679


seglo:
Known failure: https://github.com/akka/alpakka/issues/1802

PR #2058, opened 2019-12-20
Add missing metadata about storage objects to google-cloud-storage api 

DanieleSassoli:
## Purpose
The intent of this PR is to be able to retrieve all the metadata about storage objects via the alpakka api.

## Background Context

I'm writing an interface that should enable me to treat Sftp, Ftps, S3 and GCS files in the same way, I noticed that I could get information around when the document was last modified for Sftp and Ftps(still haven't got round to S3) but not for GCS, so thought I'd add it in.

I recreated this PR as substitute of #2009 as that was created of my master and that made it more difficult for people to contribute to it. Hopefully opening it from a branch should enable more collaboration.


seglo:
@DanieleSassoli Everything looks good to me.  Thanks a lot for the contribution and your patience throughout the review process!

Issue #1566, opened 2019-03-07
Elasticsearch createWithPassThrough sometimes doesn't return WriteResult [1.0-M3]


danelkotev:
Hi,

I upgraded to 1.0-M3 version of Alpakka ES and it seems like there is a bug when using ElasticsearchFlow.createWithPassThrough - I'm not receiving WriteResult for all bulk documents.
Everything worked fine when I used 1.0-M1 version of Alpakka.
I know you have made some changes in ES, and I think it's related.

ennru:
Thank you for reporting, that is definitely possible.

ennru:
Do you have retrying enabled? When retrying at least order isn't kept.

danelkotev:
Hi, I do use retrying and the problem is not with the order. It seems like some messages are missing. I have a unit test that can be used to produce the issue.
Anyways, when I not using retry mechanism, it seems like everything works as expected. On 1.0-M1  using retry mechanism yield the expected result.

ennru:
Ok, so it seems related to retrying. Can you see (eg. in the log) if it actually retries in cases where you are missing `WriteResult`s?

ennru:
Please try creating a `RetryLogic` similar to `RetryAtFixedRate` and log and exceptions it receives in `shouldRetry`. It would be interesting to improve this back to the earlier behaviour.

danelkotev:
I created a unit test with `RetryAtFixedRate`. This test reaches timeout. I don't see lost of messages in logs. Please review my test below and share your opinion.

```
"index bulk of document with passThrough" in assertAllStagesStopped {
      val indexName = "sink10"
      val sink = ElasticsearchFlow.createWithPassThrough[Book, Book](
        indexName = indexName,
        typeName = "doc",
        ElasticsearchWriteSettings() //.withBufferSize(5)
          .withRetryLogic(RetryAtFixedRate(maxRetries = 5,
          retryInterval = 1.millis))
      )

      val bookMessages = List(
        Book("Book 1"),
        Book("Book 2"),
        Book("Book 3"),
        Book("Book 4"),
        Book("Book 5"),
        Book("Book 6"),
        Book("Book 3"),
        Book("Book 8"),
        Book("Book 9"),
        Book("Book 10"),
        Book("Book 3"),
        Book("Book 11"),
        Book("Book 12"),
        Book("Book 13"),
        Book("Book 14"),
        Book("Book 3"),
        Book("Book 1"),
        Book("Book 2"),
[...]

PR #2062, opened 2019-12-27
Elasticsearch - #1566 - ElasticsearchFlowStage retry logic improvement

nivox:
- failed messages are treated independently to avoid loosing track of previously failed ones
- retry counts are kept per message so to guarantee correct number of retries and intervals
- added test case proving fix correctness

Apart for the added test, this pull request has also been tested on a live system which kept failing with the 1.1.2 release.

# Pull Request Checklist

## Purpose

This pull request corrects ElasticsearchFlowStage retry logic implementation making sure that old failed messages are not lost due to new failures.

## References

References #1566

## Changes

Only the internal implementation of ElasticsearchFlowStage has been changed to keep track of retry counts on a per message basis.

### DIFF
--- a/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchFlowStage.scala
+++ b/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchFlowStage.scala
@@ -49,76 +49,82 @@ private[elasticsearch] final class ElasticsearchFlowStage[T, C](
     private var upstreamFinished = false
     private var inflight = 0
 
-    private val failureHandler = getAsyncCallback[(immutable.Seq[WriteMessage[T, C]], Throwable)](handleFailure)
-    private val responseHandler = getAsyncCallback[(immutable.Seq[WriteMessage[T, C]], Response)](handleResponse)
-    private var failedMessages: immutable.Seq[WriteMessage[T, C]] = Nil
-    private var retryCount: Int = 0
+    private val failureHandler =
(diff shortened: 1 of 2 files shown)


seglo:
@nivox Thanks for the PR.  I'm going to park this for @ennru to review when he returns (next week).  It seems that the Elasticsearch client may support retry and backoff strategies internally.  We might get this for free without adding our own implementaiton.  IIRC Enno mentioned this at some point, but I'm not sure if it's something we're considering for 2.0.0.

ennru:
I rebased it to get passed the GC Storage error and applied scalafmt.

nivox:
@ennru Thanks for taking the time to look at the PR! 
As you correctly observed this implementation could let to out of order messages. 
More generally this implementation (and as far as I understand the general idea, the implementation based on RetryFlow) could potentially result in out of order writes.

Thinking about it the only way to ensure both message and write order is to limit a single in-flight batch of writes (i.e. bulk operation) and handle all the retries before accepting a new batch. 
This will obviously reduce the throughput but will provide much more robust guarantees that will make the logic suitable for `withContext` scenarios.

I've a rough idea on the implementation which should not require a major rewrite of the current code. If you think this could be of interest I could work on it on another PR.

ennru:
Thank you for this improvement!

My hope is that we can compose the `RetryFlow` from Akka 2.6 with the `ElasticsearchSimpleFlowStage` to achieve retrying without out-of-order responses.
`RetryFlow` allows for only one in-flight message and that reduces throughput as you noticed.
Please give it a try!

Issue #1837, opened 2019-07-24
JmsProducer memory leak


konaleksey123:
Hello. 

I am using akka-stream-alpakka-jms-1.1.0 and have intensive sending message to some queue.
A simplified version of my application is similar to:
```
  Source
    .repeat("message")
    .mapAsyncUnordered(10) { msg =>
      for {
        _ <- Source
          .single(msg)
          .runWith(
            JmsProducer.textSink(
              JmsProducerSettings(jmsProducerSettings, ibmConnFactory)
                .withQueue("queue1")
            )
          )
        // Do some other business logic. For example, save message send time in database
      } yield ()
    }
    .runWith(Sink.ignore)
```
All jms connections and sessions are closed, but there is the memory leak. I noticed that some ibm objects are in actor's buffer and don't gc.

Please help me to resolve the issue.

![memory leak](https://user-images.githubusercontent.com/14910285/61815718-8e093f00-ae53-11e9-8498-556fb96f5972.png)

2m:
Hi,

in the code example you have provided you materialize the stream ten times for every incoming message. That should not result in a memory leak, of course, but nevertheless it is sub-optimal way of using Akka Stream API.

If you want to have the message sent 10 times, duplicate that message in a stream, attach a single Sink, and materialize the stream only once.

As to the memory leak, my hunch is that those resources would be eventually released, but as the stream is being started constantly, some amount of resources is always held up.

konaleksey123:
@2m I dont want to duplicate message, its simple example for show a leak in JmsProducer.

Actually in my app I have code:
```
    JmsConsumer
      .ackSource(
        JmsConsumerSettings(jmsConsumerSettings, ibmConnFactory)
          .withQueue("topic")
      )
      .mapAsyncUnordered(10) { envelope =>
        for {
          message <- Future(envelope.message.getBody(classOf[String]))
          _ <- Source.single(message).runWith(JmsProducer.textSink(
              JmsProducerSettings(jmsProducerSettings, ibmConnFactory)
                .withQueue("queue1")
            ))
          // Do some other business logic. For example, save message send time in database
        } yield envelope.acknowledge()
      }
      .runWith(Sink.ignore)
```
and config:
```
alpakka.jms {
  consumer {
    session-count = 10
    buffer-size = 0
  }
}
```
I need a given throughput, so I want to consume message parallel in 10 ssesions, send it to other queue, save something in databaase and then ack message. 

Maybe something is wrong in my code?

[...]

PR #2067, opened 2020-01-01
JMS: Fix memory leak in connection status

### LINKED ISSUE #2066: Jms Consumer / Producer may leak memory through connection status
### Versions used 

Akka version: Commit 1805f78ad44aeecffa1a031750811ab3d5c7836e

### Expected Behavior

When completing a stream featuring a JmsConsumer or JmsProducer stage without consuming connectivity status events, all memory should be reclaimable by the JVM GC.

### Actual Behavior

Due to pre-materialization of the connection status stream and buffering of status events [here](https://github.com/akka/alpakka/blob/1805f78ad44aeecffa1a031750811ab3d5c7836e/jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala#L59), the graph stage starts a stream that isn't torn down if no subscription to connection status events happen; even completing the queue in [postStop](https://github.com/akka/alpakka/blob/1805f78ad44aeecffa1a031750811ab3d5c7836e/jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala#L73) doesn't seem to help as I determined with a heap dump.

What makes this more critical is that the connection status might contain a reference to the Jms connection. It is therefore worthwhile to think about how this leak can be fixed.

andreas-schroeder:
Fixes #2066

This change aims to make sure that references to Jms connection are cleared in both main logic and connection status publishing. Also, it clears the reference to the graph stage logic via the exception listener when the connection gets closed.

This is achieved by:

- Evicting non-consumed connection status events after a specific timeout. This uses a broadcast hub and subscribes a single discarding consumer after a configurable timeout I preliminary called `connection-status-subscription-timeout` and gave it a default of 5 seconds. There would be other ways to fix the leak, but I think this is the simplest and safest way (i.e. safe against accidental misuse).
- Making sure that the reference of the graph stage logic to its Jms connection  is cleared on stop. The code in `finishStop` didn't necessarily catch all corner cases that might exist or be introduced in the future.
- Clearing the Jms connection exception listener before closing the connection. This removes the reference from the Jms client to the Akka consumer/producer stage that can prevent it from being GCed. This for example happens when using the in-JVM ActiveMQ broker. Its buffered messages hold a reference to the connection they were received from (even if this connection is closed), and through this transitively to the Akka producer stage. While doing this, this also cleans up the stopping code of SourceStageLogic at the cost of one additional logging statement on closing Jms connections normally.

### DIFF
--- a/jms/src/main/scala/akka/stream/alpakka/jms/JmsBrowseSettings.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/JmsBrowseSettings.scala
@@ -5,7 +5,6 @@
 package akka.stream.alpakka.jms
 
 import akka.actor.ActorSystem
-import akka.util.JavaDurationConverters._
 import com.typesafe.config.{Config, ConfigValueType}
 
 /**
@@ -17,10 +16,8 @@ final class JmsBrowseSettings private (
     val destination: Option[Destination],
     val credentials: Option[Credentials],
     val selector: Option[String],
-    val acknowledgeMode: AcknowledgeMode,
-    val connectionStatusSubscriptionTimeout: scala.concurrent.duration.FiniteDuration
-) extends akka.stream.alpakka.jms.JmsSettings {
-  override val sessionCount = 1
+    val acknowledgeMode: AcknowledgeMode
+) {
 
   /** Factory to use for creating JMS connections. */
   def withConnectionFactory(value: javax.jms.ConnectionFactory): JmsBrowseSettings = copy(connectionFactory = value)
@@ -48,27 +45,20 @@ final class JmsBrowseSettings private (
   /** Set an explicit acknowledge mode. (Consumers have specific defaults.) */
   def withAcknowledgeMode(value: AcknowledgeMode): JmsBrowseSettings = copy(acknowledgeMode = value)
 
-  /** Java API: Timeout for connection status subscriber */
-  def withConnectionStatusSubscriptionTimeout(value: java.time.Duration): JmsBrowseSettings =
-    copy(connectionStatusSubscriptionTimeout = value.asScala)
-
   private def copy(
       connectionFactory: javax.jms.ConnectionFactory = connectionFactory,
       connectionRetrySettings: ConnectionRetrySettings = connectionRetrySettings,
       destination: Option[Destination] = destination,
       credentials: Option[Credentials] = credentials,
       selector: Option[String] = selector,
-      acknowledgeMode: AcknowledgeMode = acknowledgeMode,
-      connectionStatusSubscriptionTimeout: scala.concurrent.duration.FiniteDuration =
-        connectionStatusSubscriptionTimeout
+      acknowledgeMode: AcknowledgeMode = acknowledgeMode
   ): JmsBrowseSettings = new JmsBrowseSettings(
     connectionFactory = connectionFactory,
     connectionRetrySettings = connectionRetrySettings,
     destination = destination,
     credentials = credentials,
     selector = selector,
-    acknowledgeMode = acknowledgeMode,
-    connectionStatusSubscriptionTimeout = connectionStatusSubscriptionTimeout
+    acknowledgeMode = acknowledgeMode
   )
 
   override def toString =
@@ -78,8 +68,7 @@ final class JmsBrowseSettings private (
     s"destination=$destination," +
     s"credentials=$credentials," +
     s"selector=$selector," +
-    s"acknowledgeMode=${AcknowledgeMode.asString(acknowledgeMode)}," +
-    s"connectionStatusSubscriptionTimeout=${connectionStatusSubscriptionTimeout.toCoarsest}" +
+    s"acknowledgeMode=${AcknowledgeMode.asString(acknowledgeMode)}" +
     ")"
 }
 
@@ -106,15 +95,13 @@ object JmsBrowseSettings {
     val credentials = getOption("credentials", c => Credentials(c.getConfig("credentials")))
     val selector = getStringOption("selector")
     val acknowledgeMode = AcknowledgeMode.from(c.getString("acknowledge-mode"))
-    val connectionStatusSubscriptionTimeout = c.getDuration("connection-status-subscription-timeout").asScala
     new JmsBrowseSettings(
       connectionFactory,
       connectionRetrySettings,
       destination,
       credentials,
       selector,
-      acknowledgeMode,
-      connectionStatusSubscriptionTimeout
+      acknowledgeMode
     )
   }
 
--- a/jms/src/main/scala/akka/stream/alpakka/jms/JmsBrowseSettings.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/JmsBrowseSettings.scala
@@ -5,6 +5,7 @@
 package akka.stream.alpakka.jms
 
 import akka.actor.ActorSystem
+import akka.util.JavaDurationConverters._
 import com.typesafe.config.{Config, ConfigValueType}
 
 /**
@@ -16,7 +17,8 @@ final class JmsBrowseSettings private (
     val destination: Option[Destination],
     val credentials: Option[Credentials],
     val selector: Option[String],
-    val acknowledgeMode: AcknowledgeMode
+    val acknowledgeMode: AcknowledgeMode,
+    val connectionStatusSubscriptionTimeout: scala.concurrent.duration.FiniteDuration
 ) extends akka.stream.alpakka.jms.JmsSettings {
   override val sessionCount = 1
 
@@ -46,20 +48,27 @@ final class JmsBrowseSettings private (
   /** Set an explicit acknowledge mode. (Consumers have specific defaults.) */
   def withAcknowledgeMode(value: AcknowledgeMode): JmsBrowseSettings = copy(acknowledgeMode = value)
 
+  /** Java API: Timeout for connection status subscriber */
+  def withConnectionStatusSubscriptionTimeout(value: java.time.Duration): JmsBrowseSettings =
+    copy(connectionStatusSubscriptionTimeout = value.asScala)
+
   private def copy(
       connectionFactory: javax.jms.ConnectionFactory = connectionFactory,
       connectionRetrySettings: ConnectionRetrySettings = connectionRetrySettings,
       destination: Option[Destination] = destination,
       credentials: Option[Credentials] = credentials,
       selector: Option[String] = selector,
-      acknowledgeMode: AcknowledgeMode = acknowledgeMode
+      acknowledgeMode: AcknowledgeMode = acknowledgeMode,
+      connectionStatusSubscriptionTimeout: scala.concurrent.duration.FiniteDuration =
+        connectionStatusSubscriptionTimeout
   ): JmsBrowseSettings = new JmsBrowseSettings(
     connectionFactory = connectionFactory,
     connectionRetrySettings = connectionRetrySettings,
     destination = destination,
     credentials = credentials,
     selector = selector,
-    acknowledgeMode = acknowledgeMode
+    acknowledgeMode = acknowledgeMode,
+    connectionStatusSubscriptionTimeout = connectionStatusSubscriptionTimeout
   )
 
   override def toString =
@@ -69,7 +78,8 @@ final class JmsBrowseSettings private (
     s"destination=$destination," +
     s"credentials=$credentials," +
     s"selector=$selector," +
-    s"acknowledgeMode=${AcknowledgeMode.asString(acknowledgeMode)}" +
+    s"acknowledgeMode=${AcknowledgeMode.asString(acknowledgeMode)}," +
+    s"connectionStatusSubscriptionTimeout=${connectionStatusSubscriptionTimeout.toCoarsest}" +
     ")"
 }
 
@@ -96,13 +106,15 @@ object JmsBrowseSettings {
     val credentials = getOption("credentials", c => Credentials(c.getConfig("credentials")))
     val selector = getStringOption("selector")
     val acknowledgeMode = AcknowledgeMode.from(c.getString("acknowledge-mode"))
+    val connectionStatusSubscriptionTimeout = c.getDuration("connection-status-subscription-timeout").asScala
     new JmsBrowseSettings(
       connectionFactory,
       connectionRetrySettings,
       destination,
       credentials,
       selector,
-      acknowledgeMode
+      acknowledgeMode,
+      connectionStatusSubscriptionTimeout
     )
   }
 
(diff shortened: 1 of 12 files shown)


WellingR:
In order to see if this resolves the issue I mentioned in  #1837 
I ran the [this code](https://gist.github.com/WellingR/35023510c042622b5186723ad16f73aa) again on commit https://github.com/andreas-schroeder/alpakka/commit/77c187a170198fbb8de4f18aaf6daa7886cf6eda

The memory usage looks a bit better but some ActiveMqConnections remain in the heap.
In addition I see a huge load of stacktraces in the logging (see below). These seem to start appearing after about 5 seconds, so this seems to be related to the newly added timeout...

```
2020-01-02 13:04:57 ERROR akka.actor.ActorSystemImpl [akka.actor.ActorSystemImpl(default)] - exception on LARS’ timer thread
java.lang.NullPointerException: null
	at akka.actor.LightArrayRevolverScheduler$TaskHolder.executeTask(LightArrayRevolverScheduler.scala:338)
	at akka.actor.LightArrayRevolverScheduler$$anon$3.executeBucket$1(LightArrayRevolverScheduler.scala:285)
	at akka.actor.LightArrayRevolverScheduler$$anon$3.nextTick(LightArrayRevolverScheduler.scala:289)
	at akka.actor.LightArrayRevolverScheduler$$anon$3.run(LightArrayRevolverScheduler.scala:241)
	at java.base/java.lang.Thread.run(Thread.java:834)
2020-01-02 13:04:57 INFO  akka.actor.ActorSystemImpl [akka.actor.ActorSystemImpl(default)] - starting new LARS thread
2020-01-02 13:04:57 ERROR akka.actor.ActorSystemImpl [akka.actor.ActorSystemImpl(default)] - Uncaught error from thread [default-scheduler-256]
java.lang.NullPointerException: null
	at akka.actor.LightArrayRevolverScheduler$TaskHolder.executeTask(LightArrayRevolverScheduler.scala:338)
	at akka.actor.LightArrayRevolverScheduler$$anon$3.executeBucket$1(LightArrayRevolverScheduler.scala:285)
	at akka.actor.LightArrayRevolverScheduler$$anon$3.nextTick(LightArrayRevolverScheduler.scala:289)
	at akka.actor.LightArrayRevolverScheduler$$anon$3.run(LightArrayRevolverScheduler.scala:241)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
```diff
@@ -56,26 +57,40 @@ private[jms] trait JmsConnector[S <: JmsSession] {
...
+      Source
+        .queue[InternalConnectionState](2, OverflowStrategy.dropHead)
+        .toMat(BroadcastHub.sink(1))(Keep.both)
+        .run()(this.materializer)
+    connectionStateQueue = queue
+    connectionStateSourcePromise.complete(Success(source))
+
+    // add subscription to purge queued connection status events after the configured timeout.
+    val system: ActorSystem = ActorMaterializerHelper.downcast(materializer).system
+    after(jmsSettings.connectionStatusSubscriptionTimeout, system.scheduler) {
```
WellingR: The NullPointerException I menioned before seems to be caused by
```scala
  implicit protected var ec: ExecutionContext = _
````
This executionContext is null at this point and is implicitly passed here...

**Comment by lightbend-cla-validator** [comment]

File: jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala (hunk 1 above)
Comment (andreas-schroeder) [reply jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala]: Thanks, fixed now with https://github.com/akka/alpakka/pull/2067/commits/74342089e1b772df16d227463da99e40105c3270

File: jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala [hunk 2]
```diff
@@ -56,26 +57,40 @@ private[jms] trait JmsConnector[S <: JmsSession] {
   override def preStart(): Unit = {
     // keep two elements since the time between initializing and connected can be very short.
     // always drops the old state, and keeps the most current (two) state(s) in the queue.
-    val pair = Source.queue[InternalConnectionState](2, OverflowStrategy.dropHead).preMaterialize()(this.materializer)
-    connectionStateQueue = pair._1
-    connectionStateSourcePromise.complete(Success(pair._2))
+    val (queue, source) =
+      Source
+        .queue[InternalConnectionState](2, OverflowStrategy.dropHead)
+        .toMat(BroadcastHub.sink(1))(Keep.both)
```
seglo: Is it possible that a buffer of 1 could result in a strange sequence of connection states downstream?  The first subscriber should be alright, but for subsequent subscriptions it's possible they could get an old stale connection state from the `BroadcastHub.sink` buffer first before the latest messages from the queue.  

Also, if the queue overflows it's possible you could still see gaps in downstream consumers, but I guess that's how it worked before as well.

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
```diff
@@ -56,26 +57,40 @@ private[jms] trait JmsConnector[S <: JmsSession] {
   override def preStart(): Unit = {
     // keep two elements since the time between initializing and connected can be very short.
     // always drops the old state, and keeps the most current (two) state(s) in the queue.
-    val pair = Source.queue[InternalConnectionState](2, OverflowStrategy.dropHead).preMaterialize()(this.materializer)
-    connectionStateQueue = pair._1
-    connectionStateSourcePromise.complete(Success(pair._2))
+    val (queue, source) =
+      Source
+        .queue[InternalConnectionState](2, OverflowStrategy.dropHead)
```
seglo: IIUC, the behaviour of the queue (`connectionStateQueue`) is the same as before because it has its own overflow strategy.

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
```diff
@@ -232,6 +247,20 @@ private[jms] trait JmsConnector[S <: JmsSession] {
...
+    Try(connection.setExceptionListener(null))
+    try {
+      connection.close()
+      log.info("JMS connection {} closed", connection)
+    } catch {
+      case NonFatal(e) => log.error(e, "Error closing JMS connection {}", connection)
+    }
+  }
+
+  protected def closeConnectionAsync(eventualConnection: Future[jms.Connection]): Future[Done] =
```
seglo: I wonder what the consequences of closing connections asynchronously without blocking could be.  I realize that that was what was done before, but ideally, we would chain any subsequent operations after asynchronously closing the connection object with `AsyncCallbacks`.  Maybe it's not very important.

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
```diff
@@ -56,26 +57,40 @@ private[jms] trait JmsConnector[S <: JmsSession] {
...
+      case current =>
+        JmsConnectorStopped(
+          Failure(new IllegalStateException(s"BUG: completing stage stop in unexpected state ${current.getClass}"))
+        )
     }
 
     closeSessions()
     val previous = updateStateWith(update)
-    connection(previous).foreach(_.close())
+    closeConnectionAsync(connection(previous))
```
seglo: I wonder what the consequences of closing connections asynchronously without blocking could be.  I realize that that was what was done before, but ideally, we would chain any subsequent operations after asynchronously closing the connection object with `AsyncCallbacks`.  Maybe it's not very important.

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
```diff
@@ -56,26 +57,40 @@ private[jms] trait JmsConnector[S <: JmsSession] {
...
   }
 
   protected def finishStop(): Unit = {
     val update: InternalConnectionState => InternalConnectionState = {
       case JmsConnectorStopping(completion) => JmsConnectorStopped(completion)
-      case current => current
+      case stopped: JmsConnectorStopped => stopped
+      case current =>
+        JmsConnectorStopped(
+          Failure(new IllegalStateException(s"BUG: completing stage stop in unexpected state ${current.getClass}"))
```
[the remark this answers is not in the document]
andreas-schroeder: Sure thing - I wanted to stay in line with this https://github.com/akka/alpakka/blob/63346ea71a8e79669ecab5416f1fe80adadcce2a/jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala#L342 but since it potentially gets shown to clients, let's remove it

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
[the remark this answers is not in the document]
andreas-schroeder: Sure, will do

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
andreas-schroeder: Nice catch; When I designed this originally, I did this on purpose so that a state consumer would still get all the intermediate states if they subscribed "a bit late", i.e. after the connection has already been established. I also thought that the connection status stream wouldn't be one with high volume of messages, and consumers would probably easily keep up with the update rate.

The only easy way I saw feasible for draining the connection state was to use a BroadcastHub. This now means that in the case we have multiple status update consumers (which wasn't possible before), with one of those consumers being very, very slow (i.e. slower than the connection status change rate), there could be gaps in the status update sequence. 

We could increase the buffer sizes to mitigate this to a certain point, or even make the buffer size configurable. What do you think? Which option would you favor?

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
andreas-schroeder: Let me think on it for a bit, I will come back to you on this.

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
seglo: I think keeping the buffer size of the broadcast hub to 1 mitigates the problem the best (there will only ever be possibly 1 stale element).  I wrote some code to demonstrate:

```
      val (queue, source) =
        Source
          .queue[Int](2, OverflowStrategy.dropHead)
          .toMat(BroadcastHub.sink(1))(Keep.both)
          .run()(this.materializer)
      val offers = (1 to 10).map(queue.offer)
      Await.result(Future.sequence(offers), scala.concurrent.duration.Duration.Inf)
      val subscriber1 = source.map(i => s"subscriber1: $i").runForeach(println)
// prints
subscriber1: 1
subscriber1: 9
subscriber1: 10
```

I can't think of a simple way to give the broadcast hub the same semantics as the queue (only buffer the last 2 elements) using the akka streams operator DSL.

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
andreas-schroeder: I had a look at the code and I think this is mainly relevant in [maybeReconnect](https://github.com/andreas-schroeder/alpakka/blob/jms-fix-connection-status-memory-leak/jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala#L211). In other cases, the closing of the connection is within the stage teardown on successful completion or failure. In `maybeReconnect`, we could `scheduleOnce` in a `Future.andThen`. What do you think?

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
andreas-schroeder: Yes, fully agree. And to continue your example, we could mitigate the gaps if the queue had a buffer size of 9 elements. So increasing the queue connection buffer size would mitigate that effect to a certain extent.

seglo: Increasing the buffer will only serve to distribute more stale records for late subscribers.  For example, if we increased the buffer to 10, and 100 connection status events occur before a new subscriber begins consuming elements, then they will receive 10 of the earliest records, followed by whatever comes out of the queue `1,2,3,4,5,6,7,8,9,10,99,100`.

I don't know if it's a big deal or not to get stale events.  I assume it's not since the overflow strategy of the queue was `DropHead` and we could experience lossiness anyway.

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
seglo: Sounds good to me, but if you would like to work on it then it should be addressed in a new PR since your changes didn't introduce the issue in the first place.

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
andreas-schroeder: It is correct that there will always be a point where the stream will become lossy; if we buffer `n` elements in the queue with `DropHead` (that drops the oldest element from the buffer) and have a `BroadcastHub.sink(1)`, we will have one very old status update (stored in the broadcast hub), and the `n` most recent status updates. So with `10` elements and buffer size of `5`, we'll get `1,6,7,8,9,10` (with `10` being the most recent update). But still, event when we increase the buffer, eventually the stream will be losing updates. And if processing updates is slow, buffering more updates might not be the right mitigation. Therefore I was thinking that making it configurable could be an option.

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
andreas-schroeder: Alright. Then, let's leave it for the next PR.

jms/src/main/scala/akka/stream/alpakka/jms/JmsBrowseSettings.scala
```diff
@@ -16,7 +17,8 @@ final class JmsBrowseSettings private (
     val destination: Option[Destination],
     val credentials: Option[Credentials],
     val selector: Option[String],
-    val acknowledgeMode: AcknowledgeMode
+    val acknowledgeMode: AcknowledgeMode,
+    val connectionStatusSubscriptionTimeout: scala.concurrent.duration.FiniteDuration
```
ennru: `JmsBrowseStage` doesn't actually use `JmsSettings`, just hardcode it to zero here.
(We could drop the `extends` for Alpakka 2.0, actually.)

ennru:
@andreas-schroeder I'll add the changes I suggested now.

ennru:
Thank you @andreas-schroeder for improving this.

Issue #1176, opened 2018-08-29
S3 content length limit (8388608 bytes) Error with 0.20 version


pramodlahoti26:
we are getting Error as exceeded content length limit (8388608 bytes)! You can configure this by setting `akka.http.[server|client].parsing.max-content-length` 
or calling `HttpEntity.withSizeLimit` before materializing the dataBytes stream 

As per latest commits it shows this issue has been addressed with 0.20 and we are using latest version but still getting same issue ,

Is there a way we can explicitly set this property externally or any other workaround to set this property?

Code we are using to Download  - 
```s3Client.download(bucketName.get, content.key)._1.
via(Framing.delimiter(ByteString("\n"), Constant.BYTE_SIZE, byeFormatterFlag).map(_.utf8String))
```

Learner0426:
StackOverflow Details about Issue - https://stackoverflow.com/questions/52111284/alpakka-akka-http-failure-while-downloading-file-with-latest-s3-connector

gabfssilva:
You can just configure it in your application.conf inside src/main/resources:

```
akka {
  http {
    client {
      parsing {
        max-content-length = //max length limit in bytes you want to set
      }
    }
  }
}
```

A good observation is that this property does not mean the max size of data that Akka HTTP keeps in memory.

2m:
Did you try increasing the `akka.http.client.parsing.max-content-length` parameter?

ennru:
Fixed with #1036

Issue #303, opened 2017-05-15
S3 Download of file larger then 8388608 bytes throws EntityStreamSizeException


thepont:
Using the Java DSL attempting to download a file larger then 8388608 bytes returns the following exception.

I have set my `akka.http.[server|client].parsing.max-content-length` setting to a higher setting, although I am wondering if there is a way to use the s3client without a size limit, such as the `withoutSizeLimit` directive.

```
2017-05-15 12:15:34 <> ERROR ProxyRequestInvocationHandler:43 - EntityStreamSizeException: actual entity size (Some(8466195)) exceeded content length limit (8388608 bytes)! You can configure this by setting `akka.http.[server|client].parsing.max-content-length` or calling `HttpEntity.withSizeLimit` before materializing the dataBytes stream.
[ERROR] [05/15/2017 12:15:34.637] [default-akka.actor.default-dispatcher-5] [akka://default/user/StreamSupervisor-1/flow-802-0-unknown-operation] Error during preStart in [akka.http.scaladsl.model.HttpEntity$Limitable@3c2b4414]
EntityStreamSizeException: actual entity size (Some(8466195)) exceeded content length limit (8388608 bytes)! You can configure this by setting `akka.http.[server|client].parsing.max-content-length` or calling `HttpEntity.withSizeLimit` before materializing the dataBytes stream.
	at akka.http.scaladsl.model.HttpEntity$Limitable$$anon$1.preStart(HttpEntity.scala:607)
	at akka.stream.impl.fusing.GraphInterpreter.init(GraphInterpreter.scala:520)
	at akka.stream.impl.fusing.GraphInterpreterShell.init(ActorGraphInterpreter.scala:380)
	at akka.stream.impl.fusing.ActorGraphInterpreter.tryInit(ActorGraphInterpreter.scala:538)
	at akka.stream.impl.fusing.ActorGraphInterpreter.finishShellRegistration(ActorGraphInterpreter.scala:580)
	at akka.stream.impl.fusing.ActorGraphInterpreter.akka$stream$impl$fusing$ActorGraphInterpreter$$shortCircuitBatch(ActorGraphInterpreter.scala:595)
	at akka.stream.impl.fusing.ActorGraphInterpreter$$anonfun$receive$1.applyOrElse(ActorGraphInterpreter.scala:619)
	at akka.actor.Actor$class.aroundReceive(Actor.scala:497)
[...]

PR #2069, opened 2020-01-03
GCS: Remove Unexpected Download Limit

DanieleSassoli:
## Purpose

Allowing file download from GCS without having to override `akka.http.client.parsing.max-content-length`

## References

Relevant issue #303 
Relative PR #1036
Possible workout explained [here](https://github.com/akka/alpakka/issues/1176#issuecomment-422230025)

## Background Context

While investigating an issue with file downloads in GCS I came across #303, I thought the fix would be very similar for GCS as well.

### DIFF
--- a/google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/impl/GCStorageStream.scala
+++ b/google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/impl/GCStorageStream.scala
@@ -268,7 +268,7 @@ import scala.util.control.NonFatal
             Uri(settings.baseUrl)
               .withPath(Path(settings.basePath) ++ sourcePath / "rewriteTo" ++ destinationPath)
               .withQuery(Query(queryParams))
-        )
+        ).map(response => response.withEntity(response.entity.withoutSizeLimit))
       ).runWith(Sink.head)
         .flatMap(entityForSuccess)
         .flatMap(responseEntityTo[RewriteResponse])


seglo:
Hi @DanieleSassoli .  Thanks for the PR.  Is there a way we can control how large response chunks are that get returned by the GCS API?  A [cursory search suggests this can be controlled](https://googleapis.dev/python/storage/latest/blobs.html).  Perhaps it's possible to define this to something less than the Akka Http response size default.

DanieleSassoli:
hey @seglo, sure I can look into that but is there any reason why we would want a different behaviour than the S3 connector? 
As stated in #1036 description 

> Because S3 connector is using akka-http, it inherits the default 8Mb HttpRequest limit. There is no reason why an S3 download should be limited to just 8Mb, and throw an unexpected exception on large files.

Don't see how that's different for GCS? If anything we would want to set it to a higher value than the Akka http default?

seglo:
> is there any reason why we would want a different behaviour than the S3 connector?

Not specifically.  Thanks for finding that PR, I wasn't aware of it.  I agree that it doesn't make sense to impose the default (or any pre-defined setting) for max download size for a blob storage service like S3 or GCS.  There are reasons to have smaller chunk sizes in some cases (i.e. individual CRC checks).  AFAIK GCS (and S3 multi-part download, I think) all support chunking of downloads (vs downloading the whole file in one request/response) so if there was a way to control chunk size that could be useful.

I think the change in this PR serves as a reasonable default (by removing the default restriction).  I'll take a closer look on Monday.

seglo:
Thanks for the fix!

gabrielreid:
I just ran into the same problem, but noticed that the committed fix only handles calls to `GCStorage.rewrite`, but not `GCStorage.download`. This means that trying to download a file that is more than 8MB will still fail (i.e. the subject of this PR is still not resolved).

Can we re-open this ticket, or should I create a new ticket?

seglo:
@gabrielreid I see, good find.  Please go ahead and create a new issue.  If you could submit your own PR with a fix that would be greatly appreciated!

PR #2086, opened 2020-01-14
AWS S3: bucket name and object key validation

ennru:
## Purpose

Extend the validation of bucket names (introduced in #2053) and object keys so that they can't be used to refer to sub-directories by using `..` selections. 

## References

https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html
https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html

## Changes

* Move bucket name validation to `BucketAndKey` (if make it available for users to pre-validate)
* Validate object keys
* Introduce configuration to switch off object key validation (for users that consider sub-directory selection a feature)
* Use valid bucket names in tests

## Background Context

Following up on #2053 it showed that even the object keys have potential to be used to access buckets that were not intended to be accessible by using `..` selections.


s3/src/main/resources/reference.conf
```diff
@@ -85,4 +85,9 @@ alpakka.s3 {
   # Which version of the list bucket api to use. Set to 1 to use the old style version 1 API.
   # By default the newer version 2 api is used.
   list-bucket-api-version = 2
+
+  # Object keys are validated to NOT use sub-directory selection with `..` to improve security.
+  # This flag may disable the validation.
+  # See https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html
+  validate-object-key = true
```
seglo: Would it make sense to let the user toggle validating the bucket name too, for the same reason?

seglo: Maybe the config could be generalized to apply validation to both the key and bucket name.

ennru: That "feature" will not be available for DNS-based bucket names, so I think it is best to forbid that right away.

PR #2087, opened 2020-01-14
MQTT streaming: Using public PropsAdapter API

jcroig:
## References

References #2088


mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSession.scala
```diff
@@ -8,6 +8,7 @@ package scaladsl
 import java.util.concurrent.atomic.AtomicLong
 
 import akka.actor.ExtendedActorSystem
+import akka.actor.typed.Props
 import akka.actor.typed.internal.adapter.{ActorRefAdapter, PropsAdapter}
```
ennru: Instead, Alpakka should use the public API version of `PropsAdapter` and the `ActorRefAdapter` shouldn't be used, either.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSession.scala
```diff
@@ -496,7 +499,9 @@ final class ActorMqttServerSession(settings: MqttSessionSettings)(implicit mat:
                               producerPacketRouter,
                               publisherPacketRouter,
                               unpublisherPacketRouter,
-                              settings)
+                              settings),
+            Props.empty,
+            true
```
huntc: Named param?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSession.scala
jcroig: Done.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSession.scala
jcroig: After last commit these lines changed so I don't thing this applies anymore, thanks.

ennru:
We should have detected this earlier, our scheduled build is still on Akka 2.6.0...
Thank you for fixing it.

jcroig:
Pleasure.

Issue #2031, opened 2019-11-25
Elasticsearch: use `RetryFlow` as retry implementation


ennru:
### Short description

Use the new Akka `RetryFlow` (by copying it into Alpakka for now to be Akka 2.5 compatible) to offer ES write retrying with backoff.

### Details

The `ElasticsearchSimpleFlowStage` introduced in https://github.com/akka/alpakka/pull/1941 does focus on publishing to ES without built-in retrying logic.
It can be composed with the [`RetryFlow`](https://doc.akka.io/docs/akka/current/stream/operators/RetryFlow/withBackoff.html) to offer retrying which would even work in a `withContext` variant.

ennru:
Elasticsearch write retries are now implemented via Akka's `RetryFlow`. The more involved Elasticsearch stage with built-in retrying has been removed in.
Implemented with #2093.

PR #2093, opened 2020-01-15
Elasticsearch: Flow implementation using RetryFlow

nivox:
## Purpose

The purpose of this PR is to implement a retry logic for Elasticsearch flow which guarantees message ordering making it compatible with `ElasticsearchFlow.createWithContext`.

## References

References #2031

## Changes
* Backport of `RetryFlowCoordinator` from Akka 2.6
* Changed `ElasticsearchSimpleFlowStage` input/output ports types
* Changed `ElasticsearchFlow.simpleStageFlow` to wrap flow with `RetryFlowCoordinator`
* Added test mirroring the current `ElasticsearchFlowStage` retry testing logic

## Background Context
This PR adds support for retry logic to `ElasticsearchSimpleFlowStage` using a backported implementation of `RetryFlowCoordinator` from Akka 2.6.

The idea is to use `RetryFlowCoordinator` to check for failures in the bulk's `WriteResult`.
In case any failure is found, the failed messages are resubmitted to the `ElasticsearchSimpleFlowStage` with the successful results as passthrough.

Furthermore the single `WriteMessage` passtrhough value is augmented with the index of the message inside the bulk so that at the end they can be reordered to the original ordering.


nivox:
@ennru This PR is the result of the discussion we had in #2062.

The PR is in a working state. There are however some less than perfect aspects I would like some assistance on:
* how to handle the backported `RetryFlowCoordinator` (where to put it, access level)
* I needed to change access restrictions on the `RetryLogic` implementations in order to extract the needed values to pass to the `RetryFlowCoordinator`. I think these should be refactored to be simple case classes used to initialize the `RetryFlowCoordinator`. This however would change the API and I'm not feeling comfortable tackling it
* `ElasticsearchFlowStage` could probably be deleted and replaced by `ElasticsearchSimpleFlowStage` having all `ElasticsearchFlow.create*` method use it instead of keeping two distinct implementations

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/scaladsl/ElasticsearchFlow.scala
```diff
@@ -160,15 +158,62 @@ object ElasticsearchFlow {
...
+                maxRetries,
+                (_, results) => retryLogic(results)
+              )
+            )
+          val elasticsearchFlow = new impl.ElasticsearchSimpleFlowStage[T, (Int, C)](indexName,
+                                                                                     typeName,
+                                                                                     elasticsearchClient,
+                                                                                     settings,
+                                                                                     writer)
+          retryFlowCoordinator.joinMat(elasticsearchFlow)(Keep.right)
```
ennru: Copy over the `RetryFlow.withBackoff` method so the switch to the Akka 2.6 implementation is even simpler.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/backport/RetryFlowCoordinator.scala
```diff
@@ -0,0 +1,158 @@
+/*
+ * Copyright (C) 2016-2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+/*
+ * Backported from Akka 2.6.
+ * TODO: remove and refactor with official Akka implementation as soon as Akka 2.5.x support is dropped
+ */
+
+package akka.stream.alpakka.elasticsearch.impl.backport
```
ennru: Works for me.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/scaladsl/ElasticsearchFlow.scala
```diff
@@ -160,15 +158,62 @@ object ElasticsearchFlow {
                                     typeName: String,
                                     settings: ElasticsearchWriteSettings,
                                     elasticsearchClient: RestClient,
-                                    writer: MessageWriter[T]) =
-    Flow
-      .fromGraph(
-        new impl.ElasticsearchSimpleFlowStage[T, C](indexName, typeName, elasticsearchClient, settings, writer)
+                                    writer: MessageWriter[T]) = {
+    val (minBackoff, maxBackoff, maxRetries) = settings.retryLogic match {
+      case RetryNever => (Duration.Zero, Duration.Zero, 0)
```
ennru: This could return a different flow for `RetryNever` which would just use the `ElasticsearchSimpleFlowStage` almost as before.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchWriteSettings.scala
```diff
@@ -24,7 +24,7 @@ object RetryNever extends RetryLogic {
  * elastic returns error one some of the documents in a
  * bulk request.
  */
-final class RetryAtFixedRate private (maxRetries: Int, retryInterval: scala.concurrent.duration.FiniteDuration)
+final case class RetryAtFixedRate(maxRetries: Int, retryInterval: scala.concurrent.duration.FiniteDuration)
```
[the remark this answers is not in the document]
nivox: I've refactored a little bit the `RetryLogic` trait to make it more coherent with the current implementation and added a new `RetryWithBackoff` logic.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchSimpleFlowStage.scala
```diff
@@ -50,15 +52,17 @@ private[elasticsearch] final class ElasticsearchSimpleFlowStage[T, C](
 
     private var inflight = false
 
-    private val failureHandler = getAsyncCallback[Throwable](handleFailure)
-    private val responseHandler = getAsyncCallback[(immutable.Seq[WriteMessage[T, C]], Response)](handleResponse)
+    private val failureHandler =
+      getAsyncCallback[(immutable.Seq[WriteMessage[T, C]], immutable.Seq[WriteResult[T, C]], Throwable)](handleFailure)
```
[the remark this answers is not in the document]
nivox: 👍

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchSimpleFlowStage.scala
```diff
@@ -71,30 +75,40 @@ private[elasticsearch] final class ElasticsearchSimpleFlowStage[T, C](
...
 
-    private def handleFailure(exception: Throwable): Unit = {
+    private def handleFailure(
+        args: (immutable.Seq[WriteMessage[T, C]], immutable.Seq[WriteResult[T, C]], Throwable)
+    ): Unit = {
       inflight = false
-      push(out, Failure(exception))
-      if (isClosed(in)) completeStage()
-      else tryPull()
+      val (_, resultsPassthrough, exception) = args
```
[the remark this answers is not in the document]
nivox: 👍

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchSimpleFlowStage.scala
```diff
@@ -71,30 +75,40 @@ private[elasticsearch] final class ElasticsearchSimpleFlowStage[T, C](
...
+          override def onSuccess(response: Response): Unit =
+            responseHandler.invoke((messages, resultsPassthrough, response))
         },
         new BasicHeader("Content-Type", "application/x-ndjson")
       )
     }
 
-    private def handleFailure(exception: Throwable): Unit = {
+    private def handleFailure(
+        args: (immutable.Seq[WriteMessage[T, C]], immutable.Seq[WriteResult[T, C]], Throwable)
```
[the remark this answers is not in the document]
nivox: 👍

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/scaladsl/ElasticsearchFlow.scala
```diff
@@ -160,15 +158,62 @@ object ElasticsearchFlow {
...
       )
-      .map {
-        case Success(value) => value
-        case Failure(e) => throw e
+      .map { results =>
+        val orderedResults = results.sortBy(_.message.passThrough._1)
+        val finalResults = orderedResults.map { r =>
+          new WriteResult(r.message.withPassThrough(r.message.passThrough._2), r.error)
+        }
+        finalResults
```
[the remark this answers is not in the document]
nivox: Refactored as a postprocessing flow

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/scaladsl/ElasticsearchFlow.scala
```diff
@@ -160,15 +158,62 @@ object ElasticsearchFlow {
...
+        case failedResults => Some(failedResults.map(_.message) -> successful)
+      }
+    }
+
+    Flow[immutable.Seq[WriteMessage[T, C]]]
+      .map { messages =>
+        val indexedMessages = messages.zipWithIndex.map {
+          case (m, idx) =>
+            m.withPassThrough(idx -> m.passThrough)
+        }
```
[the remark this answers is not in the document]
nivox: refactored asa a preprocessing flow

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/scaladsl/ElasticsearchFlow.scala
nivox: 👍

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/scaladsl/ElasticsearchFlow.scala
nivox: I've integrated this suggestion, but I'm not sure it is clearer than always using the RetryFlow with a 0 max retries

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchFlowStage.scala
```diff
@@ -19,7 +18,9 @@ import org.elasticsearch.client.{Response, ResponseListener, RestClient}
 import scala.collection.immutable
 
 /**
- * INTERNAL API
+ * INTERNAL API.
+ *
+ * Updates Elasticsearch without any built-in retry logic.
  */
 @InternalApi
 private[elasticsearch] final class ElasticsearchFlowStage[T, C](
```
ennru: The history would be clearer if you didn't rename the `ElasticSearchSimpleFlowStage`.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchFlowStage.scala
```diff
@@ -44,137 +48,77 @@ private[elasticsearch] final class ElasticsearchFlowStage[T, C](
...
+        args: (immutable.Seq[WriteResult[T, C]], Throwable)
+    ): Unit = {
+      inflight = false
+      val (resultsPassthrough, exception) = args
 
-    override def onPull(): Unit = tryPull()
+      log.error(s"Received error from elastic after having already processed {} documents. Error: {}",
+                resultsPassthrough.size,
+                exception)
+      failStage(exception)
```
ennru: Earlier it would retry even for errors, did that ever make sense?

nivox: I initially considered it, however:
1. In order to encode the exception in a `WriteResult` I would need to abuse the `error` passing the exception message. Downstream stages would have no (clean) way to discriminate between valid Elasticsearch errors and exceptions.
2. The old code would fail the stage if the error persisted throughout all the retries. However having delegated the retry logic to the `RetryFlow` I have no way to make that decision after all the retries have been exhausted. And due to 1 I cannot easily check for this kind of failure in a downstream flow.
 
I think that its better to let the stage fail for this kind of errors happens and maybe wrap the flow in a `RestartFlow.onFailuresWithBackoff`.

Personally this is something I prefer to handle in my client code instead of having the library do it for me, but if you think its better to have this handled transparently we can do something like:

```
  @InternalApi
  private def basicStageFlow[T, C](indexName: String,
                                   typeName: String,
                                   settings: ElasticsearchWriteSettings,
                                   elasticsearchClient: RestClient,
                                   writer: MessageWriter[T]) = {
    def flowFactory() = Flow.fromGraph {
      new impl.ElasticsearchSimpleFlowStage[T, C](indexName, typeName, elasticsearchClient, settings, writer)
    }
    
    RestartFlow.withBackoff(settings.retryLogic.minBackoff, settings.retryLogic.maxBackoff, 0, settings.retryLogic.maxRetries)(flowFactory)
  }
```

ennru: Fair enough. I agree that is is better to let the user decide about that retrying.

ennru:
As the API changed for the retry settings, MiMa is upset
https://travis-ci.com/akka/alpakka/jobs/276777782#L581

Please add exclusions in the `mima-filters/2.0.0-M2.backwards.excludes` directory:
```
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.elasticsearch.RetryAtFixedRate.*")
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.elasticsearch.RetryNever.*")
ProblemFilters.exclude[Problem]("akka.stream.alpakka.elasticsearch.RetryLogic.*")
```

ennru:
Thank you @nivox for your work on Alpakka Elasticsearch!

Issue #1989, opened 2019-10-19
[Failure] ArchiveSpec, should archive file


seglo:
Failure from PR #1981 build

https://travis-ci.com/akka/alpakka/jobs/244115447#L756

seglo:
https://travis-ci.com/akka/alpakka/jobs/260670103#L519

seglo:
https://travis-ci.com/akka/alpakka/jobs/261114508#L367

seglo:
https://travis-ci.com/akka/alpakka/jobs/275777719#L370

ennru:
Most likely fixed with #2105.

PR #2105, opened 2020-01-25
Files: Compare input data in zip archive tests

michalbogacz:
## Purpose
Improve stability of File archive tests

## References
References #1989 and #1909

## Changes
- improved stability in java and scala archive zip test

## Background Context
Before, archive created by zip stream was compared with archive created with plan java.
Unfortunately, while creating zip file current time is taken https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/zip/ZipOutputStream.java#L197
So, when the tests were slow, time didn't match in result archive. 
In this fix, archives are not compared. Only input files with files after zip -> unzip.

Fix as discussed with @ennru  https://github.com/akka/alpakka/issues/1909#issuecomment-576184708

### DIFF
--- a/file/src/test/java/docs/javadsl/ArchiveHelper.java
+++ b/file/src/test/java/docs/javadsl/ArchiveHelper.java
@@ -6,45 +6,36 @@ package docs.javadsl;
 
 import akka.util.ByteString;
 
-import java.io.File;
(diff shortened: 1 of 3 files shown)


ennru:
Thank you! These test failures have been very annoying.

PR #2108, opened 2020-01-27
Unix domain socket: upgrade dependencies; remove classifier

ennru:
## Purpose

Upgrade to later versions of "jiffi" and "jnr-unixsocket" and remove the classifier on "jiffi".

## References

The classifier puts sbt-whitesource off https://travis-ci.com/akka/alpakka/jobs/280329800#L356

## Background Context

I'm not sure if the classifier is needed anymore. It works locally on OSX for me.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -440,8 +440,8 @@ object Dependencies {
 
   val UnixDomainSocket = Seq(
     libraryDependencies ++= Seq(
-        "com.github.jnr" % "jffi" % "1.2.17" classifier "complete", // ApacheV2
-        "com.github.jnr" % "jnr-unixsocket" % "0.22" // BSD/ApacheV2/CPL/MIT as per https://github.com/akka/alpakka/issues/620#issuecomment-348727265
+        "com.github.jnr" % "jffi" % "1.2.22", // classifier "complete", // Is the classifier needed anymore?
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -440,8 +440,8 @@ object Dependencies {
 
   val UnixDomainSocket = Seq(
     libraryDependencies ++= Seq(
-        "com.github.jnr" % "jffi" % "1.2.17" classifier "complete", // ApacheV2
```
huntc: Unsure if the classifier is still required. It certainly used to be. What’s the downside of having it still?

ennru: My immediate problem is that it (with Coursier) breaks the Whitesource reporting.

seglo:
Might be related to this issue in sbt-whitesource https://github.com/lightbend/sbt-whitesource/issues/58

project/Dependencies.scala
mkurz: Coursier has problems with artifacts, see e.g. https://github.com/sbt/sbt/issues/5325 (and the linked issues there as well).

seglo: The testsuite in this branch runs fine for me locally.

ennru:
@huntc I'll merge this now, please reach out in case you run into problems without the dependency classifier.

Issue #2097, opened 2020-01-16
MQTT Streaming: incompatibility with Akka 2.6.0 or Akka 2.6.1


jcroig:
### Versions used 

Akka version: 

2.6.0

### Expected Behavior

We should be able to start an MQTT Server using alpakka mqtt streaming when running last Alpakka artifact in a project along with Akka 2.6.1. 

### Actual Behavior

Get this runtime exception:
```
java.lang.NoClassDefFoundError: akka/actor/typed/scaladsl/adapter/package$UntypedActorRefOps$ at akka.stream.alpakka.mqtt.streaming.scaladsl.ActorMqttServerSession.<init>(MqttSession.scala:455)
```
The problem is that implicit class UntypedActorRefOps has been renamed to ClassicActorRefOps from Akka 2.5.27 to Akka 2.6.x so when the artifact is built compiling with Akka 2.5.27 it fails to run with Akka 2.6.x

### Reproducible Test Case

Build Alpakka with Akka 2.5.27 and then use the artifact in a project with Akka 2.6.1. Try to start a MQTT broker with akka.stream.alpakka.mqtt.streaming.scaladsl.ActorMqttServerSession.

ennru:
The implicit conversion class for `ActorRef` has been renamed between Akka 2.5 and 2.6 (there was not binary-compatibility guarantee for Akka Typed).

huntc:
There shouldn’t be an expectation of binary compatibility between 2.5 and 2.6, right? https://doc.akka.io/docs/akka/current/common/binary-compatibility-rules.html

ennru:
Exactly. Some Akka Typed API changes between 2.5 and 2.6. What is nasty for Alpakka with this is that it is source-compatible, but not binary-compatible. So we can't distribute just one jar to support both Akka versions.

huntc:
> So we can't distribute just one jar to support both Akka versions.

I think you’re going to have to have two jars.

ennru:
We will tweak the build to just build and publish an Akka 2.6 compatible jar for Alpakka MQTT streaming.

raboof:
> There shouldn’t be an expectation of binary compatibility between 2.5 and 2.6, right? https://doc.akka.io/docs/akka/current/common/binary-compatibility-rules.html

2.6 is binary compatible with 2.5, except for API's that were already deprecated in 2.5 or marked may-change or experimental. The typed API's were still marked may-change in 2.5, so that's why they could still change there.

[...]

PR #2109, opened 2020-01-27
MQTT streaming: put routers under connector actors

ennru:
## Purpose

Move the router actors to be children of the `ClientState` and `ServerState` instead of system actors as part of reducing the Akka classic vs Akka typed mixup

## Background Context

As #2097 shows, the fact that Akka Typed is not binary-compatible between Akka 2.5 and 2.6 bites us.

### DIFF
--- a/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSession.scala
+++ b/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSession.scala
@@ -5,17 +5,19 @@
 package akka.stream.alpakka.mqtt.streaming
 package scaladsl
 
+import java.util.concurrent.TimeoutException
 import java.util.concurrent.atomic.AtomicLong
 
-import akka.actor.ExtendedActorSystem
 import akka.actor.typed.scaladsl.adapter._
+import akka.actor.{ExtendedActorSystem, Scheduler}
 import akka.event.Logging
 import akka.stream._
 import akka.stream.alpakka.mqtt.streaming.impl._
-import akka.stream.scaladsl.{BroadcastHub, Flow, Keep, Source}
-import akka.util.ByteString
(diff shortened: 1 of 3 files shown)


mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/scaladsl/MqttSession.scala
```diff
@@ -181,204 +148,246 @@ final class ActorMqttClientSession(settings: MqttSessionSettings)(implicit mat:
...
+                }
+                .filter(_.nonEmpty)
+                .log("client-commandFlow", _.iterator.decodeControlPacket(settings.maxPacketSize)) // we decode here so we can see the generated packet id
+                .withAttributes(ActorAttributes.logLevels(onFailure = Logging.DebugLevel))
+          }
+          .recover {
+            case e: TimeoutException =>
+              val f: CommandFlow[A] =
+                Flow.fromSinkAndSource(Sink.ignore, Source.empty[ByteString])
+              f
```
ennru: This hack makes the shutdown test
https://github.com/akka/alpakka/blob/9a87c963dc83fae259a372b59918f70e1f57e979/mqtt-streaming/src/test/scala/docs/scaladsl/MqttSessionSpec.scala#L1256
behave as before. But the correct behaviour should be I failure when someone tries to use a session that is shut down already, right?
cc @huntc

huntc:
My first impression is that this adds complexity purely to support binary compatibility. If we didn’t have to support both 2.5 and 2.6 then we wouldn’t need this hack, right?

ennru:
I'm not sure Akka 2.5/2.6 can be done even with this. But to put the routers into the connectors looks better to me than lots of system-level actors.
I think about going Akka 2.6 only.

huntc:
Routers are a session level abstraction, so they should be able to live as children where you have them then. However, I’d look them toward hiding the routers from the outside.

ennru:
As I don't think this will solve the Akka 2.5/2.6 problem, I close this.

PR #2111, opened 2020-01-27
FTP: error on failed delete/move

dwickern:
## Purpose

`Ftp.remove` and `Ftp.move` emit the result in its materialized value of `Future[IOResult]`. Currently there's no way to know if the delete/move failed, e.g. if the file does not exist. A successful and failed delete will both materialize as a `Future.successful(IOResult(1, Success(Done)))`. After this PR, a failed delete will materialize as a `Future.successful(IOResult(1, Failure(IOException)))`.


ennru:
Thanks, very useful.

PR #2116, opened 2020-01-31
AWS S3: Retry multipart chunk uploads on transient errors using RetryFlow

### LINKED ISSUE #2114: AWS S3: retry multipart upload chunks on transient errors
### Short description

The multipart upload sink (`S3Stream.multipartUpload`) should ideally protect upstream stages from the occasional transient error that S3 is susceptible to (be they 5xx errors or connection resets) by retrying a chunk upload when such an error occurs.

### Details

This has been discussed briefly in https://discuss.lightbend.com/t/alpakka-s3-retry-multipartupload-parts-on-failure/4403.

With the recent introduction of `akka.stream.scaladsl.RetryFlow`, this should hopefully be fairly straightforward, albeit that stuff is quite new and dependent on Akka 2.6.

I can probably offer a PR for this sometime soon, though I'm not sure about the unit testing side of things and how easy/difficult it is to simulate failures in Akka HTTP.

bpiper:
##  Purpose

Fixes #2114 

## Changes

The multipart upload Sink will now automatically retry a chunk upload up to 3 (by default) times in the event of a transient request failure (either a 5xx error or any Exception).

Note: this isn't ready to merge because it requires Akka 2.6.  Not sure if it's better to wait until Alpakka fully adopts 2.6, or share the backported RetryFlow code that alpakka-elasticsearch is using.

### DIFF
--- a/s3/src/main/resources/reference.conf
+++ b/s3/src/main/resources/reference.conf
@@ -89,4 +89,20 @@ alpakka.s3 {
   # This flag may disable the validation.
   # See https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html
   validate-object-key = true
+
+  # Settings corresponding to automatic retry of chunks in a streamed multipart upload.
+  retry-settings {
+    # The maximum number of additional attempts (following transient errors) that will be made to upload a chunk
+    # before giving up.
+    max-retries-per-chunk = 3
+
+    # The minimum delay between chunk retries.
+    min-backoff = 200ms
+
+    # The maximum delay between chunk retries.
+    max-backoff = 10s
+
+    # Random jitter factor applied to retry delay calculation.
+    random-factor = 0.0
+  }
 }
--- a/s3/src/main/resources/reference.conf
+++ b/s3/src/main/resources/reference.conf
@@ -90,19 +90,25 @@ alpakka.s3 {
   # See https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html
   validate-object-key = true
 
-  # Settings corresponding to automatic retry of chunks in a streamed multipart upload.
+  # Default settings corresponding to automatic retry of requests in an S3 stream.
   retry-settings {
-    # The maximum number of additional attempts (following transient errors) that will be made to upload a chunk
-    # before giving up.
-    max-retries-per-chunk = 3
+    # The maximum number of additional attempts (following transient errors) that will be made to process a given
+    # request before giving up.
+    max-retries = 3
 
-    # The minimum delay between chunk retries.
+    # The minimum delay between request retries.
     min-backoff = 200ms
 
-    # The maximum delay between chunk retries.
+    # The maximum delay between request retries.
     max-backoff = 10s
 
     # Random jitter factor applied to retry delay calculation.
     random-factor = 0.0
+  }
+
+  # Settings specific to S3 multipart uploads.
+  multipart-upload {
+
+    retry-settings = ${alpakka.s3.retry-settings}
   }
 }
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
@@ -10,32 +10,26 @@ import java.time.{Instant, ZoneOffset, ZonedDateTime}
 import akka.actor.ActorSystem
 import akka.annotation.InternalApi
 import akka.dispatch.ExecutionContexts
-
-import scala.collection.immutable.Seq
-import scala.concurrent.Future
-import scala.util.{Failure, Success}
-import akka.{Done, NotUsed}
-import akka.http.scaladsl.{ClientTransport, Http}
 import akka.http.scaladsl.model.StatusCodes.{NoContent, NotFound, OK}
-import akka.http.scaladsl.model.headers.{
-  `Content-Length`,
-  `Content-Type`,
-  BasicHttpCredentials,
-  ByteRange,
-  CustomHeader
-}
+import akka.http.scaladsl.model.headers._
 import akka.http.scaladsl.model.{headers => http, _}
 import akka.http.scaladsl.settings.{ClientConnectionSettings, ConnectionPoolSettings}
 import akka.http.scaladsl.unmarshalling.{Unmarshal, Unmarshaller}
+import akka.http.scaladsl.{ClientTransport, Http}
 import akka.stream.alpakka.s3.BucketAccess.{AccessDenied, AccessGranted, NotExists}
-import akka.stream.{ActorMaterializer, Attributes, Materializer}
-import akka.stream.alpakka.s3.impl.auth.{CredentialScope, Signer, SigningKey}
 import akka.stream.alpakka.s3._
 import akka.stream.alpakka.s3.headers.ServerSideEncryption
-import akka.stream.scaladsl.{Flow, Keep, RunnableGraph, Sink, Source}
+import akka.stream.alpakka.s3.impl.auth.{CredentialScope, Signer, SigningKey}
+import akka.stream.scaladsl.{Flow, Keep, RetryFlow, RunnableGraph, Sink, Source}
+import akka.stream.{ActorMaterializer, Attributes, Materializer}
 import akka.util.ByteString
+import akka.{Done, NotUsed}
 
 import scala.collection.immutable
+import scala.collection.immutable.Seq
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import scala.util.{Failure, Success, Try}
 
 /** Internal Api */
 @InternalApi private[s3] final case class S3Location(bucket: String, key: String) {
@@ -136,8 +130,11 @@ import scala.collection.immutable
   }.mapMaterializedValue(_ => NotUsed)
 
   sealed trait ListBucketState
+
   case object Starting extends ListBucketState
+
   case class Running(continuationToken: String) extends ListBucketState
+
   case object Finished extends ListBucketState
 
   def listBucketCall[T](
@@ -450,9 +447,10 @@ import scala.collection.immutable
       contentType: ContentType = ContentTypes.`application/octet-stream`,
       s3Headers: S3Headers,
       chunkSize: Int = MinChunkSize,
-      chunkingParallelism: Int = 4
+      chunkingParallelism: Int = 4,
+      maxRetriesPerChunk: Int = 3
   ): Sink[ByteString, Future[MultipartUploadResult]] =
-    chunkAndRequest(s3Location, contentType, s3Headers, chunkSize)(chunkingParallelism)
+    chunkAndRequest(s3Location, contentType, s3Headers, chunkSize, maxRetriesPerChunk)(chunkingParallelism)
       .toMat(completionSink(s3Location, s3Headers.serverSideEncryption))(Keep.right)
 
   private def initiateMultipartUpload(s3Location: S3Location,
@@ -525,8 +523,11 @@ import scala.collection.immutable
   //header implementation to expose that
   private case class CustomContentTypeHeader(contentType: ContentType) extends CustomHeader {
     override def name(): String = "Content-Type"
+
     override def value(): String = contentType.value
+
     override def renderInRequests(): Boolean = true
+
     override def renderInResponses(): Boolean = true
   }
 
@@ -569,47 +570,6 @@ import scala.collection.immutable
 
   val atLeastOneByteString = Flow[ByteString].orElse(Source.single(ByteString.empty))
 
-  private def createRequests(
-      s3Location: S3Location,
-      contentType: ContentType,
-      s3Headers: S3Headers,
-      chunkSize: Int,
-      parallelism: Int
-  ): Flow[ByteString, (HttpRequest, (MultipartUpload, Int)), NotUsed] = {
-
-    assert(
-      chunkSize >= MinChunkSize,
-      s"Chunk size must be at least 5 MB = $MinChunkSize bytes (was $chunkSize bytes). See http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPart.html"
-    )
-
-    val chunkBufferSize = chunkSize * 2
-
-    // First step of the multi part upload process is made.
-    //  The response is then used to construct the subsequent individual upload part requests
-    val requestInfo: Source[(MultipartUpload, Int), NotUsed] =
-      initiateUpload(s3Location, contentType, s3Headers.headersFor(InitiateMultipartUpload))
-
-    val headers = s3Headers.serverSideEncryption.toIndexedSeq.flatMap(_.headersFor(UploadPart))
-
-    Flow
-      .setup { (mat, attr) =>
-        implicit val conf = resolveSettings(attr, mat.system)
-
-        SplitAfterSize(chunkSize, chunkBufferSize)(atLeastOneByteString)
-          .via(getChunkBuffer(chunkSize, chunkBufferSize)) //creates the chunks
-          .concatSubstreams
-          .zipWith(requestInfo) {
-            case (chunkedPayload, (uploadInfo, chunkIndex)) =>
-              //each of the payload requests are created
-              val partRequest =
-                uploadPartRequest(uploadInfo, chunkIndex, chunkedPayload.data, chunkedPayload.size, headers)
-              (partRequest, (uploadInfo, chunkIndex))
-          }
-          .flatMapConcat { case (req, info) => Signer.signedRequest(req, signingKey).zip(Source.single(info)) }
-      }
-      .mapMaterializedValue(_ => NotUsed)
-  }
-
   private def getChunkBuffer(chunkSize: Int, bufferSize: Int)(implicit settings: S3Settings) =
     settings.bufferType match {
       case MemoryBufferType =>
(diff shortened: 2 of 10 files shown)


s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -642,53 +602,99 @@ import scala.collection.immutable
...
+            .via(superPool[(MultipartUpload, Int)])
 
-        requestFlow
-          .via(superPool[(MultipartUpload, Int)])
+        SplitAfterSize(chunkSize, chunkBufferSize)(atLeastOneByteString)
+          .via(getChunkBuffer(chunkSize, chunkBufferSize)) //creates the chunks
+          .concatSubstreams
+          .zip(requestInfo)
+          // Allow requests that fail with transient errors to be retried, using the already buffered chunk.
+          .via(RetryFlow.withBackoff(200.millis, 10.seconds, 0.0, maxRetriesPerChunk, retriableFlow) {
```
bpiper: Wasn't sure if it's worth allowing the back-off settings to be configured, and whether they (and the max retries) should be configurable globally or per-upload.  Very open to opinions on this.

s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3.scala
```diff
@@ -562,17 +562,37 @@ object S3 {
...
    * @param s3Headers any headers you want to add
+   * @param maxRetriesPerChunk the maximum number of times a given chunk upload will be retried on transient errors
    * @return a [[akka.stream.javadsl.Sink Sink]] that accepts [[akka.util.ByteString ByteString]]'s and materializes to a [[java.util.concurrent.CompletionStage CompletionStage]] of [[MultipartUploadResult]]
    */
   def multipartUpload(bucket: String,
                       key: String,
                       contentType: ContentType,
-                      s3Headers: S3Headers): Sink[ByteString, CompletionStage[MultipartUploadResult]] =
+                      s3Headers: S3Headers,
+                      maxRetriesPerChunk: Int): Sink[ByteString, CompletionStage[MultipartUploadResult]] =
```
seglo: We should add `maxRetriesPerChunk` to `S3Settings` so the overload isn't necessary. I see you're already retrieving the settings in `S3Stream` by calling `resolveSettings`.

If we use `S3Settings` you should be able to revert adding this param throughout the implementation.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -642,53 +602,99 @@ import scala.collection.immutable
...
 
+  private def handleChunkResponse(response: Try[HttpResponse], upload: MultipartUpload, index: Int)(
+      implicit mat: Materializer
+  ) = {
+
+    import mat.executionContext
+
+    response match {
+      case Success(r) =>
+        if (r.status.isFailure()) {
```
seglo: [minor] I know you're just moving this code around, but adding the conditional to the pattern match would make this cleaner.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -136,8 +130,11 @@ import scala.collection.immutable
   }.mapMaterializedValue(_ => NotUsed)
 
   sealed trait ListBucketState
+
   case object Starting extends ListBucketState
+
   case class Running(continuationToken: String) extends ListBucketState
```
seglo: [minor] Make final

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
bpiper: Done.

s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3.scala
bpiper: Have now added all of the retry configurables to S3Settings.  Considered using the max-retries-per-chunk for the PutObject retries as well, but then multipart upload is kind of its own beast (and probably more likely to hit 500 errors).

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
bpiper: Done.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
bpiper: Done.

seglo:
Thanks @bpiper. I think the only thing remaining are some tests. I don't think we necessarily need the same detailed tests as in Elasticsearch, but a few retry scenarios would be a good idea.

Also, the build is failing on a style check, but I'm not sure why.  I'll look into that.

seglo:
Can you rebase on master to force another build?

bpiper:
Looks like the copyright notice was updated recently.  Fixed now.

I did extend an existing test case (that was previously only testing retries of the initiate multipart upload request) to also test the upload part request retry, but I'll have a think about how to improve it further, albeit wiremock tests can be a bit fiddly.

bpiper:
Some more tests to come (if I can figure out how to run retry tests with arbitrary Alpakka config), and I realised I need to ensure that DiskBuffer's `maxMaterializations` value is appropriately set based on the max retries... hopefully something I can also add a test for.

bpiper:
I don't feel this is RTBM yet, as I'm having issues with the happy path in real world testing.  Something along the lines of files getting uploaded to S3 successfully but the graph not completing (the HTTP request I have feeding the sink is timing out, so that seems the likely explanation).  All seems to work fine on `2.0.0-M2`.

ennru:
Thank you for trying it out more thoroughly!
`RetryFow` could be considerably slower as it leads to sequencing the requests one-by-one.

bpiper:
I don't think serialization of chunks is necessarily the issue.  In my real-world test case, I have a multipart Play request with multiple files in it (all are significantly smaller than the memory buffer size of 20 MB), each fed to its own multipart Alpakka sink.  What I'm seeing is that each one takes just marginally more (like 100 - 300 ms) than 5 seconds to complete, regardless of the file size.  Given enough of these small files buffered up on the server (e.g. 20 files buffered within a couple of seconds), the client can end up waiting long enough after its last data transmission that it times out.

5 seconds seems to be a common default timeout value for quite a few things in Akka libraries, but nothing really jumped out at me as relevant to this.  Also, if it was a timeout, it would be a soft timeout of sorts, since the graph is still completing with all elements emitted.  I'll keep investigating (e.g. litter the graph with `log` calls), but any thoughts/ideas are welcome.

bpiper:
Ok, so turns out there was a mysterious 2 second delay on each of the initiate and finalise requests (maybe also somewhat less of a delay on the part upload requests as well) that I couldn't attribute to any of the Alpakka code.  I diff'ed the list of libraries between `2.0.0-M2` and the code on my branch and found that the only difference from `2.0.0-M2` was the AWS v2 SDK version (`2.10.34` -> `2.10.47`).  After reverting to `2.10.34`, the delay disappeared, so I did a binary search and narrowed the culprit down to `2.10.43.  Unfortunately I can also confirm that the latest release, `2.10.65` still has this problem.

As to what the root cause actually is - it looks like the `AwsRegionProvider` is taking almost exactly 1 second for the first `getRegion` call it makes.  Alpakka is freshly resolving the config numerous times throughout the course of a multipart upload (for a single chunk upload, apparently around 5 times in total), hence a number of those expensive `getRegion` calls are occurring... probably `InstanceProfileRegionProvider.getRegion` to be precise.  I'm not sure exactly what change in `2.10.43` was responsible for this 1 second delay, and I may follow that up separately, but regardless of that, it's best to create a single instance of the region provider and re-use it for the lifetime of the JVM, so that probably calls for a change to `S3Stream.resolveSettings`.  Arguably this belongs in a separate issue, but I could address it under this one.

ennru:
Thank you @bpiper for investigating it this deep!
With our current routines, we can't identify this kind of regression when we update AWS library versions.
If you want to focus on the retrying first, create a new issue for caching the `Region` better.

s3/src/main/resources/reference.conf
```diff
@@ -89,4 +89,20 @@ alpakka.s3 {
   # This flag may disable the validation.
   # See https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html
   validate-object-key = true
+
+  # Settings corresponding to automatic retry of chunks in a streamed multipart upload.
+  retry-settings {
```
ennru: It makes sense to introduce `multipart-upload.retry-settings` in case we discover more needs for retry configurations.

It could rely on config inheritance:
```
multipart-upload.retry-settings = ${alpakka.s3.retry-settings}
```

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -569,53 +569,12 @@ import scala.collection.immutable
...
-  }
-
-  private def getChunkBuffer(chunkSize: Int, bufferSize: Int)(implicit settings: S3Settings) =
+  private def getChunkBuffer(chunkSize: Int, bufferSize: Int, maxRetriesPerChunk: Int)(implicit settings: S3Settings) =
     settings.bufferType match {
       case MemoryBufferType =>
         new MemoryBuffer(bufferSize)
       case d: DiskBufferType =>
-        new DiskBuffer(2, bufferSize, d.path)
+        new DiskBuffer((maxRetriesPerChunk + 1) * 2, bufferSize, d.path)
```
ennru: This new value would benefit from a proper name or explanation.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -569,53 +569,12 @@ import scala.collection.immutable
...
-                uploadPartRequest(uploadInfo, chunkIndex, chunkedPayload.data, chunkedPayload.size, headers)
-              (partRequest, (uploadInfo, chunkIndex))
-          }
-          .flatMapConcat { case (req, info) => Signer.signedRequest(req, signingKey).zip(Source.single(info)) }
-      }
-      .mapMaterializedValue(_ => NotUsed)
-  }
-
-  private def getChunkBuffer(chunkSize: Int, bufferSize: Int)(implicit settings: S3Settings) =
+  private def getChunkBuffer(chunkSize: Int, bufferSize: Int, maxRetriesPerChunk: Int)(implicit settings: S3Settings) =
```
ennru: This could become a local method in `chunkAndRequest`.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -648,47 +607,91 @@ import scala.collection.immutable
...
+        SplitAfterSize(chunkSize, chunkBufferSize)(atLeastOneByteString)
+          .via(getChunkBuffer(chunkSize, chunkBufferSize, maxRetriesPerChunk)) //creates the chunks
+          .concatSubstreams
+          .zip(requestInfo)
+          // Allow requests that fail with transient errors to be retried, using the already buffered chunk.
+          .via(RetryFlow.withBackoff(minBackoff, maxBackoff, randomFactor, maxRetriesPerChunk, retriableFlow) {
+            case (chunkAndUploadInfo, (Success(r), _)) =>
+              // 5xx errors from S3 can be treated as transient.
+              if (r.status.intValue >= 500) Some(chunkAndUploadInfo) else None
+            case (chunkAndUploadInfo, (Failure(_), _)) =>
```
ennru: What failures do happen here? Are all of them worth retrying?

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -648,47 +607,91 @@ import scala.collection.immutable
...
+    import mat.executionContext
+
+    response match {
+      case Success(r) if r.status.isFailure() =>
+        Unmarshal(r.entity).to[String].map { errorBody =>
+          FailedUploadPart(
+            upload,
+            index,
+            new RuntimeException(
+              s"Upload part $index request failed. Response header: ($r), response body: ($errorBody)."
```
ennru: Would it make sense to refer to the retry settings in the error message?

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
bpiper: So the only one I've encountered in recent memory is `akka.http.impl.engine.client.OutgoingConnectionBlueprint$UnexpectedConnectionClosureException`, which is basically a connection reset... something that S3 seems to do every so often in lieu of a 500.  My inclination is that any exception here is likely to be transient, and it's no great harm if it isn't (it'll still get thrown after a bit of a delay, and probably points to an environmental issue or something requiring human attention), particularly as UploadPart is idempotent.  I think if/when people do identify recurrent exceptions that can and do occur with some regularity, and where retrying is just a waste of time, they can add them here.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
bpiper: Yeah good idea, for a 5xx error.  For a `Failure` it's probably safest not to wrap the exception (in case anyone's relying on it not being wrapped), so may just have to leave that one as is.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
bpiper: Done.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
bpiper: Done.

s3/src/main/resources/reference.conf
bpiper: Done.

s3/src/main/scala/akka/stream/alpakka/s3/settings.scala
```diff
@@ -401,6 +408,14 @@ object S3Settings {
     }).getOrElse(ApiVersion.ListBucketVersion2)
     val validateObjectKey = c.getBoolean("validate-object-key")
 
+    val retryConfig = c.getConfig("retry-settings")
+    val retrySettings = RetrySettings(
+      retryConfig.getInt("max-retries-per-chunk"),
+      FiniteDuration(retryConfig.getDuration("min-backoff").toNanos, TimeUnit.NANOSECONDS),
+      FiniteDuration(retryConfig.getDuration("max-backoff").toNanos, TimeUnit.NANOSECONDS),
+      retryConfig.getDouble("random-factor")
+    )
```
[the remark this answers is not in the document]
bpiper: Done.

bpiper:
I've raised #2150 for the `getRegion` performance issue.  Still doing a bit of real world testing on this one, but so far it's looking ok.

@ennru if using `RetryFlow` means that chunks are uploaded sequentially rather than in parallel, should we perhaps only use it if `max-retries` is > 0? i.e. allow people to opt out of retries in order to avoid any degradation in performance.  In my own use case, the upstream is probably a little slower than the S3 upload, so chunks would generally be uploaded to S3 faster than they get buffered, but I imagine if you were uploading straight from local disk then `RetryFlow` could be quite a bottleneck?

Alternatively, can we split into sub-flows (which use RetryFlow) based on `chunkIndex % parallelism`, and then concat the substreams back into a flow of responses? I don't know if it would then make sense just to use `singleRequest` rather than `superPool` for the requests.  I may be talking total nonsense.

ennru:
Yeah, we could do a special solution that doesn't do retrying. But when would you prefer more failures over less speed?

> Alternatively, can we split into sub-flows (which use RetryFlow) based on chunkIndex % parallelism, and then concat the substreams back into a flow of responses?

We have `SubFlow`s already after the chunking. I believe it is just a matter of replacing the `.concatSubstreams` with `.mergeSubstreams(parallelism)` and put that after the `mapAsync`. (I need to sleep over it and some insight from my colleagues.)

The `superPool` is what we want in any case.

bpiper:
I think it's more a case of wanting potentially significantly better performance 99.9% of the time and copping the 0.1% on the chin in exchange.  Some people might already have their own retry logic in place at a higher layer (I know I do), and unlike me may have a much faster upstream, in which case these changes are possibly a step backwards (even though the retrying is now much more efficient in those 0.1% of failure cases).

I think your suggestion makes sense, since there doesn't seem to be anything in that flow (apart from the split of course) that requires sequential processing of chunks... so there's actually no need to merge the substreams until we have the upload responses and are ready to hand them over to the completion sink in any old order.  I think the `mapAsync` might as well be a `mapAsync(1)` as well, since realistically it'll never be handling chunk responses in parallel.

I'll test those changes in anger now.  Maybe I can also write a quick program to test uploading large files straight from my SSD to S3 (i.e. fast upstream, slow-ish downstream) just to verify that chunks are uploading in parallel.

bpiper:
Actually we do need to merge substreams before the zip... I forgot that zip is doing the InitiateMultipartUpload request and producing the chunk indices.  So then it's split and buffer chunks -> merge with parallelism -> zip in the request info and the chunk indices -> group by index % parallelism -> retriable chunk uploads -> handle responses -> merge with parallelism -> completion sink.  I've done some testing of this with large files and it seems to be working so far.  I'll push the changes for scrutiny anyway.

bpiper:
@ennru just wondering if you've had a chance to look at this again after the latest changes? My testing (albeit of relatively limited scope) hasn't revealed any problems, but I can understand if you're keen to get a 2.0.0 release out before merging this.

ennru:
Thank you for this great work! I hope it makes the upload even better for everyone.

PR #2117, opened 2020-01-31
Update copyright headers

ignasi35:
🎉 Welcome to 2020 🎉


ignasi35:
I've redone the PR to drop the year (instead of bumping to 2020) and use also moved to `https`.

PR #2118, opened 2020-01-31
MQTT streaming: target Akka 2.6

raboof:
Still failing one test: "MqttSessionSpec: MQTT client connector should receive a QoS 1 publication from a subscribed topic and ack it and then ack it again"

References #2097


ennru:
Watch out for the Alpakka testkit dependency brought in by `alpakkaProject`.

raboof:
(failure is scaladoc generation, looking into that...)

ennru:
The failure is not new to the branch https://github.com/akka/alpakka/issues/1723

ennru:
Ah, I think it should state Akka 2.6 only, instead.

raboof:
it seems obvious from the dependency table. Do you think we should spell it out further?

DeepinScreenshot_select-area_20200212154646

ennru:
Yes, it is an exception to the rule in Alpakka. 
See https://doc.akka.io/docs/alpakka/2.0/overview.html#versions

PR #2040, opened 2019-11-29
Testkit: introduce log capturing

ennru:
## Purpose

Introduce log capturing the with an Alpakka testkit which lends code from akka-actor-typed-testkit.

## References

https://doc.akka.io/docs/akka/current/typed/testing-async.html#silence-logging-output-from-tests

PR #2119, opened 2020-01-31
Fix Java Unit Test docs example in Unix Domain Sockets

seglo:
The lone Java test in unix-domain-socket used for docs has always failed, but we never noticed it because we didn't block on stream completion before the test completed.  The blocking was added #2040 (https://github.com/akka/alpakka/pull/2040/files#r368017011), but the test failure got confused with a transient test failure and the PR was merged.

Changes
* Use a temp directory instead of a file.  `jnr-unixsocket` will create the socket files on our behalf, if the file already exists then we'll observe an `Address in use` exception (https://travis-ci.com/akka/alpakka/jobs/277967218#L358)
* Actually send data through socket!
* Remove framing boilerplate and just parse as a `ByteString` like the Scala docs example does.

### DIFF
--- a/unix-domain-socket/src/test/java/docs/javadsl/UnixDomainSocketTest.java
+++ b/unix-domain-socket/src/test/java/docs/javadsl/UnixDomainSocketTest.java
@@ -4,13 +4,9 @@
 
 package docs.javadsl;
 
-import akka.Done;
 import akka.NotUsed;
 import akka.actor.ActorSystem;
-import akka.stream.ActorMaterializer;
-import akka.stream.KillSwitch;
-import akka.stream.KillSwitches;
-import akka.stream.Materializer;
+import akka.stream.*;
 import akka.stream.alpakka.testkit.javadsl.LogCapturingJunit4;
 import akka.stream.alpakka.unixdomainsocket.javadsl.UnixDomainSocket;
 import akka.stream.javadsl.*;
@@ -24,6 +20,7 @@ import org.junit.Test;
 import akka.japi.Pair;
 
 import java.nio.file.Files;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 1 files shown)


raboof:
build failure should be fixed by #2123

unix-domain-socket/src/test/java/docs/javadsl/UnixDomainSocketTest.java
```diff
@@ -80,23 +83,29 @@ public void aUnixDomainSocketShouldReceiveWhatIsSent() throws Exception {
...
                   final Flow<ByteString, ByteString, NotUsed> echo =
                       Flow.of(ByteString.class)
-                          .via(
-                              Framing.delimiter(
-                                  ByteString.fromString("\n"), 256, FramingTruncation.DISALLOW))
-                          .map(ByteString::utf8String)
-                          .map(s -> s + "!!!\n")
-                          .map(ByteString::fromString);
+                          // server logic ...
+                          // #outgoingConnection
```
raboof: Good to keep the actual implementation out of the docs

unix-domain-socket/src/test/java/docs/javadsl/UnixDomainSocketTest.java
```diff
@@ -63,15 +63,18 @@ public void aUnixDomainSocketShouldReceiveWhatIsSent() throws Exception {
...
 
     // #binding
     final Source<IncomingConnection, CompletionStage<ServerBinding>> connections =
         UnixDomainSocket.get(system).bind(path);
     // #binding
 
+    final CompletableFuture<ByteString> received = new CompletableFuture<>();
+
     // #outgoingConnection
+    ByteString sendBytes = ByteString.fromString("Hello");
```
raboof: (this line could be moved down to below the part that makes it into the docs)

seglo: :+1:

PR #2123, opened 2020-02-03
Add some missing copyright header updates

raboof:
Follow-up on #2117


ennru:
Why weren't these covered in #2117 ?

raboof:
> Why weren't these covered in #2117 ?

I'm not sure what exactly was going on, but I could reproduce the situation where `sbt headerCreate` did nothing while `sbt headerCheck` reported failures. I haven't dug deeper, but I suspect those files weren't touched after the change to the sbt config, and sbt assumed the cached versions were OK and didn't need updating. `sbt clean` before `sbt headerCreate` did the trick.

PR #2131, opened 2020-02-04
error handling for UnixDomainSocketTest

raboof:
This would have made #2129 more obvious, checking the stream
result is good practice so we should lead by example.


raboof:
(failure was a github timeout, rebuild fixed it)

PR #2138, opened 2020-02-10
Remove Memory limits for local builds

ignasi35:


.jvmopts
```diff
@@ -1,5 +1,3 @@
 -Dfile.encoding=UTF8
```
seglo: Maybe this default could be defined explicitly within any test that relies on it.

ignasi35:
> LGTM. I'm embarassed to say I didn't know this file existed ;)
> 
> WDYT about also adding it to `.gitignore`, so users can override it with the risk of committing it accidentally?

Actually, it's quite common and I've relied on it in the past:

https://github.com/lagom/lagom/commit/69aa74b91623001a6fc93bbb5630601067b99a80#diff-73387b1927e8bbf6f256745f2f547b87

I don't think ignoring is the right approach since we should be able to set some minimum settings and other recommendations or workarounds we discover help complete the build. E.g., I think the PR is wrong as it is now since we should keep `-Xms1G` so environments that can't even allocate that may just not even bother.

ignasi35:
There are other details such as `+CMSClassUnloadingEnabled` which, of course, only make sense if the user JVM defaults to CMS GC. 🤦🏼‍♂️

I think we can iterate over this.

PS: ICYMI, there's also the `.jvmopts-travis` that might need a review too (specially if we use different JDK majors and also as we bump minor versions that take new defaults)

seglo:
> I don't think ignoring is the right approach since we should be able to set some minimum settings and other recommendations or workarounds we discover help complete the build. E.g., I think the PR is wrong as it is now since we should keep `-Xms1G` so environments that can't even allocate that may just not even bother.

To clarify I meant keep the `.jvmopts` file (with your updates), but add it to `.gitignore` so that user customizations aren't committed.  I'm not crazy about that solution either, but at least it would give users something to start with.

We could also programmatically use the `javaOptions` SBT setting in the `build.sbt`.  We could set some sensible project defaults, and then add a git ignored file in the root, like `javaoptions.sbt`, where users can add customizations.

seglo:
Or a combination solution would be to programmatically add project defaults in `build.sbt`, and git ignore `.jvmopts`?  We could also integrate `.jvmopts-travis` into the `build.sbt` and only apply them them when running on Travis.

ignasi35:
Hmmm, I'm beginning to like the idea to `.gitignore` the file. We could just recommend users to get their own copy of the file locally in `CONTRIBUTING` and maybe have a `.jvmopts.sample` with some sensible defaults.

I think there are two problems to solve.

1. memory management settings (Xms, Xmx, Xss, CMSClassUnloadingEnabled) for sbt tasks.

I first needed to tweak the memory settings when noticed that a clean `sbt test:compile` was causing the task to block because it spend infinite time GC'ing. I think that points to a memory hungry sbt plugin so we are probably stuck with using `.jvmopts` and `JAVA_OPTS` (instead of `javaOptions` which might have kicked in if I had actually run the tests).
I think the problem only became apparent when compiling on my new machine (8 core) which launched too many tasks in sbt causing more memory pressure than less-beefy environments where the 2Gb limit configured in `.jvmopts` is enough. 

2. memory management settings for forked JVMs.

I am not familiar enough with the memory requirements to run the test suites but fewer restrictions don't sound wrong :-)

3. file encoding

Then there's the `-Dfile.encoding=UTF8` setting. I don't know if that's only required for tests or maybe even some sbt plugin, code generation,...

ignasi35:
I merged without waiting for travis because this change has no impact on the travis build (which overwrites `.jvmopts`  with `.jvmopts.travis`)

PR #2142, opened 2020-02-11
Adds config for Probot Settings

ignasi35:
Do not merge before https://github.com/akka/.github/pull/6

*NOTE*: we should turn on "require code owner review" for the master branch.


ignasi35:
After some more discussion, the risks caused by probot's [security implications](https://github.com/probot/settings#security-implications) may supersede the automation the bot provides.

ignasi35:
We won't be using probot settings for the time being.

PR #2143, opened 2020-02-11
[WIP] Supporting request/reply pattern with akka streams implementation

seglo:
The existing test implementation demonstrates that the request/reply pattern is possible, but it doesn't serve as a good example for users to use as a template in their own apps.  This PR iterates on the implementation to see how the JMS connector can be modified to provide a more Akka Streams idiomatic implementation.

### DIFF
--- a/jms/src/main/scala/akka/stream/alpakka/jms/JmsConsumerSettings.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/JmsConsumerSettings.scala
@@ -24,11 +24,16 @@ final class JmsConsumerSettings private (
     val acknowledgeMode: Option[AcknowledgeMode],
     val ackTimeout: scala.concurrent.duration.Duration,
     val failStreamOnAckTimeout: Boolean,
-    val connectionStatusSubscriptionTimeout: scala.concurrent.duration.FiniteDuration
+    val connectionStatusSubscriptionTimeout: scala.concurrent.duration.FiniteDuration,
+    val consumer: Option[javax.jms.MessageConsumer]
 ) extends akka.stream.alpakka.jms.JmsSettings {
 
   /** Factory to use for creating JMS connections. */
   def withConnectionFactory(value: javax.jms.ConnectionFactory): JmsConsumerSettings = copy(connectionFactory = value)
+
+  /** Use provided consumer. */
+  def withConsumer(consumer: javax.jms.MessageConsumer): JmsConsumerSettings =
+    copy(consumer = Some(consumer))
 
   /** Configure connection retrying. */
   def withConnectionRetrySettings(value: ConnectionRetrySettings): JmsConsumerSettings =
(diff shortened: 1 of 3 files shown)


jms/src/main/scala/akka/stream/alpakka/jms/JmsConsumerSettings.scala
```diff
@@ -24,12 +24,17 @@ final class JmsConsumerSettings private (
     val acknowledgeMode: Option[AcknowledgeMode],
     val ackTimeout: scala.concurrent.duration.Duration,
     val failStreamOnAckTimeout: Boolean,
-    val connectionStatusSubscriptionTimeout: scala.concurrent.duration.FiniteDuration
+    val connectionStatusSubscriptionTimeout: scala.concurrent.duration.FiniteDuration,
+    val consumer: Option[javax.jms.MessageConsumer]
```
seglo: I was not able to consume a temporary queue destination with connections created by the connection factory.  To create a `JmsConsumer` that could stream the "replies", I had to use the same consumer from the session that was used to create the temporary queue.

jms/src/test/scala/docs/scaladsl/JmsConnectorsSpec.scala
```diff
@@ -1269,4 +1269,148 @@ class JmsConnectorsSpec extends JmsSpec {
 
     connection.close()
   }
+
+  "support request/reply with temporary queues and manual reply consumer" in withConnectionFactory() { connectionFactory =>
```
seglo: This test shows my first attempt to combine a manually created `javax.jms.MessageConsumer` that consumes the replies send by the remote service.

jms/src/test/scala/docs/scaladsl/JmsConnectorsSpec.scala
```diff
@@ -1269,4 +1269,148 @@ class JmsConnectorsSpec extends JmsSpec {
...
+    msg shouldBe a[TextMessage]
+    msg.getJMSCorrelationID shouldEqual correlationId
+    msg.asInstanceOf[TextMessage].getText shouldEqual message.reverse
+
+    respondStreamControl.shutdown()
+
+    connection.close()
+  }
+
+  "support request/reply with temporary queues and zipped reply source" in withConnectionFactory() {
```
seglo: This test represents a full implementation that zips together a JMS reply stream with the request stream with a `mapAsync` stage that validates correlation IDs before resuming.

jms/src/test/scala/docs/scaladsl/JmsConnectorsSpec.scala
seglo: This example works and doesn't require us to support a custom consumer passed to `JmsConsumerSettings`, but is not very Akka Streams idiomatic.

seglo:
This documentation suggests that a temporary queue can be used by any session created by the same connection, but AFAIK the connections are created by the connection factory.

https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.1/html/configuring_messaging/temporary_queues_and_runtime_queues

ennru:
Oh, a throw-back in time: https://www.enterpriseintegrationpatterns.com/patterns/messaging/RequestReplyJmsExample.html

ennru:
I agree with you, none of the solutions you found is really convincing.

seglo:
I like the composition of the stream in the second example.  If we could find a better way to use the temporary queue, or share a session, with connections from the factory then we could promote this pattern.

seglo:
Closing since I don't have any intention at this time to develop this solution. If anyone's interested in continuing the effort feel free to branch it and/or create a new PR.

Issue #2134, opened 2020-02-07
FTP: connect with password and selfIdentity


harshalveera:
As a use case SFTP connect with username, password along with selfIdentity

PR #2144, opened 2020-02-11
FTP: Support username/password credentials and key for login

harshalveera:
References #2134


ftp/src/main/scala/akka/stream/alpakka/ftp/impl/SftpOperations.scala
```diff
@@ -41,10 +42,21 @@ private[ftp] trait SftpOperations { _: FtpLike[SSHClient, SftpSettings] =>
 
     ssh.connect(host.getHostAddress, port)
 
-    if (credentials.password != "" && sftpIdentity.isEmpty)
-      ssh.authPassword(credentials.username, credentials.password)
+    if (credentials.password != "" && sftpIdentity.isDefined) {
+      val passwordAuth: Seq[AuthPassword] = credentials.password.map { pass =>
```
ennru: This is weird, it will create an `AuthPassword` for every character in `credentials.password`.

ftp/src/main/scala/akka/stream/alpakka/ftp/impl/SftpOperations.scala
```diff
@@ -41,10 +42,21 @@ private[ftp] trait SftpOperations { _: FtpLike[SSHClient, SftpSettings] =>
 
     ssh.connect(host.getHostAddress, port)
 
-    if (credentials.password != "" && sftpIdentity.isEmpty)
-      ssh.authPassword(credentials.username, credentials.password)
+    if (credentials.password != "" && sftpIdentity.isDefined) {
```
ennru: Try and use 
```
    sftpIdentity match {
      case Some(identity) if credentials.password != "" =>
```
to get a more readable structure.

Not introduced by you, but `credentials.password != ""` could mean it is `null`?!

ftp/src/main/scala/akka/stream/alpakka/ftp/impl/SftpOperations.scala
```diff
@@ -155,6 +167,24 @@ private[ftp] trait SftpOperations { _: FtpLike[SSHClient, SftpSettings] =>
     }
   }
 
+  private[this] def authPublickey(identity: SftpIdentity)(implicit ssh: SSHClient) = {
+    def bats(array: Array[Byte]): String = new String(array, "UTF-8")
+
+    def initKey(kp: KeyProvider) = {
+      new AuthPublickey(kp)
+    }
```
ennru: Just use `new AuthPublickey` down there.

ftp/src/main/scala/akka/stream/alpakka/ftp/impl/SftpOperations.scala
```diff
@@ -155,6 +167,24 @@ private[ftp] trait SftpOperations { _: FtpLike[SSHClient, SftpSettings] =>
     }
   }
 
+  private[this] def authPublickey(identity: SftpIdentity)(implicit ssh: SSHClient) = {
+    def bats(array: Array[Byte]): String = new String(array, "UTF-8")
```
ennru:
```suggestion
    def bats(array: Array[Byte]): String = new String(array, StandardCharsets.UTF_8)
```
I'm not sure if the local method makes the code better.

ftp/src/main/scala/akka/stream/alpakka/ftp/impl/SftpOperations.scala
harshalveera: fixed

ftp/src/main/scala/akka/stream/alpakka/ftp/impl/SftpOperations.scala
harshalveera: fixed

ftp/src/main/scala/akka/stream/alpakka/ftp/impl/SftpOperations.scala
harshalveera: fixed

ftp/src/main/scala/akka/stream/alpakka/ftp/impl/SftpOperations.scala
harshalveera: Using map to convert it to seq

harshalveera:
@ennru @Falmarri  @seglo Can you guys look at this and approve it please.

ftp/src/main/scala/akka/stream/alpakka/ftp/impl/SftpOperations.scala
raboof: While it 'makes the types fit', I'm pretty sure it's wrong.

Could you add a testcase showing it's actually right?

harshalveera: @raboof I will try to add unit tests and test cases
 

Can you elaborate more on why do you think it is wrong?

raboof: As enno mentioned this looks like it'd create an `AuthPassword` for every character in the password, which I can't imagine is the right behavior.

harshalveera: okay. I changed that. it was creating the AuthPassword multiple times which I changed.

Sorry for that oversight on myside

ftp/src/main/scala/akka/stream/alpakka/ftp/impl/SftpOperations.scala
ennru: What is fixed?
I'd still prefer the match clause with guards and extractor over the ifs.

ftp/src/main/scala/akka/stream/alpakka/ftp/impl/SftpOperations.scala
```diff
@@ -41,10 +43,20 @@ private[ftp] trait SftpOperations { _: FtpLike[SSHClient, SftpSettings] =>
...
+    if (credentials.password != "" && sftpIdentity.isDefined) {
+      val passwordAuth: AuthPassword = new AuthPassword(new PasswordFinder() {
+        def reqPassword(resource: Resource[_]): Array[Char] = credentials.password.toCharArray
+
+        def shouldRetry(resource: Resource[_]) = false
+      })
 
-    sftpIdentity.foreach(setIdentity(_, credentials.username))
+      val authenticationMethods = List(authPublickey(sftpIdentity.get), passwordAuth)
+      ssh.auth(credentials.username, authenticationMethods: _*)
```
ennru: Pass the auth-methods as parameters directly.

Issue #2141, opened 2020-02-11
Provide ways to configure PubSub gRPC backpressure


jrudolph:
As was observed in https://discuss.lightbend.com/t/akka-stream-initial-backpressure-throttle-on-source-has-no-effect/5865, the PubSub connector might pull down a full buffer of messages before it starts backpressuring. This might instantly create a sizable backlog if the message size is small and the processing time long.

This is explicitly discussed in the [Google PubSub documentation](https://cloud.google.com/pubsub/docs/pull#streamingpull_dealing_with_large_backlogs_of_small_messages). It would be good if the user could be given more fine-grained control over the backlog. (Also that problem and solutions should be documented). It's not quite clear to me if it can achieved using the `streamingPull` mechanism or if there should be an alternative implementation that uses (potentially multiple concurrent) pull requests to fetch in elements with tighter control from the downstream consumer.

See also https://github.com/akka/akka-grpc/issues/501

jrudolph:
Maybe to give an additional point about the trade-offs:
 * streaming pull is great if receiving elements over the network is the bottleneck because it will try to maximize network throughput (but might use sizable buffers, maybe building up a backlog if the messages are small)
 * a sequence of explicit pulls is good if network speed is not the bottleneck and you want to keep the backlog under control

mdibaiee:
@jrudolph Thanks for filing the issue. I am willing to implement this if you can help me with an example of a similar implementation in other alpakka connectors perhaps. I would think polling is a common pattern so alpakka might already have abstractions for that?

As you pointed out, I think one solution is to use synchronous pull requests instead of streamingPull where `max_messages` has to respect backpressure.

https://cloud.google.com/pubsub/docs/reference/rpc/google.pubsub.v1#pullrequest

jrudolph:
>  I am willing to implement this if you can help me with an example of a similar implementation in other alpakka connectors perhaps. I would think polling is a common pattern so alpakka might already have abstractions for that?

Thanks, @mdibaiee. Not to familiar with what can be found in all of alpakka, maybe @ennru would know about an example about how to do efficient polling in alpakka?

> As you pointed out, I think one solution is to use synchronous pull requests instead of streamingPull where `max_messages` has to respect backpressure.

Yep, that would be the idea with `max_messages` being set to a configurable value to batch pulling. One could think of requesting the next batch even when the previous batch hasn't been consumed completely to try to optimize throughput even in that case.

seglo:
[...]

PR #2166, opened 2020-02-21
Google pub/sub gRPC: allow sync poll using PullRequest

mdibaiee:
References #2141 

Here is a simple approach to have synchronous pull. I tried this in our environment (scala) and it worked well, the only quirk being that we have to tune the `maxMessages` parameter and interval finely. Wondering if there is a better way to manage the amount of messages being pulled.

This is what the usage looks like:

```
  private val request: PullRequest =
    PullRequest().withSubscription(Settings.PubSub.pubSubSubscription).withMaxMessages(100)

  lazy val source: Source[ReceivedMessage, Future[Cancellable]] = GooglePubSub.subscribe(request, 100.millis)
```

### DIFF
--- a/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/javadsl/GooglePubSub.scala
+++ b/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/javadsl/GooglePubSub.scala
@@ -76,7 +76,7 @@ object GooglePubSub {
    * @param request the subscription FQRS field is mandatory for the request
    * @param pollInterval time between PullRequest messages are being sent
    */
-  def subscribe(
+  def subscribePolling(
       request: PullRequest,
       pollInterval: Duration
   ): Source[ReceivedMessage, CompletableFuture[Cancellable]] =
--- a/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/javadsl/GooglePubSub.scala
+++ b/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/javadsl/GooglePubSub.scala
@@ -34,7 +34,7 @@ object GooglePubSub {
       .mapMaterializedValue(japiFunction(_ => NotUsed))
 
   /**
-   * Create a source that emits messages for a given subscription.
+   * Create a source that emits messages for a given subscription using a StreamingPullRequest.
    *
    * The materialized value can be used to cancel the source.
    *
@@ -63,6 +63,34 @@ object GooglePubSub {
               )
           )
           .mapConcat(japiFunction(_.getReceivedMessagesList))
+          .mapMaterializedValue(japiFunction(_ => cancellable))
+      }
+      .mapMaterializedValue(japiFunction(flattenCs))
+      .mapMaterializedValue(japiFunction(_.toCompletableFuture))
+
+  /**
+   * Create a source that emits messages for a given subscription using a synchronous PullRequest.
+   *
+   * The materialized value can be used to cancel the source.
+   *
+   * @param request the subscription FQRS field is mandatory for the request
+   * @param pollInterval time between PullRequest messages are being sent
+   */
+  def subscribe(
+      request: PullRequest,
+      pollInterval: Duration
+  ): Source[ReceivedMessage, CompletableFuture[Cancellable]] =
+    Source
+      .setup { (mat, attr) =>
+        val cancellable = new CompletableFuture[Cancellable]()
+
+        val client = subscriber(mat, attr).client
+
+        Source
+          .tick(pollInterval, pollInterval, request)
+          .mapAsync(1, client.pull)
+          .mapConcat(japiFunction(_.getReceivedMessagesList))
+          .mapMaterializedValue(japiFunction(cancellable.complete))
           .mapMaterializedValue(japiFunction(_ => cancellable))
       }
       .mapMaterializedValue(japiFunction(flattenCs))
--- a/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/javadsl/GooglePubSub.scala
+++ b/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/javadsl/GooglePubSub.scala
@@ -58,7 +58,7 @@ object GooglePubSub {
               .single(request)
               .concat(
                 Source
-                  .tick(pollInterval, pollInterval, subsequentRequest)
+                  .tick(Duration.ZERO, pollInterval, subsequentRequest)
                   .mapMaterializedValue(japiFunction(cancellable.complete))
               )
           )
@@ -87,7 +87,7 @@ object GooglePubSub {
         val client = subscriber(mat, attr).client
 
         Source
-          .tick(pollInterval, pollInterval, request)
+          .tick(Duration.ZERO, pollInterval, request)
           .mapAsync(1, client.pull)
           .mapConcat(japiFunction(_.getReceivedMessagesList))
           .mapMaterializedValue(japiFunction(cancellable.complete))
--- a/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/scaladsl/GooglePubSub.scala
+++ b/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/scaladsl/GooglePubSub.scala
@@ -77,7 +77,7 @@ object GooglePubSub {
    * @param request the subscription FQRS field is mandatory for the request
    * @param pollInterval time between PullRequest messages are being sent
    */
-  def subscribe(
+  def subscribePolling(
       request: PullRequest,
       pollInterval: FiniteDuration
   ): Source[ReceivedMessage, Future[Cancellable]] =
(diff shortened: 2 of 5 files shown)


google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/scaladsl/GooglePubSub.scala
```diff
@@ -69,6 +69,31 @@ object GooglePubSub {
...
+      pollInterval: FiniteDuration
+  ): Source[ReceivedMessage, Future[Cancellable]] =
+    Source
+      .setup { (mat, attr) =>
+        val cancellable = Promise[Cancellable]
+        val client = subscriber(mat, attr).client
+        Source
+          .tick(0.seconds, pollInterval, request)
+          .mapMaterializedValue(cancellable.success)
+          .mapAsync(1)(client.pull(_))
```
seglo: It seems that this is the only reason this would be "blocking" (process one request/response at a time). Could this value be greater than 1?  Maybe this could be a user parameter as well.

mdibaiee: Hmm, in case the interval between pulls is less than the amount of time it takes to receive the messages (which is unlikely I would say), this can be blocking yes.

Just to be sure and flexible we can make this a user parameter with a default of 1.

jrudolph: Couldn't you just increase the number of `max_messages` instead? And maybe add a `buffer(1)` afterwards so you always already fetch the next batch while the previous batch is still processed by the `mapConcat`.

mdibaiee: @jrudolph right, I'm not sure if there is any case that parallelism would help. If you need more messages, increase maxMessages, if you need more time between messages, increase interval, etc.

Do you think I should revert this last piece then?

seglo: Yes, I think @jrudolph's idea is better :).  Do you want to revert the parallelism param and implement that instead?

mdibaiee: @seglo @jrudolph  Now that I think about it, if we have a `buffer(1)`, and someone sets an interval to N, and their ack deadline is N or close to N, then they will continuously miss their ack deadlines because the buffered message is passing its lifetime in the buffer.

PubSub is tricky because of the ack deadline, so holding onto messages is not free, it can be costly and problematic, that's why if we want to have a buffer, I think it must be configurable and not a constant. One might not want to buffer anything to avoid a message's ack deadline being passed by.

This might be an edge case but still I think it's worth avoiding it. cc

mdibaiee: Can't users just add a buffer on top of the source themselves? Wouldn't that work? :thinking:

seglo: > someone sets an interval to N, and their ack deadline is N or close to N, then they will continuously miss their ack deadlines because the buffered message is passing its lifetime in the buffer.

Wouldn't that be the case regardless of the buffer?

As long as there's demand downstream then elements will continue to move through the buffer to downstream stages.

mdibaiee: To explain the difference I see in the case of a buffer and the case without the buffer better, consider this example:

I want to pull messages in intervals of N seconds and processing time for a batch is is P. Imagine there is no parallelism in my stream, I can only process one item at a time.

_When I use the word "lifetime" I mean the message has been received from PubSub, so there is a "lease" on it, and unless it is acknowledged in time it will be replied._

In the case without buffer, this is what the timeline looks like:
First batch comes at T = 0, it takes P seconds to process it.
Batch 2 comes at T = N, it takes P seconds to process it. If P > N, then batch 2 will not be processed for P - N seconds because it's waiting for the first batch to be processed. So it will spend P - N of its lifetime waiting to be processed. As new batches come in continuously this time can grow if there is no backpressure, but if backpressure works correctly at some point we should skip a few intervals to make sure we can process more.

In case of with buffer:
First batch comes at T = 0 it takes P seconds to process it.
Second batch comes at T = 0 (almost, there is a small latency of receiving the first batch) and is saved in buffer. This batch will start being processed at T = P. That means this batch will be waiting and spending its "lifetime" of P seconds before being processed.
Third batch comes at T = N. It will be processed at 2P, it will spend 2P - N of its lifetime waiting to be processed.

In this case P > (P - N) and (2P - N) > (P - N), so it's more likely for the case with buffer to have messages exceeding their deadline.

This is an edge case but it's possible anyhow and given our experience with buffers causing issues with acknowledge deadlines, I would prefer to have the ability to turn off such buffer.

seglo: Got it. Thanks for the excellent analysis to describe the problem! I agree, let the buffer be a user concern. If you're willing you may want to add a small section to the docs describing the consequences of buffering.

jrudolph: That's the usual trade-off. 

On one hand, you want some buffer to hide some latency in the critical path.
`max_messages` is already such a kind of buffer that amortizes the pull latency to the pubsub over `max_messages` (one pull per `max_messages`). But that still leaves you with that one pull latency. If you add another buffer on top that you even hide that remaining latency. I.e. for `n` messages

 * without any buffer and `max_messages = 1` you have `n * pull latency` (O(n) high constant factor)
 * with `max_messages`, you have `n / max_messages * pull latency` (O(n) with low constant factor, or practically ~ O(1) if `max_messages` is high enough)
 * with the extra buffer, you only have the very first initial latency, `pull_latency` (O(1))

On the other hand, you want to have buffers as small as possible because as you say you want to minimize the time spent uselessly in buffers.

It's a bit harder for users to add that buffer because it will be after `mapConcat` so the size of the buffer must be manually adopted to scale with `max_messages` you chose.

mdibaiee: @jrudolph That's right, do you think by documenting this we will be able to help users decide whether they need a buffer or not, and if they do, how to add it? (i.e. set the buffer size to be the same as max_messages if you usually receive max messages and not less).

google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/scaladsl/GooglePubSub.scala
```diff
@@ -69,6 +69,31 @@ object GooglePubSub {
...
 
+  /**
+   * Create a source that emits messages for a given subscription using a synchronous PullRequest.
+   *
+   * The materialized value can be used to cancel the source.
+   *
+   * @param request the subscription FQRS field is mandatory for the request
+   * @param pollInterval time between PullRequest messages are being sent
+   */
+  def subscribe(
```
ennru: I believe it would be better to have a specific name and not use overloading. `subscribePull`, `subscribePolling`?

mdibaiee: @ennru okay then, `subscribeStream` and `subscribePoll`?

mdibaiee: And what about backward compatibility? Do we want to keep `subscribe` as it is with a deprecation note?

ennru: No, just pick a specific name for the new method.

google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/javadsl/GooglePubSub.scala
```diff
@@ -68,6 +68,34 @@ object GooglePubSub {
...
+      pollInterval: Duration
+  ): Source[ReceivedMessage, CompletableFuture[Cancellable]] =
+    Source
+      .setup { (mat, attr) =>
+        val cancellable = new CompletableFuture[Cancellable]()
+
+        val client = subscriber(mat, attr).client
+
+        Source
+          .tick(pollInterval, pollInterval, request)
```
ennru:
```suggestion
          .tick(0.seconds, pollInterval, request)
```

Can you please fix this in `subscribe` as well.

Issue #2126, opened 2020-02-04
Ftp*StageTest: move/remove failed


raboof:
https://travis-ci.com/akka/alpakka/jobs/283306069?utm_medium=notification&utm_source=email

raboof:
again https://travis-ci.com/akka/alpakka/jobs/286924274

ennru:
https://travis-ci.com/akka/alpakka/jobs/287176318#L1328 in #2147

ennru:
The new check from #2154 didn't give any new insights
https://travis-ci.com/akka/alpakka/jobs/288729882#L613

raboof:
still happening https://travis-ci.com/akka/alpakka/jobs/292955993?utm_medium=notification&utm_source=email (this time `akka.stream.alpakka.ftp.FtpsWithProxyStageTest.move`)

ennru:
https://travis-ci.com/github/akka/alpakka/jobs/298331200#L1381 in #2201 on Scala 2.13

ennru:
https://travis-ci.com/github/akka/alpakka/jobs/299822233#L1378 in #2198 on Scala 2.13

ennru:
Looks like the FTP server gave up on us

```
[error] Test akka.stream.alpakka.ftp.FtpWithProxyStageTest.remove failed: java.lang.AssertionError: expected:<IOResult(1,Success(Done))> but was:<IOResult(0,Failure(org.apache.commons.net.ftp.FTPConnectionClosedException: FTP response 421 received.  Server closed connection.))>, took 0.08 sec
[error]     at akka.stream.alpakka.ftp.CommonFtpStageTest.remove(CommonFtpStageTest.java:125)
[error]     at akka.stream.alpakka.ftp.FtpWithProxyStageTest.remove(FtpWithProxyStageTest.java:48)
[error]     ...
```
https://travis-ci.com/github/akka/alpakka/jobs/328162262#L1384 in #2286

raboof:
again in https://travis-ci.com/github/akka/alpakka/jobs/396859271

ennru:
With #2444 these tests are now ignored.

PR #2173, opened 2020-03-02
FTP: Harden move/remove tests

raboof:
Refs #2126

More meaningfor IOResult, check whether the file is available from the
source before removing/moving

### DIFF
--- a/ftp/src/test/scala/akka/stream/alpakka/ftp/CommonFtpStageSpec.scala
+++ b/ftp/src/test/scala/akka/stream/alpakka/ftp/CommonFtpStageSpec.scala
@@ -342,6 +342,10 @@ trait CommonFtpStageSpec extends BaseSpec with Eventually {
 
       val source = listFiles("/")
 
+      eventually {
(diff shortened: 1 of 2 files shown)


ftp/src/test/scala/akka/stream/alpakka/ftp/CommonFtpStageSpec.scala
```diff
@@ -342,6 +342,10 @@ trait CommonFtpStageSpec extends BaseSpec with Eventually {
 
       val source = listFiles("/")
 
+      eventually {
+        source.map(_.name).runWith(Sink.head).futureValue shouldBe fileName
+      }
```
ennru: A follow-up to #2154, did you see any indications this a problem? I didn't notice an improvement for the Java tests, but still good to make sure this part is correct.

raboof: I saw this go wrong on my machine while dealing with some permissions problems, but it's pretty likely that that doesn't have anything to do with the failures on travis. Still, good to take it out of the equation.

PR #2176, opened 2020-03-02
GCS - Owner entityId should be optional

francisdb:
Another field that should be optional
Somehow we ended up with main.scala.akka.stream... packages

References #2163

PR #2177, opened 2020-03-03
Google Cloud Storage: fix package names

ennru:
Something went wrong when these classes were created and their package names got the `main.scala` prefix.


seglo:
Oops, thanks.

francisdb:
was already fixed in #2176 but rebased that one

PR #2180, opened 2020-03-09
Add module for BigQuery Storage API

aperkins1310:
# Purpose
This adds support for the [BigQuery Storage API](https://cloud.google.com/bigquery/docs/reference/storage) which now seems to be the preferred method of reading data from BigQuery, due to its built in streaming support, and support for parallel reads. It doesn't offer any resource management capabilities however.

This PR references #2106, I'd probably see it as an alternative rather than a replacement, in a similar way to the PubSub HTTP/gRPC modules.

# Background Context 
It is quite similar in setup to the PubSub gRPC module, and similarly makes use of akka-gRPC to reduce the complexity in this module. A lot of the auth and configuration is very similar, and could be shared if there was a need/want for a base module.

In order to test this I've created an implementation of the gRPC server, which is not the most thorough, but at least allows local tests with a reasonable representation of the real behaviour, and an exact representation of the API format. I've currently only added support for Avro GenericRecords as a starting point, but Apache Arrow is also supported by the API itself, and I would imagine you could extend this to also support Avro Specific Records for generated classes in the future.


project/Dependencies.scala
```diff
@@ -215,6 +215,19 @@ object Dependencies {
...
+  val GoogleBigQueryStorage = Seq(
+    // see Akka gRPC version in plugins.sbt
+    libraryDependencies ++= Seq(
+        // https://github.com/googleapis/java-bigquerystorage/tree/master/proto-google-cloud-bigquerystorage-v1
+        "com.google.api.grpc" % "proto-google-cloud-bigquerystorage-v1" % "0.90.0" % "protobuf", // ApacheV2
+        "io.grpc" % "grpc-auth" % "1.25.0", // ApacheV2
+        "com.google.auth" % "google-auth-library-oauth2-http" % "0.19.0", // BSD 3-clause
+        // pull in Akka Discovery for our Akka version
+        "com.typesafe.akka" %% "akka-discovery" % AkkaVersion,
+        "org.apache.avro" % "avro" % "1.9.2"
```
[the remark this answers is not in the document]
aperkins1310: Yep so you can ask for either Avro or Arrow DataFormats from BigQuery - maybe I'm trying to be "too helpful" here and deserialise into GenericRecords as well, as the google Java API just returns the ReadStream response and lets the user deserialise the rows. My idea was to start with just supporting Avro GenericRecords, and then to extend from there to support Apache Arrow, and then potentially SpecificRecords, maybe via a typeclass, but I'm not too sure whether that's required/wanted at all, so just started with the use case I know we have need for

DanieleSassoli:
Any updates on this?

docs/src/main/paradox/google-cloud-bigquery-storage.md
```diff
@@ -0,0 +1,146 @@
+# Google Cloud BigQuery Storage
+
+@@@ note
+The BigQuery Storage API offers fast access to BigQuery-managed storage using an [rpc-based](https://cloud.google.com/bigquery/docs/reference/storage/rpc) protocol.
+It is seen as an improvement over the REST API, and bulk data `extract` jobs for accessing BigQuery-managed table data, but doesn't offer any functionality around managing BigQuery resources.
+Further information at the official [Google Cloud documentation website](https://cloud.google.com/bigquery/docs/reference/storage).
+@@@
+
+This connector communicates to the BigQuery Storage API via the gRPC protocol. The integration between Akka Stream and gRPC is handled by the
+[Akka gRPC library](https://github.com/akka/akka-grpc). Currently this connector only supports returning each row as an Avro GenericRecord.
```
ennru, line 10:
```suggestion
@extref:[Akka gRPC library](akka-grpc:/). Currently, this connector only supports returning each row as an Avro GenericRecord.
```

docs/src/main/paradox/google-cloud-bigquery-storage.md
```diff
@@ -0,0 +1,146 @@
...
+Add the Java agent to the runtime configuration.
+
+build.sbt
+:   ```scala
+    javaAgents += "org.mortbay.jetty.alpn" % "jetty-alpn-agent" % "2.0.9"
+    ```
+
+### Gradle
+
+When using JDK 8: Configure your project to use the Java agent for ALPN and add `-javaagent:...` to your startup scripts as described in the @extref:[Akka gRPC documentation](akka-grpc:/buildtools/gradle.html#starting-your-akka-grpc-server-from-gradle).
```
ennru, line 72: This is not required anymore (since JDK 8 update 252).

google-cloud-bigquery-storage/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/storage/BigQueryStorageSettings.scala
```diff
@@ -0,0 +1,118 @@
...
+   */
+  def withCallCredentials(callCredentials: CallCredentials): BigQueryStorageSettings =
+    copy(callCredentials = Some(callCredentials))
+
+  private def copy(host: String = host,
+                   port: Int = port,
+                   rootCa: Option[String] = rootCa,
+                   callCredentials: Option[CallCredentials] = callCredentials) =
+    new BigQueryStorageSettings(host, port, rootCa, callCredentials)
+}
```
ennru, line 49: add a `toString`

google-cloud-bigquery-storage/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/storage/BigQueryStorageSettings.scala
```diff
@@ -0,0 +1,118 @@
...
+
+    Seq(setRootCa, setCallCredentials).foldLeft(bigQueryConfig) {
+      case (c, f) => f(c)
+    }
+  }
+
+  /**
+   * Create settings from ActorSystem's config.
+   */
+  def apply(system: ActorSystem): BigQueryStorageSettings =
```
ennru, line 91: Use `ClassicActorSystemProvider` to be compatible with both the classic and the typed APIs.

google-cloud-bigquery-storage/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/storage/javadsl/GoogleBigQueryStorage.scala
```diff
@@ -0,0 +1,169 @@
...
+      .getOrElse(GrpcBigQueryStorageReaderExt()(mat.system).reader)
+
+  /**
+   * Helper for creating akka.japi.function.Function instances from Scala
+   * functions as Scala 2.11 does not know about SAMs.
+   */
+  private def japiFunction[A, B](f: A => B): akka.japi.function.Function[A, B] =
+    new akka.japi.function.Function[A, B]() {
+      override def apply(a: A): B = f(a)
+    }
```
ennru, line 168: Scala 2.11 is not necessary for new modules.

google-cloud-bigquery-storage/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/storage/javadsl/GoogleBigQueryStorage.scala
```diff
@@ -0,0 +1,169 @@
...
+  private val RequestParamsHeader = "x-goog-request-params"
+
+  private[this] def read(
+      projectId: String,
+      datasetId: String,
+      tableId: String,
+      readOptions: Option[TableReadOptions],
+      maxNumStreams: Int
+  ): Source[Source[GenericRecord, NotUsed], CompletionStage[NotUsed]] =
+    Source.setup[Source[GenericRecord, NotUsed], NotUsed] { (mat: ActorMaterializer, attr: Attributes) =>
```
ennru, line 101: What is the reason for a duplication of the implementation between scaladsl and javadsl?
In most cases, it should be possible to use the scaladsl implementation and amend it with conversions to they Java counterparts.

ennru (commented):
Sorry for leaving this hanging for so long. The Akka Team has had different priorities recently.
It would be great if the community engaged more with reviewing each other's PRs...

I had a quick read-through, some parts can be removed as other libraries have evolved.

docs/src/main/paradox/google-cloud-bigquery-storage.md
ennru: I just merged https://github.com/akka/alpakka/pull/2312 which upgrades Akka gRPC and the Google Pub/Sub gRPC docs with this info.

gkatzioura:
Hi! I am thinking to pick this up (mainly because I have to implement this functionality for a project I work for).
Will go for it if none else is interested.

ennru:
Please go ahead. I can't promise snappy reviewing at this time either, I'm afraid.

seglo:
@aperkins1310 @gkatzioura Are either of you interested in continuing this PR? 

On a related note, a new Google BigQuery connector is being written in #2548 if you're looking for inspiration, but AFAIK this is meant to be a separate connector to a separate BigQuery API.

Tagging @armanbilge for his thoughts.

gkatzioura:
Hi @seglo! I am happy to start looking at it. Will also check #2548 . If there is anything extra please let me know!

seglo:
That's great @gkatzioura! Look forward to seeing this effort moved forward. LMK if you need anything.

armanbilge:
Had a look through this, this is definitely an excellent complement to the BigQuery connector! A couple thoughts/questions:

1. I noticed that this connector adds dependencies to Google's Java APIs, isn't it preferable to avoid this? Not sure if its necessitated by gRPC stuff. The forthcoming `alpakka-google-auth` module could help avoid the dependency to `google-auth-library-oauth2-http`.
2. IMO the current API is a little too opinionated, ideally it should be up to the user whether to use Avro or Arrow and their choice of library implementation (e.g. [avro4s](https://github.com/sksamuel/avro4s)). We could keep the [AvroDecoder](https://github.com/aperkins1310/alpakka/blob/bigquery-storage/google-cloud-bigquery-storage/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/storage/impl/AvroDecoder.scala) as an optional utility method/`Flow` and avoid the dependency by marking it as `provided`.
3. Not quite sure how to implement it, but would be awesome to add a [`split(fraction: Double)`](https://cloud.google.com/bigquery/docs/reference/storage/rpc/google.cloud.bigquery.storage.v1#google.cloud.bigquery.storage.v1.BigQueryRead.SplitReadStream) method to `Source` via an implicit conversion.

aperkins1310:
Sorry guys I completely forgot about this, we have way too many GitHub notifications at work so these get a bit lost. Let me know if you need a hand @gkatzioura, but looks like you've got it all in hand! I agree with all the comments above 🙂

seglo:
I'm going to close this PR in favour of @gkatzioura's #2619

PR #2183, opened 2020-03-10
Build: Mima 0.7.0 and sbt-scalafmt 2.3.2

ennru:


seglo:
All the HBase tests failed with a timeout.  I don't think it's related, but I'll re-run the tests.

https://github.com/akka/alpakka/issues/2185

seglo:
It happened in master too.

s3/src/main/mima-filters/2.0.0-M3.backwards.excludes
```diff
@@ -0,0 +1,2 @@
+# MiMa upgrade
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.s3.S3Settings.this")
```
raboof: So this is a binary incompatibility that wasn't detected on 0.6.4, but which is OK since we're starting a new major version anyway? Might be better to explain the change in the comment rather than just 'mima upgrade', but doesn't matter much since we're not promising binary compatibility here anyway.

ennru: You're right. (This one is even `private[alpakka]`, so it is not an API change anyway.)

PR #2186, opened 2020-03-10
Build: Use stage job name instead of DIR var on travis

seglo:
Alpakka stage jobs are hard to read now that additional env vars push out the `DIR` env var in the env var field of jobs.  We can use job name instead of `DIR`.


seglo:
We could also update the docker-compose to match the job names and reduce complexity further.  For jobs that require a container, just add an env var `DOCKER=true`.

Issue #2100, opened 2020-01-17
Mongo API: replaceOne


ovu:
Hi all,

I am using the version 2.0.0-M2 and I realized that updateOne can update just one attribute in the collection in the Sink.

Here I have two questions:
1. Why updateOne does not support a list of attributes to be updated? 
2. I am missing is the replaceOne method that is available in the scala driver. Are you planning to implement it?

Thanks for your help!

ovu:
I realized that actully the update of several attributes can be done in the following way:
        val updates = new BasicDBObject()
        updates.append("attr1", "New value")
        updates.append("attr2", "New value")
        updates.append("attr3", "New value"s)
        updates.append("attr4", "New value")
Sink.updateOne(filter, updates)

That was the answer to my first question. 
Regards

seglo:
Hi @ovu.  I don't think there's a particular reason that a wrapper fir `replaceOne` has not been implemented.  Would you be interested in contributing it?

jewertow:
@seglo I will implement this method.

seglo:
Thanks @jewertow!

jewertow:
@ovu
You can simply update multiple attributes using the builder method `Updates.combine`. For example:
```scala
Source(testRange).map(
        i =>
          DocumentUpdate(filter = Filters.eq("_id", i),
                         update = Updates.combine(Updates.set("attr1", "new value 1"), Updates.set("attr2", "new value 2")))
      )
```

ennru:
Implemented with #2187

PR #2187, opened 2020-03-10
MongoDB: replaceOne operation

jewertow:
References #2100

### DIFF
--- a/mongodb/src/test/scala/docs/scaladsl/MongoSinkSpec.scala
+++ b/mongodb/src/test/scala/docs/scaladsl/MongoSinkSpec.scala
@@ -49,15 +49,18 @@ class MongoSinkSpec
   private val db = client.getDatabase("MongoSinkSpec").withCodecRegistry(codecRegistry)
   private val numbersColl: MongoCollection[Number] =
     db.getCollection("numbersSink", classOf[Number]).withCodecRegistry(codecRegistry)
+  private val numbersDocumentColl = db.getCollection("numbersSink")
   private val domainObjectsColl: MongoCollection[DomainObject] =
     db.getCollection("domainObjectsSink", classOf[DomainObject]).withCodecRegistry(codecRegistry)
-  private val numbersDocumentColl = db.getCollection("numbersSink")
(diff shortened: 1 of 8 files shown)


mongodb/src/test/scala/docs/scaladsl/MongoSinkSpec.scala
```diff
@@ -204,6 +217,29 @@ class MongoSinkSpec
...
+    "replace with replaceOne and codec support" in assertAllStagesStopped {
+      insertDomainObjectsRange()
+      val updatedObjects =
+        testRange.map(i => DomainObject(i, s"updated-first-property-$i", s"updated-second-property-$i"))
+
+      // #replace-one
+      val source = Source(testRange).map(
+        i =>
+          DocumentReplace[DomainObject](
+            filter = Filters.eq("_id", i),
```
jewertow: In case of `replaceOne` such filter is probably the only one filter that makes sense, so it would be nice to have some predefined filter which wraps this expression. It would be something like:
```
object FilterId {
  def apply[T](id: T): Bson = Filters.eq("_id", id)
}
```
What do you think?

seglo: I would guess that's the usual use case as well, but the current `updateOne` or `replaceOne` APIs still expect a generic filter.  This API should be consistent.

jewertow:
@seglo what do you think about using in tests and docs object `DomainObject`(https://github.com/akka/alpakka/pull/2187/files#diff-670af5a1ee8a19bd56d7ed678837ecc6) instead of the existing `Number`.
In case of `replaceOne` object `Number` is useless. 
Do you want to have all tests and examples based on `DomainObject` or only `replaceOne`?

jewertow:
Ok, the implementation is done. You can review the code.

seglo:
@jewertow Thanks. I'm on vacation ATM, but @ennru or someone else from the team will take a look next week.

ennru:
Thank you for another great contribution!

Issue #2171, opened 2020-02-27
AWS S3: Virtual host style requests fails witch custom endpoint (Minio)


WojciechMazur:
Version 2.0.M3

When using virtual-host-style (which is now enabled by default) resulting endpoint does not include bucket name.
Expected request URI: http://some-bucket-name.minio-domain/some-key
Actual request URI: http://minio-domain/some-key

Bug is probably created in  `akka.stream.alpakka.s3.impl.HttpRequests` `requestAuthority` function.
With current logic if endpoint is specified in config (when using minio it is always specified) then host is not enriched by bucket sub-domain.

Also it would be better if logging about deprecation would appear only 1 instead of logging in each usage.

seglo:
Hi @WojciechMazur. Can you provide some more context about what API you're calling and the configuration/parameters.  There's a test in `HttpRequestsSpec` that asserts the request is using hostname style.

https://github.com/akka/alpakka/blob/master/s3/src/test/scala/akka/stream/alpakka/s3/impl/HttpRequestsSpec.scala#L81

WojciechMazur:
I've prepared spec which would fail in this example. 
In my project I'm not using AWS S3, but Minio inside Kubernetes cluster. In this case I'm providing as endpoint for S3 node/service address.  

```scala
    it should "create checkIfExits bucket request with custom endpoint" in {
    implicit val settings: S3Settings = getSettings().withEndpointUrl("http://custom.s3.storage:9000")

    val request: HttpRequest = HttpRequests.bucketManagementRequest(location, method = HttpMethods.HEAD)

    //Date is added by akka by default
    request.uri.authority.host.toString should equal("bucket.custom.s3.storage")
    request.uri.authority.port shouldEqual 9000
    request.entity.contentLengthOption should equal(Some(0))
    request.uri.queryString() should equal(None)
    request.method should equal(HttpMethods.HEAD)
  }
```

seglo:
Yes, when the endpoint URL is explicitly defined it will use it as-is.

https://github.com/akka/alpakka/blob/master/s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala#L181

If you encode the bucket into the endpoint URL itself everything seems to ok.

```
  it should "create checkIfExits bucket request with custom endpoint" in {
    implicit val settings: S3Settings = getSettings().withEndpointUrl("http://bucket.custom.s3.storage:9000")

[...]

Issue #2139, opened 2020-02-10
Use more explicit settings for S3 operations


kag0:
It would be nice if the s3 connector didn't default to the `DefaultAWSCredentialsProviderChain` and instead took an implicit `S3Settings` for each operation. Similar to how standard library operations take an implicit `ExecutionContext`, users could simply import something like `akka.stream.alpakka.s3.S3Settings.implicit.defaultChain` if they want the default. 

If not, this makes it easier to pass custom settings like `S3.download(bucket, bucketKey)(customSettings)` or define their own implicit (in my case, settings from `application.conf`.

Either way, an explicit action makes it more clear what's happening from reading the code without having to go to the documentation.

ennru:
Did you notice the documentation for passing explicit `S3Settings`?
https://doc.akka.io/docs/alpakka/current/s3.html#apply-s3-settings-to-a-part-of-the-stream

So the default is to use the settings specified in the config for all Alpakka S3 actions, but you can choose to use different settings when you need to.

kag0:
I did see that. What I'm suggesting is that that default not be applied automatically.
For different use cases, the change would be like:
**Default settings desired all the time:**
    Now: do nothing, defaults used automatically
    After: add an import for the defaults
**Different settings desired throughout:**
    Now: add `withAttributes` to every call, but if you forget one then the code will become incorrect
    After: pass settings to each call in second parameter list, if you forget one the code won't compile

ennru:
The current API (introduced for Alpakka 1.0) assumes that most users would just have one setup for S3 which can be specified in `application.conf`.

PR #2193, opened 2020-03-12
AWS S3: virtual-host-style access with endpoint URL

ennru:
## Purpose 

Enable use of DNS-style access while setting an endpoint URL so to use other S3 compatible solutions such as [MinIO](https://min.io/).

Introduce a new flag `path-style-access-warning` which can be used to switch off the logged warning when path-style access is enabled.

References #2171

## Solution

Support a placeholder `{bucket}` in the `endpoint-url` setting which is replaced with the bucket name.

The second commit makes use of Akka HTTP `ClientTransport` to force requests to `localhost` even if the request authority is something different, to allow the dynamic hostnames to contain the bucket names. This affects the `S3IntegrationSpec` quite significantly, as now all requests need to get the `S3Settings` passed explicitly. (This shows the motivation for #2139.)

## Background

As #2099 (issue #1679) switched the default bucket access back to virtual host-style, non-AWS endpoints did not work with Alpakka Kafka by default anymore.

### DIFF
--- a/s3/src/main/mima-filters/2.0.0-M3.backwards.excludes/PR2183
+++ b/s3/src/main/mima-filters/2.0.0-M3.backwards.excludes/PR2183
@@ -1,4 +1,4 @@
-# MiMa upgrade
+# MiMa upgrade shows new error, method is private
 ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.s3.S3Settings.this")
 
 # override of apply in extension with the concrete type instead of the generic type
--- a/s3/src/main/resources/reference.conf
+++ b/s3/src/main/resources/reference.conf
@@ -74,7 +74,7 @@ alpakka.s3 {
   }
 
   # DEPRECATED (since Alpakka 2.0.0)
-  # AWS S3 is going to retire path-style access, thus prefer the virtutal-host-style access
+  # AWS S3 is going to retire path-style access, thus prefer the virtual-host-style access
   # https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/
   # To switch off the logged warning: use `force`
   path-style-access = false
--- a/s3/src/main/resources/reference.conf
+++ b/s3/src/main/resources/reference.conf
@@ -18,6 +18,7 @@ alpakka.s3 {
 
   # An address of a proxy that will be used for all connections using HTTP CONNECT tunnel.
   # forward-proxy {
+  #   scheme = "https"
   #   host = "proxy"
   #   port = 8080
   #   credentials {
--- a/s3/src/main/resources/reference.conf
+++ b/s3/src/main/resources/reference.conf
@@ -77,8 +77,13 @@ alpakka.s3 {
   # AWS S3 is going to retire path-style access, thus prefer the virutal-host-style access
   # https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/
   path-style-access = false
+  # Don't use! Possibility to switch off the logged warning when using `path-style-access`.
+  path-style-access-warning = true
 
   # Custom endpoint url, used for alternate s3 implementations
+  # To enable DNS-style access with Alpakka S3 use the placeholder `{bucket}` in the URL
+  # eg. enpoint-url = "http://{bucket}.s3minio.alpakka:9000"
+  #
   # endpoint-url = null
 
   # Which version of the list bucket api to use. Set to 1 to use the old style version 1 API.
--- a/s3/src/main/resources/reference.conf
+++ b/s3/src/main/resources/reference.conf
@@ -74,15 +74,14 @@ alpakka.s3 {
   }
 
   # DEPRECATED (since Alpakka 2.0.0)
-  # AWS S3 is going to retire path-style access, thus prefer the virutal-host-style access
+  # AWS S3 is going to retire path-style access, thus prefer the virtutal-host-style access
   # https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/
+  # To switch off the logged warning: use `force`
   path-style-access = false
-  # Don't use! Possibility to switch off the logged warning when using `path-style-access`.
-  path-style-access-warning = true
 
   # Custom endpoint url, used for alternate s3 implementations
-  # To enable DNS-style access with Alpakka S3 use the placeholder `{bucket}` in the URL
-  # eg. enpoint-url = "http://{bucket}.s3minio.alpakka:9000"
+  # To enable virtual-host-style access with Alpakka S3 use the placeholder `{bucket}` in the URL
+  # eg. endpoint-url = "http://{bucket}.s3minio.alpakka:9000"
   #
   # endpoint-url = null
 
--- a/s3/src/main/resources/reference.conf
+++ b/s3/src/main/resources/reference.conf
@@ -77,7 +77,11 @@ alpakka.s3 {
   # DEPRECATED (since Alpakka 2.0.0)
   # AWS S3 is going to retire path-style access, thus prefer the virtual-host-style access
   # https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/
-  # To switch off the logged warning: use `force`
+  #
+  # Possible values:
+  #  - false: use virtual-host style access
+  #  - true: use legacy path-style access, warnings will be logged
+  #  - force: use legacy path-style access, disable warnings
   path-style-access = false
 
   # Custom endpoint url, used for alternate s3 implementations
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
@@ -28,6 +28,7 @@ import scala.concurrent.{ExecutionContext, Future}
 @InternalApi private[impl] object HttpRequests {
 
   private final val log = LoggerFactory.getLogger(getClass)
+  private final val PatternBucket = "{bucket}"
 
   def listBucket(
       bucket: String,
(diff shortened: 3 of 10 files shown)


s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3IntegrationSpec.scala
```diff
@@ -156,17 +171,23 @@ trait S3IntegrationSpec
...
 
   it should "download with real credentials" in {
     val Some((source, meta)) =
-      Await.ready(S3.download(defaultRegionBucket, objectKey).runWith(Sink.head), 5.seconds).futureValue
+      Await
+        .ready(S3.download(defaultBucket, objectKey)
+                 .withAttributes(attributes)
+                 .runWith(Sink.head),
+               5.seconds)
+        .futureValue
```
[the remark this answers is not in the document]
ennru: IDK. Removed all use of `Await` now, as I touched almost all tests anyway.

s3/src/main/scala/akka/stream/alpakka/s3/settings.scala
```diff
@@ -258,6 +270,7 @@ final class S3Settings private (
     credentialsProvider = credentialsProvider,
     s3RegionProvider = s3RegionProvider,
     pathStyleAccess = pathStyleAccess,
+    pathStyleAccessWarning,
```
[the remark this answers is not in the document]
ennru: I turned it around. The paramter names do not add anything here.

s3/src/main/resources/reference.conf
```diff
@@ -77,8 +77,13 @@ alpakka.s3 {
   # AWS S3 is going to retire path-style access, thus prefer the virutal-host-style access
```
jrudolph:
```suggestion
  # AWS S3 is going to retire path-style access, thus prefer the virtual-host-style access
```

s3/src/main/resources/reference.conf
```diff
@@ -77,8 +77,13 @@ alpakka.s3 {
...
   # https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/
   path-style-access = false
+  # Don't use! Possibility to switch off the logged warning when using `path-style-access`.
```
jrudolph: A bit surprising that it starts with "Don't use!".

An alternative solution to introducing a new flag could be a new value to the old flag that signifies that the user has understood what they are doing. Like `path-style-access = force` or `path-style-access = yes-I-know-what-I-am-doing`.

s3/src/main/resources/reference.conf
```diff
@@ -77,8 +77,13 @@ alpakka.s3 {
...
+  path-style-access-warning = true
 
   # Custom endpoint url, used for alternate s3 implementations
+  # To enable DNS-style access with Alpakka S3 use the placeholder `{bucket}` in the URL
+  # eg. enpoint-url = "http://{bucket}.s3minio.alpakka:9000"
```
jrudolph:
```suggestion
  # eg. endpoint-url = "http://{bucket}.s3minio.alpakka:9000"
```

s3/src/main/resources/reference.conf
```diff
@@ -77,8 +77,13 @@ alpakka.s3 {
   # AWS S3 is going to retire path-style access, thus prefer the virutal-host-style access
   # https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/
   path-style-access = false
+  # Don't use! Possibility to switch off the logged warning when using `path-style-access`.
+  path-style-access-warning = true
 
   # Custom endpoint url, used for alternate s3 implementations
+  # To enable DNS-style access with Alpakka S3 use the placeholder `{bucket}` in the URL
```
jrudolph: Maybe use "virtual-host-style" access or whatever the S3 documentation uses consistently here.

s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
```diff
@@ -28,6 +28,7 @@ import scala.concurrent.{ExecutionContext, Future}
 @InternalApi private[impl] object HttpRequests {
 
   private final val log = LoggerFactory.getLogger(getClass)
+  private final val PatternBucket = "{bucket}"
```
jrudolph: ? :)

```suggestion
  private final val BucketPattern = "{bucket}"
```

s3/src/main/scala/akka/stream/alpakka/s3/settings.scala
```diff
@@ -115,7 +116,10 @@ object ForwardProxyCredentials {
 
 }
 
-final class ForwardProxy private (val host: String, val port: Int, val credentials: Option[ForwardProxyCredentials]) {
+final class ForwardProxy private (val host: String,
+                                  val port: Int,
+                                  val credentials: Option[ForwardProxyCredentials],
+                                  val clientTransport: Option[ClientTransport]) {
```
jrudolph: This seems like a bit weird way to configure things. In the end, host/port/credentials are one way to specify a `ClientTransport.httpsProxy` and `clientTransport` is a complete alternative. Not sure how this is used exactly and what the compatibility constraints etc are but wouldn't it be better to have a way to make those two ways mutually exclusive?

ennru: Yes, that would be better. This is mostly a hack to run the tests, but there might be real use cases. I'll have another look.

s3/src/main/mima-filters/2.0.0-M3.backwards.excludes/PR2183
```diff
@@ -1,4 +1,4 @@
-# MiMa upgrade
+# MiMa upgrade shows new error, method is private
 ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.s3.S3Settings.this")
```
jrudolph: That's a bit unexpected. I checked and it seems if you have a scala private constructor, it will usually also be private in bytecode. However, if there's an access to the constructor from the companion object, the constructor might be made public in bytecode, otherwise it couldn't be accessed.

jrudolph: In any case, it's private from Scala, so there shouldn't be any compat issues (I don't know if we even care about those currently).

s3/src/main/resources/reference.conf
```diff
@@ -18,6 +18,7 @@ alpakka.s3 {
 
   # An address of a proxy that will be used for all connections using HTTP CONNECT tunnel.
   # forward-proxy {
+  #   scheme = "https"
```
jrudolph: Might make sense to better document those options now or later.

s3/src/main/resources/reference.conf
```diff
@@ -74,11 +75,15 @@ alpakka.s3 {
   }
 
   # DEPRECATED (since Alpakka 2.0.0)
-  # AWS S3 is going to retire path-style access, thus prefer the virutal-host-style access
+  # AWS S3 is going to retire path-style access, thus prefer the virtual-host-style access
   # https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/
+  # To switch off the logged warning: use `force`
```
jrudolph: Slightly more verbose docs:

```suggestion
  # Possible values:
  #
  #  - false: use virtual-host style access
  #  - true: use legacy path-style access, warnings will be logged
  #  - force: use legacy path-style access, disable warnings
```

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -572,16 +573,32 @@ import scala.util.{Failure, Success, Try}
...
+          TCPTransport(address)
+      }
+      ConnectionPoolSettings(system).withConnectionSettings(ClientConnectionSettings(system).withTransport(transport))
+    })
 
-      ConnectionPoolSettings(system)
-        .withConnectionSettings(
-          ClientConnectionSettings(system)
-            .withTransport(transport)
+  private case class TCPTransport(address: InetSocketAddress) extends ClientTransport {
```
jrudolph: Maybe rename to `ProxyTransport` or `ChangeTargetEndpointTransport` or something.

s3/src/main/scala/akka/stream/alpakka/s3/settings.scala
```diff
@@ -333,18 +371,13 @@ object S3Settings {
...
+    val maybeForwardProxy =
+      if (c.hasPath("forward-proxy")) Some(ForwardProxy(c.getConfig("forward-proxy")))
+      else None
 
-      Option(ForwardProxy(c.getString("forward-proxy.host"), c.getInt("forward-proxy.port"), maybeCredentials))
-    } else None
-
-    val pathStyleAccess = c.getBoolean("path-style-access")
+    val (pathStyleAccess, pathStyleAccessWarning) =
+      if (c.getString("path-style-access") == "force") (true, false)
```
jrudolph: :+1: good enough

laszlovandenhoek:
@ennru now there is no way to disable the `pathStyleAccessWarning` value on `S3Settings` programmatically. That could be remedied by adding something like this:

```
  def withPathStyleAccessWarning(value: Boolean): S3Settings =
    if (pathStyleAccessWarning == value) this else copy(pathStyleAccessWarning = value)
```

Would you be open to such an addition?

FYI, I understand why there's an `@deprecation` annotation on `withPathStyleAccess`, but we do use it and intend to continue to use it, because we use an on-prem S3 implementation (Rados Gateway). So I hope this option won't go away in the future.

ennru:
Yes, I underestimated how common non-AWS use of the S3 protocol is out there. See eg. #2364 

What about introducing `withPathStyleAccessForce(Boolean)` for a deliberate developer decision to enable path-style access and disable the warning? Or does that become to freaky?

laszlovandenhoek:
I think I'd rather have a non-boolean property that rolls up all options:

`withAccessStyle(AccessStyle)`

where AccessStyle can be one of:

- `PathStyle`
- `VirtualHostStyle`
- `DefaultAccessStyle` - maybe not needed. This could be the default in `reference.conf`, otherwise `VirtualHostStyle`.

`with-path-style-access` should be unset in `reference.conf`. If it is set manually and gets a value, it takes precedence over `access-style`, but triggers a warning in the log to that effect, prompting developers to rectify their settings. They would then make a conscious choice for either `PathStyle` or `VirtualHostStyle`, and there would be no need for warnings or "forcing" or anything.

This is probably more far-reaching than you had in mind, but imagine what would happen if the need ever arose to add a fourth value next to `true`, `false` and `force`? If we had `withPathStyleAccessForce`, you would have to be consistent and add another boolean method. Or, with my proposal, you could add just one extra `AccessStyle`. To me, the latter seems to be clearer and also the more maintainable option.

laszlovandenhoek:
PR is in the mail :)

Issue #2194, opened 2020-03-13
Update APIs to be Akka classic and Akka typed compatible / hide materializer


ennru:
### Short description

Make it easier to use the new actor APIs with Alpakka and hide the `Materializer` where possible.

### Details

Akka 2.5.30 includes the `ClassicActorSystemProvider` trait so that APIs can accept both class `akka.actor.ActorSystem` and `akka.actor.typed.ActorSystem`.

The `SystemMaterializer` makes it easy to access the system-wide materializer so that users do not need to provide it implicitly or explicitly anymore.

ennru:
I went through a couple of connectors to make the integration with the new actors API easier.
Sadly, it turned out that the current Akka 2.5.30 version does contain the trait without the actor systems classes implementing it. https://github.com/akka/akka/issues/28764

There might be more places where we could offer better support for the new actors API and even hide the materializer.

PR #2195, opened 2020-03-13
Cassandra: add methods for new actors API (Akka 2.6.4)

ennru:
Complement the `CassandraSessionRegistry` extension with access methods for `ClassicActorSystemProvider` which makes it usable without changed for the new actors API's `akka.actor.typed.ActorSystem` without depending on that module.

This can't be compiled against Akka 2.6 as `ExtensionId` in that version contains the `apply(ClassicActorSystemProvider)` method so it requires an `override` modifier in `CasssandraSessionRegistry`.

Use the `SystemMaterializer` to get hold of a `Materializer`.

**Upgrades to Akka 2.6.4 and Akka 2.5.30.**

See #2194

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -11,7 +11,7 @@ object Dependencies {
   val ScalaVersions = Seq(Scala212, Scala211, Scala213).filterNot(_ == Scala211 && Nightly)
 
   val Akka25Version = "2.5.30"
-  val Akka26Version = "2.6.1"
+  val Akka26Version = "2.6.4"
   val AkkaVersion = if (Nightly) Akka26Version else Akka25Version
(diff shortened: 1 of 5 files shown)


project/Dependencies.scala
```diff
@@ -10,8 +10,8 @@ object Dependencies {
   val Scala213 = "2.13.1"
   val ScalaVersions = Seq(Scala212, Scala211, Scala213).filterNot(_ == Scala211 && Nightly)
 
-  val Akka26Version = "2.6.1"
-  val AkkaVersion = if (Nightly) Akka26Version else "2.5.27"
+  val Akka26Version = "2.6.4"
```
patriknw: not just yet

PR #2196, opened 2020-03-13
MQTT streaming: ClassicActorSystemProvider in the API

ennru:
Changes the factory methods and constructors of the `MqttSession`s to just require a `ClassicActorSystemProvider` instead of `akka.actor.ActorSystem` and `akka.stream.Materializer`.

This is possible as Alpakka MQTT Streaming is only compatible with Akka 2.6.

This makes it easier to use with the new actor APIs and hides the existence of the materializer form the users.

References #2118

### DIFF
--- a/mqtt-streaming/src/main/mima-filters/2.0.0-M3.backwards.excludes/ClassicActorSystemProvider
+++ b/mqtt-streaming/src/main/mima-filters/2.0.0-M3.backwards.excludes/ClassicActorSystemProvider
@@ -1,0 +1,11 @@
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.mqtt.streaming.javadsl.ActorMqttServerSession.this")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.mqtt.streaming.javadsl.ActorMqttServerSession.create")
+
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.mqtt.streaming.javadsl.ActorMqttClientSession.this")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.mqtt.streaming.javadsl.ActorMqttClientSession.create")
+
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.mqtt.streaming.scaladsl.ActorMqttServerSession.apply")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.mqtt.streaming.scaladsl.ActorMqttServerSession.this")
+
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.mqtt.streaming.scaladsl.ActorMqttClientSession.apply")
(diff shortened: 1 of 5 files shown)


mqtt-streaming/src/main/mima-filters/2.0.0-M3.backwards.excludes/ClassicActorSystemProvider
```diff
@@ -0,0 +1,11 @@
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.mqtt.streaming.javadsl.ActorMqttServerSession.this")
```
jrudolph: Maybe add a comment why we don't care. Do we just not care about binary compat here?

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/javadsl/MqttSession.scala
```diff
@@ -52,19 +51,19 @@ abstract class MqttClientSession extends MqttSession {
 }
 
 object ActorMqttClientSession {
-  def create(settings: MqttSessionSettings, mat: Materializer, system: ActorSystem): ActorMqttClientSession =
-    new ActorMqttClientSession(settings, mat, system)
+  def create(settings: MqttSessionSettings, system: ClassicActorSystemProvider): ActorMqttClientSession =
```
jrudolph: Is that already in anticipation that `untyped.ActorSystem extends ClassActorSystemProvider` in 2.5 or does it already work with 2.5.30?

ennru: You're right, I did this before I found out that it doesn't work as expected. https://github.com/akka/akka/issues/28764

I need to allow both `akka.actor.ActorSystem` and the provider.

ennru: No. Good that I had lunch. This connector is Akka 2.6 only.

mqtt-streaming/src/main/mima-filters/2.0.0-M3.backwards.excludes/ClassicActorSystemProvider
ennru: We are going for a major release and this connector is Akka 2.6 only now.

ennru:
I added tests that which prove by compilation that both classic and typed actor systems are accepted as implicits.

mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/javadsl/MqttSession.scala
jrudolph: Ah, that explains it!

mqtt-streaming/src/main/mima-filters/2.0.0-M3.backwards.excludes/ClassicActorSystemProvider
jrudolph:
```suggestion
# Incompatible changes for major release, now Akka 2.6 only
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.mqtt.streaming.javadsl.ActorMqttServerSession.this")
```

PR #2197, opened 2020-03-13
Couchbase: add methods for new actors API

ennru:
Complement the `CouchbaseSessionRegistry` extension with access methods for `ClassicActorSystemProvider` which makes it usable without changed for the new actors API's `akka.actor.typed.ActorSystem` without depending on that module.

This can't be compiled against Akka 2.6 as `ExtensionId` in that version contains the `apply(ClassicActorSystemProvider)` method so it requires an `override` modifier in `CasssandraSessionRegistry`.

See #2194, #2195

### DIFF
--- a/couchbase/src/main/scala/akka/stream/alpakka/couchbase/CouchbaseSessionRegistry.scala
+++ b/couchbase/src/main/scala/akka/stream/alpakka/couchbase/CouchbaseSessionRegistry.scala
@@ -7,7 +7,7 @@ package akka.stream.alpakka.couchbase
 import java.util.concurrent.CompletionStage
 import java.util.concurrent.atomic.AtomicReference
 
-import akka.actor.{ActorSystem, ExtendedActorSystem, Extension, ExtensionId, ExtensionIdProvider}
+import akka.actor.{ClassicActorSystemProvider, ExtendedActorSystem, Extension, ExtensionId, ExtensionIdProvider}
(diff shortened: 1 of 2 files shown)


couchbase/src/main/scala/akka/stream/alpakka/couchbase/CouchbaseSessionRegistry.scala
```diff
@@ -27,10 +27,30 @@ object CouchbaseSessionRegistry extends ExtensionId[CouchbaseSessionRegistry] wi
     new CouchbaseSessionRegistry(system)
 
   /**
-   * Java API: get the session registry
+   * Get the session registry with new actors API.
    */
-  override def get(system: ActorSystem): CouchbaseSessionRegistry =
-    super.get(system)
+  // This is not source compatible with Akka 2.6 as it lacks `overrride`
```
jrudolph: I think you talked about that. So, that means it will still work on the bytecode level as long as alpakka is compiled against Akka 2.5 but needs changing once Akka 2.5 support is removed?

ennru: Yes. The byte-code-level method signature is unaffected by the `override` so it works as expected with Akka 2.6.
But compilation with Akka 2.6 fails as the compiler requires the `override`.

PR #2198, opened 2020-03-13
Google Cloud Pub/Sub gRPC: hide materializer, support new actors API

ennru:
Complement the factory methods with methods for `ClassicActorSystemProvider` which makes it usable without changed for the new actors API's `akka.actor.typed.ActorSystem` without depending on that module.

This can't be compiled against Akka 2.6 as `ExtensionId` in that version contains the `get(ClassicActorSystemProvider)` method so it requires an `override`.

**Upgrades to Akka 2.5.30 and Akka gRPC 0.8.0**

See #2194, #2195, #2197

### DIFF
--- a/docs/src/main/paradox/google-cloud-pub-sub-grpc.md
+++ b/docs/src/main/paradox/google-cloud-pub-sub-grpc.md
@@ -6,8 +6,8 @@ Google Cloud Pub/Sub provides many-to-many, asynchronous messaging that decouple
 Further information at the official [Google Cloud documentation website](https://cloud.google.com/pubsub/docs/overview).
 @@@
 
-This connector communicates to Pub/Sub via the gRPC protocol. The integration between Akka Stream and gRPC is handled by the
-[Akka gRPC library](https://github.com/akka/akka-grpc). For a connector that uses HTTP for the communication, take a
+This connector communicates to Pub/Sub via the gRPC protocol. The integration between Akka Stream and gRPC is handled by
+@extref[Akka gRPC $akka-grpc.version$](akka-grpc:). For a connector that uses HTTP for the communication, take a
 look at the alternative @ref[Alpakka Google Cloud Pub/Sub](google-cloud-pub-sub.md) connector.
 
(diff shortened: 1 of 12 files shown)


google-cloud-pub-sub-grpc/src/main/mima-filters/2.0.0-M3.backwards.excludes/Akka-gRPC-upgrade
```diff
@@ -0,0 +1,11 @@
+ProblemFilters.exclude[Problem]("com.google.*")
```
jrudolph: Maybe add comment why we don't care about bin compat.

docs/src/main/paradox/google-cloud-pub-sub-grpc.md
```diff
@@ -34,6 +34,15 @@ The table below shows direct dependencies of this module and the second tab show
 
 @@dependencies { projectId="google-cloud-pub-sub-grpc" }
 
+## Binary compatibility
+
+@@@warning
+
+This connector contains code generated from Protobuf files which is bound to @extref:[Akka gRPC $akka-grpc.version$](akka-grpc:). This makes it @extref:[NOT binary-compatible](akka-grpc:/binary-compatibility.html) with later versions of Akka gRPC.
+You can not use a different version of Akka gRPC within the same JVM instance.
```
ennru: @raboof Would you add anything more to this note?

google-cloud-pub-sub-grpc/src/main/mima-filters/2.0.0-M3.backwards.excludes/Akka-gRPC-upgrade
ennru: Thanks, I added it to the docs now.

docs/src/main/paradox/google-cloud-pub-sub-grpc.md
raboof: Sounds good to me

Issue #2200, opened 2020-03-13
Cassandra: Override gremlin-core to 3.5.0


ignasi35:
In order to fix #2199 `gremlin-core` was excluded from the transitive dependencies of the cassandra java driver.

When `gremlin-core` `3.5.0` is out instead of excluding we could override the transitive dependency (`gremlin-core` `3.4.6`).

Alternatively, we can wait until the cassandra java driver depends on `gremlin-core` `3.5.0`.

ennru:
Solved for now with #2201

ignasi35:
Ah, I didn't know about https://github.com/akka/alpakka/pull/2201#issuecomment-599454639

If Grmlin is only for non-OSS features of Cassandra, then this issue which aimed to restore the transitive dependency is not required.

PR #2201, opened 2020-03-13
Cassandra: exclude gremlin-core from transitive deps

### LINKED ISSUE #2199: [Failed] Scheduled build fails on Cassandra
https://travis-ci.com/github/akka/alpakka/jobs/297794147#L1957

ignasi35:
Fixes #2199 

References #2200


ennru:
Our understanding is that the `"org.apache.tinkerpop" % "gremlin-core"` dependency used for features that are not available in the OSS version of the Cassandra Java driver.

PR #2202, opened 2020-03-15
AWS connectos: mark com.github.matsluni.aws-spi-akka-http as test

Ophirr33:
Trying to construct an aws client when depending on alpakka and the aws sdk v2 library fails due to the following exception:

> software.amazon.awssdk.core.exception.SdkClientException: Multiple HTTP implementations were found on the classpath. To avoid non-deterministic loading implementations, please explicitly provide an HTTP client via the client builders, set the software.amazon.awssdk.http.async.service.impl system property with the FQCN of the HTTP service to use as the default, or remove all but one
HTTP implementation from the classpath

Excluding the akka spi http transitive dependency fixes the problem:

```
"com.lightbend.akka" %% "akka-stream-alpakka-dynamodb" % versions.alpakka exclude("com.github.matsluni", "aws-spi-akka-http")
```

Since this library is only used in examples, I suggest that we mark it as a test dependency so that users must explicitly include the akka spi http client if they wish to use it, and only those users will have to handle this exception (by specifying the akka spi http client within the aws client builder).

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -87,7 +87,7 @@ object Dependencies {
   val AwsLambda = Seq(
     libraryDependencies ++= Seq(
         "com.typesafe.akka" %% "akka-http" % AkkaHttpVersion, // ApacheV2
-        "com.github.matsluni" %% "aws-spi-akka-http" % AwsSpiAkkaHttpVersion excludeAll // ApacheV2
+        "com.github.matsluni" %% "aws-spi-akka-http" % AwsSpiAkkaHttpVersion % Test excludeAll // ApacheV2
         (
           ExclusionRule(organization = "com.typesafe.akka")
         ),
@@ -145,7 +145,7 @@ object Dependencies {
 
   val DynamoDB = Seq(
     libraryDependencies ++= Seq(
-        "com.github.matsluni" %% "aws-spi-akka-http" % AwsSpiAkkaHttpVersion excludeAll // ApacheV2
+        "com.github.matsluni" %% "aws-spi-akka-http" % AwsSpiAkkaHttpVersion % Test excludeAll // ApacheV2
         (
           ExclusionRule(organization = "com.typesafe.akka")
         ),
@@ -304,7 +304,7 @@ object Dependencies {
   val Kinesis = Seq(
     libraryDependencies ++= Seq(
         "com.typesafe.akka" %% "akka-http" % AkkaHttpVersion, // ApacheV2
-        "com.github.matsluni" %% "aws-spi-akka-http" % AwsSpiAkkaHttpVersion excludeAll ExclusionRule(
+        "com.github.matsluni" %% "aws-spi-akka-http" % AwsSpiAkkaHttpVersion % Test excludeAll ExclusionRule(
           organization = "com.typesafe.akka"
         )
       ) ++ Seq(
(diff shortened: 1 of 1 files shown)


Ophirr33:
Is the test failure in MqttFlowSpec unrelated?

ennru:
Hi @Ophirr33 
Thank you for engaging with Alpakka. The library is not "just" used in tests, our recommendation is to use Akka HTTP as the underlying HTTP client and that is why Alpakka excludes the transitive dependency on the AWS provided HTTP client (Apache and Netty).

I don't know which version of Alpakka you use, but for the upcoming 2.0 release, I've added more details about the HTTP client: https://doc.akka.io/docs/alpakka/2.0/aws-shared-configuration.html

Ophirr33:
> our recommendation is to use Akka HTTP as the underlying HTTP client

I was not aware of that, is that recommendation documented anywhere? I looked at https://github.com/matsluni/aws-spi-akka-http#aws-akka-http-spi-implementation and noticed this statement in the README:

> This library is not production ready and early alpha. Use at your own risk.

Does that mean getting that library to a 1.0.0 release is part of the alpakka 2.0 release goals? Or at the very least to a version that doesn't have such a dire warning

> I don't know which version of Alpakka you use

Currently `2.0.0-M3`

ennru:
I value your concerns about @matsluni's library and the warning about its state.
The Alpakka AWS connectors have been using it for quite a while and I'm not aware of any reported problems.

The library is not in our control and we don't plan to suggest changes to it right now.

>  I suppose users in this scenario that want to use the default builder should make a choice of whether to exclude the akka http spi client from alpakka, or to exclude the netty and apache client from the sdk v2 dependency.

Yes. It would be great if you could suggest an improvement to the docs for that.

matsluni:
@Ophirr33 and @ennru I am the author of [aws-spi-akka-http](matsluni/aws-spi-akka-http). 
I created the readme and the mentioned warning quite some time ago, before anybody and especially alpakka decided to use it. 
As I currently don't use the library myself and don't have any production experience with it, I can't comment on any issues with it. The only issues with the library were reported shortly after alpakka decided to use, but this was solve some time ago already and since then I haven't heard anything.
So, maybe it is time now to remove this warning, as with the used connectors in alpakka we probably can be confident, that it works as expected. I am happy to remove the warning.

ennru:
Thank you @matsluni, good to hear no issues were reported on your side either.
Please go ahead and remove the warning.

Ophirr33:
>  It would be great if you could suggest an improvement to the docs for that.

I can look into doing that.

Thanks for chiming in @matsluni, glad to hear that the warning is likely stale.

> As I currently don't use the library myself and don't have any production experience with it, I can't comment on any issues with it.

If alpakka is the primary user of this library, I wonder if it makes sense for akka to lend a hand in the stewardship of it? Either by being added as a contributor, maintaining an official fork of it, or by inlining its three source files into alpakka?

ennru:
> If alpakka is the primary user of this library, I wonder if it makes sense for akka to lend a hand in the stewardship of it? Either by being added as a contributor, maintaining an official fork of it, or by inlining its three source files into alpakka?

For the time being, the Akka team has too many things to keep track of already. We'll need to continue with the current setup.

ennru:
I'll close this PR now. It would be great if you can suggest some additions to the AWS shared docs.

Issue #2054, opened 2019-12-19
Drop Akka 2.5 support


ennru:
### Short description

When is a good time for Alpakka to go on and stop supporting Akka 2.5?

### References

#624
https://github.com/akka/akka-grpc/issues/762

### Details

Alpakka dropped Akka 2.4 more than 2 years ago (#633).
Akka 2.6 was [released](https://akka.io/blog/news/2019/11/06/akka-2.6.0-released) on 2019-11-06.

ennru:
When dropping Akka 2.5, the backported copies of `RetryFlow` need to be dropped from Alpakka. (eg. #2146)

ennru:
With #2457 Akka 2.6.10 is the only version supported.
Leaving this open as Alpakka needs to leverage to improvements in Akka Streams from Akka 2.6

seglo:
Implemented in #2457

PR #2205, opened 2020-03-16
API docs: exclude MQTT streaming

### LINKED ISSUE #2203: Akka 2.6 in MQTT streaming fails Unidoc
https://github.com/akka/alpakka/pull/2118 made MQTT Streaming use Akka 2.6.

This makes Unidoc fail as it tries to compile against Akka 2.5:

```
[error] /home/travis/build/akka/alpakka/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala:510:18: value futureSource is not a member of object akka.stream.scaladsl.Source
[error] possible cause: maybe a semicolon is missing before `value futureSource'?
[error]                 .futureSource(reply.future)
[error]                  ^
[error] /home/travis/build/akka/alpakka/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/impl/ClientState.scala:532:18: value futureSource is not a member of object akka.stream.scaladsl.Source
[error] possible cause: maybe a semicolon is missing before `value futureSource'?
[error]                 .futureSource(reply.future)
[error]                  ^

```

https://travis-ci.com/github/akka/alpakka/jobs/297873059#L578

Maybe using Akka 2.6 during Unidoc is a viable solution.

ennru:
Fixes #2203

The API doc generation can be added back to Unidoc when we drop Akka 2.5 #2054

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -76,12 +76,14 @@ lazy val alpakka = project
         .filterNot(_.data.getAbsolutePath.contains("guava-28.1-android.jar"))
         .filterNot(_.data.getAbsolutePath.contains("commons-net-3.1.jar"))
         .filterNot(_.data.getAbsolutePath.contains("protobuf-java-2.6.1.jar"))
-        // Some projects require (and introduce) Akka 2.6:
-        .filterNot(_.data.getAbsolutePath.contains("akka-stream_2.12-2.5.27.jar"))
     },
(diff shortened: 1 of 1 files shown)


ennru:
Right, #2206 is the correct fix.

PR #2212, opened 2020-03-18
Unix Domain Socket: add access with new actors API

ennru:
Complement the `UnixDomainSocket` extensions with access methods for `ClassicActorSystemProvider` which makes it usable without changed for the new actors API's `akka.actor.typed.ActorSystem` without depending on that module.

This can't be compiled against Akka 2.6 as `ExtensionId` in that version contains the `apply(ClassicActorSystemProvider)` method so it requires an `override` modifier in `UnixDomainSocket`.

See #2194, #2195, #2197, #2211

### DIFF
--- a/unix-domain-socket/src/main/scala/akka/stream/alpakka/unixdomainsocket/javadsl/UnixDomainSocket.scala
+++ b/unix-domain-socket/src/main/scala/akka/stream/alpakka/unixdomainsocket/javadsl/UnixDomainSocket.scala
@@ -12,8 +12,7 @@ import java.util.concurrent.CompletionStage
 import scala.compat.java8.OptionConverters._
 import scala.compat.java8.FutureConverters._
 import akka.NotUsed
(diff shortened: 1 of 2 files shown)


unix-domain-socket/src/main/scala/akka/stream/alpakka/unixdomainsocket/javadsl/UnixDomainSocket.scala
```diff
@@ -25,7 +24,7 @@ object UnixDomainSocket extends ExtensionId[UnixDomainSocket] with ExtensionIdPr
   /**
    * Represents a prospective UnixDomainSocket server binding.
    */
-  final class ServerBinding private[akka] (delegate: ScalaUnixDomainSocket.ServerBinding) {
+  final class ServerBinding private[akka] (delegate: scaladsl.UnixDomainSocket.ServerBinding) {
```
raboof: :+1:

PR #2213, opened 2020-03-18
Google Cloud Storage: add access with new actors API

ennru:
Complement the `GCStorageExt` extension with access methods for `ClassicActorSystemProvider` which makes it usable without changed for the new actors API's `akka.actor.typed.ActorSystem` without depending on that module.

This can't be compiled against Akka 2.6 as `ExtensionId` in that version contains the `apply(ClassicActorSystemProvider)` method so it requires an `override` modifier in `GCStorageExt`.

See #2194, #2195, #2197, #2211, #2212


google-cloud-storage/src/main/mima-filters/2.0.0-M3.backwards.excludes
```diff
@@ -15,3 +15,6 @@ ProblemFilters.exclude[MissingClassProblem]("main.scala.*")
 ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.stream.alpakka.googlecloud.storage.StorageObject.*")
 ProblemFilters.exclude[IncompatibleMethTypeProblem]("akka.stream.alpakka.googlecloud.storage.StorageObject.withCustomerEncryption")
 ProblemFilters.exclude[IncompatibleMethTypeProblem]("akka.stream.alpakka.googlecloud.storage.StorageObject.withOwner")
+
+# override of apply in extension with the concrete type instead of the generic type
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.stream.alpakka.googlecloud.storage.GCStorageExt.apply")
```
jrudolph: Has this really changed? Maybe it's just https://github.com/lightbend/mima/issues/433

Issue #1840, opened 2019-07-26
Handle no subscription instead of throwing "Unsupported Content-Type"


2m:
When there is no subscription available, the following exception is logged:

```
akka.http.scaladsl.unmarshalling.Unmarshaller$UnsupportedContentTypeException: Unsupported Content-Type, supported: application/json
```

<details>
<pre>
[info] akka.http.scaladsl.unmarshalling.Unmarshaller$UnsupportedContentTypeException: Unsupported Content-Type, supported: application/json
[info] at akka.http.scaladsl.unmarshalling.Unmarshaller$UnsupportedContentTypeException$.apply(Unmarshaller.scala:158)
[info] at akka.http.scaladsl.unmarshalling.Unmarshaller$EnhancedFromEntityUnmarshaller$.$anonfun$forContentTypes$3(Unmarshaller.scala:114)
[info] at akka.http.scaladsl.unmarshalling.Unmarshaller$$anon$1.apply(Unmarshaller.scala:58)
[info] at akka.http.scaladsl.unmarshalling.Unmarshaller.$anonfun$transform$3(Unmarshaller.scala:23)
[info] at akka.http.scaladsl.unmarshalling.Unmarshaller$$anon$1.apply(Unmarshaller.scala:58)
[info] at akka.http.scaladsl.unmarshalling.Unmarshaller.$anonfun$transform$3(Unmarshaller.scala:23)
[info] at akka.http.scaladsl.unmarshalling.Unmarshaller$$anon$1.apply(Unmarshaller.scala:58)
[info] at akka.http.scaladsl.unmarshalling.LowerPriorityGenericUnmarshallers.$anonfun$messageUnmarshallerFromEntityUnmarshaller$3(GenericUnmarshallers.scala:25)
[info] at akka.http.scaladsl.unmarshalling.Unmarshaller$$anon$1.apply(Unmarshaller.scala:58)
[info] at akka.http.scaladsl.unmarshalling.Unmarshal.to(Unmarshal.scala:25)
[info] at akka.stream.alpakka.googlecloud.pubsub.impl.PubSubApi.$anonfun$pull$2(PubSubApi.scala:101)
[info] at scala.concurrent.Future.$anonfun$flatMap$1(Future.scala:307)
[info] at scala.concurrent.impl.Promise.$anonfun$transformWith$1(Promise.scala:41)
[info] at scala.concurrent.impl.CallbackRunnable.run$$original(Promise.scala:64)
[...]

PR #2217, opened 2020-03-18
Google Cloud Pub/Sub: require JSON responses

ennru:
Guard the HTTP response to be JSON before trying to deserialize it so that missing subscriptions aren't reported as marshalling problems.

References #1840


google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApi.scala
```diff
@@ -164,7 +164,7 @@ private[pubsub] trait PubSubApi {
       .mapAsync(1) {
         case Success(response) =>
           response.status match {
-            case StatusCodes.Success(_) =>
+            case StatusCodes.Success(_) if response.entity.contentType == ContentTypes.`application/json` =>
```
raboof: Maybe also include the contentType in the "Unexpected response" exception below, to help debugging when the response is rejected because the content-type is wrong?

PR #2224, opened 2020-03-22
Kinesis: more descriptive Source error log

MrRexZ:
In one of my project, I'd like to further understand easily what are the cause of the errors when using `KinesisSourceStage`, further in my downstream handler. 
When failing to fetch the Kinesis shard for example, it will now tell you more detailed error message (such as when the Kinesis HTTP server fails to respond)


ennru:
Thank you, that's very helpful.

Issue #2219, opened 2020-03-19
KinesisSourceSpec: KinesisSchedulerSource must not drop messages in case of Shard end


ennru:
```
KinesisSchedulerSource
- must not drop messages in case of Shard end *** FAILED *** (3 seconds, 750 milliseconds)
  The code passed to eventually never returned normally. Attempted 18 times over 612.41528 milliseconds. Last failure message:
  Wanted but not invoked:
  recordProcessorCheckpointer.checkpoint();
  -> at akka.stream.alpakka.kinesis.KinesisSchedulerSourceSpec$$anonfun$1$$anonfun$apply$mcV$sp$8$$anonfun$apply$8$$anon$5$$anonfun$11.apply$mcV$sp(KinesisSchedulerSourceSpec.scala:209)
  Actually, there were zero interactions with this mock.
  . (KinesisSchedulerSourceSpec.scala:208)
  org.scalatest.exceptions.TestFailedDueToTimeoutException:
  at org.scalatest.enablers.Retrying$$anon$4.tryTryAgain$2(Retrying.scala:174)
  at org.scalatest.enablers.Retrying$$anon$4.retry(Retrying.scala:192)
  at org.scalatest.concurrent.Eventually$class.eventually(Eventually.scala:402)
  at akka.stream.alpakka.kinesis.KinesisSchedulerSourceSpec.eventually(KinesisSchedulerSourceSpec.scala:43)
  at akka.stream.alpakka.kinesis.KinesisSchedulerSourceSpec$$anonfun$1$$anonfun$apply$mcV$sp$8$$anonfun$apply$8$$anon$5.<init>(KinesisSchedulerSourceSpec.scala:208)
  at akka.stream.alpakka.kinesis.KinesisSchedulerSourceSpec$$anonfun$1$$anonfun$apply$mcV$sp$8$$anonfun$apply$8.apply(KinesisSchedulerSourceSpec.scala:176)
```

https://travis-ci.com/github/akka/alpakka/jobs/299822244#L437 in #2198

ennru:
https://travis-ci.com/github/akka/alpakka/jobs/301695913#L440 in #2227

seglo:
Hopefully resolved by #2229.  I can no longer reproduce this issue locally.

PR #2229, opened 2020-03-25
Fix flicker test in Kinesis KCL

aserrallerios:
Fixes (or tries to) https://github.com/akka/alpakka/issues/2219

I cannot reproduce the error locally, can you try it @seglo ?


aserrallerios:
Sorry for my ignorance, but why not use `org.scalatest.Retries`?

seglo:
> Sorry for my ignorance, but why not use `org.scalatest.Retries`?

I use the `Repeated` trait to force many test runs to identify and fix transiently failing tests.  AFAIK there's nothing available in ScalaTest that can do that.  `Retries` is to retry a test if it fails, with some rules.

Issue #2034, opened 2019-11-26
Add tarball flow


choffmeister:
### Short description

In the same way that there already is a flow to generate a ZIP archive it would be nice to have the same for a simple (uncompressed) TAR ball.

### Details

I already have some working code taken from one of my personal projects (see https://github.com/choffmeister/alpakka/tree/tarball-flow). The implementation is pretty straight forward. When one does not need compression (which still could be applied afterwards, like `.tar` -> `.tar.gz`), it has one big benefit over the ZIP archive stream:

It does not rely on the blocking Java Input/OutputStream API, but instead it directly uses akka.util.ByteString. So it seems to be way more efficient in terms of CPU usage (in some tests of mine I found 50% usage vs 3% CPU usage when comparing uncompressed ZIP stream vs my TAR stream).

Of course this should be integrated into the already existing `Archive` class. But before I invest more time to make the code fit nicely into the existing code base, I would like to know if such a feature would be merged.

ennru:
Thank you for suggesting the work from https://github.com/akka/akka-stream-contrib/pull/176 here instead.
I think this would fit nicely in Alpakka. Please go ahead.

ennru:
Did you get to start on this? It would be great to get tar support in Alpakka.

choffmeister:
@ennru Sorry, I did completely forgot about this. Took some time to rewrite my code to hopefully fit nicely next to `Archive.zip()` (see #2241)

ennru:
Creating TAR archives is now supported with #2241.

Reading TAR archives would be great to have...

PR #2241, opened 2020-03-27
TAR archive generation

choffmeister:
This PR provides very similar functionality to `Archive.zip()`, just for uncompressed tarballs. The biggest benefit compared to using the zip flow is, that this does not rely on any blocking Java IO APIs.

References #2034


choffmeister:
Still want to improve two things:

1. Write the `tar` bin integration differently
2. Provide a `Archive.tarGz()` convenient function that wraps `Archive.tar()` with gzip compression

file/src/main/scala/akka/stream/alpakka/file/impl/archive/TarArchiveHeader.scala
```diff
@@ -0,0 +1,87 @@
...
+  private val filePathPrefix = Option(filePathSegments.init).filter(_.nonEmpty).map(_.mkString("/"))
+  private val filePathName = filePathSegments.last
+
+  require(
+    filePathPrefix.forall(fnp => fnp.length > 0 && fnp.length <= 154),
+    "File path prefix must be between 1 and 154 characters long"
+  )
+  require(filePathName.length > 0 && filePathName.length <= 99,
+          "File path name must be between 1 and 99 characters long")
+
```
ennru: This validation could be moved to `ArchiveMetadataWithSize` so that it can fail as early as possible.

file/src/test/scala/docs/scaladsl/ArchiveSpec.scala
```diff
@@ -116,6 +117,97 @@ class ArchiveSpec
...
+
+      "archive file" in {
+        val filePath1 = getPathFromResources("akka_full_color.svg")
+        val filePath2 = getPathFromResources("akka_icon_reverse.svg")
+        val fileStream1: Source[ByteString, Any] = FileIO.fromPath(filePath1)
+        val fileStream2: Source[ByteString, Any] = FileIO.fromPath(filePath2)
+        val fileContent1 = ByteString(Files.readAllBytes(filePath1))
+        val fileContent2 = ByteString(Files.readAllBytes(filePath2))
+        val fileSize1 = Files.size(filePath1)
+        val fileSize2 = Files.size(filePath2)
```
ennru: I wonder if we can provide helpers to get the size from the stream, but that requires to process the source twice.

For example
```scala
            .map { path =>
              path.getFileName.toString -> FileIO.fromPath(path)
            }
            .mapAsync(1) {
              case (name, fileSource) =>
                val size = fileSource.runWith(Sink.fold(0)(_ + _.size))
                size.map { s =>
                  ArchiveMetadataWithSize(name, s) -> fileSource
                }(system.dispatcher)
            }
```

file/src/main/scala/akka/stream/alpakka/file/impl/archive/TarArchiveHeader.scala
```diff
@@ -0,0 +1,87 @@
...
+import java.lang.Long.toOctalString
+import java.time.Instant
+
+import akka.annotation.InternalApi
+import akka.util.ByteString
+
+/**
+ * INTERNAL API
+ */
+@InternalApi private[file] final class TarballHeader(val filePath: String,
```
ennru: Align the filename and the classname, please.

file/src/main/scala/akka/stream/alpakka/file/impl/archive/TarArchiveManager.scala
```diff
@@ -0,0 +1,30 @@
...
+    Flow[(ArchiveMetadataWithSize, Source[ByteString, Any])]
+      .flatMapConcat {
+        case (metadata, stream) =>
+          val header = new TarballHeader(metadata.filePath, metadata.size)
+          Source.single(header.bytes).concat(stream).concat(Source.single(padding(metadata.size)))
+      }
+  }
+
+  private def padding(fileSize: Long): ByteString =
+    ByteString(new Array[Byte](if (fileSize % 512 > 0) (512 - fileSize % 512).toInt else 0))
```
ennru: `else ByteString.empty`

file/src/main/scala/akka/stream/alpakka/file/scaladsl/Archive.scala
```diff
@@ -21,4 +21,10 @@ object Archive {
   def zip(): Flow[(ArchiveMetadata, Source[ByteString, Any]), ByteString, NotUsed] =
     ZipArchiveManager.zipFlow()
 
+  /**
+   * Flow for packaging multiple files into one TAR file.
+   */
+  def tar(): Flow[(ArchiveMetadataWithSize, Source[ByteString, Any]), ByteString, NotUsed] =
```
ennru:
```suggestion
  def tar(): Flow[(ArchiveMetadataWithSize, Source[ByteString, _]), ByteString, NotUsed] =
```

file/src/main/scala/akka/stream/alpakka/file/impl/archive/TarArchiveHeader.scala
choffmeister: @ennru Was thinking about that too. But since this a highly TAR file specific constraint, I would have to name the class smth like `TarArchiveMetadata` and that looked bad next to `ArchvieMetadata`, which is only used for the ZIP archive. And I did not want to rename `ArchiveMetadata` -> `ZipArchiveMetadata`.

So I see three alternatives:

1. Stick with what is in this PR
2. Do what you suggested, but then rename
    a. `ArchiveMetadata` -> `ZipArchiveMetadata` and
    b. `ArchiveMetadataWithSize` -> `TarArchiveMetadata`

What do you prefer?

file/src/test/scala/docs/scaladsl/ArchiveSpec.scala
choffmeister: Yeah it is unfortunate, that for TAR files you need to know the size of the files upfront. But I think  finding out the size is not in the scope of this.

But what I maybe should add is a check, that one actually provides a stream of matching size. So Running

```
Source((ArchiveMetadataWithSize("file.txt", 1), Source.single(ByteString("too long"))).runWith(...)
```

should fail.

file/src/main/scala/akka/stream/alpakka/file/scaladsl/Archive.scala
choffmeister: I had this before, but I wanted to be consistent with the `Archive.zip` API. Is it okay to also change the API for `.zip()` so that both use `Source[ByteString, _]`?

file/src/test/scala/docs/scaladsl/ArchiveSpec.scala
choffmeister: @ennru What you think about this for verifying the actual stream size matches the header:

```scala
  private def ensureFileSize(header: TarArchiveHeader): Flow[ByteString, ByteString, NotUsed] =
    Flow[ByteString]
    .map(Option.apply)
    .concat(Source.single(None))
    .statefulMapConcat(() => {
      var size = 0L
      elem => elem match {
        case Some(chunk) =>
          size = size + chunk.size
          chunk :: Nil
        case None =>
          if (size == header.size) Nil
          else throw new IllegalStateException(s"Byte stream for ${header.filePath} had $size bytes but expected were ${header.size} bytes")
      }
    })
```

choffmeister: Wrote the size checking as a custom stage (so I don't have to wrap every single chunk in a `Some(_)`).

file/src/main/scala/akka/stream/alpakka/file/impl/archive/TarArchiveHeader.scala
ennru: Making these independent of each other seems best.
So make it `TarArchiveMetadata` and leave the existing `ArchiveMetadata` unchanged.

file/src/main/scala/akka/stream/alpakka/file/scaladsl/Archive.scala
ennru: Changing `zip` wouldn't be fully backwards compatible. But you can do better on the new method.

file/src/test/scala/docs/scaladsl/ArchiveSpec.scala
ennru: That's a great addition.
In many situations, the size might be known. Let's leave the auto-detection for now.

choffmeister:
@ennru Thanks for all your feedback. Should now all be integrated in this PR. A little extra change that I made: I renamed `TarArchiveHeader` -> `TarArchiveEntry` and this now provides both, the header bytes as well as the trailing padding bytes. This makes the `TarArchiveManager` cleaner.

ennru:
Thank you! Great addition.

PR #2254, opened 2020-04-01
Akka 2.5.31 and prefer `ClassicActorSystemProvider`

ennru:
In Akka 2.5.31 the `ClassicActorSystemProvider` is actually implemented by the `akka.actor.ActorSystem` and `akka.actor.typed.ActorSystem[_]` so that it can be used to accept both types of actor systems.

With this PR the Alpakka APIs prefer the provider so that integration with the new actor APIs doesn't need any adapter in user code.


ennru:
When this is merged, the scheduled build with Akka 2.6 can be enabled again.

PR #2269, opened 2020-04-14
Google Cloud Storage: emit more chunks when input is closed

### LINKED ISSUE #2267: Last element of GCS chunker can exceed chunk size
The GCS chunker emits the entire buffer when the input is closed: https://github.com/akka/alpakka/blob/master/google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/impl/Chunker.scala#L52-L54

But this buffer might be larger than the chunk size if the chunker receives an individual byte string larger than the chunk size. This can eg. have the effect that GCS multipart uploads fail due to it creating too large HTTP entities for the final entity.

Instead, the chunker should ensure that the output chunks after the input has closed are smaller than the chunk size.

mpdn:
Fixes #2267


lightbend-cla-validator:
Hi @mpdn,

PR #2270, opened 2020-04-16
AWS S3: Allow source keys with non-ASCII characters

mpdn:
Adds missing URL encoding for source keys when copying.

### DIFF
--- a/s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3IntegrationSpec.scala
+++ b/s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3IntegrationSpec.scala
@@ -292,48 +292,15 @@ trait S3IntegrationSpec
     "1 + 2 = 3"
   )
 
-  it should "upload, copy, download the copy, and delete" in {
-    val sourceKey = "original/file.txt"
-    val targetKey = "copy/file.txt"
-    val source: Source[ByteString, Any] = Source(ByteString(objectValue) :: Nil)
+  it should "upload, copy, download the copy, and delete" in uploadCopyDownload(
+    "original/file.txt",
+    "copy/file.txt"
+  )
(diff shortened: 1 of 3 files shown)


s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3IntegrationSpec.scala
```diff
@@ -508,6 +475,47 @@ trait S3IntegrationSpec
       .runWith(Sink.head)
       .futureValue shouldEqual akka.Done
   }
+
+  private def uploadCopyDownload(sourceKey: String, targetKey: String): Assertion = {
+    val source: Source[ByteString, Any] = Source(ByteString(objectValue) :: Nil)
```
ennru, line 480:
```suggestion
    val source: Source[ByteString, Any] = Source.single(ByteString(objectValue))
```

ennru:
Some tests fail because of this change as the `/` now gets encoded.
https://travis-ci.com/github/akka/alpakka/jobs/322886269#L605

mpdn:
I corrected HttpRequestSpec to assume urlencoded keys. Alternatively, we could have chosen not to urlencode `/`, but this seems like a simpler solution. There shouldn't be any problem with always urlencoding `/`.

ennru:
Please run `scalafmtAll`.

ennru:
There is another test failure
https://travis-ci.com/github/akka/alpakka/jobs/323013018#L675

mpdn:
That seems odd. I suspect that might actually be a bug in MinIO as it seems like the space is causing problems. Anyway, just removing the space makes the test work at least. Alpakka still supports more characters than before, and AFAIK this should work against S3.

ennru:
Thank you, great fix!

PR #2273, opened 2020-04-17
Avroparquet: Lower bound higher kinded 

paualarco:
Hello!
There is no issue related with this PR.

While looking at how to implement protobuf parquet connector, I realised that the current implementation of avroparquet was statically typead as `GenericRecord`, which for the Akka user is not nice.

On continuation with that the tests were simplified and improved by not just checking the number of records written but also the actual content of the file.

### DIFF
--- a/avroparquet/src/main/mima-filters/2.0.0-RC2.backwards.excludes/PR2273-generics.excludes
+++ b/avroparquet/src/main/mima-filters/2.0.0-RC2.backwards.excludes/PR2273-generics.excludes
@@ -1,0 +1,7 @@
+# added generics type to the factory methods
+ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.stream.alpakka.avroparquet.javadsl.AvroParquetFlow.create")
+ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.stream.alpakka.avroparquet.javadsl.AvroParquetSource.create")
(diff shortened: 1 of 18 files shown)


paualarco:
I have noticed that it breaks the site creation with paradox, once I got some feedback I would fix that by creating a new tests that supports them.

ennru:
Thank you for this PR. I rebased it (as we had some problem with tests not running), fixed Paradox and added the required Mima exclusions.

Please improve the docs and examples to show that you may use a more specific type than Generic Record now.

We plan to release Alpakka 2.0 next week, it would be great to get this in.

paualarco:
> Thank you for this PR. I rebased it (as we had some problem with tests not running), fixed Paradox and added the required Mima exclusions.
> 
> Please improve the docs and examples to show that you may use a more specific type than Generic Record now.
> 
> We plan to release Alpakka 2.0 next week, it would be great to get this in.

Have pushed the changes, note that I have not added any java example that shows the improvement, but I have explicitly mentioned that in the documentation. 
Hope it will be enough for now :)

I think that for the time being this is a good improvement, but still could be even better to completely erase the type constrain on GenericRecord, in which therefore any implementation of `ParquetWriter` or `ParquetReader` could be used and not just the `Avro` one.

I have not the time to implement it now, but may be in the future... that would also imply a wider test coverage as well as renaming the whole sub-module.

ennru:
Thank you, I've pushed some updates as I'd like to see the PR for the release.

PR #2275, opened 2020-04-18
[WIP] Debezium connector

### LINKED ISSUE #2110: Debezium connector
### Short description

Implement Debezium connector

### Details

Debezium is a Change Data Capture (CDC) kafka connector that integrates with several databases. Although Debezium is a kafka connector, it also works as a standalone tool: https://debezium.io/documentation/reference/1.0/operations/embedded.html

The idea here is to use Debezium as a CDC source. The only issue is understand how backpressure would work here.

I also noticied there's already a PR with a CDC connector for PostgreSQL (#891), but, it'd be a nice to have if we could use databases too.

As I create this request, the lastest final version supports the following databases:

Database | Versions
-- | --
MySQL | 5.7, 8.0.13
MongoDB | 3.2, 3.4, 3.6, 4.0, 4.2
PostgreSQL | 9.6, 10, 11, 12
Oracle | 11g, 12c
SQL Server | 2017, 2019
Cassandra | 3.11.4

Apache Camel already has a component for it (actually, there are a few: one for each database). We could check on how they implement it and see if it's a viable solution for Alpakka.

- https://camel.apache.org/components/latest/debezium-mongodb-component.html
- https://camel.apache.org/components/latest/debezium-mongodb-component.html
- https://camel.apache.org/components/latest/debezium-postgres-component.html
- https://camel.apache.org/components/latest/debezium-sqlserver-component.html

gabfssilva:
Fixes #2110

Still under development.

What's left to do:

- Test against PostgreSQL and MongoDB (I have just tested against MySQL)
- Integrated tests with at least one database
- Documentation

Basic usage:

```scala
object DebeziumExample extends App {
  implicit val sys: ActorSystem = ActorSystem()
  import sys._

  val config = DebeziumSettings(
    name = "test-connector",
    connectorClass = classOf[MySqlConnector],
    offsetStorage = classOf[FileOffsetBackingStore],
    offsetCommitPolicy = classOf[PeriodicCommitOffsetPolicy],
    offsetStorageFileName = "/tmp/offsets.dat",
    offsetFlushInterval = 2.seconds,
    offsetFlushTimeout = 5.seconds,
    customSettings = Map(
      "database.hostname" -> "localhost",
      "database.port" -> 3306,
      "database.user" -> "debezium",
      "database.password" -> "dbz",
      "database.server.id" -> 184054,
      "database.server.name" -> "dbserver1",
      "database.whitelist" -> "inventory",
      "database.history" -> classOf[FileDatabaseHistory].getName,
      "database.history.file.filename" -> "dbhistory.dat"
    ),
    internalKeyConverter = classOf[JsonConverter],
    internalValueConverter = classOf[JsonConverter],
    maxInMemoryElements = 100
  )

  DebeziumSource(config) //auto commit
    .map(_.payload)
    .map(new String(_))
    .runWith(Sink.foreach(println))

  DebeziumSource
    .committable(config)
    .map { committableRecord =>
      println(new String(committableRecord.payload))
      committableRecord
    }
    .runWith(DebeziumSink.committer())
}
```

Waiting for feedbacks.

### DIFF
--- a/debezium/src/main/java/akka/stream/alpakka/debezium/impl/LazyEmbeddedEngine.java
+++ b/debezium/src/main/java/akka/stream/alpakka/debezium/impl/LazyEmbeddedEngine.java
@@ -1,0 +1,569 @@
+/*
+ * Copyright (C) 2016-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.debezium.impl;
+
+import io.debezium.annotation.ThreadSafe;
+import io.debezium.config.Configuration;
+import io.debezium.config.Field;
+import io.debezium.engine.DebeziumEngine;
+import io.debezium.engine.spi.OffsetCommitPolicy;
+import io.debezium.util.Clock;
+import io.debezium.util.VariableLatch;
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.connect.connector.ConnectorContext;
+import org.apache.kafka.connect.connector.Task;
+import org.apache.kafka.connect.errors.RetriableException;
+import org.apache.kafka.connect.json.JsonConverter;
+import org.apache.kafka.connect.runtime.WorkerConfig;
+import org.apache.kafka.connect.runtime.distributed.DistributedConfig;
+import org.apache.kafka.connect.runtime.standalone.StandaloneConfig;
+import org.apache.kafka.connect.source.SourceConnector;
+import org.apache.kafka.connect.source.SourceRecord;
+import org.apache.kafka.connect.source.SourceTask;
+import org.apache.kafka.connect.source.SourceTaskContext;
+import org.apache.kafka.connect.storage.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.time.Duration;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+
+public final class LazyEmbeddedEngine implements LazyDebeziumEngine<SourceRecord> {
+  public static final Field ENGINE_NAME =
+      Field.create("name")
+          .withDescription("Unique name for this connector instance.")
+          .withValidation(Field::isRequired);
+
+  public static final Field CONNECTOR_CLASS =
+      Field.create("connector.class")
+          .withDescription("The Java class for the connector")
+          .withValidation(Field::isRequired);
+
+  public static final Field OFFSET_STORAGE =
+      Field.create("offset.storage")
+          .withDescription(
+              "The Java class that implements the `OffsetBackingStore` "
+                  + "interface, used to periodically store offsets so that, upon "
+                  + "restart, the connector can resume where it last left off.")
+          .withDefault(FileOffsetBackingStore.class.getName());
+
+  public static final Field OFFSET_STORAGE_FILE_FILENAME =
+      Field.create(StandaloneConfig.OFFSET_STORAGE_FILE_FILENAME_CONFIG)
+          .withDescription(
+              "The file where offsets are to be stored. Required when "
+                  + "'offset.storage' is set to the "
+                  + FileOffsetBackingStore.class.getName()
+                  + " class.")
+          .withDefault("");
+
+  public static final Field OFFSET_STORAGE_KAFKA_TOPIC =
+      Field.create(DistributedConfig.OFFSET_STORAGE_TOPIC_CONFIG)
+          .withDescription(
+              "The name of the Kafka topic where offsets are to be stored. "
+                  + "Required with other properties when 'offset.storage' is set to the "
+                  + KafkaOffsetBackingStore.class.getName()
+                  + " class.")
+          .withDefault("");
+
+  public static final Field OFFSET_STORAGE_KAFKA_PARTITIONS =
+      Field.create(DistributedConfig.OFFSET_STORAGE_PARTITIONS_CONFIG)
+          .withType(ConfigDef.Type.INT)
+          .withDescription(
+              "The number of partitions used when creating the offset storage topic. "
+                  + "Required with other properties when 'offset.storage' is set to the "
+                  + KafkaOffsetBackingStore.class.getName()
+                  + " class.");
+
+  public static final Field OFFSET_STORAGE_KAFKA_REPLICATION_FACTOR =
+      Field.create(DistributedConfig.OFFSET_STORAGE_REPLICATION_FACTOR_CONFIG)
+          .withType(ConfigDef.Type.SHORT)
+          .withDescription(
+              "Replication factor used when creating the offset storage topic. "
+                  + "Required with other properties when 'offset.storage' is set to the "
+                  + KafkaOffsetBackingStore.class.getName()
+                  + " class.");
+
+  public static final Field OFFSET_FLUSH_INTERVAL_MS =
+      Field.create("offset.flush.interval.ms")
+          .withDescription("Interval at which to try committing offsets. The default is 1 minute.")
+          .withDefault(60000L)
+          .withValidation(Field::isNonNegativeInteger);
+
+  public static final Field OFFSET_COMMIT_TIMEOUT_MS =
+      Field.create("offset.flush.timeout.ms")
+          .withDescription(
+              "Maximum number of milliseconds to wait for records to flush and partition offset data to be"
+                  + " committed to offset storage before cancelling the process and restoring the offset "
+                  + "data to be committed in a future attempt.")
+          .withDefault(5000L)
+          .withValidation(Field::isPositiveInteger);
+
+  public static final Field OFFSET_COMMIT_POLICY =
+      Field.create("offset.commit.policy")
+          .withDescription(
+              "The fully-qualified class name of the commit policy type. This class must implement the interface "
+                  + OffsetCommitPolicy.class.getName()
+                  + ". The default is a periodic commit policy based upon time intervals.")
+          .withDefault(
+              io.debezium.embedded.spi.OffsetCommitPolicy.PeriodicCommitOffsetPolicy.class
+                  .getName())
+          .withValidation(Field::isClassName);
+
+  protected static final Field INTERNAL_KEY_CONVERTER_CLASS =
+      Field.create("internal.key.converter")
+          .withDescription(
+              "The Converter class that should be used to serialize and deserialize key data for offsets.")
+          .withDefault(JsonConverter.class.getName());
+
+  protected static final Field INTERNAL_VALUE_CONVERTER_CLASS =
+      Field.create("internal.value.converter")
+          .withDescription(
+              "The Converter class that should be used to serialize and deserialize value data for offsets.")
+          .withDefault(JsonConverter.class.getName());
+
+  /** The array of fields that are required by each connectors. */
+  public static final Field.Set CONNECTOR_FIELDS = Field.setOf(ENGINE_NAME, CONNECTOR_CLASS);
+
+  /** The array of all exposed fields. */
+  protected static final Field.Set ALL_FIELDS =
+      CONNECTOR_FIELDS.with(
+          OFFSET_STORAGE,
+          OFFSET_STORAGE_FILE_FILENAME,
+          OFFSET_FLUSH_INTERVAL_MS,
+          OFFSET_COMMIT_TIMEOUT_MS,
+          INTERNAL_KEY_CONVERTER_CLASS,
+          INTERNAL_VALUE_CONVERTER_CLASS);
+
+  private static final Duration WAIT_FOR_COMPLETION_BEFORE_INTERRUPT_DEFAULT =
+      Duration.ofSeconds(2);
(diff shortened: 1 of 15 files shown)


gabfssilva:
I've ran into an issue:

Debezium has no asynchronous backpressure mechanism, so, this code won't backpressure anything and Debezium will fetch all data into memory:

```scala
DebeziumEngine
  .create[SourceRecord](classOf[Connect])
  .using(configurations)
  .notifying((records: util.List[SourceRecord], committer: DebeziumEngine.RecordCommitter[SourceRecord]) => {
    //pushEventAsynchronouslyToSource
  })
  .using((success: Boolean, message: String, error: Throwable) => 
    if (!success) queue.fail(error)
    else queue.complete()
  )
  .build()
```

What we could do is, instead of pushing the data asynchronously into our `Source`, we could block the `notifying` function and only return when all records were committed properly. This way Debezium will backpressure, since we can limit how many threads it will use to fetch data, we could backpressure per thread.
But, I'm a little skeptical about this one, since we're kind of cheating Akka Streams backpressure.
I've searched their api and it seems that we cannot pull data in a decent way. I guess there's something we could do but we would be rewriting a lot of their core code.

So, what do you people think?

gabfssilva:
Just to add: blocking `notifying` works as a backpressure mechanism and this is part of the api. Not sure if it's all that bad to do it instead of rewrite most of the engine code.

seglo:
Thanks for working on this @gabfssilva! 

> we could block the `notifying` function and only return when all records were committed properly.

Isn't this the purpose of `DebeziumSink|Flow.committer()`?

I'm curious why you added an async boundary in its implementation:

```scala
    Flow[CommittableRecord].async
      .map(_.markAsProcessed())
```

debezium/src/main/scala/akka/stream/alpakka/debezium/DebeziumRecord.scala
```diff
@@ -0,0 +1,29 @@
...
+private[debezium] trait DebeziumRecord {
+  private[debezium] val converter: Converter
+
+  val record: SourceRecord
+
+  lazy val payload: Array[Byte] = {
+    val schemaBuilder = SchemaBuilder.struct.field("key", record.keySchema)
+    if (record.valueSchema != null) schemaBuilder.field("value", record.valueSchema)
+
+    val schema = schemaBuilder.build
```
seglo, line 20: Building a schema seems like something we only need to do once. Perhaps it could be triggered lazily and stored in `DebeziumSettings`, or in a companion object of `DebeziumRecord`.

debezium/src/main/scala/akka/stream/alpakka/debezium/DebeziumSettings.scala
```diff
@@ -0,0 +1,124 @@
...
+    properties.put("internal.value.converter", internalValueConverter.getName)
+
+    val custom: util.Map[Object, Object] = customSettings.asJava.asInstanceOf[util.Map[Object, Object]]
+    properties.putAll(custom)
+
+    properties
+  }
+}
+
+object DebeziumSettings {
```
seglo, line 76: We'll need to elaborate this to include defaults in a `reference.conf` and a fluent API for Scala and Java users to call at runtime.

debezium/src/main/scala/akka/stream/alpakka/debezium/CommittableRecord.scala
```diff
@@ -0,0 +1,23 @@
...
+import org.apache.kafka.connect.storage.Converter
+
+final class CommittableRecord(val record: SourceRecord,
+                              private[debezium] val committer: ActorRef,
+                              private[debezium] val converter: Converter)
+    extends DebeziumRecord {
+
+  private[debezium] def markAsProcessed(): Done = {
+    committer ! record
+    Done
```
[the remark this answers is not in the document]
gabfssilva: The thing is: I was using the ask pattern there, but, actually it makes no difference since `markProcessed` and `markBatchFinished` just marks the message to be committed and not really commit it.
Also, what's preventing back-pressure to work properly is actually how DebeziumEngine works: it expects a blocking `notifying` function. No matter how we commit offsets, it has nothing to do with Debezium back-pressuring whatsoever. So, if commit or not commit anything, Debezium will keep to emit new notifications.
I've done a few changes and I forgot to push, sorry, I'm going to push it and so you can take a look at it. This way I was able to simulate a "fair enough" back-pressure.

debezium/src/main/scala/akka/stream/alpakka/debezium/DebeziumRecord.scala
gabfssilva: Actually, Debezium emits events for multiple tables and tables have evolution too, the schema can change at any time.

gabfssilva:
> I'm curious why you added an async boundary in its implementation:

Actually I was trying something and I forgot to remove it, sorry. haha

debezium/src/main/scala/akka/stream/alpakka/debezium/scaladsl/DebeziumSource.scala
```diff
@@ -0,0 +1,103 @@
...
+          }(ExecutionContexts.sameThreadExecutionContext)
+      })
+      .using(
+        (success: Boolean, message: String, error: Throwable) =>
+          if (!success) queue.fail(error)
+          else queue.complete()
+      )
+      .build();
+
+    executor.execute(engine)
```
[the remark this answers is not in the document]
gabfssilva: Source as a custom graph stage probably would be better, but, I'm not quite sure how I'd fit since it's Debezium that pushes data into the Source, and not the Source that tries to pull it. That's why I thought a queue it'd fit a little better.
I guess I'll try to hack Debezium by creating a new custom engine that does not run by itself but rather expose a method for us to pulls data. I'm just not quite sure if this is straightforward tho.

debezium/src/main/scala/akka/stream/alpakka/debezium/scaladsl/DebeziumSource.scala
```diff
@@ -0,0 +1,108 @@
...
+          queue
+            .offer(list.map(r => new CommittableRecord(r, committerActor, valueConverter)))
+            .flatMap {
+              case QueueOfferResult.Enqueued =>
+                p.future
+              case _ =>
+                Future.failed(new InterruptedException("error while enqueueing records"))
+            }(ExecutionContexts.sameThreadExecutionContext)
+
+        Await.result(result, settings.acknowledgmentTimeout)
```
gabfssilva, line 57: The only thing preventing Debezium to emit new events is this line of code. If we don't block here, it will continue to emit new elements, even if we don't commit the offset.

debezium/src/main/scala/akka/stream/alpakka/debezium/scaladsl/DebeziumSource.scala
```diff
@@ -0,0 +1,142 @@
...
+import akka.stream.{OverflowStrategy, QueueOfferResult}
+import io.debezium.embedded.Connect
+import io.debezium.engine.DebeziumEngine
+import org.apache.kafka.connect.source.SourceRecord
+
+import scala.collection.JavaConverters._
+import scala.concurrent.{Await, ExecutionContext, Future, Promise}
+
+object DebeziumSource {
+  def custom(settings: DebeziumSettings)(implicit s: ActorSystem): Source[CommittableRecord, NotUsed] = {
```
gabfssilva, line 25: Ok, so, I created this one using an engine that I tried to create based on the `EmbeddedEngine` class. It's basically `EmbeddedEngine`, but without the automatic polling, so I exposed a method `poll` so we can control it within Akka Streams.
I didn't see why create a `GraphStage` here, so I tried to keep it simple.

debezium/src/main/scala/akka/stream/alpakka/debezium/scaladsl/DebeziumSource.scala
```diff
@@ -0,0 +1,142 @@
...
+
+    val engine = new LazyEmbeddedEngine.BuilderImpl().using(configurations).build()
+
+    lazy val committer = engine.committer()
+
+    val singleThreadEc = ExecutionContext.fromExecutorService(Executors.newSingleThreadExecutor())
+
+    Source
+      .repeat(engine)
+      .mapAsync(1) { engine =>
```
gabfssilva, line 39: AFAIK, Debezium polling is single threaded most time, but this parameter could be customised I guess.

debezium/src/main/scala/akka/stream/alpakka/debezium/scaladsl/DebeziumSource.scala
```diff
@@ -0,0 +1,142 @@
...
+          val records = engine.poll()
+          if (records == null) List.empty else records.asScala.toList
+        }(singleThreadEc)
+      }
+      .mapConcat { list =>
+        val p = Promise[Unit]()
+        val committerActor = s.actorOf(Props(new CommitterActor(list, p, committer)))
+
+        list.map { record =>
+          new CommittableRecord(record, committerActor, valueConverter)
```
gabfssilva, line 50: Still Debezium that controlls offset committing, that's why I kept the actor to commit the whole offset as before.

debezium/src/main/java/akka/stream/alpakka/debezium/impl/LazyEmbeddedEngine.java
```diff
@@ -0,0 +1,569 @@
...
+import java.util.Map;
+import java.util.Properties;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+
+public final class LazyEmbeddedEngine implements LazyDebeziumEngine<SourceRecord> {
```
gabfssilva, line 43: This is a adaptation of `EmbeddedEngine`. Mostly it's `EmbeddedEngine` code, but I changed how polling works: now it does not poll data by itself but rather expose a `poll` method, this way it's easy for us to control in Akka Streams.

gabfssilva:
So, I created the custom engine, but, as I said: I needed to rewrite most of the engine code. 

Pros: now it's easier for Akka Streams to control backpressure
Cons: many lines of code rewritten, to keep it up to date to Debezium could be a bit painful.

Debezium definitely wasn't designed for reactive streams integration. =/

debezium/src/main/java/akka/stream/alpakka/debezium/impl/LazyEmbeddedEngine.java
gabfssilva: Also, this is just a draft. I'll try to remove more useless code later. I just want to see from you if I should go with this strategy or not.

debezium/src/main/scala/akka/stream/alpakka/debezium/CommittableRecord.scala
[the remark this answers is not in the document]
seglo: That's interesting. I see your `notifiying` implementation now.

debezium/src/main/scala/akka/stream/alpakka/debezium/DebeziumRecord.scala
seglo: Got it.  I guess we could cache schemas, but probably not something we need to consider for now.

debezium/src/main/scala/akka/stream/alpakka/debezium/scaladsl/DebeziumSource.scala
seglo: Seems reasonable.  It's too bad how low-level we need to get to make this work.

debezium/src/main/scala/akka/stream/alpakka/debezium/scaladsl/DebeziumSource.scala
seglo: Yes, I don't see much value in a graph stage either.

debezium/src/main/java/akka/stream/alpakka/debezium/impl/LazyEmbeddedEngine.java
seglo: It certainly does make the implementation more of a burden to maintain.  Perhaps it would be beneficial to reach out to the Debezium team and ask them to review this PR. I'm sure they would be open to an upstream contribution to make life easier.

seglo:
I'm really impressed with this work so far, @gabfssilva.  I left some comments.  WDYT about reaching out to the Debezium team for advice, or perhaps a contribution upstream to make integration easier?

gabfssilva:
> I'm really impressed with this work so far, @gabfssilva. I left some comments. WDYT about reaching out to the Debezium team for advice, or perhaps a contribution upstream to make integration easier?

I talked to them, they even have a WIP Reactive Streams integration: debezium/debezium#458, but as far I could see they had other issues to be resolved before merging it into master, such as: the `SourceRecord` is a Kafka Connect class, so they probably want to make Debezium more agnostic before adding Reactive Streams support since there's no point to use Kafka Connect API if you're using Debezium standalone.

I guess I could try to reach them out again and maybe help them with that.

seglo:
It's probably worth a discussion at least before proceeding with the custom engine approach.

gabfssilva:
Ok, so I talked to the Debezium team and I'll send a PR to Debezium repository with a Reactive Streams implementation.

I guess we can close this PR for now until I finish the implementation on their side, after that it'll be much easier for Alpakka to integrate with Debezium, what do you think?

seglo:
Sounds good @gabfssilva. Thanks for taking on this effort!  I'll close this PR for now.

Issue #2098, opened 2020-01-17
Elasticsearch: make use of Testcontainers


ennru:
Elasticsearch changes its REST API between major versions.
Alpakka could make use of [Testcontainers Elasticsearch](https://www.testcontainers.org/modules/elasticsearch/) to run against different ES versions in our tests.

## References

#945
#1232
#2025

ennru:
When @r-glyde tried using Testcontainers in #2277, they ran into issues with too slow startup/shutdown. Especially the multi-Scala-version builds showed issues, as the containers would be started and stopped multiple times within one sbt session.
So Alpakka resorted to docker-compose.

PR #2277, opened 2020-04-21
Elasticsearch: add support for version 7

r-glyde:
* Support v7 API
* ~Use testcontainers to run tests against multiple elasticsearch versions~ Edit: testcontainers did not seem to work very consistently on travis and other machines at the moment. Different versions are now tested with docker-compose

References #1232 
References #2025
~References #2098~ Edit - attempted as part of this but reverted to using docker-compose in the end

If this looks ok (or along the right lines at least) I don't mind having a look at adding version 6 as well if it is needed (although I think users would probably be fine using either V5 or V7).


docs/src/main/paradox/elasticsearch.md
```diff
@@ -30,7 +30,7 @@ Sources, Flows and Sinks provided by this connector need a prepared `org.elastic
 access to Elasticsearch.
```
ennru, line 30: Please add a section about the implemented API versions with a note that it does not need to excatly match the Elasticsearch server version.

docs/src/main/paradox/elasticsearch.md
```diff
@@ -111,17 +111,18 @@ Java
...
-| versionType         | None    | If set, `ElasticsearchSink` uses the chosen versionType to index documents. See [Version types](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#_version_types) for accepted settings. |
-| retryLogic | No retries | See below |
+| Parameter           | Default    | Description                                                                                            |
+| ------------------- | -------    | ------------------------------------------------------------------------------------------------------ |
+| bufferSize          | 10         | Flow and Sink batch messages to bulk requests when back-pressure applies.                             |
+| versionType         | None       | If set, `ElasticsearchSink` uses the chosen versionType to index documents. See [Version types](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#_version_types) for accepted settings. |
+| retryLogic          | No retries | See below |
+| apiVersion          | V5         | See [supported versions](/elasticsearch/src/main/java/akka/stream/alpakka/elasticsearch/ApiVersion.java) |
```
ennru, line 125: The link to the source won't work in the docs. We might need to live with a potential mismatch here. State the available API versions.

r-glyde:
Thanks for the review @ennru. I'll go through and address soon.

On testcontainers not working for you - any ideas on what might fix that? Do both the java and scala tests fail? Might be able to add a wait strategy to the containers but will be hard to see if that works properly I think. I guess once #2278 is done and this is rebased (assuming you want to do it in that order) we can see if these tests run properly in travis.

ennru:
Thank you for noticing the blunder in the test script setup. Please rebase now to run the tests, let's see what Travis says.

The Java tests time out when the "other" round starts:
```
[debug] Test docs.javadsl.ElasticsearchTest.testMultipleOperations[0: version=6.8.0 api=V5] finished, took 0.392 sec
[error] Test [1: version=7.6.0 api=V7] failed: java.io.IOException: listener timeout after waiting for [30000] ms, took 0.0 sec
```

The Scala tests seem to be broken in another way:
```
[info] ElasticsearchSpec:
[info] docs.scaladsl.ElasticsearchSpec *** ABORTED *** (30 milliseconds)
[info]   Assertion should be put inside it or they clause, not should, must, when, that, which or can clause. (ElasticsearchConnectorBehaviour.scala:1073)
...
[info]   Cause: org.scalatest.exceptions.TestFailedException: "ElasticsearchUpdateSettings(bufferSize=10,retryLogic=akka.stream.alpakka.elasticsearch.RetryAtFixedRate@5259561f,versionType=Some(internal),apiVersion=V5)" did not start with substring "ElasticsearchWriteSettings("
[info]   at org.scalatest.matchers.MatchersHelper$.indicateFailure(MatchersHelper.scala:339)
[info]   at org.scalatest.matchers.should.Matchers$ShouldMethodHelperClass.shouldMatcher(Matchers.scala:6783)
[info]   at org.scalatest.matchers.should.Matchers$AnyShouldWrapper.should(Matchers.scala:6822)
[info]   at docs.scaladsl.ElasticsearchConnectorBehaviour.elasticsearchConnector(ElasticsearchConnectorBehaviour.scala:1073)
```

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/RestBulkApi.scala
```diff
@@ -14,7 +16,36 @@ import scala.collection.immutable
...
+      case (item, message) =>
+        val command = message.operation.command
+        val res = item.asJsObject.fields(command).asJsObject
+        val error: Option[String] = res.fields.get("error").map(_.toString())
+        new WriteResult(message, error)
+    }
+    messageResults
+  }
+
+  def optionalString(value: Option[String], fieldName: String): Option[(String, JsString)] =
```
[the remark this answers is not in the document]
r-glyde: Yep, that would probably make more sense.

docs/src/main/paradox/elasticsearch.md
r-glyde: Added a brief bit into the `Sink and flow configuration` section of the docs. Added links to the corresponding API docs but can go into more detail here if needed.

r-glyde:
@ennru I think I've addressed the comments.

I still need to investigate some issues in the travis build. It now runs properly and all the java tests appear to run and pass but the scala 2.13 tests fails to startup the container (other scala versions seem fine). Running locally everything appears good for all versions so I think I'll need to add some debug to see what is happening with testcontainers on travis.

The build also fails for the `amqp` and `mqtt` builds are failing - I assume they ran because dependencies changed.

ennru:
> The build also fails for the amqp and mqtt builds are failing - I assume they ran because dependencies changed.

Yes, and we might have flaky tests there. So these failures are most likely not connected to your changes.

ennru:
The container startup looks surprisingly slow. Could you make it start both ES instances and reuse them for all tests instead of a new container per test?

r-glyde:
> The container startup looks surprisingly slow. Could you make it start both ES instances and reuse them for all tests instead of a new container per test?

For the java tests I think we could although I'm not sure if that is the issue in travis at the moment. Are they still failing for you locally?

The scala tests currently start all containers first and should stop them afterwards so they are only restarted for different scala versions which I'm not sure if we could stop (at least I'm not sure how right now). I think the issue on travis at the moment is the final container for scala 2.13 - it looks like most containers start in 15-30 seconds but the final one times out after 240 which seems odd. I was wondering if the previous containers are not being shutdown properly or something is left running and by the time the final tests start travis can't handle everything.

I'll have a look at just starting up containers once and also checking that things get shutdown properly. I think I should be able to confirm this stuff locally without having to run a load of travis builds. Probably won't be able to get to this until this evening or the weekend though I don't think.

ennru:
It still fails for me locally: The first round of Java tests succeed, the 7.6.0 startup is too slow so that the second round fails. For the Scala tests the 6.8.0 container doesn't start in time.

Maybe resort to docker-compose to get the changes into Alpakka 2 which we'll release next week?

ennru:
I set `Test / fork := true`, but it still fails locally (now running Scala tests first). I believe that will help on Travis with the cross Scala version build.

r-glyde:
Ah, didn't realise alpakka 2 would be releasing that soon. Yeah, I think just using docker-compose for now might be the best solution. It should be quite straight forward so I'll try to get it done today.

r-glyde:
@ennru I've removed testcontainers and added services for the different elasticsearch versions into the docker-compose yaml and the build now looks good as far as elasticsearch is concerned (at least on my machine and on travis) so I think this is ready for another review.

To be honest, I'm not sure we were gaining much by using testcontainers - the tests now run faster and more reproducibly and the setup is simpler and more consistent with all the other modules (maybe just my opinion).

ennru:
Thank you! This bothered me since a while back, but I never got to work on it.

Issue #2209, opened 2020-03-17
S3SinkSpec: S3Sink should retry part upload after a transient downstream connection error


raboof:
Failed at https://travis-ci.com/github/akka/alpakka/jobs/298526637#L916

```
[info] S3Sink
[info] - should retry part upload after a transient downstream connection error *** FAILED *** (1 minute)
[info]   A timeout occurred waiting for a future to complete. Queried 101 times, sleeping 600 milliseconds between each query. (S3SinkSpec.scala:75)
[info]   org.scalatest.concurrent.Futures$FutureConcept$$anon$1:
[info]   at org.scalatest.concurrent.Futures$FutureConcept.tryTryAgain$1(Futures.scala:538)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValueImpl(Futures.scala:550)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValueImpl$(Futures.scala:479)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:275)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue(Futures.scala:476)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue$(Futures.scala:475)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:275)
[info]   at docs.scaladsl.S3SinkSpec.$anonfun$new$4(S3SinkSpec.scala:75)
```

Possibly related to #2114 / #2116

bpiper:
I've tried replicating this failure locally on latest master and the commit that the build failed on, but without success.  It doesn't appear to have failed since the build mentioned above, but it did also fail on a couple of older builds, and on at least one occasion, `retry part upload no more than the configured number of times` also failed in the same build.  Both tests are quite reliant on Wiremock doing its thing (receiving multiple requests in quick succession), so it is conceivable that there is some sort of fragility/race-condition there.

I'll try to keep an eye on this and if it fails again, perhaps I can tweak the test cases to add delay between HTTP requests or something like that.

seglo:
[...]

PR #2282, opened 2020-04-28
S3: Add some backoff delay to multipart upload retries for tests

bpiper:
…to try and address intermittent failures (#2209).

No guarantees this will address the failures, but I can't replicate them locally, so will have to rely on Travis.


bpiper:
Just going to trigger Travis builds until either the test failure recurs or I'm satisfied that the issue may be resolved.  Please don't merge until I've squashed/rebased to remove the superfluous commits.

bpiper:
Ok, I've run something like 24 Travis builds and the failure hasn't recurred (two of the builds did fail, but for unrelated reasons), so I'm confident that in the worst case this change hasn't made things any worse, and has quite possibly addressed the intermittent failures, so it's RTBM from my perspective.

bpiper:
@ennru would you mind giving this a quick review when you have time (or flick it to someone else)? Thanks.

ennru:
Thank you for following up on this!

PR #2283, opened 2020-04-28
Update silencer plugin to 1.6.0

raboof:
which supports scala 2.13.2


ennru:
Seems like version 1.7.0 is fresh off the press.

raboof:
> Seems like version 1.7.0 is fresh off the press.

Yes, but that's not built for 2.12.10 (only 2.12.11) https://repo1.maven.org/maven2/com/github/ghik/silencer-lib_2.12.10/ ... I'll just amend #2281 and do all the updates :)

raboof:
> I'll just amend #2281 and do all the updates :)

Ah that PR doesn't accept pushes, so created #2286

PR #2286, opened 2020-04-28
Scala to 2.12.11 and 2.13.2 and silencer 1.7.0

raboof:


ennru:
This works now (#2126 excluded).

ennru:
> Curious exactly what our policy is regarding incrementing Scala versions.

You're right. We might want to leave this hanging until those versions can be assumed more common.

Issue #2288, opened 2020-04-28
[Failed] Mqtt source should automatically reconnect


ennru:
Promise already completed.

```
- should automatically reconnect *** FAILED *** (1 second, 257 milliseconds)
  java.lang.AssertionError: expected OnNext, found OnError(java.lang.IllegalStateException: Promise already completed.
	at scala.concurrent.Promise$class.complete(Promise.scala:55)
	at scala.concurrent.impl.Promise$DefaultPromise.complete(Promise.scala:157)
	at akka.stream.alpakka.mqtt.impl.MqttFlowStageLogic$$anonfun$1.apply(MqttFlowStage.scala:97)
	at akka.stream.alpakka.mqtt.impl.MqttFlowStageLogic$$anonfun$1.apply(MqttFlowStage.scala:96)
	at akka.stream.impl.fusing.GraphInterpreter.runAsyncInput(GraphInterpreter.scala:452)
	at akka.stream.impl.fusing.GraphInterpreterShell$AsyncInput.execute(ActorGraphInterpreter.scala:481)
	at akka.stream.impl.fusing.GraphInterpreterShell.processEvent(ActorGraphInterpreter.scala:581)
	...

  at akka.stream.testkit.TestSubscriber$ManualProbe.expectNext(StreamTestKit.scala:349)
  at akka.stream.testkit.TestSubscriber$Probe.requestNext(StreamTestKit.scala:802)
  at docs.scaladsl.MqttSourceSpec$$anonfun$2$$anonfun$apply$mcV$sp$1.apply$mcV$sp(MqttSourceSpec.scala:356)
  at docs.scaladsl.MqttSourceSpec$$anonfun$2$$anonfun$apply$mcV$sp$1.apply(MqttSourceSpec.scala:302)
  at docs.scaladsl.MqttSourceSpec$$anonfun$2$$anonfun$apply$mcV$sp$1.apply(MqttSourceSpec.scala:302)
  
```

https://travis-ci.com/github/akka/alpakka/jobs/324193079#L640 in #2284

ennru:
And even in #2285

https://travis-ci.com/github/akka/alpakka/jobs/324197704#L559

ennru:
This is a real error, even failing locally for me.

seglo:
Fixed in #2289

PR #2289, opened 2020-04-28
MQTT: avoid completing promise on reconnect

ennru:
The materialized subscription promise should not be completed again when a reconnect happens.

(I'm not sure yet why this problem didn't show earlier.)

References #2288


ennru:
The behaviour changed with the upgrade of the Paho client library in #2236

PR #2297, opened 2020-05-01
Kinesis: Avoid Any in KinesisSchedulerSourceStage

YikSanChan:
I would prefer avoiding `Any` if possible, but not sure if Any has any benefit that I might haven't noticed in this specific use case. To me, this is internal and should be safe to change for better type safety


YikSanChan:
@aserrallerios Do you mind reviewing this? Thanks!

YikSanChan:
@ennru Got Albert's approval, do you mind taking a look?

PR #2302, opened 2020-05-02
[kinesis] Get rid of unnecessary broadcast-zip

YikSanChan:
The broadcast aims to carry along commitable records, while at the same time, try to checkpoint based on the last record per batch. Since the checkpointing runs synchronously (told from the return type) there isn't a need to have the broadcast-then-zip pattern. Instead, simply run the try-checkpoint side effect in a map stage suffice.


YikSanChan:
@ennru May I get some eyes on this?

ennru:
I'm afraid I won't get to it for a while. Possibly @aserrallerios can have a look at your PRs?

ennru:
Thank you, this made it significantly simpler.

PR #2307, opened 2020-05-08
Mongo: Change to mongo-scala-driver 4.1.1

mikela:
Update mongo-scala-driver to latest version (4.0.3)


lightbend-cla-validator:

mikela:
CLA signed.

mikela:
CLA signed.

mikela:
CLA signed.

ennru:
Thank you @mikela for opening this PR. I'm sorry, but it came a little bit too late to make it into Alpakka 2.0 and we don't want to change dependencies within patch releases. 
As this is [quite a jump with changed APIs](https://doc.akka.io/docs/alpakka/current/other-docs/versioning.html#third-party-dependencies) and everything, we'll have to wait until Alpakka 3.0 before merging. 
But I believe we'll see more library upgrades piling up and might release 3.0 by end of 2020.

mikela:
Sounds good! Thanks!

agemooij:
Hi @ennru . Any chance this will be released earlier than end of year? I'd like to contribute some extra work on top of this PR, e.g. to add additional overloads of the APIs that take the new Scala-specific classes, which also implement the Reactive Streams APIs so I use the newer, Scala-specific APIs without running into this:

Screen Shot 2020-06-09 at 23 02 25

Thanks,
Age

ennru:
Hi @agemooij,

I'm afraid that's unlikely. We'll do the major release when enough things accumulate which would break APIs or include major dependency changes. As we just did Alpakka 2.0 it will take a while.

agemooij:
Thanks @ennru . I'll try with the non-Scala APIs but still with the 4.x dependencies and see what happens.

mikela:
Hi @ennru, I updated this PR with the newer version of the Mongo driver. Not sure if [85cf76c](https://github.com/akka/alpakka/pull/2307/commits/85cf76cef2cafc8406cc88537db7d47d93a02d7a) (I pulled the changes from the Alpakka project to my fork) affects the approval in any way. If so I can just discard this PR and make a new one. Thanks!

project/Dependencies.scala
```diff
@@ -340,8 +340,7 @@ object Dependencies {
 
   val MongoDb = Seq(
     libraryDependencies ++= Seq(
-        "org.mongodb" % "mongodb-driver-reactivestreams" % "1.12.0", // ApacheV2
-        "org.mongodb.scala" %% "mongo-scala-bson" % "2.7.0" % "test" // ApacheV2
+        "org.mongodb.scala" %% "mongo-scala-driver" % "4.1.0" // ApacheV2
```
ennru, line 343: We are ready for Alpakka 3.0 merges now!
There is a later release of the Mongo driver (4.1.1).
What's the value-add of `mongo-scala-driver` compared to `mongodb-driver-reactivestreams` and `mongo-scala-bson` as used before?

agemooij: @ennru AFAIK, the main 4.x Java driver lib now has all APIs in the same distribution package. The separate Scala driver and reactivestreams driver have been discontinued during the release of their 4.x driver.

The main driver page describes that here: https://mongodb.github.io/mongo-java-driver/

It's been a while again now that I looked at the detailed differences but judging from my own comments above the migration process is mostly one of changing the imported packages.

mikela: The old driver has been superseded by the new one. According to the [driver compatibility table](https://docs.mongodb.com/drivers/scala/#mongodb-compatibility), only the new driver is compatible with MongoDB 4.4+.

ennru: Thank you for this references. Please add them in a comment for later reference.
Please update even the docs page in `mongodb.md` with appropriate links to give information about the driver and compatible MongoDB versions.

mikela:
The new [MongoDB Java Driver](http://mongodb.github.io/mongo-java-driver/) is [compatible](https://docs.mongodb.com/drivers/scala#mongodb-compatibility) with all MongoDB releases from versions 2.6 to 4.4, unlike the old one, which is not compatible with MongoDB 4.4.

project/Dependencies.scala
mikela: -  Not sure if by _add them in a comment_ you mean here on this PR or somewhere else. 
-  `mongodb.md` updated with information about the new driver.
Regarding this doc, I'd remove the _Alternative connector_ section as it is not [compatible](http://reactivemongo.org/releases/1.0/documentation/release-details.html#compatibility) with MongoDB version 4.4, at least for now. I haven't removed it just in case you want to keep it for other reasons. 
Thanks!

ennru (approved):
LGTM.

ennru:
Thank you! We're ramping up for Alpakka 3 now.

robert-blankenship:
Any timeline on when Alpakka 3 will be released? My team is struggling with some incompatibility issues related to the outdated Mongo driver

mikela:
@robert-blankenship I'm not part of Lightbend so I have no idea but if you need to use the new Mongo driver now you can use a snapshot version of Alpakka, like `2.0.2+46-a1f2cf02`. I suggest you test it first but the Mongo code in that snapshot version (or above) is most likely the same code Alpakka 3 will have. You can see more versions here https://bintray.com/akka/snapshots/alpakka (perhaps you knew already but also for other users). Cheers!

ennru:
Sorry for the delay, we at Lightbend don't have much time to spend on Alpakka for the time being.

I hope the snapshot works for you, we might be able to do a M1.

PR #2311, opened 2020-05-12
CSV: handle UTF-8 BOM correctly

tst60:
When the input to CSVParser.offer starts with a UTF-8 byte order marker, the cursor is advanced 4 places (but the size of the BOM is 3 bytes). This causes an edge case if the first value is quoted, where it gets wrapped in additional quotes:

```
      val in = ByteString("\"one\",\"two\",\"three\"\n", StandardCharsets.UTF_8.name())
      val inWithBom = ByteOrderMark.UTF_8 ++ in
      val parser = new CsvParser(',', '"', '\\', maximumLineLength)

      println("Without BOM")
      parser.offer(in)
      parser.poll(false).value.map(_.utf8String).foreach(s => println("- " + s))

      println("With BOM")
      parser.offer(inWithBom)
      parser.poll(false).value.map(_.utf8String).foreach(s => println("- " + s))
```

Output prior to this change (see the extra quotes around "one" in the second case):

```
Without BOM
- one
- two
- three
With BOM
- ﻿"one"
- two
- three
```

### DIFF
--- a/csv/src/main/scala/akka/stream/alpakka/csv/impl/CsvParser.scala
+++ b/csv/src/main/scala/akka/stream/alpakka/csv/impl/CsvParser.scala
@@ -182,7 +182,7 @@ import scala.collection.mutable
   private[this] def checkForByteOrderMark(): Unit =
     if (buffer.length >= 2) {
       if (buffer.startsWith(ByteOrderMark.UTF_8)) {
-        advance(4)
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
Hi @tst60,

ennru:
Thank you for spotting and fixing it!

PR #2312, opened 2020-05-13
Update Akka gRPC to use native openssl

raboof:
Even without specifying a CA

Tested against google cloud, also documenting how to do that in `application.conf`.


google-cloud-pub-sub-grpc/src/main/mima-filters/2.0.0.backwards.excludes/update-to-akka-grpc-1.0.0-M1.backwards.excludes
```diff
@@ -0,0 +1,2 @@
+# Private ctor
```
ignasi35, line 1:
```suggestion
# Private constructor
```

project/Dependencies.scala
```diff
@@ -225,14 +225,14 @@ object Dependencies {
...
-        "io.grpc" % "grpc-auth" % "1.28.0", // ApacheV2
+        "io.grpc" % "grpc-auth" % "1.29.0", // ApacheV2
         "com.google.auth" % "google-auth-library-oauth2-http" % "0.20.0", // BSD 3-clause
         // pull in Akka Discovery for our Akka version
         "com.typesafe.akka" %% "akka-discovery" % AkkaVersion
       ) ++ Silencer
   )
 
-  val GooglePubSubGrpcAlpnAgent = "org.mortbay.jetty.alpn" % "jetty-alpn-agent" % "2.0.9"
+  val GooglePubSubGrpcAlpnAgent = "org.mortbay.jetty.alpn" % "jetty-alpn-agent" % "2.0.10"
```
ennru, line 235: Do we still need the ALPN agent?

project/plugins.sbt
```diff
@@ -15,7 +15,7 @@ addSbtPlugin("com.eed3si9n" % "sbt-unidoc" % "0.4.3")
 addSbtPlugin("com.thoughtworks.sbt-api-mappings" % "sbt-api-mappings" % "3.0.0")
 addSbtPlugin("com.typesafe.sbt" % "sbt-site" % "1.4.0")
 // Akka gRPC -- sync with version in Dependencies.scala
-addSbtPlugin("com.lightbend.akka.grpc" % "sbt-akka-grpc" % "0.8.4")
+addSbtPlugin("com.lightbend.akka.grpc" % "sbt-akka-grpc" % "1.0.0")
```
ennru, line 18: Or 1.0.1?

project/Dependencies.scala
raboof: We do for testing on jdk8 before 8u251.

We probably don't do that, so I'll drop it.

project/plugins.sbt
raboof: 1.0.1 doesn't (easily) allow adding external generators like play-grpc, so I figured it'd be a bit safer to stick with 1.0.0 for now.

ennru: It would be great to update [the docs](https://doc.akka.io/docs/alpakka/2.0.1/google-cloud-pub-sub-grpc.html#build-setup), as well.

ennru:
A left-over reference to `JavaAgent` breaks the build.

ennru (approved):

PR #2316, opened 2020-05-15
"Canonicalizing" text flow

Voltir:
@ennru I just quickly through together this PR to give a starting point for discussion about what a canonicalizing text flow might look like. This is effectively just a direct port of what we are doing right now, but we made some choices which might not be the best choices for a wider audience.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -471,6 +471,13 @@ object Dependencies {
       )
   )
 
+  val Text = Seq(
+    libraryDependencies ++= Seq(
+      "com.ibm.icu" % "icu4j" % "67.1",
+      "com.ibm.icu" % "icu4j" % "67.1",
+    )
+  )
+
   val UnixDomainSocket = Seq(
     libraryDependencies ++= Seq(
         "com.github.jnr" % "jffi" % "1.2.23", // classifier "complete", // Is the classifier needed anymore?
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -473,8 +473,7 @@ object Dependencies {
 
   val Text = Seq(
     libraryDependencies ++= Seq(
-      "com.ibm.icu" % "icu4j" % "67.1",
-      "com.ibm.icu" % "icu4j" % "67.1",
+      "com.ibm.icu" % "icu4j" % "67.1"
     )
   )
 
--- a/text/src/main/scala/akka/stream/alpakka/text/impl/ByteOrderMark.scala
+++ b/text/src/main/scala/akka/stream/alpakka/text/impl/ByteOrderMark.scala
@@ -1,0 +1,24 @@
+package akka.stream.alpakka.text.impl
+
+import akka.util.ByteString
+
+// Todo: this is copied from alpakka-csv - need to decide what to do with it
+object ByteOrderMark {
+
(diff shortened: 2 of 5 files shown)


lightbend-cla-validator:

project/Dependencies.scala
```diff
@@ -471,6 +471,12 @@ object Dependencies {
       )
   )
 
+  val Text = Seq(
+    libraryDependencies ++= Seq(
+      "com.ibm.icu" % "icu4j" % "67.1"
```
ennru, line 476: Do you know which license applies to this?
How "big" is it, would it make sense to require users to add it explicitly by marking it `provided`?

text/src/main/scala/akka/stream/alpakka/text/impl/ByteOrderMark.scala
```diff
@@ -0,0 +1,24 @@
+package akka.stream.alpakka.text.impl
+
+import akka.util.ByteString
+
+// Todo: this is copied from alpakka-csv - need to decide what to do with it
```
ennru, line 5: We can have this in `akka.stream.alpakka.text` and the same in Alpakka CSV.

project/Dependencies.scala
Voltir: According to maven, the licences is: https://raw.githubusercontent.com/unicode-org/icu/master/icu4c/LICENSE
Which, admittedly, I have not read.

Looking at https://mvnrepository.com/artifact/com.ibm.icu/icu4j/67.1, icu4j itself has no additional dependencies, which is good, and the entire api seems reasonably small: https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/

Given that, I don't think I would want to mark it as provided, but maybe there is argument to be made.

text/src/main/scala/akka/stream/alpakka/text/impl/CharsetCanonicalizingFlow.scala
```diff
@@ -0,0 +1,75 @@
...
+import akka.annotation.InternalApi
+import akka.stream.scaladsl.{Flow, Source}
+import akka.util.ByteString
+
+@InternalApi
+private[text] class CharsetCanonicalizingFlow(outgoing: Charset) {
+
+  // see: https://en.wikipedia.org/wiki/Byte_order_mark
+  private val ensureByteOrderMark: Flow[ByteString, ByteString, NotUsed] =
+    Flow[ByteString].prefixAndTail(1).flatMapConcat {
```
Voltir, line 19: @ennru How do you feel about using the scaladsl flows here? I see the other implementations are implemented in GraphStageLogic and was wondering if we should rewrite this in terms of graphstage logic directly.

ennru: Use the least powerful tool that does the job. This looks good to me.

project/Dependencies.scala
ennru: The license is OK for us. The jar is 12 MB in size. Alpakka Text is 23 kB.
I'm OK with depending on it, people concerned about library sizes can exclude it if needed.

Voltir:
Going to revisit this laster

PR #2318, opened 2020-05-18
Slick/JDBC: Support `PreparedStatement` use in Java DSL

ihostage:
Extend Java DSL for using `java.sql.PreparedStatement`

References #2247

### DIFF
--- a/slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
+++ b/slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
@@ -4,20 +4,25 @@
 
 package akka.stream.alpakka.slick.javadsl
 
+import java.sql.Connection
+import java.sql.PreparedStatement
 import java.util.concurrent.CompletionStage
-import java.util.function.{BiFunction => JBiFunction, Function => JFunction}
+import java.util.function.{Function => JFunction}
+import java.util.function.{BiFunction => JBiFunction}
 
-import scala.compat.java8.FunctionConverters._
-import scala.compat.java8.FutureConverters._
 import akka.Done
 import akka.NotUsed
+import akka.japi.function.Function2
+import akka.stream.alpakka.slick.scaladsl.{Slick => ScalaSlick}
 import akka.stream.javadsl._
 import slick.dbio.DBIO
 import slick.jdbc.GetResult
 import slick.jdbc.SQLActionBuilder
 import slick.jdbc.SetParameter
-import akka.stream.alpakka.slick.scaladsl.{Slick => ScalaSlick}
+import slick.jdbc.SimpleJdbcAction
 
+import scala.compat.java8.FunctionConverters._
+import scala.compat.java8.FutureConverters._
 import scala.concurrent.ExecutionContext
 
 object Slick {
@@ -54,7 +59,7 @@ object Slick {
    *           those statements against the specified Slick database.
    *
    * @param session The database session to use.
-   * @param toStatement A function that creeates the SQL statement to
+   * @param toStatement A function that creates the SQL statement to
    *                    execute from the current element. Any DML or
    *                    DDL statement is acceptable.
    */
@@ -71,10 +76,27 @@ object Slick {
    *           those statements against the specified Slick database.
    *
    * @param session The database session to use.
+   * @param toStatement A function that creates the SQL statement to
+   *                    execute from the current element. Any DML or
+   *                    DDL statement is acceptable.
+   */
+  def flow[T](
+      session: SlickSession,
+      toStatement: Function2[T, Connection, PreparedStatement]
+  ): Flow[T, java.lang.Integer, NotUsed] =
+    flow(session, 1, toStatement)
+
+  /**
+   * Java API: creates a Flow that takes a stream of elements of
+   *           type T, transforms each element to a SQL statement
+   *           using the specified function, and then executes
+   *           those statements against the specified Slick database.
+   *
+   * @param session The database session to use.
    * @param parallelism How many parallel asynchronous streams should be
    *                    used to send statements to the database. Use a
    *                    value of 1 for sequential execution.
-   * @param toStatement A function that creeates the SQL statement to
+   * @param toStatement A function that creates the SQL statement to
    *                    execute from the current element. Any DML or
    *                    DDL statement is acceptable.
    */
@@ -85,7 +107,31 @@ object Slick {
   ): Flow[T, java.lang.Integer, NotUsed] =
     ScalaSlick
       .flow[T](parallelism, toDBIO(toStatement))(session)
-      .map(Int.box(_))
+      .map(Int.box)
+      .asJava
+
+  /**
+   * Java API: creates a Flow that takes a stream of elements of
+   *           type T, transforms each element to a SQL statement
+   *           using the specified function, and then executes
+   *           those statements against the specified Slick database.
+   *
+   * @param session The database session to use.
+   * @param parallelism How many parallel asynchronous streams should be
+   *                    used to send statements to the database. Use a
+   *                    value of 1 for sequential execution.
+   * @param toStatement A function that creates the SQL statement to
+   *                    execute from the current element. Any DML or
+   *                    DDL statement is acceptable.
+   */
+  def flow[T](
+      session: SlickSession,
+      parallelism: Int,
+      toStatement: Function2[T, Connection, PreparedStatement]
+  ): Flow[T, java.lang.Integer, NotUsed] =
+    ScalaSlick
+      .flow[T](parallelism, toDBIO(toStatement))(session)
+      .map(Int.box)
       .asJava
 
   /**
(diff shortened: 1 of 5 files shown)


slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
```diff
@@ -221,4 +365,11 @@ object Slick {
   private def toDBIO[T](javaDml: JFunction[T, String]): T => DBIO[Int] = { t =>
     SQLActionBuilder(javaDml.asScala(t), SetParameter.SetUnit).asUpdate
   }
+
+  private def toDBIO[T](javaDml: JBiFunction[T, Connection, PreparedStatement]): T => DBIO[Int] = { t =>
+    SimpleJdbcAction { ctx =>
```
ihostage, line 370: I need help with this. I do not very know a Slick and I'm not sure about transactionality every `SimpleJdbcAction`.
It works in tests and our application, but I'm not sure that it's absolutely correct.

ennru:
Thank you for suggesting this, it's a great improvement for the Java DSL.

I wonder if there is a way to create the prepared statement only once and update it for every element.

Maybe @renatocaval can have a look?

ihostage:
> I wonder if there is a way to create the prepared statement only once and update it for every element.

Yep, I thought about that too. But `PreparedStatement` bound to `Connection` and It's a really problem for us 😞 
https://stackoverflow.com/questions/4647476/reusing-of-a-preparedstatement-between-methods

Reuse `PreparedStatement` very easy, when we have a loop and one connection 😄 But it's not our case 🤷‍♂️ 
Besides, the user can use an external `Map<Connection, PreparedStatement>` for storing statements and reuse them if it really wants to do that.

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
```diff
@@ -88,6 +109,30 @@ object Slick {
...
+   *                    DDL statement is acceptable.
+   */
+  def flow[T](
+      session: SlickSession,
+      parallelism: Int,
+      toStatement: JBiFunction[T, Connection, PreparedStatement]
+  ): Flow[T, java.lang.Integer, NotUsed] =
+    ScalaSlick
+      .flow[T](parallelism, toDBIO(toStatement))(session)
+      .map(Int.box(_))
```
renatocaval, line 133:
```suggestion
      .map(Int.box)
```

(same as above)

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
```diff
@@ -88,6 +109,30 @@ object Slick {
       .map(Int.box(_))
```
renatocaval, line 109:
```suggestion
      .map(Int.box)
```

This is giving warning in IntelliJ, not essential though

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
renatocaval: I don't get your question about transactionality. 

This DBIO will have a single statement and will run in its own transaction and not participating to any other transaction. If the user had access to it they could `flatMap` with another DBIO. In that case, they will need to add `.transactionally` to wrap both DBIO on the same tx. 

Anyway, that's not applicable here. 

However, what could be an addition is a `BiFunction[T, Connection, List[PreparedStatement]]`. 

The user could then return a few prepared statements, all bound to the same connection. 

In such a case, we would convert each of them to a DBIO and indeed execute them in the same transaction using `.transactionally`.

Can be added on another PR though.

ihostage: > This DBIO will have a single statement and will run in its own transaction and not participating to any other transaction.

Exactly this is what I wanted to clarify. Thank you! 👍

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
ihostage: ✅ Done

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
ihostage: ✅ Done

docs/src/main/paradox/slick.md
```diff
@@ -158,6 +158,8 @@ Scala
 Java
 : @@snip [snip](/slick/src/test/java/docs/javadsl/DocSnippetFlow.java) { #flow-example }
 
+Java (JDBC)
```
ennru, line 161: Please remove the old Java example. It is very bad to show this dangerous way of construction SQL statements.

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
```diff
@@ -64,6 +68,23 @@ object Slick {
...
     flow(session, 1, toStatement)
 
+  /**
+   * Java API: creates a Flow that takes a stream of elements of
+   *           type T, transforms each element to a SQL statement
+   *           using the specified function, and then executes
+   *           those statements against the specified Slick database.
+   *
+   * @param session The database session to use.
+   * @param toStatement A function that creeates the SQL statement to
```
ennru, line 78:
```suggestion
   * @param toStatement A function that creates the SQL statement to
```
(same error on the method above BTW)

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
```diff
@@ -221,4 +365,11 @@ object Slick {
   private def toDBIO[T](javaDml: JFunction[T, String]): T => DBIO[Int] = { t =>
     SQLActionBuilder(javaDml.asScala(t), SetParameter.SetUnit).asUpdate
   }
+
+  private def toDBIO[T](javaDml: JBiFunction[T, Connection, PreparedStatement]): T => DBIO[Int] = { t =>
+    SimpleJdbcAction { ctx =>
+      val statement = javaDml(t, ctx.connection)
+      if (statement != null) statement.executeUpdate() else 0
```
ennru, line 372: This becomes a hidden `filter` feature, but makes sense as a safe-guard.

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
```diff
@@ -64,6 +68,23 @@ object Slick {
...
+   *           those statements against the specified Slick database.
+   *
+   * @param session The database session to use.
+   * @param toStatement A function that creeates the SQL statement to
+   *                    execute from the current element. Any DML or
+   *                    DDL statement is acceptable.
+   */
+  def flow[T](
+      session: SlickSession,
+      toStatement: JBiFunction[T, Connection, PreparedStatement]
```
ennru, line 84: Use `akka.japi.Function2` instead, so the `SQLException` can be thrown and fail the stream properly.

docs/src/main/paradox/slick.md
ihostage: If we want not to show this approach in docs, maybe needly also to mark these API as deprecated? WDYT?

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
ihostage: OMG. I fixed this in other places, but itself copy&past incorrect version 🤦

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
ihostage: Yep. I thought about NPE-safety and I haven’t come up with anything better like this 🤷‍♂️

docs/src/main/paradox/slick.md
ennru: It is still fine if the user knows what they are doing. Eg. when the data is properly typed or validated.
We can leave it as is, but not recommend it.

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
ennru: But it becomes a dangerous mix with the failure handling you suggested in the example...

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
ihostage: ✅ Done

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
ihostage: ✅ Done

docs/src/main/paradox/slick.md
ihostage: ✅ Done

slick/src/main/scala/akka/stream/alpakka/slick/javadsl/Slick.scala
ihostage: I removed this check, rewritten examples in docs, and added test with `recover` function.
Please, review.

ennru:
Thank you @ihostage!

PR #2319, opened 2020-05-18
Fix intermittent issues with Wiremock connection reset simulation

bpiper:
References #2209

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -390,7 +390,7 @@ object Dependencies {
         "software.amazon.awssdk" % "auth" % AwsSdk2Version,
         // in-memory filesystem for file related tests
         "com.google.jimfs" % "jimfs" % "1.1" % Test, // ApacheV2
-        "com.github.tomakehurst" % "wiremock" % "2.25.1" % Test // ApacheV2
+        "com.github.tomakehurst" % "wiremock-jre8" % "2.26.3" % Test // ApacheV2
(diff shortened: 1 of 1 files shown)


bpiper:
As with previous PR, I'll run a bunch of Travis builds before considering this RTBM (though there seems to be some special property of `master` builds that causes the issue in question to manifest, unlike branch builds).

s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3WireMockBase.scala
```diff
@@ -955,8 +955,8 @@ object S3WireMockBase {
     |  path-style-access = false
     |  endpoint-url = "http://localhost:$proxyPort"
     |  retry-settings {
-    |    min-backoff = 0s
-    |    max-backoff = 0s
+    |    min-backoff = 100ms
+    |    max-backoff = 300ms
```
raboof, line 959: Was this intentionally part of this PR?

bpiper: Oops, my bad... forgot to rebase after the last PR was merged.  Fixed.

bpiper:
Probably, but I'll run a few more builds and then reset to remove the trigger commits.

bpiper:
RTBM now from my perspective.

raboof:
Thanks! Let's see how this fares on master ;)

Issue #2226, opened 2020-03-23
Cassandra: remove Netty dependency override


ennru:
As there is some change in Netty 4.1.45 which made throughput drop, Alpakka Cassandra downgrades Netty to 4.1.39.

References 
* https://github.com/akka/akka-persistence-cassandra/pull/730
* https://datastax-oss.atlassian.net/browse/JAVA-2676

patriknw:
Until this has been fixed we will have a security vulnerability reported by Whitesource on the Netty 4.1.39 dependency.

chbatey:
https://github.com/akka/alpakka/pull/2320

ennru:
Fixed with #2320

PR #2320, opened 2020-05-19
Cassandra: Driver 4.6.1 (Remove netty override)

chbatey:
Underlying perf issue has been fixed


ennru:
@chbatey What do we know about bin-compatibility between Cassandra 4.5 and 4.6?
Should we consider this change for a patch version?

chbatey:
Only major version break binary compatibility AFAIK

ennru:
Fixes #2226

PR #2325, opened 2020-05-22
file/tar: Allow space as terminal character in tar header

### LINKED ISSUE #2324: Handle space character as termination for header fields in tar parser
### Short description

The current tar header parser only allows for header fields to be terminated by the null character (`\u0000`). However, [tar documentation](https://www.freebsd.org/cgi/man.cgi?query=tar&apropos=0&sektion=5&manpath=FreeBSD+7.0-RELEASE&arch=default&format=html) states that some fields may either be terminated by a space character or a null character. 

When calling `Archive.tarReader()` on a tar file using space as terminal character for some fields (specifically `fileSizeLength` and `lastModificationLength`), I get a `NumberFormatException` as a string on the form `00000001760 ` is sent to `parseUnsignedLong`.

### Details

Should be easy to fix by letting `TarArchiveEntry.getString()` stop at _either_ a null character or a space character, instead of only a null character.

sberglann:
Fixes #2324


ennru:
Thank you for fixing this.

PR #1844, opened 2019-07-27
[elasticsearch] Prepend the index name to the url in _bulk operations when necessary

Tommmster:
## Purpose
Enables the use of the Elasticsearch Sink in non-default environments where url-based access control is performed.

## References

References #1843 

## Changes

* Adding two fields to `ElasticsearchFlowStage [T,C]`: The actual endpoint to which the `_bulk` request will be performed and a method that includes (or not ) the index name on each document. The Stage contains all necessary information at construct-time.
 
## Background Context

Current implementation of the Elastic search Sink will fail (requests rejected) if the cluster configuration does not hold the default setting for the `rest.action.multi.allow_explicit_index` property, which is common when url-based access control is performed (multi-tenancy)

These changes enabled the elasticsearch sink to work in these conditions, while preserving the current behaviour by default.

PR #2326, opened 2020-05-22
[elasticsearch] Support `allow_explicit_index: false`

mrooding:
Purpose
---
Enables the use of the Elasticsearch Sink in non-default environments where url-based access control is performed.

This PR is an update of #1844 which seems to be abandoned.

References
---
References #1843 

Changes
---
`ElasticsearchWriteSettings`: Added an aditional setting `allowExplicitIndex`
`ElasticsearchSimpleFlowStage[T, C]`: The actual endpoint to which the _bulk request is determined based on the newly introduced `ElasticSearchWriteSetting.allowExplicitIndex` 
`RestBulkApi` and `RestBulkApiV5` and `RestBulkApiV7`: Support writing the shared fields to the API based on the `allowExplicitIndex` setting

Background Context
---
Current implementation of the Elastic search Sink will fail (requests rejected) if the cluster configuration does not hold the default setting for the rest.action.multi.allow_explicit_index property, which is common when url-based access control is performed (multi-tenancy)

These changes enabled the elasticsearch sink to work in these conditions, while preserving the current behaviour by default.


ennru:
Thank you for warming this up and making it happen!

Issue #2314, opened 2020-05-14
[Elasticsearch] Make ElasticsearchSourceStage Delete Its Scroll On Stage Completion


michaeljmarshall:
### Short description

Add support for the `ElasticsearchSourceStage` to call Elasticsearch to clear its scroll upon stream completion. As shown in the [Elasticsearch scroll documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#_clear_scroll_api) for versions 5, 6, and 7:

> Search context are automatically removed when the scroll timeout has been exceeded. However keeping scrolls open has a cost, as discussed in the previous section so scrolls should be explicitly cleared as soon as the scroll is not being used anymore using the clear-scroll API

This feature request amounts to asynchronous resource cleanup where the scroll is the resource and calling the delete scroll endpoint is the cleanup.

### Details

The call to delete a scroll is simple: `DELETE /_search/scroll/${scroll_id}`.

I see only 3 events that cause the `ElasticsearchSourceStage` to terminate:

1. The scroll completes normally.
1. Downstream finishes.
1. The call to get the next set of scroll hits fails for one of two reasons:
   a. The scroll no longer exists.
   b. Some other exception.

In the case of 1, 2, and 3.b, the stage should attempt to clear the scroll. In the case of 3.a, the scroll is already cleared, so there is no need to delete it.

In second case, I'm not sure how the graph framework handles asynchronous cleanup. I know of the `postStop` lifecycle method, but based on some brief reading through the source code, that does not look like it is designed to handle asynchronous cleanup. Is there a common pattern for asynchronous resource clean up in the case of down stream finish? Even if the scroll has to live in that case, I would still want to delete the scroll for the other cases. Note that in the case of the scroll completing normally (1) or the Elasticsearch client returning an exception (3.b), it'd be easy enough to attempt the async cleanup before completing the stage.

Because the `ElasticsearchSourceStage` privately maintains the `scroll_id`, there is currently no good way for end users to clear the scroll. Further, when the stage completes, the `scroll_id` is gone, so there is no case where the scroll should persist beyond the stage's lifecycle.

If this new feature gains traction, I'm be happy to help implement it.

ennru:
Fixed with #2330.

PR #2330, opened 2020-05-28
ElasticsearchSource: Clear Scroll On Exhaustion or Completion

michaeljmarshall:
References #2314

In the referenced issue, I had asked about the proper way to keep the stage alive for clean up when downstream has already finished. I noticed that some other stages were doing this by calling `setKeepGoing(true)`, so I used that here. The documentation seems to support the use case, but please let me know if it looks incorrect.

I plan to add tests and documentation in the coming days. I wanted to get this PR up first to see if there was any feedback on the initial implementation.

### DIFF
--- a/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchSourceStage.scala
+++ b/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchSourceStage.scala
@@ -18,6 +18,7 @@ import spray.json.DefaultJsonProtocol._
 import spray.json._
 
 import scala.collection.JavaConverters._
+import scala.util.{Failure, Success, Try}
 
 /**
  * INTERNAL API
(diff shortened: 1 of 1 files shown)


ennru:
Thank you for trying it out.
Your use of `setKeepGoing` is correct.
But I wonder if there is a good reason to do await the response from the delete request, we can't do anything but log it anyway. So alternatively this would be just a fire and forget and the stage can complete immediately.

michaeljmarshall:
My primary motivation for awaiting the response from the delete request is a clean shutdown. When I stop the stream early with a `KillSwitch`, I want to make sure that the source stage properly makes the delete call before letting the application shutdown. Given that there is a real cost to keeping a scroll open on the Elasticsearch side (and that the Elasticsearch documentation explicitly recommends clearing scrolls when they're no longer needed), it seems worth the slight delay to ensure that the call is made.

Regarding the logging, I do think it could be valuable for troubleshooting.

If you think awaiting the result of the delete request is too opinionated, I could add a setting to the `ElasticsearchSourceSettings` to allow end users to choose the right behavior for their use case.

ennru:
Thank you @michaeljmarshall, great to see Alpakka Elasticsearch being improved!

Issue #2315, opened 2020-05-14
ShardSettings API doesn't allow for unordered configuration


vaslabs:
### alpakka kinesis
If you want to set the iterator to AFTER_SEQUENCE_NUMBER, AT_SEQUENCE_NUMBER or
AT_TIMESTAMP you have to set first the value of the sequence number (or timestamp). Which
means that simple mistakes can be made which can be avoided.

### Details

We can define a ShardIterator ADT which accepts its dependencies if any and doesn't require subsequent methods to set it up.
E.g. `withShardIterator(ShardIterator.AfterSequenceNumber(sequence))` which will avoid this runtime issue.

If you agree please assign the ticket to me and I can follow up with a pull request.

Source: https://github.com/akka/alpakka/blob/master/kinesis/src/main/scala/akka/stream/alpakka/kinesis/ShardSettings.scala#L25

ennru:
Fixed with #2331

PR #2331, opened 2020-05-29
Kinesis: Construction safe API

vaslabs:
References #2315 

I'd also suggest to mark the following methods as deprecated:
```scala
def withShardIteratorType
def withStartingSequenceNumber
def withStartingAfterSequenceNumber
def withAtTimestamp
```

Although technically the latter 3 are safe, it is not visible in their type that they are also changing the shard iterator type and I've observed people making the mistake of setting the shard iterator type first anyway  (including myself :P)

Please let me know what you think.

Thanks


vaslabs:
thanks, just noticed the deprecation fatal warnings, thanks for fixing it :+1:

PR #2332, opened 2020-06-03
Fail LogRotatorSink on failures in the sinks

shagoon:
I would expect the stream to be terminated with an exception, if the sink created by `sinkFactory` of `LogRotatorSink.withSinkFactory` throws an exception. This is not the case. I added a test case in `LogRotatorSinkSpec` that demonstrates this.

The issue initially occured while trying to use `LogRotatorSink` together with `S3.multipartUpload`. My aws-config was not correct, leading to Exceptions being thrown in `S3.multipartUpload`-Sink, which never showed up and left the stream in a non-functional, but not terminated condition. I had to remove `LogRotatorSink` and use `S3.multipartUpload` directly to see those exceptions and have the stream terminated.

### DIFF
--- a/file/src/main/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSink.scala
+++ b/file/src/main/scala/akka/stream/alpakka/file/scaladsl/LogRotatorSink.scala
@@ -156,6 +156,14 @@ final private class LogRotatorSink[T, C, R](triggerGeneratorCreator: () => T => 
       pull(in)
     }
 
+    override def postStop(): Unit =
+      promise.tryCompleteWith {
+        implicit val ec = materializer.executionContext
+        Future
+          .sequence(sinkCompletions)
+          .map(_ => Done)(akka.dispatch.ExecutionContexts.sameThreadExecutionContext)
+      }
+
     def futureCB(newFuture: Future[R]) =
       getAsyncCallback[Holder[R]](sinkCompletionCallbackHandler(newFuture))
 
--- a/file/src/test/scala/docs/scaladsl/LogRotatorSinkSpec.scala
+++ b/file/src/test/scala/docs/scaladsl/LogRotatorSinkSpec.scala
@@ -358,7 +358,7 @@ class LogRotatorSinkSpec
             sinkFactory = (_: Path) =>
               Flow[ByteString]
                 .map { data =>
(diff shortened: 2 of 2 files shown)


ennru:
Thank you, that is nasty indeed. Will you have a try on fixing it?

shagoon:
I already had a look at this. I'm not sure, if I can handle that, though.

By using a debugger, I could see my exception arriving at `LogRotatorSink`'s `newFuture` and propagated through its `onComplete` method to the `holder` (`apply`). The `Holder`'s `cb.invoke(this)` did _not_ invoke the callback `sinkCompletionCallbackHandler`, though.

I could _not_ use the debugger to step through the internals of `ConcurrentAsyncCallback`'s `invoke` method, since the debugger's lines did not match the code (i.e. it stopped inside large comments above a method).

`LogRotatorSink`'s usage of `futureCb` is quite different than those of other usages of a `Holder` throughout the code. Since the value of `newFuture` is being propagated into the holder itself, I don't see a reason why `sinkCompletionCallbackHandler` needs to bother with _both_, the holder and `newFuture`. But since that code never got called, I couldn't debug through that either.

ennru:
You're right, something gets stuck so the async callback doesn't get executed. It is not obvious to me what that might be.

jrudolph:
I had a quick look. I think the issue is that the stage is already completed when the async callback is delivered. The whole setup seems quite prone to race conditions with the `LogRotatorSink` being subscribed to both the beginning (via substream OutHandler) and the end of the user-supplied sinks (via materialized completion futures). From the scaladoc / documentation it's not clear to me why it was done that way, probably to be able to report errors in the sink at all?

With Akka 2.6, a better approach is to use the new cancellation propagation facilities to report that error.

Fixing it here requires to make sure that all the outstanding `sinkCompletions` are directly bound to the promise (using `promise.completeWith`) when the stage shuts down.

jrudolph:
E.g. this `postStop` fixes the issue:

```scala
    override def postStop(): Unit =
      promise.tryCompleteWith {
        implicit val ec = materializer.executionContext
        Future
          .sequence(sinkCompletions)
          .map(_ => Done)(akka.dispatch.ExecutionContexts.sameThreadExecutionContext)
      }
```

ennru:
Thank you for identifying the problem!

PR #2333, opened 2020-06-03
Removed obsolete reference config for Alpakka DynamoDB

vigoo:
The configuration keys in the DynamoDB library's `reference.conf` are no longer used since the move to the DynamoDB v2 API in Alpakka 2.0. Keeping them is confusing for the library users trying to figure out available options.


ennru:
Thanks for keeping it tidy.

Issue #2310, opened 2020-05-11
S3Exception MalFormed : akka.stream.alpakka.s3.S3Exception:  (Code: , RequestID: , HostID: )


GhaileneSouissi:
Akka version:
```
"com.typesafe.akka" %% "akka-http" % "10.1.10",
"com.typesafe.akka" %% "akka-stream" % "2.5.25",
"com.typesafe.akka" %% "akka-http-xml" % "10.1.10"
```

Alpakka version :
`"com.lightbend.akka" %% "akka-stream-alpakka-s3" % "2.0.0"`

Hello, I am using alpakka S3 to upload files into amazon S3 storage system. The implementation I made is simple :

```
 def s3Settings(implicit system: ActorSystem) = S3Ext.get(system).settings
    .withCredentialsProvider(credentielProvider)
    .withS3RegionProvider(regionProvider)
    .withBufferType(MemoryBufferType)
    .withListBucketApiVersion(ListBucketVersion2)
    .withForwardProxy(ForwardProxy("endpoint",Port , Some(ForwardProxyCredentials("username", "password"))))
    .withEndpointUrl(endpointUrl)
    .withPathStyleAccess(true)
```

I am using docker to simulate an S3 system (minio) and everything worked just fine.

After adding a proxy and changing the url to connect to my S3 system in production,  it works when i download files and delete files from S3. but when i try to upload files, it throws this exception :

`akka.stream.alpakka.s3.S3Exception: (Code: , RequestID: , HostID: )`

Is there any way to have more information in the exception ?

ennru:
Looking at `S3Exception` I get the impression it will never parse the returned error correctly:

https://github.com/akka/alpakka/blob/5534c616c3c654d26b7f535287943bd1d38b398b/s3/src/main/scala/akka/stream/alpakka/s3/S3Exception.scala#L15-L33

Please suggest an improvement to this error handling.

ihrimech:
Hi @ennru 

I would like to contribute to this issue but I'll need some guidance since doing it for the first time. What exactly made you think that the parsing would never be made correctly ?

ennru:
Thank you, I looked into it and created #2341.
Can you try it with your setup and see if it improves your situation, please?

ennru:
Fixed with #2341

PR #2341, opened 2020-06-09
AWS S3: parse error replies

ennru:
* Simplify construction of the exception when the XML can't be parsed.
* Always keep the HTTP status code.
* According to the [docs](https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html) it should be `RequestId` not `RequestID` as the earlier code did.
* `HostID` doesn't exist anymore(?), but `Resource` is there

References #2310

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
@@ -234,13 +234,18 @@ import scala.util.{Failure, Success, Try}
               Source.fromFuture(entity.discardBytes().future().map(_ => None))
             case HttpResponse(code, _, entity, _) =>
               Source.fromFuture {
-                Unmarshal(entity).to[String].map { err =>
-                  throw new S3Exception(err, code)
-                }
+                unmarshalError(code, entity)
               }
           }
       }
       .mapMaterializedValue(_ => NotUsed)
+
+  private def unmarshalError(code: StatusCode, entity: ResponseEntity)(implicit mat: Materializer) = {
+    import mat.executionContext
+    Unmarshal(entity).to[String].map { err =>
+      throw S3Exception(err, code)
+    }
+  }
 
   def deleteObject(s3Location: S3Location, versionId: Option[String], s3Headers: S3Headers): Source[Done, NotUsed] =
     Source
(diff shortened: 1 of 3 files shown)


s3/src/test/scala/akka/stream/alpakka/s3/S3ExceptionSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+              |  <Code>NoSuchKey</Code>
+              |  <Message>The resource you requested does not exist</Message>
+              |  <Resource>/mybucket/myfoto.jpg</Resource>
+              |  <RequestId>4442587FB7D0A2F9</RequestId>
+              |</Error>""".stripMargin
+    val e = S3Exception(s, StatusCodes.NotFound)
+    e.code shouldBe "NoSuchKey"
+    e.message shouldBe "The resource you requested does not exist"
+    e.requestId shouldBe "4442587FB7D0A2F9"
+    e.resource shouldBe "/mybucket/myfoto.jpg"
```
jrudolph, line 30: Maybe check the toString here as well.

s3/src/test/scala/akka/stream/alpakka/s3/S3ExceptionSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+ * Copyright (C) 2016-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.s3
+
+import akka.http.scaladsl.model.StatusCodes
+import org.scalatest.flatspec.AnyFlatSpecLike
+import org.scalatest.matchers.should.Matchers
+
+class S3ExceptionSpec extends AnyFlatSpecLike with Matchers {
```
jrudolph, line 11: :+1:

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -234,14 +234,19 @@ import scala.util.{Failure, Success, Try}
...
-                }
+                unmarshalError(code, entity)
               }
           }
       }
       .mapMaterializedValue(_ => NotUsed)
 
+  private def unmarshalError(code: StatusCode, entity: ResponseEntity)(implicit mat: Materializer) = {
+    import mat.executionContext
+    Unmarshal(entity).to[String].map { err =>
```
jrudolph, line 245: A bit weird way to use the marshaller but still better than all the duplication from before.

Issue #2338, opened 2020-06-08
alpakka-json-streaming cannot be used from Maven as-is


laszlovandenhoek:
### Versions used 

2.0.0

### Expected Behavior

Adding `akka-stream-alpakka-json-streaming_2.xx` as a Maven dependency works out of the box.

### Actual Behavior

Given a `pom.xml` like this:

```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>
    <groupId>test</groupId>
    <artifactId>test</artifactId>
    <version>0.0.1-SNAPSHOT</version>

    <dependencies>
        <dependency>
            <groupId>com.lightbend.akka</groupId>
            <artifactId>akka-stream-alpakka-json-streaming_2.13</artifactId>
            <version>2.0.0</version>
        </dependency>
    </dependencies>
</project>
```
Running `mvn verify` from the command line yields this error:

`[ERROR] Failed to execute goal on project test: Could not resolve dependencies for project test:test:jar:0.0.1-SNAPSHOT: Could not find artifact com.github.jsurfer:jsurfer:jar:1.6.0 in nexus`

The issue is that there __is__ no JAR, only a POM. However, in the published `pom.xml` for `alpakka-json-streaming`, the dependency on `com.github.jsurfer:jsurfer:jar:1.6.0` is not qualified as a `pom` dependency, and thus assumed to be a JAR. It is specified here: 

https://github.com/akka/alpakka/blob/master/project/Dependencies.scala#L301

ennru:
I'm afraid sbt can not express this correctly.

ennru:
Can we depend on the libraries it points to directly?

laszlovandenhoek:
Until the PR is merged and a new version is released, you can work around the problem by excluding the erroneous dependency like so:

```
<dependency>
    <groupId>com.lightbend.akka</groupId>
    <artifactId>akka-stream-alpakka-json-streaming_2.13</artifactId>
    <version>2.0.0</version>
    <exclusions>
        <exclusion>
            <groupId>com.github.jsurfer</groupId>
            <artifactId>jsurfer</artifactId>
        </exclusion>
[...]

PR #2344, opened 2020-06-11
json-streaming: fix generated POM

laszlovandenhoek:
Removes the explicit dependency on `jsurfer`, which is of type `pom`, but is implicitly specified as being of type `jar` in the generated POM for `alpakka-json-streaming`, making the latter unusable from Maven (#2338).


ennru:
Thank you for noticing and solving it!

Issue #2308, opened 2020-05-10
Elasticsearch - Method signature of performRequest and performRequestAsync changed in ES 7.x


renphi:
### Versions used 

Akka version: 2.6.5
Alpakka Verison: 2.0.0
Elasticsearch Version: 7.6.2

### Expected Behavior

Can perform request against elasticsearch.

### Actual Behavior

Can't perform request against elasticsearch.

#### ES Documentation
There is a new way of using the performRequestAsync Method (I guess it was changed in ES 6.4 - marked as depricated and dropped in ES 7.0) https://www.elastic.co/guide/en/elasticsearch/client/java-rest/7.x/java-rest-low-usage-requests.html vs. https://www.elastic.co/guide/en/elasticsearch/client/java-rest/6.3/java-rest-low-usage-requests.html

#### ES Breaking Changes (Version 6.4)
https://www.elastic.co/guide/en/elasticsearch/reference/6.4/breaking-changes-6.4.html#breaking_64_rest_client_changes

#### PR when old request methods were removed (Version 7.0) 
https://github.com/elastic/elasticsearch/pull/33223/commits/36a3e9c1c2ea109cba6ff1055703e42e7522b75f

### Possible solution
We identify this occurrences:
https://github.com/akka/alpakka/blob/fefc985d535f0f03ee699371139bb8543d7b1392/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchSourceStage.scala#L125
https://github.com/akka/alpakka/blob/fefc985d535f0f03ee699371139bb8543d7b1392/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchSourceStage.scala#L136
https://github.com/akka/alpakka/blob/fefc985d535f0f03ee699371139bb8543d7b1392/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchSimpleFlowStage.scala#L75 

Implement version check like used for RestBulkApi(V5|V7)

ES >=6.4 & 7.x version:
 https://www.elastic.co/guide/en/elasticsearch/client/java-rest/7.x/java-rest-low-usage-requests.html
```scala
val req = new Request("POST", endpoint)
        req.setOptions(RequestOptions.DEFAULT.toBuilder.addHeader("Content-Type", "application/json"))
        req.setEntity(new StringEntity(searchBody, StandardCharsets.UTF_8))
[...]

PR #2346, opened 2020-06-18
Update the elasticsearch-rest-client to 7.7.1

mrooding:
# Purpose

Update the elasticsearch-rest-client to the latest 7.x release

# References

This'll solve the method signature change introduces in elasticsearch-rest-client 6.4 as described in #2308

# Changes

- Updated all `performRequest` and `performRequestAsync` calls in both the source and tests
- Updated the ElasticSearch Docker images used in the `docker-compose.yml`

# Context

I've updated the elasticsearch-rest-client to the latest available version, being 7.7.1. I think there is no harm doing so since the ES low-level client in no way binds itself to a particular ES version.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -74,13 +74,13 @@ services:
     ports:
       - "8001:8000"
   elasticsearch6:
-    image: docker.elastic.co/elasticsearch/elasticsearch:6.8.0
+    image: docker.elastic.co/elasticsearch/elasticsearch:6.8.10
     ports:
       - "9201:9200"
     environment:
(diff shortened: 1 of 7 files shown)


mrooding:
The PR is labeled with mqqt-streaming because Intellij automatically added an empty line to one of the files related to mqqt-streaming. I've reverted that change so the label can be removed

ennru:
As the Elasticsearch client library changes the API, we can't merge this into the Alpakka 2.x line.

ennru:
With #2347 shaping up nicely, I'll close this.

PR #2347, opened 2020-06-22
ElasticSearch: Replace RestClient with Akka HTTP client

mrooding:
# Purpose

Perform all HTTP calls to ElasticSearch using an Akka HTTP client instead of using the low level elasticsearch-rest-client.

# References

#2308

# Changes

- Removed the dependency on elasticsearch-rest-client and introduced a dependency on akka-http and akka-http-spray-json
- Updated all API calls to ES to use akka-http

# Context

As proposed by @ennru, this PR replaces the elasticsearch-rest-client with an Akka HTTP implementation. The `ElasticsearchSimpleFlowStage` and `ElasticsearchSourceStage` have been updated but I'd like to get some early feedback on the way I've tackled it.

Some outstanding thoughts:

1. Currently, I've added the required settings to connect to ES in both the `ElasticsearchSourceSettings` and `ElasticsearchWriteSettings`. I'm not too happy about that. What about having 1 `ElasticsearchSettings` instance that looks like this:

```
ElasticsearchSettings(
  baseUrl: String,
  credentials: ElasticsearchCredentials,
  sourceSettings: ElasticsearchSourceSettings,
  writeSettings: ElasticsearchWriteSettings
)
```
2. Do we want to have a single `baseUrl` property which includes the scheme, host and port for ES or should we allow configuring them separately?
3. I've used `Source.setup` and `Flow.setup` to gain access to the materializer, actor system and execution context. Is this the best way to do so or is there an alternative?
4. I couldn't find a `ContentType` for `application/x-ndjson` in akka-http which should be used for the bulk API

I'll hold on with updating all the tests until I've got some feedback to prevent rework.

Looking forward to hearing what you think @ennru


ennru:
Thank you for giving this a try, it looks promising.

Please observe that even this approach is not compatible to go into Alpakka 2.0 releases, but is of course a great improvement to be independent in the future.

1. What about creating `ElasticsearchConnectionSettings` which becomes part of the source- and write-settings?

2. A single `endpoint-url` seems enough, AFAICS. (An extension to this would be to integrate with Akka Discovery, but that would go into a different PR.)

3. Yes, `.setup` is our best way to get hold of the `ActorSystem` without requiring it on the factory methods.

4. That will require a [custom content type](https://doc.akka.io/docs/akka-http/current/routing-dsl/source-streaming-support.html#supporting-custom-content-types).

mrooding:
Hi @ennru 

Thanks for the quick feedback. I'm aware of the fact that it won't be available in Alpakka 2 but it's the best long term solution there is so definitely worth it.

I'll take your comments into account and continue working on the PR 👍

mrooding:
@ennru I finally found some time to wrap this up. The biggest issue I ran into was the `require` checks which would get swallowed by wrapping the instantiation of the SourceStage and FlowStage with `Source.setup`. I liked how the API threw an exception without forcing the user to use `recover` on the stream itself. To maintain the immediate feedback, I decided to wrap the index and type in a case class which is instantiated outside of the `Source.setup` blocks. 

Since ES is deprecating type mappings, and since the FlowStage was already using `ApiVersion` to optionally add the `typeName`, I decided to also use this for the `ElasticsearchSourceSettings`. The `ElasticsearchSourceStage` was creating the search API endpoint based on the optional type. That's now being done based on the `ApiVersion` making both reading and writing consistent in API usage.

Let me know what you think about this approach

mrooding:
Hi @ennru 

Would you be able to take a look at the progress I made? It would be a shame if this would miss the major release planned for the end of the year.

Thank you!

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchConnectionSettings.scala
```diff
@@ -0,0 +1,43 @@
...
+    val baseUrl: String,
+    val username: Option[String],
+    val password: Option[String]
+) {
+
+  def withBaseUrl(value: String): ElasticsearchConnectionSettings = copy(baseUrl = value)
+
+  def withUsername(value: String): ElasticsearchConnectionSettings = copy(username = Option(value))
+
+  def withPassword(value: String): ElasticsearchConnectionSettings = copy(password = Option(value))
```
ennru, line 17: Set both in a single `withCredentials`?

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/javadsl/ElasticsearchSource.scala
```diff
@@ -168,4 +210,22 @@ object ElasticsearchSource {
     }
   }
 
+  /**
+   * Helper for creating BiFunction instances from Scala
+   * functions as Scala 2.11 does not know about SAMs.
```
ennru, line 215: We'll drop Scala 2.11 for Alpakka 3.0. You can switch it off for Elasticsearch.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchIndexType.scala
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (C) 2016-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.elasticsearch
+
+case class ElasticsearchIndexType(indexName: String, typeName: Option[String], apiVersion: ApiVersion) {
```
ennru, line 7: Should this become the type to set index and type instead?

ennru (commented):
This looks very good now. Actually simpler than with the REST client.

* As this breaks the API, are there other changes that are good to do now?
* For Alpakka 3.0 we should switch the API version default to V7, WDYT? (in a separate PR)

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchConnectionSettings.scala
mrooding: 👍

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchConnectionSettings.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  def hasCredentialsDefined: Boolean = username.isDefined && password.isDefined
+
+  def copy(baseUrl: String = baseUrl,
+           username: Option[String] = username,
+           password: Option[String] = password): ElasticsearchConnectionSettings =
+    new ElasticsearchConnectionSettings(baseUrl = baseUrl, username = username, password = password)
+
+  override def toString =
+    s"""ElasticsearchSourceSettings(baseUrl=$baseUrl,username=$username,password=$password)"""
```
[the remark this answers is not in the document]
mrooding: 👍

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/javadsl/ElasticsearchSource.scala
mrooding: Added `crossScalaVersions -= Dependencies.Scala211,`

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchIndexType.scala
mrooding: I'm not sure that I understand what you're asking here. Could you clarify this for me?

mrooding:
I've updated the PR to reflect the requested changes (except for the last outstanding one). After wrapping this up, I can definitely do another PR to set the default to ES 7.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchIndexType.scala
ennru: The current API takes `indexName: String, typeName: String`. For ES 7 users can ignore `typeName`.
I wonder if this class should become the parameter to pass in these fields in an API version aware way. Roughly `ElasticsearchFlow.create(Index("indexName"), settings)` instead of ``ElasticsearchFlow.create("indexName", null, settings)`.

mrooding: Thanks for clarifying. That's actually something that crossed my mind back when I added the type. Maybe it's a nice idea to add 2 builder methods to the `ElasticsearchIndexType` to make it even more transparent to users what properties are actually required for their ES version:

```
object ElasticsearchIndexType {
  def forElasticsearchV7(indexName: String): ElasticsearchIndexType =
    ElasticsearchIndexType(indexName, None, ApiVersion.V7)

  def forElasticsearchV5(indexName: String, typeName: String): ElasticsearchIndexType =
    ElasticsearchIndexType(indexName, Some(typeName), ApiVersion.V5)
}
```

What do you think?

ennru: Can we come up with something shorter?

eg `ElasticsearchFlow.create(EsParams.v5("indexName", "typeName"), settings)`

This might help if later versions would allow/require even more things.

mrooding: Sounds good 👍 Will update a.s.a.p.

mrooding:
@ennru the PR has been updated as discussed. I've set up `EsParams` in a similar way as the other setting classes. If we agree on this approach then I'll add some documentation with examples.

There's some tests failing in a different subproject but that doesn't seem to be related to my changes.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/EsParams.scala
```diff
@@ -0,0 +1,45 @@
...
+  def withIndexName(value: String): EsParams = {
+    require(value != null, "You must define an index name")
+
+    copy(indexName = value)
+  }
+
+  def withTypeName(value: String): EsParams = {
+    if (apiVersion == ApiVersion.V5) {
+      require(value != null && value.trim.nonEmpty, "You must define a type name for ElasticSearch API version V5")
+    }
```
ennru, line 17: To make this check useful an API version should always be selected prior to calling the other setters.
I'd remove the default variants (an index name must always be given) and just keep the `V5` and `V7` constructors which can use `new` directly.
That's more invasive in our tests, but normal users will not switch between different versions in their code.

Or do you see a way that we don't need to have the API version on both `EsParams` and the source settings?

mrooding: I've changed `EsParams` to only have the 2 creation methods `V5` and `V7`. The only place where `EsParams.apiVersion` was used was in the `ElasticsearchSourceStage` and since that also has access to the `ElasticsearchSourceSettings` I've decided to remove `EsParams.apiVersion` all together.

Since both Java and Scala tests run for both V5 and V7 I did have to add a helper function in the tests to construct the right `EsParams` in the tests. The biggest issue with this setup is that all the documentation examples (`run-typed`, `run-jsobject`, `run-flow`, `string`, `kafka-example`, `custom-index-name-example` and `custom-search-params`  are taken from tests that now use the helper method to enable having a single test that runs against both API versions.

I'd prefer to have all examples documented with only 1 of the 2 `EsParam` versions (V7 since that'll be the default for Alpakka 3?). Do you see a better alternative to setting this up besides duplicating all these tests so that they work with the explicit `EsParams.V5` and `EsParams.V7`?

I did already add some generic documentation on `EsParams`.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchConnectionSettings.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  def hasCredentialsDefined: Boolean = username.isDefined && password.isDefined
+
+  def copy(baseUrl: String = baseUrl,
+           username: Option[String] = username,
+           password: Option[String] = password): ElasticsearchConnectionSettings =
+    new ElasticsearchConnectionSettings(baseUrl = baseUrl, username = username, password = password)
+
+  override def toString =
+    s"""ElasticsearchSourceSettings(baseUrl=$baseUrl,username=$username,password=${password.fold("")(_ => "***")}"""
```
ennru, line 26:
```suggestion
    s"""ElasticsearchConnectionSettings(baseUrl=$baseUrl,username=$username,password=${password.fold("")(_ => "***")}"""
```

docs/src/main/paradox/elasticsearch.md
```diff
@@ -116,15 +117,32 @@ Scala
...
 | Parameter           | Default    | Description                                                                                            |
 | ------------------- | -------    | ------------------------------------------------------------------------------------------------------ |
+| connection          |            | The connection details and credentials to authenticate against ElasticSearch. See `ElasticsearchConnectionSettings` |
 | bufferSize          | 10         | Flow and Sink batch messages to bulk requests when back-pressure applies.                              |
 | versionType         | None       | If set, `ElasticsearchSink` uses the chosen versionType to index documents. See [Version types](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#_version_types) for accepted settings. |
 | retryLogic          | No retries | See below |
 | apiVersion          | V5         | Currently supports `V5` and `V7` (see below) |
 | allowExplicitIndex  | True       | When set to False, the index name will be included in the URL instead of on each document (see below) | 
 
+### ElasticSearch connection
```
ennru, line 129:
```suggestion
### Elasticsearch connection
```

docs/src/main/paradox/elasticsearch.md
```diff
@@ -116,15 +117,32 @@ Scala
...
+
+The connection and credentials to authenticate with are configured with `ElasticSearchConnectionSettings`.
```
ennru, line 131:
```suggestion
The connection and credentials to authenticate with are configured with `ElasticsearchConnectionSettings`.
```

docs/src/main/paradox/elasticsearch.md
```diff
@@ -24,20 +24,21 @@ The table below shows direct dependencies of this module and the second tab show
 @@dependencies { projectId="elasticsearch" }
 
 
-## Set up REST client
+## ElasticSearch Parameters
```
ennru, line 27: I believe starting with the connection settings feels more natural and is the proper replacement for the REST client.

The branded spelling does surprisingly use a lower-case `s`:
```suggestion
## Elasticsearch Parameters
```

docs/src/main/paradox/elasticsearch.md
```diff
@@ -24,20 +24,21 @@ The table below shows direct dependencies of this module and the second tab show
...
 
-Sources, Flows and Sinks provided by this connector need a prepared `org.elasticsearch.client.RestClient` to
-access to Elasticsearch.
+Any API method that allows reading from and writing to ElasticSearch takes an instance of @apidoc[EsParams$].
```
ennru, line 29:
```suggestion
Any API method that allows reading from and writing to Elasticsearch takes an instance of @apidoc[EsParams$].
```

elasticsearch/src/main/mima-filters/3.0.0.backwards.excludes
```diff
@@ -0,0 +1,30 @@
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.elasticsearch.ElasticsearchSourceSettings.this")
```
ennru, line 1: You can do away with
```
ProblemFilters.exclude[Problem]("akka.stream.alpakka.elasticsearch.*)
```

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/EsParams.scala
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (C) 2016-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.elasticsearch
+
+final class EsParams private (val indexName: String, val typeName: Option[String])
```
ennru, line 7: Add a `toString`.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/impl/ElasticsearchSimpleFlowStage.scala
```diff
@@ -58,35 +57,43 @@ private[elasticsearch] final class ElasticsearchSimpleFlowStage[T, C](
...
-        new ResponseListener() {
-          override def onFailure(exception: Exception): Unit =
-            failureHandler.invoke((resultsPassthrough, exception))
-
-          override def onSuccess(response: Response): Unit =
-            responseHandler.invoke((messages, resultsPassthrough, response))
-        },
-        new BasicHeader("Content-Type", "application/x-ndjson")
-      )
+      val uri = Uri(settings.connection.baseUrl).withPath(Path(endpoint))
```
ennru, line 74: The URI is static and can be created once in the constructor.

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/scaladsl/ElasticsearchSink.scala
```diff
@@ -20,12 +19,9 @@ object ElasticsearchSink {
   /**
    * Create a sink to update Elasticsearch with [[akka.stream.alpakka.elasticsearch.WriteMessage WriteMessage]]s containing type `T`.
    */
-  def create[T](indexName: String,
-                typeName: String,
-                settings: ElasticsearchWriteSettings = ElasticsearchWriteSettings.Default)(
-      implicit elasticsearchClient: RestClient,
-      sprayJsonWriter: JsonWriter[T]
+  def create[T](esParams: EsParams, settings: ElasticsearchWriteSettings = ElasticsearchWriteSettings.Default)(
```
ennru, line 22: Please remove the settings default parameters from all methods, they won't work anyway as connection must be given.

```suggestion
  def create[T](esParams: EsParams, settings: ElasticsearchWriteSettings)(
```

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchConnectionSettings.scala
```diff
@@ -0,0 +1,42 @@
...
+object ElasticsearchConnectionSettings {
+
+  val Default = new ElasticsearchConnectionSettings(
+    baseUrl = "",
+    username = None,
+    password = None
+  )
+
+  /** Scala API */
+  def apply(): ElasticsearchConnectionSettings = Default
```
ennru, line 38: Require the base URL to be passed in:

```suggestion
  def apply(baseUrl: String) = new ElasticsearchConnectionSettings(baseUrl, None, None)
```

ennru (commented):
I found a few more details.

We'll soon open for Alpakka 3 merges!

The test code showing in the docs: Could we just duplicate the ones that show in the docs? Just showing V7 is enough. As this breaks the API substantially, I wonder if V7 should become the default.

elasticsearch/src/main/mima-filters/3.0.0.backwards.excludes
mrooding: Awesome!

elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/ElasticsearchConnectionSettings.scala
mrooding: This did force me to also remove the Default for ElasticsearchSourceSettings and ElasticsearchWriteSettings since they relied on ElasticsearchConnectionSettings.Default. Are you ok with how it is now?

mrooding:
I've updated it once again. All comments have been resolved and I've split up the tests that were used in the documentation between V5 and V7. I've tried to reduce duplication where possible. V5 tests are being used in the docs.

ennru:
I've rebased and fixed formatting.
`ElasticsearchV5Spec` fails on Travis but I don't have time to investigate this.
https://travis-ci.com/github/akka/alpakka/jobs/432021406#L648

mrooding:
Hi @ennru, thank you. I've updated the PR and fixed the test. Any test was running flaky due to the 600ms default future timeout. I've added IntegrationPatience to prevent this.

Let me know if this is good to go. I wanted to suggest to align naming of `EsParams` with `ElasticsearchParams` to keep naming consistent with the other configuration case classes. What do you think?

ennru:
> I wanted to suggest to align naming of `EsParams` with `ElasticsearchParams` to keep naming consistent with the other configuration case classes.

I agree, making it short is the wrong optimisation.

mrooding:
I've renamed all instances of EsParams. Only the mqtt tests seem to be failing but that seems unrelated.

Let me know if there's anything else to pick up!

ennru (approved):
LGTM.

ennru:
Thank you! Great to become independent of the Elasticsearch client library.

If you're up for more, https://github.com/akka/alpakka/issues/2472 would be great.

PR #2351, opened 2020-07-01
build: plugin updates

ennru:
- Akka Paradox (for apidoc)
- Dynver
- sbt scalafmt


ignasi35:
Do you dare doing #2352  in this PR too?

ennru:
Most changes look good, but some require diving into the scalafmt configuration, which I don't want to spend time on right now.

PR #2362, opened 2020-07-23
AWS S3: Avoid uploading a empty chunk after splitting

mpdn:
Fixes https://github.com/akka/alpakka/issues/2363

### DIFF
--- a/s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3IntegrationSpec.scala
+++ b/s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3IntegrationSpec.scala
@@ -443,6 +443,25 @@ trait S3IntegrationSpec
     exception.code shouldBe StatusCodes.Forbidden.toString()
   }
 
+  it should "only upload single chunk when size of the ByteString equals chunk size" in {
+    val source: Source[ByteString, Any] = Source.single(ByteString.fromArray(Array.fill(S3.MinChunkSize)(0.toByte)))
+
+    val metadata =
+      for {
+        _ <- source.runWith(
+          S3.multipartUpload(defaultBucket, objectKey, chunkSize = S3.MinChunkSize)
+            .withAttributes(attributes)
+        )
+        metadata <- S3
+          .getObjectMetadata(defaultBucket, objectKey)
+          .withAttributes(attributes)
+          .runWith(Sink.head)
+      } yield metadata
+
+    val etag = metadata.futureValue.get.eTag.get
+    etag.substring(etag.indexOf('-') + 1) shouldBe "1"
+  }
+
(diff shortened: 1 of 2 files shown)


s3/src/test/scala/akka/stream/alpakka/s3/impl/DiskBufferSpec.scala
```diff
@@ -84,4 +84,12 @@ class DiskBufferSpec(_system: ActorSystem)
     }
 
   }
+
+  it should "not emit rest chunk if rest chunk is empty" in {
```
seglo, line 88: Does this test validate that the extra element is only emitted if the `MemoryBuffer` is used for chunking?

s3/src/main/scala/akka/stream/alpakka/s3/impl/MemoryBuffer.scala
```diff
@@ -42,11 +43,14 @@ import akka.util.ByteString
...
       override def onUpstreamFinish(): Unit = {
-        if (isAvailable(out)) emit()
+        if (isAvailable(out) && (!hasEmitted || buffer.nonEmpty)) emit()
         completeStage()
       }
 
-      private def emit(): Unit = emit(out, Chunk(Source.single(buffer), buffer.size), () => completeStage())
+      private def emit(): Unit = {
+        hasEmitted = true
+        emit(out, Chunk(Source.single(buffer), buffer.size), () => completeStage())
```
seglo, line 52: An alternative impl. would be to empty the buffer as part of the  `andThen` thunk.  Then we don't need `hasEmitted` and just rely on the buffer to signal if there's anything to emit.

seglo:
Can you take a look at the test failure?

https://travis-ci.com/github/akka/alpakka/jobs/364241675#L593

mpdn:
On second thought, I am not really happy with the way I tried to fix this. I'll mark the PR as draft again and change my approach.

mpdn:
There we go. I think this is a better way of solving it instead of changing the behavior of the buffers.

s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3IntegrationSpec.scala
```diff
@@ -443,6 +443,59 @@ trait S3IntegrationSpec
...
+
+  it should "only upload single chunk when exact chunk is followed by an empty ByteString" in {
+    val source: Source[ByteString, Any] = Source[ByteString](
+      chunk :: ByteString.empty :: Nil
+    )
+
+    uploadAndAndCheckParts(source, 1)
+  }
+
+  it should "upload two chunks size of ByteSTrings equals chunk size" in {
```
ennru, line 461:
```suggestion
  it should "upload two chunks size of ByteStrings equals chunk size" in {
```

ennru:
Thank you @mpdn!

PR #2371, opened 2020-08-04
AMQP: Fixes handling null replyTo header in AmqpReplyToSinkStage

Grandys:
Fixes situation when `properties` of the message are present, but `replyTo` header in this properties is `null`. In that case `elem.properties.map(_.getReplyTo)` evaluated to `Some(null)` and sending message resulted with the error:
```
java.lang.IllegalStateException: Invalid configuration: 'routingKey' must be non-null.
	at com.rabbitmq.client.impl.AMQImpl$Basic$Publish.<init>(AMQImpl.java:2280)
	at com.rabbitmq.client.AMQP$Basic$Publish$Builder.build(AMQP.java:1219)
	at com.rabbitmq.client.impl.ChannelN.basicPublish(ChannelN.java:696)
	at com.rabbitmq.client.impl.recovery.AutorecoveringChannel.basicPublish(AutorecoveringChannel.java:202)
	at akka.stream.alpakka.amqp.impl.AmqpReplyToSinkStage$$anon$1$$anon$2.onPush(AmqpReplyToSinkStage.scala:74)
```


ennru:
Thank you @Grandys!

PR #2376, opened 2020-08-11
fix typo

ldrygala:


ldrygala:
looks like it duplicates https://github.com/akka/alpakka/pull/2375

PR #2377, opened 2020-08-12
Kinesis KCL source backpressure

### LINKED ISSUE #2370: Alpakka kinesis does not backpressure correctly
### Versions used 

alpakka-kinesis 2.0.1

### Akka version: 

2.6.1

### Expected Behavior

I started using alpakka kinesis scheduled source (with kcl) and I feel like it does not backpressure correctly.

### Actual Behavior

I create the source via the api (and custom scheduler) like this:
```
        final KinesisSchedulerSourceSettings schedulerSettings =
                KinesisSchedulerSourceSettings.create(1000, Duration.of(1L, ChronoUnit.SECONDS));

        Source<CommittableRecord, CompletionStage<Scheduler>> source = KinesisSchedulerSource.create(schedulerBuilder, schedulerSettings);
```

and the application is growing memory eventhough my later stages are extremely slow. I still see debug logs from kinesis, informing me that processRecords has been called eventhough it is in the moments when the later stages are completely blocked (waiting for some io to finish).

### Relevant logs

With the buffer of 1000 I still see (example shard 11) still accepting records when it should be waiting for the slow downstream.

```
2020-08-03 18:56:57,164 [ShardRecordProcessor-0011] DEBUG s.a.kinesis.lifecycle.ProcessTask - Calling application processRecords() with 1185 records from shardId-000000000011
2020-08-03 18:56:59,804 [ShardRecordProcessor-0011] DEBUG s.a.kinesis.lifecycle.ProcessTask - Calling application processRecords() with 1699 records from shardId-000000000013
2020-08-03 18:57:01,434 [ShardRecordProcessor-0011] DEBUG s.a.kinesis.lifecycle.ProcessTask - Calling application processRecords() with 2106 records from shardId-000000000008
2020-08-03 18:57:02,050 [ShardRecordProcessor-0011] DEBUG s.a.kinesis.lifecycle.ProcessTask - Calling application processRecords() with 125 records from shardId-000000000014
2020-08-03 18:57:02,991 [ShardRecordProcessor-0011] DEBUG s.a.kinesis.lifecycle.ProcessTask - Calling application processRecords() with 1842 records from shardId-000000000012
2020-08-03 18:57:05,764 [ShardRecordProcessor-0011] DEBUG s.a.kinesis.lifecycle.ProcessTask - Calling application processRecords() with 1823 records from shardId-000000000010
2020-08-03 18:57:05,784 [ShardRecordProcessor-0012] DEBUG s.a.kinesis.lifecycle.ProcessTask - Calling application processRecords() with 2276 records from shardId-000000000011
```

### Possible cause

After some debugging I see that the internal Semaphore in the KinesisSchedulerSourceStage has more permits than 1000, even tens of thousands after a while. It led me to this code in the library:

https://github.com/akka/alpakka/blob/master/kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala#L81-L84

Where it seems to me that acquire is called just once when the records processor is initialized and release is called after every records pump to the downstream. It looks like that the acquire call should be part of the lambda.

aserrallerios:
Fixes  #2370

### DIFF
--- a/kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
+++ b/kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
@@ -78,7 +78,7 @@ private[kinesis] class KinesisSchedulerSourceStage(
       matValue.success(scheduler)
     }
     private val callback: AsyncCallback[Command] = getAsyncCallback(awaitingRecords)
-    private val newRecordCallback: CommittableRecord => Unit = record => {
+    private def newRecordCallback(record: CommittableRecord): Unit = {
       backpressureSemaphore.tryAcquire(backpressureTimeout.length, backpressureTimeout.unit)
       callback.invoke(NewRecord(record))
     }
--- a/kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
+++ b/kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
@@ -78,9 +78,9 @@ private[kinesis] class KinesisSchedulerSourceStage(
       matValue.success(scheduler)
     }
     private val callback: AsyncCallback[Command] = getAsyncCallback(awaitingRecords)
-    private val newRecordCallback: CommittableRecord => Unit = {
+    private val newRecordCallback: CommittableRecord => Unit = record => {
       backpressureSemaphore.tryAcquire(backpressureTimeout.length, backpressureTimeout.unit)
-      record => callback.invoke(NewRecord(record))
+      callback.invoke(NewRecord(record))
     }
     override def onPull(): Unit = awaitingRecords(Pump)
     override def onDownstreamFinish(): Unit = awaitingRecords(Complete)
--- a/kinesis/src/test/scala/akka/stream/alpakka/kinesis/Valve.scala
+++ b/kinesis/src/test/scala/akka/stream/alpakka/kinesis/Valve.scala
@@ -1,0 +1,162 @@
+/*
+ * Copyright (C) 2016-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.kinesis
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+import akka.stream.stage.{GraphStageLogic, GraphStageWithMaterializedValue, InHandler, OutHandler}
+import akka.stream._
+import SwitchMode.{Close, Open}
+
+import scala.concurrent.{Future, Promise}
+
+/**
+ * Pause/ Resume a Flow
+ */
+sealed trait ValveSwitch {
+
+  /**
+   * Change the state of the valve
+   *
+   * @param mode expected mode to switch on
+   * @return A future that completes with true if the mode did change and false if it already was in the requested mode
+   */
+  def flip(mode: SwitchMode): Future[Boolean]
+
+  /**
+   * Obtain the state of the valve
+   *
+   * @return A future that completes with [[SwitchMode]] to indicate the current state of the valve
+   */
+  def getMode(): Future[SwitchMode]
+}
+
+object Valve {
+
+  /**
+   * Factory for [[Valve]] instances.
+   *
+   */
+  def apply[A](): Valve[A] = Valve[A](SwitchMode.Open)
+
+  /**
+   * Java API: Factory for [[Valve]] instances.
+   *
+   */
+  def create[A](): Valve[A] = Valve[A](SwitchMode.Open)
+
+  /**
+   * Factory for [[Valve]] instances.
+   *
+   */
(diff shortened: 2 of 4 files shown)


aserrallerios:
This test was not originally included because I found it impossible to tweak the internal buffers so I can check the Kinesis KCL Source buffer.

Now it looks like it must be implemented somehow (a bug was present in the release because this test was missing) so I made a draft of the test, but I still find impossible to control the demand to the upstream.

The "numbers" of the test are found by trial-and-error. Can you help me with this @ennru ? How should I configure my stream to precisely control the demand to the Kinesis KCL Source? Or maybe I should totally change the approach to test the backpressure?

The idea was:

```
1. bufferSize = 4
2. push 5 records, it blocks after pushing 4 records
3. request 1 record downstream
4. check that all 5 records have been pushed
```

ennru:
Sorry for leaving you hanging here, I spent my efforts in other parts of Akka right now.

This might be what you need it you want to expect things message by message

https://github.com/akka/akka/blob/418d6d3ec0a8c16f3bc3f54b770ba60dfd0c5b65/akka-stream-tests/src/test/scala/akka/stream/scaladsl/RetryFlowSpec.scala#L17-L20

aserrallerios:
Thanks @ennru.

I've tried those attributes but still the result is not consistent. I think the problem resides in that the stream is not a simple, 1 component stream, so there's more than one stream component with buffer = 1.

https://github.com/akka/alpakka/blob/master/kinesis/src/main/scala/akka/stream/alpakka/kinesis/scaladsl/KinesisSchedulerSource.scala#L39-L45
https://github.com/akka/alpakka/blob/master/kinesis/src/test/scala/akka/stream/alpakka/kinesis/KinesisSchedulerSourceSpec.scala#L245-L251

Can I put something here so we can force backpressure?

```scala
﻿﻿      KinesisSchedulerSource(schedulerBuilder,
                             KinesisSchedulerSourceSettings(bufferSize = bufferSize,
                                                            backpressureTimeout = backpressureTimeout))
        // HERE
        .viaMat(KillSwitches.single)(Keep.right)
        .watchTermination()(Keep.both)
        .toMat(TestSink.probe)(Keep.both)
        .run()
```

aserrallerios:
Well, let me be more precise, the results **are** consistent. The problem is that even when I configure the Source with buffer 1, the stream consumes 3 events before transmitting backpressure to the Source. 

But we can leave it like that if you think that it won't be affected by unrelated implementation changes in Akka Streams.

ennru:
Maybe a [`Valve`](https://github.com/akka/akka-stream-contrib/blob/master/src/main/scala/akka/stream/contrib/Valve.scala) from Akka Stream contrib could help?

aserrallerios:
I think this should be good enough. It looks like `StreamSpec` is not available in the project, and `akka-stream-contrib` is not a dependency. Should we change that before finishing this issue?

ennru:
You're right `StreamSpec` is local to `akka-stream-tests` and not part of the testkit.
Akka Stream Contrib is not to be used as a library, we won't publish more releases. If code from it is useful, copy it.

aserrallerios:
I already copied the Valve :). This can be reviewed now.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
```diff
@@ -78,9 +78,9 @@ private[kinesis] class KinesisSchedulerSourceStage(
       matValue.success(scheduler)
     }
     private val callback: AsyncCallback[Command] = getAsyncCallback(awaitingRecords)
-    private val newRecordCallback: CommittableRecord => Unit = {
+    private val newRecordCallback: CommittableRecord => Unit = record => {
       backpressureSemaphore.tryAcquire(backpressureTimeout.length, backpressureTimeout.unit)
-      record => callback.invoke(NewRecord(record))
+      callback.invoke(NewRecord(record))
     }
```
ennru, line 84: That was sneaky indeed.
I believe regular `def` notation would have prevented this problem:
```
    private def newRecordCallback(record: CommittableRecord): Unit = {
      backpressureSemaphore.tryAcquire(backpressureTimeout.length, backpressureTimeout.unit)
      callback.invoke(NewRecord(record))
    }
```

kinesis/src/test/scala/akka/stream/alpakka/kinesis/Valve.scala
```diff
@@ -0,0 +1,162 @@
+/*
+ * Copyright (C) 2016-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.kinesis
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
```
ennru, line 8: Too much copy-paste.

ennru (approved):
Great to ensure this behaviour with tests!
LGTM, with a suggestion and a clean-up.

ennru:
...and the compiler is unhappy about this warning: https://travis-ci.com/github/akka/alpakka/jobs/378009226#L1014

aserrallerios:
Thanks for the reviews @ennru! Hope is fine now.

ennru:
Thank you @aserrallerios!

Issue #2361, opened 2020-07-23
DelaySpec: should work as passthrough if should delay returns false


raboof:
and others

https://travis-ci.com/github/akka/alpakka/jobs/364151701

```
info] Delay

[info] - should work as passthrough if should delay returns false *** FAILED *** (35 milliseconds)

[info]   java.lang.AssertionError: expected OnNext, found OnComplete

[info]   at akka.stream.testkit.TestSubscriber$ManualProbe.expectNext(StreamTestKit.scala:370)

[info]   at akka.stream.alpakka.googlecloud.bigquery.impl.util.DelaySpec.$anonfun$new$6(DelaySpec.scala:42)

[info]   at akka.stream.alpakka.googlecloud.bigquery.impl.util.DelaySpec.$anonfun$new$6$adapted(DelaySpec.scala:39)

[info]   at scala.collection.immutable.Range.foreach(Range.scala:158)

[info]   at akka.stream.alpakka.googlecloud.bigquery.impl.util.DelaySpec.$anonfun$new$2(DelaySpec.scala:39)

[info]   at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)

[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)

[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)

[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)

[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)

[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)

[info]   at org.scalatest.wordspec.AnyWordSpecLike$$anon$3.apply(AnyWordSpecLike.scala:1076)

[info]   at org.scalatest.TestSuite.withFixture(TestSuite.scala:196)

[info]   at org.scalatest.TestSuite.withFixture$(TestSuite.scala:195)

[info]   at akka.stream.alpakka.googlecloud.bigquery.impl.util.DelaySpec.withFixture(DelaySpec.scala:19)

[info]   at org.scalatest.wordspec.AnyWordSpecLike.invokeWithFixture$1(AnyWordSpecLike.scala:1074)

[info]   at org.scalatest.wordspec.AnyWordSpecLike.$anonfun$runTest$1(AnyWordSpecLike.scala:1086)

[...]

PR #2380, opened 2020-08-17
Make DelaySpec complete only after processing all elements

raboof:
When the input is completed, the 'delayed' side of the split and the 'non-delayed' side of the split are both completed, and whichever completes first will complete the output.

This means that when there are no delayed elements, and the input completes, this completion may overtake elements still on the 'non-delaying' side, as is visible in test failures.

One solution might be to set eagerComplete to 'false'. This will mean that when the input completes, the stage will wait for both branches to complete before completing itself. It's not completely obvious if that's desirable: do we want to wait for delayed elements in this case? If not, perhaps we should adapt the DelayFlow stage accordingly.

Refs #2361


raboof:
(marking as draft pending discussion in #2361)

ennru:
This is now solved with #2381

PR #2382, opened 2020-08-19
Check client instance for nulls

ennru:
I got reports from users that ended up with an NPE without understanding where it came from.

This PR covers some random connectors which get a client instance passed (implicitly).


sqs/src/main/scala/akka/stream/alpakka/sqs/scaladsl/SqsAckFlow.scala
```diff
@@ -202,4 +208,8 @@ object SqsAckFlow {
             }(sameThreadExecutionContext)
       }
       .mapConcat(identity)
+
+  @InternalApi
```
seglo, line 212: How come this is an `@InternalApi` when the implementations in other connectors are not?

ennru: Fixed now. (They started off as just `private`...)

PR #2388, opened 2020-08-27
Make GCS md5Hash optional

K-Adnan:
### Purpose
The intent of this PR is to allow composite objects in GCS, which do not have an MD5 Hash, to be processed through alpakka.

### Background Context
According to the [GCS Documentation](https://cloud.google.com/storage/docs/composite-objects), composite objects do not get a MD5 Hash metadata field. Since some objects we'd like to process through our system are composites, we are currently getting an exception: ```j.u.NoSuchElementException: key not found: md5Hash```. Turning the `md5Hash` field into an optional field would resolve this issue and would allow us to process these objects.


lightbend-cla-validator:

seglo:
@K-Adnan Do you mind signing the Lightbend CLA?

K-Adnan:
> @K-Adnan Do you mind signing the Lightbend CLA?

I've just signed that.

ennru:
Thank you for this improvement. We'd need to find a way to make this change in a binary-compatible way. Methods in the public API need to stay as before, but can be complemented with overloads.

lightbend-cla-validator:

lightbend-cla-validator:

lightbend-cla-validator:

K-Adnan:
OK to test

lightbend-cla-validator:

K-Adnan:
> Thank you for this improvement. We'd need to find a way to make this change in a binary-compatible way. Methods in the public API need to stay as before, but can be complemented with overloads.

Thanks for the feedback, I've just made the changes to keep the public API unchanged.

seglo:
We may want to consider making an exception to breaking binary compatibility changes to `StorageObject`. I recall when @DanieleSassoli worked on updating this type pre-2.0.0 there were a number of fields that needed to be added, some were considered optional, some were not. 

In this [API doc](https://cloud.google.com/storage/docs/json_api/v1/objects#resource-representations) it indicates that `md5hash` is a _writeable_ field.  I'm not sure precisely what that means, but fields marked writeable seem to be optional in `StorageObject` (for the most part).  I think that we just missed this field.  

I did a quick cross-reference with the API doc.

Writeable fields that we do not have as optional: 

* `contentType`
* `crc32c`
* **`md5hash`**
* `storageClass`

Non-writeable fields that we indicate are optional. I don't think there's any harm in having required fields as optional, but I thought I would include them anyway: 

* `kmsKeyName` (docs explain only applicable when object is encrypted)
* `owner` (docs explain is optional under some circumstances)
* `retentionExpirationTime`
* `timeDeleted` (docs explain is optional under some circumstances)

Fields we do not include in `StorageClass`. We should add these:

* `customTime` (writeable)

K-Adnan:
@seglo Are you happy for me to change the named fields into optional fields as well or would you prefer that change to be part of a separate PR?

seglo:
@K-Adnan If you're open to it I think it would be great to update all the identified fields as optional.  However, as @ennru pointed out it's important that we do so without breaking binary compatibility.  I would suggest that we add additional optional fields and then set both the old and new fields during deserialization.  If the field is not present or null, then set the non-optional field to `""` (or null, or some default value in the case of primitive types, but I think they're all `String` anyway).  

The deserialization for `StorageObject` can be tweaked in [`Formats.scala`](https://github.com/akka/alpakka/blob/master/google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/impl/Formats.scala#L212).

google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/StorageObject.scala
```diff
@@ -109,7 +116,7 @@ final class StorageObject private (
 
   /** Java API */
   def withContentType(value: akka.http.javadsl.model.ContentType): StorageObject =
-    copy(contentType = value.asInstanceOf[ContentType])
+    copy(contentType = Option(value).asInstanceOf[ContentType])
```
seglo, line 119: This looks like it would fail at runtime. The cast should be on `value`.  The `with` methods should set either both fields, `contentType` and `maybeContentType`, or just the optional field and then in the copy method do a `getOrElse` with some default value.

google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/StorageObject.scala
```diff
@@ -109,7 +116,7 @@ final class StorageObject private (
...
   def withSize(value: Long): StorageObject = copy(size = value)
   def withEtag(value: String): StorageObject = copy(etag = value)
   def withMd5Hash(value: String): StorageObject = copy(md5Hash = value)
```
seglo, line 122: Need to update `withCrc32c` too.

google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/StorageObject.scala
```diff
@@ -129,6 +136,7 @@ final class StorageObject private (
   def withRetentionExpirationTime(value: OffsetDateTime): StorageObject = copy(retentionExpirationTime = Option(value))
   def withTimeStorageClassUpdated(value: OffsetDateTime): StorageObject = copy(timeStorageClassUpdated = value)
   def withCacheControl(value: String): StorageObject = copy(cacheControl = Option(value))
+  def withCustomTime(value: OffsetDateTime): StorageObject = copy(customTime = Option(value))
```
seglo, line 139: Same as above.

google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/StorageObject.scala
```diff
@@ -229,13 +243,14 @@ final class StorageObject private (
...
-    acl.fold("")(acls => acls.mkString("[", ",", "]")) +
-    ")"
+    s"customTime = $customTime"
+  metadata.fold("")(m => s"metadata = $m,") +
+  componentCount.fold("")(cc => s"componentCount = $cc,") +
+  kmsKeyName.fold("")(kkn => s"kmsKeyName = $kkn,") +
+  customerEncryption.fold("")(ce => s"customerEncryption = $ce,") +
+  owner.fold("")(o => s"owner = $o,") +
+  acl.fold("")(acls => acls.mkString("[", ",", "],")) +
+  ")"
```
seglo, line 253: Not sure why this change was needed, but you should add any missing fields to the `toString`.

seglo (commented):
Thanks for working on this.  Left some comments.

K-Adnan:
> Thanks for working on this. Left some comments.

Thanks for the comments. I've just made the changes based on those.

google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/StorageObject.scala
```diff
@@ -229,13 +247,14 @@ final class StorageObject private (
...
     s"timeStorageClassUpdated = $timeStorageClassUpdated," +
     s"cacheControl = $cacheControl," +
-    metadata.fold("")(m => s"metadata = $m,") +
-    componentCount.fold("")(cc => s"componentCount = $cc,") +
-    kmsKeyName.fold("")(kkn => s"kmsKeyName = $kkn,") +
-    customerEncryption.fold("")(ce => s"customerEncryption = $ce,") +
-    owner.fold("")(o => s"owner = $o,") +
-    acl.fold("")(acls => acls.mkString("[", ",", "]")) +
-    ")"
+    s"customTime = $customTime"
```
seglo, line 250:
```suggestion
    s"customTime = $customTime,"
```

seglo (commented):
Almost there. Just need to add some getters for the Java API.

It would also be nice if you could update `GCStorageWiremockBase` with any missing fields from `StorageObject`.

Have you tested that this PR works for your use case?

google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/StorageObject.scala
```diff
@@ -91,6 +97,7 @@ final class StorageObject private (
   def getEventBasedHold: Optional[Boolean] = eventBasedHold.asJava
   def getRetentionExpirationTime: Optional[OffsetDateTime] = retentionExpirationTime.asJava
   def getCacheControl: Optional[String] = cacheControl.asJava
+  def getCustomTime: Optional[OffsetDateTime] = customTime.asJava
```
seglo, line 100: Add Java getters for the new optional fields.

K-Adnan:
> Almost there. Just need to add some getters for the Java API.
> 
> It would also be nice if you could update `GCStorageWiremockBase` with any missing fields from `StorageObject`.
> 
> Have you tested that this PR works for your use case?

Thanks for the comments. I've addressed the changes and am just checking this locally to see if it works for our use case.

seglo (approved):
LGTM.  We can get this into the release we're planning tomorrow.  I'll merge it at end of my day, but I want to give you a chance to test it first.

seglo:
On second thought. I'll just merge this now so the release is not blocked.

PR #2392, opened 2020-09-03
S3: Add access-style property

laszlovandenhoek:
References #2171 . In the pull request https://github.com/akka/alpakka/pull/2193#issuecomment-686304485 it was noted that it would be better to have a separate `access-style` property instead of the current tripartite `path-style-access` one. This PR attempts to improve that situation.

With this change, `access-style` is the preferred way of setting the property, using either `path` or `virtual`.

If the old, deprecated `path-style-access` is used, that takes precedence, but a warning is logged to instruct the developer to change to the new property.

If the effective access style is `path`, and `endpoint-url` is not set (i.e. the S3 provider is AWS), one warning is issued. This is not configurable, but it shouldn't have to be, because using path-style access on AWS S3 is likely going to cause problems sooner or later.


ennru (approved):
LGTM.

ennru:
Thank you for suggesting and providing a clean solution to this!

PR #2393, opened 2020-09-04
Build against Akka 2.6.8 on Travis nightlies

raboof:
Needs to be at least 2.6.5 because the grpc-based pub-sub
implementation needs https://github.com/akka/akka/issues/28818


ennru (approved):
LGTM.

PR #2396, opened 2020-09-07
docs: show Akka 2.6.5 as least 2.6 version

ennru:
follow-up to #2393

Or should it say Akka 2.6.8 as that's what we validate in the scheduled build?


ennru:
Whoops, this came from my main branch.

PR #2397, opened 2020-09-09
docs: least Akka 2.6 version from the scheduled build

ennru:
follow-up to #2393

Or should it say Akka 2.6.8 as that's what we validate in the scheduled build?


seglo:
> Or should it say Akka 2.6.8 as that's what we validate in the scheduled build?

Since the text preceding this list says

> The code in this documentation is compiled against:

I think it should always be the version we're actually compiling against.  Though maybe you could say in a sub bullet that anything since `2.6.5` should work.

seglo (approved):

Issue #2296, opened 2020-04-30
Akka GCP Pub Sub 2.0.0 not discarding entity bytes


cullender-skew:
### Versions used 

om.lightbend.akka', name: 'akka-stream-alpakka-google-cloud-pub-sub_2.12 version 2.0.0

Akka version: 

2.6.1

### Expected Behavior

Pull messages from GCP Pub Sub

### Actual Behavior

exception is thrown: cp pub sub team - I recently upgraded to version 2.0.0 and now I am getting java.util.concurrent.TimeoutException: Response entity was not subscribed after 1 second. Make sure to read the response entity body or calldiscardBytes()on it.

on the :pull request

Habitats:
Can confirm that this is happening, though I can still pull messages.

It also seem to only happen on busy topics. For topics with only a few events now and then, this does _not_ happen. However, for my busy ones (~30-50 messages/s) it happens basically every other second.

I've tried setting this in `application.conf` but it didn't change anything
```
stream {
  materializer {
     subscription-timeout {
        mode = warn
        timeout = 5s
      }
  }
}
```

ennru:
More details by @Habitats on [Gitter](https://gitter.im/akka/alpakka-kafka?at=5eaa94b7a9de3d01b1dc32cb)

ennru:
One thing I can think of when staring at the code is the interaction between unmarshalling in `mapAsync` and the `mapConcat`. This might lead to backpressure which makes the `mapAsync` wait too long so the no entity subscription timer kicks in.

Habitats:
But should the `subscription-timeout` config still not apply? Or am I confusing that flag with something else?

ennru:
Incidentally, a similar problem was just reported to Akka HTTP https://github.com/akka/akka-http/issues/3134

jrudolph:
> But should the `subscription-timeout` config still not apply? Or am I confusing that flag with something else?

It's another setting in akka-http which is responsible in that case, `akka.http.host-connection-pool.response-entity-subscription-timeout`.

In akka/akka-http#3134, I have outlined a possibly path forward.

jrudolph:
> In [akka/akka-http#3134](https://github.com/akka/akka-http/issues/3134), I have outlined a possibly path forward.

Currently, the implementation uses 

[...]

PR #2403, opened 2020-09-14
Google Cloud Pub/Sub: use SingleRequest

karlssonsimon:
Use `Http().SingleRequest` for pull and acknowledge and read response as it comes back to avoid timeout on reponse entity.

References #2296


jrudolph (approved):
AFAICT this seems pretty reasonable. A somewhat smaller change would have been to just replace `poolSimple` with a flow using `singleRequest` in a `mapAsync` making sure that `toStrict` would be called right afterwards, that would be somewhat of an extra step but since the unmarshalling is done anyway afterwards and json and string unmarshallers basically do something equivalent to `toStrict` it would also be fine.

In general, I wonder if the flow based API is required here at all? A cleaner solution for the future might be trying to move away from the stream-based HTTP client API as much as possible.

ennru:
Thank you for improving this!

PR #2404, opened 2020-09-14
tar with directory support and EOF block detection

ennru:
* Support directory entries in tar files (writing and reading)
* Accept closing EOF marker blocks of 512 0-bytes

The somewhat loose tar standard allows for more variants than Alpakka File can handle. This adds support for the "link indicator" flag, which also indicates directory entries.

Tar recommends files end with two 512 byte blocks of 0-bytes. This makes Alpakka File handle those and read to the end of the incoming streams.

### DIFF
--- a/file/src/main/scala/akka/stream/alpakka/file/impl/archive/TarArchiveEntry.scala
+++ b/file/src/main/scala/akka/stream/alpakka/file/impl/archive/TarArchiveEntry.scala
@@ -17,6 +17,31 @@ import akka.util.ByteString
  */
 @InternalApi private[file] object TarArchiveEntry {
 
+  /* The C structure for a Tar Entry's header is:
+   *
+   * struct header {
+   * char name[100];     // TarConstants.NAMELEN    - offset   0
+   * char mode[8];       // TarConstants.MODELEN    - offset 100
+   * char uid[8];        // TarConstants.UIDLEN     - offset 108
+   * char gid[8];        // TarConstants.GIDLEN     - offset 116
+   * char size[12];      // TarConstants.SIZELEN    - offset 124
+   * char mtime[12];     // TarConstants.MODTIMELEN - offset 136
+   * char chksum[8];     // TarConstants.CHKSUMLEN  - offset 148
+   * char linkflag[1];   //                         - offset 156
+   * char linkname[100]; // TarConstants.NAMELEN    - offset 157
+   * The following fields are only present in new-style POSIX tar archives:
+   * char magic[6];      // TarConstants.MAGICLEN   - offset 257
+   * char version[2];    // TarConstants.VERSIONLEN - offset 263
+   * char uname[32];     // TarConstants.UNAMELEN   - offset 265
+   * char gname[32];     // TarConstants.GNAMELEN   - offset 297
+   * char devmajor[8];   // TarConstants.DEVLEN     - offset 329
+   * char devminor[8];   // TarConstants.DEVLEN     - offset 337
+   * char prefix[155];   // TarConstants.PREFIXLEN  - offset 345
+   * // Used if "name" field is not long enough to hold the path
+   * char pad[12];       // NULs                    - offset 500
+   * } header;
+   */
+
   val fileNameLength = 100
   val fileModeLength = 8
   val ownerIdLength = 8
@@ -49,10 +74,14 @@ import akka.util.ByteString
 
   private val fixedData2 = {
     // [148, 156)
-    ByteString("        ") ++ // headerChecksumLength
+    ByteString("        ") // headerChecksumLength
+  }
+
+  // [156, 157)
+  // linkIndicatorLength
+
+  private val fixedData3 = {
     empty(
-      // [156, 157)
-      linkIndicatorLength +
       // [157, 257)
       linkFileNameLength
     ) ++
@@ -86,6 +115,7 @@ import akka.util.ByteString
 
   def parse(bs: ByteString): TarArchiveMetadata = {
     require(bs.length >= headerLength, s"the tar archive header is expected to be at least 512 bytes")
+    require(bs.head != 0, "the file name may not be empty")
     val filename = getString(bs, 0, fileNameLength)
     val fileSizeString =
       getString(bs, fileNameLength + fileModeLength + ownerIdLength + groupIdLength, fileSizeLength)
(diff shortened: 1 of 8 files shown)


file/src/main/scala/akka/stream/alpakka/file/impl/archive/TarReaderStage.scala
```diff
@@ -68,18 +68,26 @@ private[file] class TarReaderStage
...
+          val metadata = TarArchiveEntry.parse(headerBuffer)
+          val buffer = headerBuffer.drop(TarArchiveEntry.headerLength)
+          if (isAvailable(flowOut)) {
+            pushSource(metadata, buffer)
+          } else {
+            // await flow demand
+            setHandler(flowOut, new OutHandler {
+              override def onPull(): Unit = {
+                val handler = pushSource(metadata, buffer)
+                log.debug(s"setting $handler")
```
jrudolph, line 85: Want to leave that in?

file/src/main/scala/akka/stream/alpakka/file/impl/archive/TarReaderStage.scala
```diff
@@ -68,18 +68,26 @@ private[file] class TarReaderStage
           } else new CollectFile(metadata, buffer)
         }
 
-        val metadata = TarArchiveEntry.parse(headerBuffer)
-        val buffer = headerBuffer.drop(TarArchiveEntry.headerLength)
-        if (isAvailable(flowOut)) {
-          pushSource(metadata, buffer)
+        if (headerBuffer.head == 0) {
+          log.debug("empty filename, detected EOF padding, completing")
```
jrudolph, line 72: Hard to find any information about that. I guess lots of historic reasons and tape stuff...

file/src/main/scala/akka/stream/alpakka/file/model.scala
```diff
@@ -21,72 +21,127 @@ final class TarArchiveMetadata private (
     val filePathPrefix: Option[String],
     val filePathName: String,
     val size: Long,
-    val lastModification: Instant
+    val lastModification: Instant,
+    val linkIndicator: Byte
```
jrudolph, line 25: Maybe rename it to `linkIndicatorByte` so we could later add more modelling for that value?

```
/* The linkflag defines the type of file */
#define  LF_OLDNORMAL '\0'       /* Normal disk file, Unix compatible */
#define  LF_NORMAL    '0'        /* Normal disk file */
#define  LF_LINK      '1'        /* Link to previously dumped file */
#define  LF_SYMLINK   '2'        /* Symbolic link */
#define  LF_CHR       '3'        /* Character special file */
#define  LF_BLK       '4'        /* Block special file */
#define  LF_DIR       '5'        /* Directory */
#define  LF_FIFO      '6'        /* FIFO special file */
#define  LF_CONTIG    '7'        /* Contiguous file */
```

file/src/main/scala/akka/stream/alpakka/file/impl/archive/TarArchiveEntry.scala
```diff
@@ -17,6 +17,31 @@ import akka.util.ByteString
  */
 @InternalApi private[file] object TarArchiveEntry {
 
+  /* The C structure for a Tar Entry's header is:
+   *
+   * struct header {
+   * char name[100];     // TarConstants.NAMELEN    - offset   0
```
jrudolph, line 23: So, weird that this is all ascii but I've read that it was done that way by purpose to allow operators to fix archives manually...

file/src/main/scala/akka/stream/alpakka/file/model.scala
jrudolph: Ah, I see you have that table down below already.

file/src/main/scala/akka/stream/alpakka/file/model.scala
```diff
@@ -21,72 +21,127 @@ final class TarArchiveMetadata private (
...
+    s"lastModification=$lastModification," +
+    s"linkIndicator=${linkIndicator.toChar})"
 }
 
 object TarArchiveMetadata {
+
+  /**
+   * Constants for the `linkIndicator` flag.
+   */
+  val linkIndicatorNormal: Byte = '0'
```
jrudolph, line 64: Small enum(-ish) for that?

file/src/test/java/docs/javadsl/ArchiveTest.java
```diff
@@ -188,22 +200,61 @@ public void tarReader() throws Exception {
...
+
+  @Test
+  public void tarReaderWithEofBlocks() throws Exception {
+    ByteString tenDigits = ByteString.fromString("1234567890");
+    TarArchiveMetadata metadata1 = TarArchiveMetadata.create("file5.txt", tenDigits.length());
+    CompletionStage<ByteString> oneFileArchive =
+        Source.single(Pair.create(metadata1, Source.single(tenDigits)))
+            .via(Archive.tar())
+            // tar standard suggests two empty 512 byte blocks as EOF marker
+            .concat(Source.single(ByteString.fromArray(new byte[1024])))
```
jrudolph, line 237: I checked what tar does on my machine and there the padding seems to be 20 x 512 which is somewhat suggested by this spec: https://www.gnu.org/software/tar/manual/html_node/Blocking-Factor.html.

Anyway, it shouldn't matter since any amount of data is skipped at the end.

Would be good to add an actual file created by (some version of) tar itself, or am I missing that?

file/src/test/scala/akka/stream/alpakka/file/impl/archive/ZipArchiveFlowTest.scala
```diff
@@ -23,7 +23,7 @@ class ZipArchiveFlowTest
   implicit val mat = ActorMaterializer()
 
   "ZipArchiveFlowStage" when {
-    "steam ends" should {
+    "stream ends" should {
```
jrudolph, line 26: :steam_locomotive: Out of steam? - end stream :+1:

jrudolph (approved):
LGTM (though, didn't check tests in anger)

file/src/main/scala/akka/stream/alpakka/file/model.scala
ennru: There exist custom extensions to these values. For now, we just care about normal and directory.

file/src/test/java/docs/javadsl/ArchiveTest.java
ennru: A user shared a file which blew up in Alpakka with me (it padded with 6x512 bytes), but I can't add that file to the repository.

PR #2410, opened 2020-09-14
CI renovation and Patch version upgrades accross the board

ennru:
* Reduce steps in check stage
* Cache Jabba installer
* Latest JDK 8 (and JDK 11 in the scheduled build)
* sbt 1.3.13
* Scala 2.13.3
* Upgraded to latest patch versions for most libraries
* Latest version of ScalaTest and Mockito (those are not exposed outside of Alpakka)
* Avoid deprecations in Java code (and fail for new deprecations)


raboof:
TarArchiveSpec succeeds for me locally (and shouldn't really be touched by these changes), but seems to fail pretty consistently on travis...

ennru:
This became a bit overwhelming. I split it into #2412, #2413 and #2414.

PR #2411, opened 2020-09-15
tests: disable parallel execution

ennru:
To make the log capturing more useful and possibly understand what's off in eg #1723

Most modules did switch it off even before this.


raboof (approved):

raboof:
(failure is mqtt-streaming)

PR #2413, opened 2020-09-16
Less use of deprecated APIs in Java code

ennru:
tests: use Hamcrest's assertThat 
JMS tests: expect connection to be closed eventually 
File tests: Check that the file exists, eventually on Travis 
Slick test: recover -> recoverWithRetries

A part of #2410

PR #2418, opened 2020-09-21
CI: allow Java warnings in the schdeduled build

ennru:
Since enabling more Java warnings in #2413 the scheduled build, which runs on JKD11, failed.
This allows warnings in the scheduled build, just as Scala warnings are allowed.

### DIFF
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -98,7 +98,8 @@ object Common extends AutoPlugin {
           "-Xlint:deprecation"
         ),
       compile / javacOptions ++= (scalaVersion.value match {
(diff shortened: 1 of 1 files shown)


project/Common.scala
```diff
@@ -98,7 +98,8 @@ object Common extends AutoPlugin {
           "-Xlint:deprecation"
         ),
       compile / javacOptions ++= (scalaVersion.value match {
-          case Dependencies.Scala212 if insideCI.value && fatalWarnings.value => Seq("-Werror")
+          case Dependencies.Scala212 if insideCI.value && fatalWarnings.value && !Dependencies.CronBuild =>
```
ennru, line 101: See line 59 for the Scala counterpart.

raboof (approved):

Issue #1571, opened 2019-03-11
FTP: In some tests replace sleeping with different wait


ennru:
`akka.stream.alpakka.ftp.BaseSpec` defines `extraWaitForStageShutdown` which can be replaced with configuration when Akka 2.5.22 is out.

See https://github.com/akka/akka/issues/26410

ennru:
Fixed with #2421

PR #2421, opened 2020-09-21
FTP test: use all stages stopped timeout

ennru:
Refs #1571﻿


raboof (approved):
Looks nicer indeed ;)

Issue #2167, opened 2020-02-24
Use google/cloud-sdk emulator to add integration tests for Google Pub Sub


seglo:
There's no integration testing currently done in the google-pub-sub connector.  This issue was raised in https://github.com/akka/alpakka/issues/2163 .

Some posts about using the `google/cloud-sdk` Docker image in a test suite:

https://medium.com/@mrkrijgsman/integration-testing-pub-sub-interactions-in-your-application-45e35476de23 
https://medium.com/google-cloud/using-google-cloud-emulators-for-integration-tests-7812890ebe0d

ennru:
Both Pub/Sub connectors use the official Google cloud emulator for testing now.

Issue #1459, opened 2019-01-28
Google Pub/Sub gRPC - Expose acknowledge Sink as a Flow


josefelixh:
Google Pub/Sub gRPC exposes a Sink to acknowledge processed messages, it would be convenient to expose the acknowledge functionality also as a Flow.

Happy to contribute 😄

2m:
Sounds good! Just a heads up that quite soon I will do some refactoring where GooglePubSubExternal will go away and GrpcPublisher will be able to be injected via attributes. Same fate that struck S3 and DynamoDb connectors so far.

2m:
@josefelixh this is the PR I mentioned before #1466

josefelixh:
Cheers @2m are you suggesting I wait for #1466 to be completed/merged? or shall I branch from `2m:wip-pubsub-grpc-attr-2m`?

2m:
It will be merged quite soon. But if you want to start immediately you can start working from my branch and then rebase when the PR gets merged.

josefelixh:
I see #1466 is merged 🎉

ennru:
Fixed with #2422 and #2425

PR #2425, opened 2020-09-22
Google Pub/Sub: tests using the gcloud emulator 

ennru:
* use the google/cloud-sdk emulator via docker compose
* add `IntegrationSpec` for some end-to-end tests using the emulator
* add `acknowledgeFlow` as suggested for the gRPC variant in #1459

References #2167 

A continuation of this would be to show the code from IntegrationSpec as sample code in the docs.

### DIFF
--- a/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/javadsl/GooglePubSub.scala
+++ b/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/javadsl/GooglePubSub.scala
@@ -16,43 +16,76 @@ import scala.collection.JavaConverters._
 import scala.compat.java8.FutureConverters._
 import scala.concurrent.Future
 
+/**
+ * Java DSL for Google Pub/Sub
+ */
 object GooglePubSub {
 
+  /**
+   * Creates a flow to that publishes messages to a topic and emits the message ids.
+   */
   def publish(topic: String,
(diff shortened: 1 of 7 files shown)


google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/javadsl/GooglePubSub.scala
```diff
@@ -16,43 +16,76 @@ import scala.collection.JavaConverters._
...
-      .subscribeFlow(subscription = subscription, config = config)
+      .subscribeFlow(subscription, config)
       .asJava
 
+  /**
+   * Creates a sink for acknowledging messages on a subscription.
+   *
+   * @deprecated Use `acknowledge` without `parallelism` param (since 2.0.0)
+   */
+  @Deprecated
```
seglo, line 68: Is this redundant?

seglo (approved):
LGTM

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/javadsl/GooglePubSub.scala
ennru: Most likely, I don't remember how much deprecation one should add to mark it deprecated for Java users...

PR #2427, opened 2020-09-22
Inherit io.grpc version from Akka gRPC

raboof:
Refs https://github.com/akka/akka-grpc/issues/1125


ennru (approved):
Great, exactly what we need here.
LGTM.

Issue #2441, opened 2020-10-06
Pravega setting builder issue.


cheleb:
### Versions used 

Akka version: 

### Expected Behavior
keyExtractor is never set by the builder.

Please describe the expected behavior of the issue, starting from the first action.

### Actual Behavior

Please provide a description of what actually happens, working from the same starting point.

routing key extractor is never set from the Builder

### Relevant logs

If you identified a section in your logs that explains the bug or might be important to understand it, please add it.

### Reproducible Test Case

Please provide a PR with a failing test.

If the issue is more complex or requires configuration, please provide a link to a project that reproduces the issue.

ennru:
Fixed with #2442

PR #2442, opened 2020-10-06
Pravega: keyExtractor not set

cheleb:
Builder didn't forward the keyExtractor

References #2441


ignasi35 (approved):
nice catch. 

Thanks @cheleb

Issue #2352, opened 2020-07-01
Update to latest scalafmt


ignasi35:
### Short description

Bump [scalafmt version](https://github.com/scalameta/scalafmt/releases) in `.scalafmt.conf`.

### Details

This version bump may introduce multiple format changes and spark style changes because of new defaults or improvements. We may want to dry-run it first to see the impact since the version bump is high (from `2.1.0` to `2.6.1.` as of this writing)

ignasi35:
We decided we can stay on `2.1.0`

PR #2447, opened 2020-10-13
Update to latest scalafmt

jcasadella:
Hi there! This is our first public PR! 🐣 🐣 

- Update `scalafmt` and `sbt-scalafmt` in the **first commit**.
- Apply the result of executing `sbt scalafmtAll` in the **second commit**.

Co-authored-by: Sol Lerma <mlermabrines@gmail.com>
References #2352

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,4 +1,4 @@
-version = 2.1.0
+version = 2.7.4
 
 style = defaultWithAlign
 
--- a/amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AbstractAmqpAsyncFlowStageLogic.scala
+++ b/amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AbstractAmqpAsyncFlowStageLogic.scala
@@ -61,8 +61,8 @@ import scala.concurrent.Promise
   }
 
   private def asAsyncCallback(confirmCallback: (DeliveryTag, Boolean) => Unit): ConfirmCallback = {
-    val callback = getAsyncCallback[(DeliveryTag, Boolean)] {
-      case (tag: DeliveryTag, multiple: Boolean) => confirmCallback(tag, multiple)
+    val callback = getAsyncCallback[(DeliveryTag, Boolean)] { case (tag: DeliveryTag, multiple: Boolean) =>
+      confirmCallback(tag, multiple)
     }
     new ConfirmCallback { // cant use function literal because it doesn't work with 2.11
       override def handle(tag: DeliveryTag, multiple: Boolean): Unit = callback.invoke((tag, multiple))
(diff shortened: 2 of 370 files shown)


amqp/src/main/scala/akka/stream/alpakka/amqp/impl/AbstractAmqpAsyncFlowStageLogic.scala
```diff
@@ -61,8 +61,8 @@ import scala.concurrent.Promise
   }
 
   private def asAsyncCallback(confirmCallback: (DeliveryTag, Boolean) => Unit): ConfirmCallback = {
-    val callback = getAsyncCallback[(DeliveryTag, Boolean)] {
-      case (tag: DeliveryTag, multiple: Boolean) => confirmCallback(tag, multiple)
+    val callback = getAsyncCallback[(DeliveryTag, Boolean)] { case (tag: DeliveryTag, multiple: Boolean) =>
+      confirmCallback(tag, multiple)
     }
```
ignasi35, line 66: This change looks like a step back.

.scalafmt.conf
```diff
@@ -1,4 +1,4 @@
-version = 2.1.0
```
ignasi35, line 1: I'm in favor of keeping this config in sync across multiple repos of the Akka organization. For example, copy/paste the conf from https://github.com/akka/akka-grpc/blob/master/.scalafmt.conf

(I just realized akka/akka is still stuck in `2.1.0` so we can't use that as a baseline)

ignasi35 (commented):
+1 to not reinvent the wheel/diverge across the Akka organization.

ignasi35:
Hi @jcasadella and Sol! 

Thanks for this contribution.

@jcasadella in order to add Sol as a Co-Author, I think it only works if you add the info in a commit message (not in the PR description): `Co-authored-by: Sol Lerma mlermabrines@gmail.com` :-)

.scalafmt.conf
seglo: I wonder if there is some new scalafmt configuration we're missing to avoid too much churn in the codebase.  It's quite a jump in versions so I expect some rules have changed in various ways.  For example, to not push parens to the next line.

ignasi35:
Hi @jcasadella and Sol, unfortuntely we've changed the plan and would rather not bump the scalafmt version as it introduces a huge amount of changes, and navigating the git history is more important than keeping up to date with some arbitrary formatting decisions.

Thanks for looking into this. You work sparked som internal discussion and review of the formatters in other repos (e.g. https://github.com/akka/akka-grpc/pull/1148).

I will close this PR and the issue that caused it.

jcasadella:
That's totally understandable. Also would like to say that Sol and I have learned some basics of `scalafmt` so the time wasn't wasted! ;) 
Thank you @ignasi35 for your guidance and support, we'll find another issue to contribute on!

PR #2454, opened 2020-10-22
Google Pub-Sub: Switch to older emulator image

raboof:
The last successful CRON build was 26/9, this is the most
recent tag that was uploaded before that date.


ennru (approved):
LGTM.

Issue #2456, opened 2020-10-23
Re-enable failing the build on compiler warnings


ennru:
With the switch to Akka 2.6, Alpakka can now make use of the simpler Akka Streams API which doesn't require the user to provide a `Materializer` any more.

To enable working on the new deprecations in Akka 2.6 connector-by-connector, we disable failing on warnings.

seglo:
Implemented in #2628

PR #2457, opened 2020-10-23
Require Akka 2.6 (drop Akka 2.5)

ennru:
Set Akka 2.6.10 as the only Akka version used in Alpakka.

To allow the deprecation warnings in CI, `fatalWarnings` is now `false` by default. https://github.com/akka/alpakka/issues/2456

References #2054


raboof (approved):

ennru:
Wow, an all modules build that went through! (As Scala 2.11 isn't there anymore 1/3 fewer chances to fail, I guess.)

PR #2458, opened 2020-10-23
scalafmtSbt

raboof:
To fix the build. #2457 also has these changes I think, so merging that instead
is also an option.


ennru:
I'll merge #2457 later today.

PR #2460, opened 2020-10-26
CSV: remove materializer

ennru:
Update the CSV tests to pass the actor system instead of an explicit materializer.

References #2054, #2468

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -121,7 +121,7 @@ lazy val cassandra =
 lazy val couchbase =
   alpakkaProject("couchbase", "couchbase", Dependencies.Couchbase, whitesourceGroup := Whitesource.Group.Supported)
 
-lazy val csv = alpakkaProject("csv", "csv", whitesourceGroup := Whitesource.Group.Supported, fatalWarnings := false)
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -121,7 +121,7 @@ lazy val cassandra =
 lazy val couchbase =
   alpakkaProject("couchbase", "couchbase", Dependencies.Couchbase, whitesourceGroup := Whitesource.Group.Supported)
 
-lazy val csv = alpakkaProject("csv", "csv", whitesourceGroup := Whitesource.Group.Supported)
+lazy val csv = alpakkaProject("csv", "csv", whitesourceGroup := Whitesource.Group.Supported, fatalWarnings := false)
```
ignasi35, line 124: Do we want `fatalWarnings=false` in a supported module?

ignasi35 (approved):
Left the comment re `fatalWarnings`  above but otherwise LGTM

build.sbt
ennru: Boolean logic it too much for me.

Issue #2453, opened 2020-10-21
PravegaGraphSpec: Pravega connector should runs sources


raboof:
https://travis-ci.com/github/akka/alpakka/jobs/403083496

```
[info] PravegaGraphSpec:
[info] Pravega connector
[info] - should runs sources *** FAILED *** (10 seconds, 294 milliseconds)
[info]   java.util.concurrent.TimeoutException: Futures timed out after [10 seconds]
[info]   at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:259)
[info]   at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:187)
[info]   at scala.concurrent.Await$.$anonfun$ready$1(package.scala:195)
[info]   at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:57)
[info]   at scala.concurrent.Await$.ready(package.scala:146)
[info]   at akka.stream.alpakka.pravega.PravegaGraphSpec.$anonfun$new$2(PravegaGraphSpec.scala:65)
```

ennru:
This seems to fail very often.
https://travis-ci.com/github/akka/alpakka/jobs/408586344#L390 in #2508

cheleb:
Is this issue still open ?

seglo:
This issue hasn't been observed (at least, it hasn't been logged here) since the PR, so I think it's safe to close. We can reopen it if necessary.

PR #2509, opened 2020-10-30
Pravega: Re-factor tests

seglo:
Initially I was only attempting to recreate #2453. I noticed when I ran the same test twice I received a timeout on the same Future as #2453.  I added some explicit timeouts to the data population code and got a little carried away.

* Always use unique group, scope, and stream name per test
* Java API: overload to load `ReaderBasicSettings` from supplied config
* Remove usages of Materializer


ennru (approved):
LGTM.

ennru:
This covers a part of #2492 (system instead of materializer).

PR #2519, opened 2020-11-03
akka 2.6.

dotbg:
References #2473

### DIFF
--- a/file/src/test/java/docs/javadsl/DirectoryTest.java
+++ b/file/src/test/java/docs/javadsl/DirectoryTest.java
@@ -7,10 +7,9 @@ package docs.javadsl;
 import akka.NotUsed;
 import akka.actor.ActorSystem;
 import akka.japi.Pair;
-import akka.stream.ActorMaterializer;
-import akka.stream.Materializer;
 // #walk
 // #ls
+import akka.stream.Materializer;
 import akka.stream.alpakka.file.javadsl.Directory;
 // #ls
 import java.nio.file.FileVisitOption;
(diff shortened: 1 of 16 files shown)


file/src/test/java/docs/javadsl/DirectoryTest.java
```diff
@@ -46,7 +45,7 @@
   @BeforeClass
   public static void beforeAll() throws Exception {
     system = ActorSystem.create();
-    materializer = ActorMaterializer.create(system);
+    materializer = Materializer.apply(system);
```
seglo, line 48: We want to drop all the materializer references in the tests (where possible).

seglo (commented):
We still need to replace all the materializer references with system in the Java and Scala tests. The easiest way to see if the Akka 2.6 deprecations are fixed is by enabling `fatalWarnings` on the sub project.

file/src/test/java/docs/javadsl/DirectoryTest.java
dotbg: @ennru sadly, there is a [method](https://github.com/akka/akka/blob/master/akka-stream-testkit/src/main/scala/akka/stream/testkit/javadsl/StreamTestKit.scala#L18) in the tearDown, requiring the Materializer instance. 

```java
  @After
  public void tearDown() throws Exception {
    fs.close();
    fs = null;
    StreamTestKit.assertAllStagesStopped(materializer);
  }
```

seglo: You can get the materialized from an actor system, `Materializer.matFromSystem(system)`

dotbg: just pushed it changed

ennru (approved):
LGTM.

ennru:
Thank you so much for taking these forward!

PR #2526, opened 2020-11-09
spring-web: Akka 2.6 features

dotbg:
References #2499

### DIFF
--- a/spring-web/src/main/java/akka/stream/alpakka/spring/web/AkkaStreamsRegistrar.java
+++ b/spring-web/src/main/java/akka/stream/alpakka/spring/web/AkkaStreamsRegistrar.java
@@ -4,6 +4,7 @@
 
 package akka.stream.alpakka.spring.web;
 
+import akka.actor.ActorSystem;
 import akka.stream.Materializer;
 import akka.stream.javadsl.AsPublisher;
 import org.springframework.core.ReactiveAdapterRegistry;
@@ -13,10 +14,16 @@ import static org.springframework.core.ReactiveTypeDescriptor.multiValue;
 
 public class AkkaStreamsRegistrar {
 
-  private final Materializer materializer;
+  private final ActorSystem system;
 
-  public AkkaStreamsRegistrar(Materializer mat) {
-    materializer = mat;
+  /** @deprecated use {@link #AkkaStreamsRegistrar(ActorSystem)}. */
+  @Deprecated
+  public AkkaStreamsRegistrar(Materializer materializer) {
+    this(materializer.system());
+  }
+
+  public AkkaStreamsRegistrar(ActorSystem system) {
+    this.system = system;
   }
 
   public void registerAdapters(ReactiveAdapterRegistry registry) {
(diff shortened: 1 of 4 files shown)


ennru:
This changes the API a bit - which is OK for Alpakka 3.0.

But MiMa requires an exclusion rule. See https://github.com/akka/alpakka/blob/master/CONTRIBUTING.md#binary-compatibility-mimahttps://assets.grammarly.com/emoji/v1/1f628.svg

dotbg:
> 
> 
> This changes the API a bit - which is OK for Alpakka 3.0.
> 
> But MiMa requires an exclusion rule. See https://github.com/akka/alpakka/blob/master/CONTRIBUTING.md#binary-compatibility-mima https://assets.grammarly.com/emoji/v1/1f628.svg

@ennru Actually, I can make the API compatible. Should I? May not play well with the `fatalWarnings `, though

spring-web/src/main/java/akka/stream/alpakka/spring/web/AkkaStreamsRegistrar.java
```diff
@@ -25,15 +26,16 @@ public void registerAdapters(ReactiveAdapterRegistry registry) {
...
         akka.stream.javadsl.Source::fromPublisher);
 
     registry.registerReactiveType(
         multiValue(akka.stream.scaladsl.Source.class, akka.stream.scaladsl.Source::empty),
         source ->
             ((akka.stream.scaladsl.Source<?, ?>) source)
-                .runWith(akka.stream.scaladsl.Sink.asPublisher(true), materializer),
+                .runWith(
+                    akka.stream.scaladsl.Sink.asPublisher(true),
+                    Materializer.matFromSystem(system)),
```
ennru, line 38: Shouldn't this be just `system`?

ennru (commented):
Adding back a deprecated `public AkkaStreamsRegistrar(Materializer mat)` which does `mat.system` and a corresponding getter would be enough, right?
Than go for it.

spring-web/src/main/java/akka/stream/alpakka/spring/web/AkkaStreamsRegistrar.java
dotbg: sadly, I have the compile-time error in that case :( Perhaps, I should do the relevant change in akka-streams.

ennru: No there are no changes required. The problem here is that it is using the Scala DSL from Java code. Your explicit use of `matFromSystem` is correct.

dotbg: yes, indeed. 
However, the logic does require scala dsl. Hence, my thought on having the corresponding method (accepting `system`, not `materializer`)

dotbg: I can wrap the scaladsl.Source with the javadsl.Source.
Besides, the constructor may look as follows 

```java
    /**
     * @deprecated use {@link #AkkaStreamsRegistrar(ActorSystem)}.
     */
  @Deprecated
  public AkkaStreamsRegistrar(Materializer materializer) {
      this(materializer.system());
  }
```

Shall I ?

ennru:
As the signature had to change from `ActorMaterializer` anyway, I just removed it altogether.

dotbg:
Rebased the branch.

ennru (approved):
LGTM.

ennru:
Thank you!

dotbg:
Thank you, @ennru

PR #2530, opened 2020-11-17
support for cloudSecureConnectBundle

patriknw:
* DataStax Astra

### DIFF
--- a/cassandra/src/main/resources/reference.conf
+++ b/cassandra/src/main/resources/reference.conf
@@ -13,6 +13,25 @@ alpakka.cassandra {
   # The ExecutionContext to use for the session tasks and future composition.
   session-dispatcher = "akka.actor.default-dispatcher"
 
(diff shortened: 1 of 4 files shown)


patriknw:
We need this for release-2.0, so should be backported.

patriknw:
I have verified that it works with the shopping-cart example, which includes both akka-persistence-cassandra and akka-projection-cassandra.

patriknw:
hm, I forgot that we already had documentation of how to do this with plain datastax-java-driver configuration
https://doc.akka.io/docs/akka-persistence-cassandra/current/astra.html

Meaning that this will not be needed, but I could add the documentation here also.

Let me confirm that it works.

patriknw:
It works with the config only, but requires the `session-keyspace` property, which is slightly unfortunate but I guess alright.

patriknw:
This wasn't needed. Can be done with ordinary `datastax-java-driver` config

Issue #2535, opened 2020-11-26
AWS S3: discrete requests should also be retried on exceptions


bpiper:
### Versions used 

2.0.2

Akka version: 2.6.7

### Expected Behavior

S3 requests that use S3Stream.signAndRequest should retry on exceptions (e.g. connection resets, which are not uncommon with Amazon S3 - or perhaps could be symptomatic of some obscure Akka HTTP issue) as well as 5xx HTTP responses.

### Actual Behavior

Requests fail immediately in the event of an exception.

I had been wondering why on some occasions my multipart uploads would fail with an akka.http.impl.engine.client.OutgoingConnectionBlueprint$UnexpectedConnectionClosureException (basically a connection reset) despite the part uploads using retry logic that covers both 5xx HTTP errors and IO exceptions. I then realised that it could actually be the multipart upload initiation or even completion requests that were failing, which was more or less confirmed by the fact that I couldn't see any incomplete multipart upload requests in the bucket (because they'd never been successfully initiated).

I think it makes sense to just use RetryFlow in signAndRequest, similar to how it's used for the part uploads.  I'm just testing this change now and will likely submit a PR in the not too distant future.

seglo:
Implemented in #2536

PR #2536, opened 2020-11-26
S3: Updated retry logic for S3Stream.signAndRequest

bpiper:
References #2535


seglo (commented):
Looking good. Just some notes on `Settings` boilerplate.

s3/src/main/scala/akka/stream/alpakka/s3/settings.scala
```diff
@@ -238,6 +238,7 @@ final class S3Settings private (
     val listBucketApiVersion: ApiVersion,
     val forwardProxy: Option[ForwardProxy],
     val validateObjectKey: Boolean,
+    val retrySettings: RetrySettings,
```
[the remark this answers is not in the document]
bpiper: Oops, looks like I missed that stuff in a previous PR as well.  Should hopefully all be there now.

s3/src/main/scala/akka/stream/alpakka/s3/settings.scala
```diff
@@ -229,6 +230,119 @@ object AccessStyle {
...
+  def getMaxBackoff: JavaDuration = JavaDuration.ofNanos(maxBackoff.toNanos)
+
+  /** Java API */
+  def getRandomFactor: Double = randomFactor
+
+  def withMaxRetries(value: Int): RetrySettings = copy(maxRetries = value)
+
+  def withMinBackoff(value: FiniteDuration): RetrySettings = copy(minBackoff = value)
+
+  def withMinBackoff(value: JavaDuration): RetrySettings =
```
seglo, line 254: Can you decorate these java setters with `/** Java API */` too?

seglo (commented):
Some minor docs updates.

s3/src/main/scala/akka/stream/alpakka/s3/settings.scala
bpiper: Sure.  I just added it to the ones that take `java.time.Duration` but let me know if I should add it to the others too.

seglo: I think that's all of them!

seglo:
Looks like there's a pile of MiMa exceptions that need to be added for the new APIs. The build should tell you the exact rules you need to add. LMK if you need help creating these.

https://travis-ci.com/github/akka/alpakka/jobs/449213845#L1126

bpiper:
@seglo any advice on resolving the binary compatibility issues? They're kind of a non issue since nobody's likely to be using those case class functions on the settings classes, but not sure if it's better to suppress MiMa somehow or preserve binary compatibility in some not-too-ugly way.

seglo (approved):
LGTM

seglo:
Thanks!

PR #2538, opened 2020-12-08
Tar reader: avoid deadlocks by synching in- and out-handlers

ennru:
The tar reader stage has situations where it would deadlock or get stuck in repeating the same file over and over again. This became particularly apparent when trying to extract nested tar files from within tar file in a single stream.

This PR changes the implementation to synchronize the switching of the `InHandler` and `OutHandler` for most parts, still switching the `OutHandler` when there is enough information to process from upstream.

To make theses changes I have a larger externally created tar file at hand which showed multiple flaws in the earlier tar reader implementation.


file/src/main/scala/akka/stream/alpakka/file/impl/archive/TarReaderStage.scala
```diff
@@ -37,70 +37,61 @@ private[file] class TarReaderStage
...
+              new OutHandler with InHandler {
+                override def onPull(): Unit = {
+                  setHandler(flowOut, IgnoreDownstreamPull)
+                  pushSource(metadata, buffer)
+                }
+
+                // fail on upstream push
+                override def onPush(): Unit = failStage(new TarReaderException("upstream pushed unexpectedly"))
+                override def onUpstreamFinish(): Unit = setKeepGoing(true)
               }
```
seglo, line 80: I suggest giving this handler a name like we have for the others.

file/src/test/java/FileProcessingUtil.java
```diff
@@ -0,0 +1,114 @@
+/*
```
seglo, line 1: Should we transform this into a test and/or include in the docs?

file/src/test/java/FileProcessingUtil.java
```diff
@@ -0,0 +1,114 @@
...
+  private static final int MAX_GUNZIP_CHUNK_SIZE = 64000;
+  private static final String TARGZ_EXT = "tar.gz";
+  private static final Logger log = LoggerFactory.getLogger(FileProcessingUtil.class);
+
+  public static void main(String[] args) throws Exception {
+
+    String filenameOk = "src/test/resources/container/lbktag_eakte_0000000015.tar";
+
+    // String filename=filenameOk;
+    String filename = "file/src/test/resources/nested-sample.tar";
```
seglo, line 34: Should we include this artifacts?

file/src/test/java/docs/javadsl/NestedTarReaderTest.java
```diff
@@ -0,0 +1,123 @@
...
+              } else if (targetFile.getFileName().toString().endsWith(TARGZ_EXT)) {
+                Path targetSubDir =
+                    targetFile
+                        .getParent()
+                        .resolve(
+                            Paths.get(
+                                targetFile
+                                    .getFileName()
+                                    .toString()
+                                    .substring(0, TARGZ_EXT.length() - 2)));
```
seglo, line 106: Seems a little hacky. Maybe we can make this path parsing easier for the user?

seglo (commented):
Looking good so far.

file/src/main/scala/akka/stream/alpakka/file/impl/archive/TarReaderStage.scala
ennru: 👍

file/src/test/java/FileProcessingUtil.java
ennru: This shouldn't have been there. It became `NestedTarReaderTest`.

file/src/test/java/docs/javadsl/NestedTarReaderTest.java
ennru: This is not part of our API and I'd rather not add something for it.

seglo (approved):
LGTM

seglo:
We don't have a date for the next release, but I'm hoping to release `3.0.0-M1` sometime in Feb.

Scarysize:
Hey, sorry to chime in on this closed PR. We are also pretty anxious getting access to this fix. Either with a v3 release or a patch for v2. Any chance we can help with a back port?

seglo:
Yes, things haven't gone as quickly as I hoped, but I plan to release `3.0.0-M1` tomorrow or over the weekend.

PR #2540, opened 2020-12-08
Run link validator and release drafter (as GitHub workflows)

ennru:
...and fix links that were reported to require an updated.

References https://github.com/akka/alpakka-kafka/pull/1267

### DIFF
--- a/.github/workflows/link-validator.yml
+++ b/.github/workflows/link-validator.yml
@@ -1,8 +1,7 @@
 name: Link Validator
 
 on:
-  pull_request_target:
-    branches: [master, main]
+  pull_request:
   schedule:
     - cron:  '0 6 * * 1'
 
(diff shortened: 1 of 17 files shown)


.github/workflows/link-validator.yml
```diff
@@ -0,0 +1,38 @@
+name: Link Validator
+
+on:
+  pull_request_target:
+    branches: [master, main]
```
raboof: @ennru if I understand correctly from different PR's this is not the right way, right?

ennru: Hm, you looked at an outdated version. I've updated this to `pull_request`.

raboof: oh, weird, sorry

raboof (approved):

PR #2543, opened 2020-12-11
SSE: Akka 2.6

ennru:
References #2502


sse/src/main/scala/akka/stream/alpakka/sse/javadsl/EventSource.scala
```diff
@@ -81,7 +106,7 @@ object EventSource {
           uri.asScala,
           send(_).toScala.map(_.asInstanceOf[SHttpResponse])(mat.executionContext),
           lastEventId.asScala
-        )(mat)
+        )(mat.system)
```
seglo, line 109: Should we deprecate this overload?

seglo (approved):
LGTM. One consideration for deprecation.

sse/src/main/scala/akka/stream/alpakka/sse/javadsl/EventSource.scala
ennru: Marked it as deprecated.

Issue #773, opened 2018-02-13
Sharing code between google cloud modules


francisdb:
Something to handle after https://github.com/akka/alpakka/pull/764 and https://github.com/akka/alpakka/pull/650 have been merged

This would unify configuration / session handling for all google cloud related modules. For amazon we are using their sdk but @tg44 pointed out that it would be a bad idea for the google cloud sdk

We should still keep the possibility to configure manually if you want to use multiple service accounts for example

Previous discussion:
https://github.com/akka/alpakka/pull/650#issuecomment-364968964

tg44:
Related to this: 
Use the jwt-core, or do it by hand? (fcm uses jwt-core, pubsub uses byhand)
Pro: less code
Con: more deps (signed jar cause of the bouncycastle dep)

francisdb:
@ennru would you be open for a shared (internal?) module google-common that is used by fcm/storage/pubsub? Otherwise we should close this.

2m:
I think shared module sounds good. It will still have to be published as a jar, but all of the classes in that jar can be inside the `impl` package.

ennru:
https://github.com/akka/alpakka/pull/2451 copied in code from another Google connector.

A shared module would make sense to not duplicate code but on the other hand, users will need to use the exact same Alpakka version for different connectors, which is not required this far.

tg44:
Do we need that? Can't we sbt magic the module files to the given submodules at compile time? There are pros and cons with this solution, but in that case we could write/test code as a module, but we don't need to publish it, and all submodules would use their own implementations.

gkatzioura:
What it seems to be very valuable would be to extract the http code that happens behind the scenes accompanied with the re-authentication/token-refresh, and have a library or even a http-connector (that handles the token refresh) so that if you cannot find a connector for a GCP component you can roll one out on your own easily.
For example if I want to use a StackDriver connector right now I would copy the code from p:google-cloud-bigquery or p:google-cloud-storage because they already have the code needed and the GCP api does not differ a-lot. 
If an http base code is out there will make it easier, like a toolkit for GCP connectors. 
This could apply to new GCP features since you expect them to use the usual GRPC and HTTP APIs

tg44:
[...]

PR #2548, opened 2020-12-20
BigQuery: New BigQuery connector

armanbilge:
References #2353, #2533; cc @tg44

I began working on a PR adding sinks to the BigQuery connector but this evolved into something much bigger. I hope such an extensive PR is not unwelcome! Perhaps targeting Alpakka 3?

Although much is borrowed from the existing connector, my goal was to mimic the APIs of the more mature connectors in Alpakka (I looked to S3 and Google Cloud Storage as my examples).

The features of this connector are currently best demonstrated in the [end-to-end spec](https://github.com/armanbilge/alpakka/blob/wip-new-bigquery/google-cloud-bigquery/src/test/scala/akka/stream/alpakka/googlecloud/bigquery/e2e/scaladsl/BigQueryEndToEndSpec.scala):
- Methods for [retrieving/listing/creating/deleting](https://github.com/armanbilge/alpakka/blob/wip-new-bigquery/google-cloud-bigquery/src/test/scala/docs/scaladsl/BigQueryDoc.scala#L83) datasets and tables
- Automatic schema generation for case classes when [creating tables](https://github.com/armanbilge/alpakka/blob/wip-new-bigquery/google-cloud-bigquery/src/test/scala/docs/scaladsl/BigQueryDoc.scala#L96)
- [A sink](https://github.com/armanbilge/alpakka/blob/wip-new-bigquery/google-cloud-bigquery/src/test/scala/docs/scaladsl/BigQueryDoc.scala#L58) for real-time streaming inserts to tables with configurable de-duplication
- [A flow](https://github.com/armanbilge/alpakka/blob/wip-new-bigquery/google-cloud-bigquery/src/test/scala/docs/scaladsl/BigQueryDoc.scala#L63) for "poor man's" inserts via asynchronous load jobs (load jobs are free-of-charge)
- [A source](https://github.com/armanbilge/alpakka/blob/wip-new-bigquery/google-cloud-bigquery/src/test/scala/docs/scaladsl/BigQueryDoc.scala#L53) for all rows/columns of a table (a fast, cost-saving alternative to running a `SELECT *` query)
- [A source](https://github.com/armanbilge/alpakka/blob/wip-new-bigquery/google-cloud-bigquery/src/test/scala/docs/scaladsl/BigQueryDoc.scala#L42) for results of SQL queries that also [materializes metadata](https://github.com/armanbilge/alpakka/blob/wip-new-bigquery/google-cloud-bigquery/src/test/scala/docs/scaladsl/BigQueryDoc.scala#L48) (e.g., bytes processed)
- Automatic marshalling/unmarshalling to/from case classes and [even tuples](https://github.com/armanbilge/alpakka/blob/wip-new-bigquery/google-cloud-bigquery/src/test/scala/docs/scaladsl/BigQueryDoc.scala#L42)
- Sufficient abstraction (IMHO) to let a power user easily make [custom API calls](https://github.com/armanbilge/alpakka/blob/wip-new-bigquery/google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/BigQueryRest.scala#L30) beyond what our implementation supports
- [Fully configurable](https://github.com/armanbilge/alpakka/blob/wip-new-bigquery/google-cloud-bigquery/src/main/resources/reference.conf#L30) retrying-with-backoff of failed requests
- All configuration managed via `BigQuerySettings` either as an implicit parameter or a stream attribute
- Automatic retrieval of credentials when running on Google Compute Engine

Behind the scenes, I simplified the implementation by replacing all custom graph stages used in the current connector with higher level DSLs.

**There is one frustrating thorn in my side:** apparently the BigQuery REST API for upload jobs responds with two `Content-Type` headers, which causes Akka HTTP to fail with `HTTP message must not contain more than one Content-Type header` errors. I filed akka/akka-http#3712 and [176002651](https://issuetracker.google.com/176002651) about this, but I'd be grateful for any suggestions for how to work around this in the meantime.

Obviously there's still quite a bit of work left to be done, but I'd be grateful for any and all feedback. Apologies if this PR is overly aggressive or ambitious; if so I can try breaking it up into a series of smaller PRs perhaps. Thanks in advance for your time and reviews.

TODO:
- [x] Resolve double `Content-Type` header issue (workaround ~pending~ merged akka/akka-http#3727)
- [x] Improve test coverage
- [x] Implement Java DSL
- [x] Write lots of documentation and examples


google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/BigQuery.scala
```diff
@@ -0,0 +1,513 @@
...
+
+        Flow[TableDataInsertAllRequest[In]]
+          .mapAsync(1)(Marshal(_).to[RequestEntity])
+          .map(request.withEntity)
+          .mapAsync(1)(requestWithOAuth)
+          .mapAsync(1)(Unmarshal(_).to[TableDataInsertAllResponse])
+      }
+      .mapMaterializedValue(_ => NotUsed)
+
+  def insertAllAsync[In](datasetId: String,
```
armanbilge, line 358: Thoughts on method name? It is similar in functionality to `insertAll` (adding rows to an existing table). However `insertAll` implements [(near) real-time streaming inserts](https://cloud.google.com/bigquery/streaming-data-into-bigquery), whereas `insertAllAsync` creates load jobs that ["execute asynchronously and can be polled for their status."](https://cloud.google.com/bigquery/docs/jobs-overview)

project/Dependencies.scala
```diff
@@ -210,13 +210,13 @@ object Dependencies {
     libraryDependencies ++= Seq(
         "com.typesafe.akka" %% "akka-http" % AkkaHttpVersion,
         "com.typesafe.akka" %% "akka-http-spray-json" % AkkaHttpVersion,
-        "com.pauldijou" %% "jwt-core" % JwtCoreVersion, //ApacheV2
+        "com.pauldijou" %% "jwt-core" % "3.1.0", //ApacheV2
```
armanbilge, line 213: Consider bumping the version for the entire Alpakka project? The latest is 4.2.0. The jump from 3 to 4 is apparently ["not really a breaking change release."](https://github.com/pauldijou/jwt-scala/releases/tag/v4.0.0)

tg44:
I didn't checked the code yet, but;

> Behind the scenes, I simplified the implementation by replacing all custom graph stages used in the current connector with higher level DSLs.

This could break things. Lots of things. We have a cycle in the graph which is highly not advised, but we need it anyway. Managing backpressure in cycles without making deadlocks is hard. I know bcs I did it, and every time sb asks why we don't just use a cycle, I show them some of the [mentions from the docs](https://doc.akka.io/docs/akka/current/stream/stream-graphs.html#graph-cycles-liveness-and-deadlocks). Even so "sometimes" some stages inner implementation changes (#2379).

armanbilge:
Oops, I should have assumed there were good reasons for these custom stages. I will investigate forthwith.

armanbilge:
@tg44 thanks for pointing out the complexities of graph cycles, it was a nudge in the right direction: I was able to use `RestartSource` and `Source.unfoldAsync` to re-implement the paginated request flow without any cycles :) bye-bye deadlocks!

tg44:
One (probably last) thing;
We used exponential backoff when the next page was not ready yet. I'm not sure if we still have this functionality. If the next page created slowly you can get back responses with "please wait more" kind of thing. And as the docs said when you get one of these, you should wait more and more.

We implemented this anno, bcs we had rather slow queries, and the initial tests on the live data needed it.

armanbilge:
Yup, the RestartSource implements an exponential backoff with unlimited restarts.

armanbilge:
Update: Google has acknowledged the duplicate `Content-Type` header problem (issue [176002651](https://issuetracker.google.com/176002651)) although there is no ETA on when it may be fixed. I also have PR akka/akka-http#3727 pending merge as a workaround. In any case, this problem only afflicts a small part of this connector.

If there's still interest, I'd love to see this PR merged; I'm already using it in my own project. I was hoping for a quick review just to make sure you'd be happy with these extensive changes before I continue investing work into creating a Java DSL and writing the docs/examples. Thanks in advance!

ennru:
Thank you for this effort!
Alpakka is moving very slowly for the time being, but `master` is targeting Alpakka 3 already.

To bring this forward we'd need to make sure Travis CI is happy with it.

armanbilge:
No problem, thanks for the reply! Travis CI is rejecting because this PR is still WIP (need to write Java DSL and docs). I'm more than happy to fix this all up, I just wanted to be sure you'd accept a PR with such extensive changes before I continue working on it. Sounds like this is okay?

ennru:
Would be great if @tg44 can continue reviewing your work so it gets to a mergeable state.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/BigQueryAttributes.scala
```diff
@@ -0,0 +1,44 @@
+/*
```
tg44, line 1: It is possile that it is just me, but when I have multiple classes/objects in the same file, I split them to multiple files, or place the "sub"-classes into the "main" object of the file... So I would move the last 3 classes into the `BigQueryAttributes`.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/InsertAllRetryPolicy.scala
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package akka.stream.alpakka.googlecloud.bigquery
+
+sealed abstract class InsertAllRetryPolicy {
+  def retry: Boolean
+  def deduplicate: Boolean
+}
+
+case object NoRetries extends InsertAllRetryPolicy {
```
tg44, line 12: I would move these to an `InsertAllRetryPolicy` object too.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/impl/auth/ServiceAccountCredentials.scala
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright (C) 2016-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.googlecloud.bigquery.impl.auth
```
tg44, line 5: We agreed multiple times, that this package should be a separate project/module. Probably if this get merged, I will sbt-hack it to use this in every google related project and stop duplicating code... (But I like the way as it gets bigger and bigger :D)

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/impl/http/ForwardProxyPoolSettings.scala
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright (C) 2016-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.googlecloud.bigquery.impl.http
```
tg44, line 5: I'm not entirely sure, but this forward proxy mechanism is also shared code among other connectors is it?

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/impl/util/AnnotateLast.scala
```diff
@@ -0,0 +1,54 @@
...
+}
+
+@InternalApi
+private[impl] final case class Last[+T](value: T) extends MaybeLast[T] {
+  override def isLast: Boolean = true
+  override def map[S](f: T => S): Last[S] = Last(f(value))
+}
+
+@InternalApi
+private[impl] object AnnotateLast {
```
tg44, line 36: would pack all of the above inside this

tg44 (commented):
I think overall it is a good direction, and now you officially know more about this connector than I am :D

I commented on some nitpicking, those can be totally skipped if you do not agree with them. I would also split the BigQuery.scala into a query and nonquery functionality to make it smaller (500+ lines hurts my OCD), and more manageable.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/BigQueryAttributes.scala
armanbilge: I think it's just you ;) e.g. `Some` and `None` are not inside of `Option`, same for `Either`/`Left`/`Right`.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/InsertAllRetryPolicy.scala
armanbilge: Actually I agree here! `InsertAllRetryPolicy` is basically an enum.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/impl/http/ForwardProxyPoolSettings.scala
armanbilge: Yes that's right. Actually, I think my `BigQueryHttp` class should become a common `GoogleHttp` class that handles the proxy stuff, requests with OAuth, retries, etc.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/impl/auth/ServiceAccountCredentials.scala
armanbilge: Yes, I saw this mentioned many times, the main discussion is in #773? I would fully support this!

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/BigQueryAttributes.scala
armanbilge: BTW, I just copied class structure is taken from the other connectors, `S3` IIRC.

armanbilge:
Thanks as always for the prompt and thorough review @tg44! I split the `BigQuery` object into traits based on the different resources: `BigQueryRest`, `BigQueryDatasets`, `BigQueryTables`, `BigQueryTableData`, `BigQueryJobs`, hopefully this should be nicer although `BigQueryJobs` is still on the larger side.

I've implemented most of the Java DSL, all that's left is to figure out how to handle parsing with Jackson (this is Akka's preferred JSON library for Java DSLs) and write up the examples/docs.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/BigQueryAttributes.scala
tg44: hmm...

https://docs.scala-lang.org/style/files.html says that `class/trait + object` should be in a file, and if you start to introduce adts you should lowercase the filename (like `option.scala`). No idea when I started to pack every subclass of a sealed trait to the companion object, but I think it started to develop when we tried out tagless final in some projects :D I told you it is nitpicking, I would pack it, you would not, and the scala codestyle says we should lowercase the filename which I would personally never do.

I can let it go :P

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/InsertAllRetryPolicy.scala
armanbilge: I ended up reverting this. Still not good at Scala/Java interop but I had problems accessing from Java when they were nested inside.

armanbilge:
@ennru This PR is now in a mergeable state, pending final reviews: Travis build passes and Scala+Java DSLs have feature parity, complete with docs, examples, and tests.

armanbilge:
I'm pleased to report that akka/akka-http#3727 was just merged, which provides a workaround for the duplicated `Content-Type` header problem until Google fixes their API.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/impl/PaginatedRequest.scala
```diff
@@ -0,0 +1,62 @@
...
+      implicit paginated: Paginated[Out]
+  ): Source[Out, NotUsed] =
+    Source
+      .fromMaterializer { (mat, attr) =>
+        implicit val system = mat.system
+        implicit val settings = BigQueryAttributes.resolveSettings(attr, mat)
+
+        Source.unfoldAsync[Either[Done, Option[String]], Out](Right(initialPageToken)) {
+          case Left(Done) =>
+            FastFuture.successful(None)
```
seglo, line 35: There's probably a reason for `FastFuture` in akka-http, but it seems we could just use  `Future.successful(None)` here. We could avoid multiple allocations by defining it at the object scope and just returning it when the pages are done.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/impl/auth/ServiceAccountCredentials.scala
seglo: The consensus in #773 seems to be to factor out Google auth things into an external library. If either of you are interested in starting a project we can bring it under the akka organization (perhaps `akka/alpakka-google-auth`) and setup a release for it.

Realistically this is probably something the akka team will get to soon, but I'm willing to help facilitate the integration/library/release/etc.

seglo (commented):
We should see an akka-http release shortly so I think it makes sense to resume reviewing this PR. This looks like amazing work. Thanks for all the time you've put into it. I did a first pass, but this is enormous so it will take some time.

General comments:

* There's a [Google Cloud BigTable emulator](https://cloud.google.com/bigtable/docs/emulator). Have you considered using this in place or in addition to your existing e2e testing suite? The Alpakka Google Pub Sub connector uses it (see how it's setup in `docker-compose.yml`).
* Creating a separate Alpakka Google Auth/Http module seems like a good idea. Definitely in support of that effort if you or anyone else would like to take the lead. We could host and release the library from the akka or lightbend organization.
* There are some travis build failures that need to be addressed.

armanbilge:
Thanks for the initial review @seglo! Much appreciated.

- Re emulator: unless I'm confused about something, BigTable != BigQuery so I don't think it can help us here.
- I'd be happy to help make an Alpakka Google Auth/Http module (I assume as a part of this project, or do you mean another project/repo?) ... if someone could help me with the basic module setup (sbt etc.) I am more than happy to hack away at it.
- Re travis: sorry! I accidentally introduced Java APIs not available in the target.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/BigQueryQueries.scala
```diff
@@ -0,0 +1,157 @@
...
+                val restartSettings = RestartSettings(minBackoff, maxBackoff, randomFactor)
+                RestartSource.onFailuresWithBackoff(restartSettings)(() => pages).map(_.get)
+              }
+
+            head.concat(tail)
+          }
+
+        }
+      }
+      .wireTapMat(Sink.head)(Keep.right)
```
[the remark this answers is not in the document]
armanbilge: Yes, that's exactly what it does. My initial instinct was to use `alsoTo` but because `Sink.head` cancels after receiving its first element, this would cause the entire stream to cancel. On the other hand `wireTap` is unaffected by the secondary sink (the tap).

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/BigQueryDatasets.scala
```diff
@@ -0,0 +1,128 @@
...
+   * @param filter a key, value [[scala.collection.immutable.Map]] for filtering the results of the request by label
+   * @return a [[akka.stream.scaladsl.Source]] that emits each [[akka.stream.alpakka.googlecloud.bigquery.model.DatasetJsonProtocol.Dataset]]
+   */
+  def datasets(maxResults: Option[Int] = None,
+               all: Option[Boolean] = None,
+               filter: Map[String, String] = Map.empty): Source[Dataset, NotUsed] =
+    source { settings =>
+      import SprayJsonSupport._
+      val uri = BigQueryEndpoints.datasets(settings.projectId)
+      val query = ("maxResults" -> maxResults) ?+:
```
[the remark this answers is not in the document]
armanbilge: Sorry! Just a convenience method for my own sake, happy to change.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/BigQueryQueries.scala
```diff
@@ -0,0 +1,157 @@
...
+        import SprayJsonSupport._
+        import mat.executionContext
+        implicit val system = mat.system
+        implicit val settings = BigQueryAttributes.resolveSettings(attr, mat)
+
+        Source.lazyFutureSource { () =>
+          for {
+            entity <- Marshal(query).to[RequestEntity]
+            initialRequest = HttpRequest(POST, BigQueryEndpoints.queries(settings.projectId), entity = entity)
+            response <- BigQueryHttp().retryRequestWithOAuth(initialRequest)
```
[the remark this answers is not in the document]
armanbilge: The source should fail right? Is this not the desired behavior?

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/impl/auth/ServiceAccountCredentials.scala
tg44: I'm happy to help! Actually, I started to play with it at some point, but the main question is "where?" remained unclear. I think it would be convenient to do it here, but also this project starting to be huge. If I get a green lamp, I'm happy to start to refactor the common google part to a subproject.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/BigQueryQueries.scala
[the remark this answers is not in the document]
seglo: It should fail, but I think because you're using the for comprehension to resolve this intermediate `Future` any failure will just get swallowed and the end result would be the `Source` getting stuck.

I reproduced that behaviour with this sample.

```
        def fail: Future[String] = Future(throw new Exception("error"))

        val source: Source[String, Future[NotUsed]] = Source.lazyFutureSource { () =>
          for {
            r1 <- Future.successful("f1")
            r2 <- fail
          } yield Source(List(r1, r2))
        }

        val f = source.runWith(Sink.collection)

        val strs = Await.result(f, 5.seconds)
```

I could be missing something though..

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/BigQueryQueries.scala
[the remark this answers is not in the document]
seglo: :+1:

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/BigQueryDatasets.scala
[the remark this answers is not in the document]
seglo: No, that's ok. It seems like a good fit with akka-http's `+:`.

seglo:
> Re emulator: unless I'm confused about something, BigTable != BigQuery so I don't think it can help us here.

I got my Google BigThings confused. Nevermind :)

> I'd be happy to help make an Alpakka Google Auth/Http module (I assume as a part of this project, or do you mean another project/repo?) ... if someone could help me with the basic module setup (sbt etc.) I am more than happy to hack away at it.

@armanbilge 

> I'm happy to help! Actually, I started to play with it at some point, but the main question is "where?" remained unclear. I think it would be convenient to do it here, but also this project starting to be huge. If I get a green lamp, I'm happy to start to refactor the common google part to a subproject.

@tg44 

My understanding from #773 is @ennru was proposing a separate library (in a separate repository) so it could have a separate version and release process. Then we can manage different versions of the library across the google connectors and we're not forced to update them all at once.  @ennru is that correct or did you have something else in mind?

> A separate library to plug in the authentication seems attractive, especially if it gets improved to what we have today.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/BigQueryQueries.scala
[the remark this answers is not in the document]
armanbilge: Really? I just ran your code snippet and it throws the exception.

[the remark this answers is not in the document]
seglo: That's strange. I added the snippet as a test in `BigQueryQueriesSpec` and [pushed a branch](https://github.com/seglo/alpakka/tree/seglo/wip-new-bigquery).

https://github.com/seglo/alpakka/commit/8cff3db2465a1abeb19fcbe45c64912abd859334

```
sbt:alpakka> google-cloud-bigquery/testOnly *.BigQueryQueriesSpec -- -z "should fail when failed future for comprehension"
[info] BigQueryQueriesSpec:
14:18:24.150 [pool-1-thread-1] INFO io.specto.hoverfly.junit.core.TempFileManager - Selecting the following binary based on the current operating system: hoverfly_linux_amd64
14:18:24.152 [pool-1-thread-1] INFO io.specto.hoverfly.junit.core.TempFileManager - Storing binary in temporary directory /tmp/hoverfly.2557597409281694884/hoverfly_linux_amd64
14:18:24.230 [pool-1-thread-1] INFO io.specto.hoverfly.junit.core.Hoverfly - Executing binary at /tmp/hoverfly.2557597409281694884/hoverfly_linux_amd64
14:18:24.234 [pool-1-thread-1] DEBUG org.zeroturnaround.exec.ProcessExecutor - Executing [/tmp/hoverfly.2557597409281694884/hoverfly_linux_amd64, -pp, 8500, -ap, 8888, -logs, json] in /tmp/hoverfly.2557597409281694884.
14:18:24.238 [pool-1-thread-1] DEBUG org.zeroturnaround.exec.ProcessExecutor - Started Process[pid=56875, exitValue="not exited"]
14:18:24.248 [Thread-3] INFO hoverfly - Default proxy port has been overwritten port=8500
14:18:24.250 [Thread-3] INFO hoverfly - Default admin port has been overwritten port=8888
14:18:24.250 [Thread-3] INFO hoverfly - Using memory backend
14:18:24.250 [Thread-3] INFO hoverfly - Proxy prepared... Destination=. Mode=simulate ProxyPort=8500
14:18:24.251 [Thread-3] INFO hoverfly - current proxy configuration destination=. mode=simulate port=8500
14:18:24.251 [Thread-3] INFO hoverfly - Admin interface is starting... AdminPort=8888
14:18:24.251 [Thread-3] INFO hoverfly - serving proxy
14:18:24.257 [Thread-3] INFO hoverfly - Mode has been changed mode=simulate
14:18:24.262 [pool-1-thread-1] INFO io.specto.hoverfly.junit.core.ProxyConfigurer - Setting proxy host to localhost
14:18:24.262 [pool-1-thread-1] INFO io.specto.hoverfly.junit.core.ProxyConfigurer - Setting proxy proxyPort to 8500
[info] BigQueryQueries
[info]   should get query results
[info]   should fail
[info]   - when failed future for comprehension *** FAILED *** (5 seconds, 53 milliseconds)
[info]     java.util.concurrent.TimeoutException: Futures timed out after [5 seconds]
[info]     at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:259)
[info]     at scala.concurrent.impl.Promise$DefaultPromise.result(Promise.scala:263)
[info]     at scala.concurrent.Await$.$anonfun$result$1(package.scala:220)
[info]     at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:57)
[info]     at scala.concurrent.Await$.result(package.scala:146)
[info]     at akka.stream.alpakka.googlecloud.bigquery.scaladsl.BigQueryQueriesSpec.$anonfun$new$13(BigQueryQueriesSpec.scala:216)
[info]     at org.scalatest.wordspec.AsyncWordSpecLike.transformToOutcomeParam$1(AsyncWordSpecLike.scala:134)
[info]     at org.scalatest.wordspec.AsyncWordSpecLike.$anonfun$registerTestToRun$1(AsyncWordSpecLike.scala:135)
[info]     at org.scalatest.AsyncTestSuite.$anonfun$transformToOutcome$1(AsyncTestSuite.scala:240)
[info]     at org.scalatest.wordspec.AsyncWordSpecLike$$anon$3.apply(AsyncWordSpecLike.scala:1061)
[info]     at org.scalatest.AsyncTestSuite.withFixture(AsyncTestSuite.scala:313)
[info]     at org.scalatest.AsyncTestSuite.withFixture$(AsyncTestSuite.scala:312)
[info]     at akka.stream.alpakka.googlecloud.bigquery.scaladsl.BigQueryQueriesSpec.withFixture(BigQueryQueriesSpec.scala:33)
[info]     at org.scalatest.wordspec.AsyncWordSpecLike.invokeWithAsyncFixture$1(AsyncWordSpecLike.scala:1059)
[info]     at org.scalatest.wordspec.AsyncWordSpecLike.$anonfun$runTest$1(AsyncWordSpecLike.scala:1073)
[info]     at org.scalatest.AsyncSuperEngine.runTestImpl(AsyncEngine.scala:374)
[info]     at org.scalatest.wordspec.AsyncWordSpecLike.runTest(AsyncWordSpecLike.scala:1073)
[info]     at org.scalatest.wordspec.AsyncWordSpecLike.runTest$(AsyncWordSpecLike.scala:1053)
[info]     at akka.stream.alpakka.googlecloud.bigquery.scaladsl.BigQueryQueriesSpec.runTest(BigQueryQueriesSpec.scala:33)
[info]     at org.scalatest.wordspec.AsyncWordSpecLike.$anonfun$runTests$1(AsyncWordSpecLike.scala:1132)
[info]     at org.scalatest.AsyncSuperEngine.$anonfun$runTestsInBranch$1(AsyncEngine.scala:432)
[info]     at scala.collection.LinearSeqOptimized.foldLeft(LinearSeqOptimized.scala:126)
[info]     at scala.collection.LinearSeqOptimized.foldLeft$(LinearSeqOptimized.scala:122)
[info]     at scala.collection.immutable.List.foldLeft(List.scala:89)
[info]     at org.scalatest.AsyncSuperEngine.traverseSubNodes$1(AsyncEngine.scala:406)
[info]     at org.scalatest.AsyncSuperEngine.runTestsInBranch(AsyncEngine.scala:479)
[info]     at org.scalatest.AsyncSuperEngine.$anonfun$runTestsInBranch$1(AsyncEngine.scala:460)
[info]     at scala.collection.LinearSeqOptimized.foldLeft(LinearSeqOptimized.scala:126)
[info]     at scala.collection.LinearSeqOptimized.foldLeft$(LinearSeqOptimized.scala:122)
[info]     at scala.collection.immutable.List.foldLeft(List.scala:89)
[info]     at org.scalatest.AsyncSuperEngine.traverseSubNodes$1(AsyncEngine.scala:406)
[info]     at org.scalatest.AsyncSuperEngine.runTestsInBranch(AsyncEngine.scala:479)
[info]     at org.scalatest.AsyncSuperEngine.$anonfun$runTestsInBranch$1(AsyncEngine.scala:460)
[info]     at scala.collection.LinearSeqOptimized.foldLeft(LinearSeqOptimized.scala:126)
[info]     at scala.collection.LinearSeqOptimized.foldLeft$(LinearSeqOptimized.scala:122)
[info]     at scala.collection.immutable.List.foldLeft(List.scala:89)
[info]     at org.scalatest.AsyncSuperEngine.traverseSubNodes$1(AsyncEngine.scala:406)
[info]     at org.scalatest.AsyncSuperEngine.runTestsInBranch(AsyncEngine.scala:487)
[info]     at org.scalatest.AsyncSuperEngine.runTestsImpl(AsyncEngine.scala:555)
[info]     at org.scalatest.wordspec.AsyncWordSpecLike.runTests(AsyncWordSpecLike.scala:1132)
[info]     at org.scalatest.wordspec.AsyncWordSpecLike.runTests$(AsyncWordSpecLike.scala:1131)
[info]     at akka.stream.alpakka.googlecloud.bigquery.scaladsl.BigQueryQueriesSpec.runTests(BigQueryQueriesSpec.scala:33)
[info]     at org.scalatest.Suite.run(Suite.scala:1112)
[info]     at org.scalatest.Suite.run$(Suite.scala:1094)
[info]     at akka.stream.alpakka.googlecloud.bigquery.scaladsl.BigQueryQueriesSpec.org$scalatest$wordspec$AsyncWordSpecLike$$super$run(BigQueryQueriesSpec.scala:33)
[info]     at org.scalatest.wordspec.AsyncWordSpecLike.$anonfun$run$1(AsyncWordSpecLike.scala:1177)
[info]     at org.scalatest.AsyncSuperEngine.runImpl(AsyncEngine.scala:625)
[info]     at org.scalatest.wordspec.AsyncWordSpecLike.run(AsyncWordSpecLike.scala:1177)
[info]     at org.scalatest.wordspec.AsyncWordSpecLike.run$(AsyncWordSpecLike.scala:1175)
[info]     at akka.stream.alpakka.googlecloud.bigquery.scaladsl.BigQueryQueriesSpec.org$scalatest$BeforeAndAfterAll$$super$run(BigQueryQueriesSpec.scala:33)
[info]     at org.scalatest.BeforeAndAfterAll.liftedTree1$1(BeforeAndAfterAll.scala:213)
[info]     at org.scalatest.BeforeAndAfterAll.run(BeforeAndAfterAll.scala:210)
[info]     at org.scalatest.BeforeAndAfterAll.run$(BeforeAndAfterAll.scala:208)
[info]     at akka.stream.alpakka.googlecloud.bigquery.scaladsl.BigQueryQueriesSpec.run(BigQueryQueriesSpec.scala:33)
[info]     at org.scalatest.tools.Framework.org$scalatest$tools$Framework$$runSuite(Framework.scala:318)
[info]     at org.scalatest.tools.Framework$ScalaTestTask.execute(Framework.scala:513)
[info]     at sbt.ForkMain$Run.lambda$runTest$1(ForkMain.java:413)
[info]     at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[info]     at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
[info]     at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
[info]     at java.base/java.lang.Thread.run(Thread.java:834)
14:18:29.384 [pool-1-thread-1] INFO io.specto.hoverfly.junit.core.Hoverfly - Destroying hoverfly process
14:18:29.385 [WaitForProcess-Process[pid=56875, exitValue="not exited"]] DEBUG org.zeroturnaround.exec.WaitForProcess - Process[pid=56875, exitValue=143] stopped with exit code 143
[info] ScalaTest
[info] Run completed in 6 seconds, 473 milliseconds.
[info] Total number of tests run: 1
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 0, failed 1, canceled 0, ignored 0, pending 0
[info] *** 1 TEST FAILED ***
[error] Failed: Total 1, Failed 1, Errors 0, Passed 0
[error] Failed tests:
[error]         akka.stream.alpakka.googlecloud.bigquery.scaladsl.BigQueryQueriesSpec
[error] (google-cloud-bigquery / Test / testOnly) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 7 s, completed Feb. 23, 2021, 2:18:29 p.m.
sbt:alpakka>
```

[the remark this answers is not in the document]
armanbilge: Ok thanks, when I run your test it times out as you say. Still confused why this doesn't happen when I run your snippet in a new project or why it happens at all. I'm going to poke around with this a bit.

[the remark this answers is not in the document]
armanbilge: Ok, I can get your test the work by adding `import system.dispatcher`. The timeout you observe seems to occur when using the `SerialExecutionContext` provided by `AsyncTestSuite`.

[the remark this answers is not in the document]
seglo: Got it. Thanks for the explanation.

armanbilge:
> @ennru was proposing a separate library (in a separate repository) so it could have a separate version and release process. Then we can manage different versions of the library across the google connectors and we're not forced to update them all at once.

If you don't mind me asking—I am struggling to comprehend how the Alpakka project brings multiple disparate libraries together under a single roof, yet on one of the (few?) occasions that these libraries do share some common code, this code may get shunted off to its own repository (and with excellent rationale!)?

seglo:
> If you don't mind me asking—I am struggling to comprehend how the Alpakka project brings multiple disparate libraries together under a single roof, 

It's a challenge! But the alternative is to maintain 40+ repositories so this is the lesser of two evils.

> yet on one of the (few?) occasions that these libraries do share some common code, this code may get shuttered off to its own repository (and with excellent rationale!)?

The testkit is one example of shared code. Aside from that, code has just been duplicated across projects, like in this case. We could host the shared code here of course, but I think in this case there's enough shared code to warrant a separate library to reduce the maintenance burden of alpakka in general. I'm not 100% committed to that direction though.

I'm not sure I answered your question. Are you saying you would prefer a subproject in this repo?

armanbilge:
> But the alternative is to maintain 40+ repositories so this is the lesser of two evils.

You did answer my question, thanks for the explanation!

> Are you saying you would prefer a subproject in this repo?

In any other case, I would say a subproject makes the most sense. After all, Alpakka itself will likely be the primary user/driver of `alpakka-google-auth` development. _But_ in this case, between the maintenance burden and versioning issues, I see the appeal of a separate library.

Probably a non-option, but my vote would be for a separate `alpakka-google` project altogether. I think there's enough connectors to warrant it  (I count 5 and growing e.g. #2180).

tg44:
Yapp, when I saw this discussion this was my idea too. Kafka has a separate repo, probably the google connectors could have one too. It is 5 from 48 right now. 12%, I'm not sure...

seglo:
> Yapp, when I saw this discussion this was my idea too. Kafka has a separate repo, probably the google connectors could have one too. It is 5 from 48 right now. 12%, I'm not sure...

Kafka has a separate repo for historical reasons (it pre-dates the Alpakka project). It's also by far the most used and complex of the connectors, so I think it makes sense.

I'm not sure what added benefit there would be to break out the google connectors into their own repo. I'm not too fond of creating the separate repo just for the google shared code library either. Creating a separate sub project in this repo will add a nominal amount of complexity to the release process, but I think it's the best way forward all around.

docs/src/main/paradox/google-cloud-bigquery.md
```diff
@@ -15,114 +17,177 @@ Alpakka Google Cloud BigQuery was added in Alpakka 2.0.2 in July 2020 and is mar
...
-: @@snip [snip](/google-cloud-bigquery/src/test/scala/docs/scaladsl/GoogleBigQuerySourceDoc.scala) { #imports }
+The table below shows direct dependencies of this module and the second tab shows all libraries that it depends on transitively.
 
-Java
-: @@snip [snip](/google-cloud-bigquery/src/test/java/docs/javadsl/GoogleBigQuerySourceDoc.java) { #imports }
+@@dependencies { projectId="google-cloud-bigquery" }
+
+## Configuration
+
+The settings for the BigQuery connector are read by default from `alpakka.google.bigquery` configuration section.
```
seglo, line 52:
```suggestion
The settings for the BigQuery connector are read by default from the `alpakka.google.bigquery` configuration section.
```

docs/src/main/paradox/google-cloud-bigquery.md
```diff
@@ -15,114 +17,177 @@ Alpakka Google Cloud BigQuery was added in Alpakka 2.0.2 in July 2020 and is mar
...
+The frequency with which new load jobs are created is controlled by the `alpakka.google.bigquery.load-job.per-table-quota` configuration setting.
 
-The configuration will contain the session (which includes your service-token).
+@@@warning
 
-If you create multiple requests to the same source (likely to happen) you should create a single `BigQueryConfig` instance and reuse it.
+Pending the resolution of [Google BigQuery issue 176002651](https://issuetracker.google.com/176002651), the `BigQuery.insertAllAsync` API may not work as expected.
 
-If you call multiple bigquery sources (not likely to happen) it is worth to cache the configs, so you can save a lot of unneeded authorization requests.
+As a workaround, you can use the config setting `akka.http.parsing.conflicting-content-type-header-processing-mode = first` with Akka HTTP v2.6.4 or later (not yet released at time of writing but available as a [snapshot build](https://doc.akka.io/docs/akka-http/snapshot/contributing.html#snapshots)).
```
seglo, line 144: Update/delete with Akka http version bump.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/spray/BigQueryRestJsonProtocol.scala
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+package akka.stream.alpakka.googlecloud.bigquery.scaladsl.spray
+
+import spray.json.{AdditionalFormats, CollectionFormats, ProductFormats, StandardFormats}
+
+/**
+ * Provides all the predefined JsonFormats for the BigQuery REST API.
+ */
+trait BigQueryRestJsonProtocol
```
seglo, line 12: There are lots of datatypes to support. I noticed that the date/datetime/timestamp related fields don't have a formats representation. Should we add those?

Are all the provided formats under test?

seglo (commented):
Another review round.

I noticed the link validator build action failed because this is a new connector.  I'll take a look at that.

armanbilge:
Thanks for taking another pass!

Regarding the link validator, actually I have a guess why it may be broken. I don't think it's because this is a new connector, because strictly speaking it's not (just a complete rewrite of the old one) and also, it was working fine many commits ago. The error message in the link validator's SBT build is that it was `Unable to determine previous version`. This is based on git tags, but since it is a shallow clone of the repository, I wonder if the number of commits pushed the last version tag beyond the shallow clone into the "deep waters" causing this to break. Anyway just a guess!

> Creating a separate sub project in this repo will add a nominal amount of complexity to the release process, but I think it's the best way forward all around.

Sounds good to me!

seglo:
> Regarding the link validator, actually I have a guess why it may be broken. I don't think it's because this is a new connector, because strictly speaking it's not (just a complete rewrite of the old one) and also, it was working fine many commits ago. The error message in the link validator's SBT build is that it was `Unable to determine previous version`. This is based on git tags, but since it is a shallow clone of the repository, I wonder if the number of commits pushed the last version tag beyond the shallow clone into the "deep waters" causing this to break. Anyway just a guess!

Good point. I ran into a similar issue on another repo.

Your distance from the last tag is over 100

```
git rev-list v2.0.2..HEAD --count
117
```

Maybe we should just remove the depth limit.

project/Dependencies.scala
```diff
@@ -209,14 +209,13 @@ object Dependencies {
   val GoogleBigQuery = Seq(
     libraryDependencies ++= Seq(
         "com.typesafe.akka" %% "akka-http" % AkkaHttpVersion,
```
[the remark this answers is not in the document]
armanbilge: I'm happy surprised about this! I was told it might be a while until release.

google-cloud-bigquery/src/test/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/schema/BigQuerySchemasSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+  case class A(integer: Int, long: Long, float: Float, double: Double, string: String, boolean: Boolean, record: B)
+  case class B(nullable: Option[String], repeated: Seq[C])
+  case class C(numeric: BigDecimal)
+
+  val schema = TableSchema(
+    List(
+      TableFieldSchema("integer", IntegerType, Some(RequiredMode), None),
+      TableFieldSchema("long", IntegerType, Some(RequiredMode), None),
+      TableFieldSchema("float", FloatType, Some(RequiredMode), None),
+      TableFieldSchema("double", FloatType, Some(RequiredMode), None),
```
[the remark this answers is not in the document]
armanbilge: Hmm, the aliases are not mentioned in that part of the docs. In the [TableFieldSchema schema](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#TableFieldSchema) they indicate

> INTEGER, INT64 (same as INTEGER), FLOAT, FLOAT64 (same as FLOAT)

which led me to think `INTEGER` and `FLOAT` are the primary names and `INT64` and `FLOAT64` are aliases, hence why I named these `IntegerType` and `FloatType`. But maybe in JVM world we should prefer `INT64` and `FLOAT64` to emphasize the precision.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/spray/BigQueryRestJsonProtocol.scala
armanbilge: There are! I noticed a new preview datatype `BIGNUMERIC` as well.

I can definitely add the date/time stuff, I think I avoided this because people tend to prefer 3rd-party libs for this? E.g. Joda time.

But yes, all the formats should be covered by the [end-to-end test](https://github.com/armanbilge/alpakka/blob/wip-new-bigquery/google-cloud-bigquery/src/test/scala/akka/stream/alpakka/googlecloud/bigquery/e2e/A.scala).

docs/src/main/paradox/google-cloud-bigquery.md
armanbilge: Here's a question: should we leave it up to users decide to set `akka.http.parsing.conflicting-content-type-header-processing-mode = first` or should we set this ourselves in the `reference.conf` so that this connector works out-of-the-box?

armanbilge: Oops, nevermind! The [Akka docs](https://doc.akka.io/docs/akka/current/general/configuration.html#where-configuration-is-read-from) say
> It’s not supported to override a config property owned by one library in a reference.conf of another library

project/Dependencies.scala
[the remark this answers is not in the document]
armanbilge: Bumped the Akka HTTP version, just for the BigQuery connector. Hope this is okay?

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/spray/BigQueryRestJsonProtocol.scala
armanbilge: Decided to add these! They are covered by the E2E test.

armanbilge:
> Maybe we should just remove the depth limit.

I just did this, and now the link validator passes :) I also bumped the Akka HTTP version for the BigQuery module. Hope these are okay, I don't like messing around with the build/CI too much.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/BigQueryQueries.scala
```diff
@@ -0,0 +1,164 @@
...
+            initialQueryResponse <- Unmarshal(response.entity).to[QueryResponse[Out]]
+          } yield {
+
+            val jobReference = initialQueryResponse.jobReference
+            val jobId = jobReference.jobId.getOrElse {
+              throw BigQueryException("Query response did not contain job id.")
```
PepRoll, line 76: A response may not have a job id if the request sends with "dryRun" param. Example
```
{
  "kind": "bigquery#queryResponse",
  "schema": {
    "fields": [
      {
        "name": "other",
        "type": "STRING",
        "mode": "NULLABLE"
      }
    ]
  },
  "jobReference": {
    "projectId": "analytics",
    "location": "europe-west2"
  },
  "totalBytesProcessed": "495297590",
  "jobComplete": true,
  "cacheHit": false
}
```

armanbilge: Thanks for catching this! Just pushed a fix and added test coverage.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/QueryJsonProtocol.scala
```diff
@@ -0,0 +1,235 @@
...
+   * @param numDmlAffectedRows the number of rows affected by a DML statement
+   * @tparam T the data model for each row
+   */
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  final case class QueryResponse[T](schema: Option[TableSchema],
+                                    jobReference: JobReference,
+                                    totalRows: Option[Long],
+                                    pageToken: Option[String],
+                                    rows: Option[Seq[T]],
+                                    totalBytesProcessed: Option[Long],
```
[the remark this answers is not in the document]
armanbilge: That's strange. It's why I wrote [BigQueryLongJsonFormat](https://github.com/armanbilge/alpakka/blob/wip-new-bigquery/google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/spray/BigQueryRestBasicFormats.scala) specifically to handle this custom deser, gotta figure out why it's not picking up.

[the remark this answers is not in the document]
armanbilge: @PepRoll do you mind providing an example with this problem? As far as I can tell the dry run test in the e2e is able to handle long encoded as string.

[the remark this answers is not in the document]
armanbilge: Aha, yes that format is intended just for testing the paging logic, not for deserialization. Thanks for following up!

You should use the [`responseReader`](https://github.com/armanbilge/alpakka/blob/wip-new-bigquery/google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/QueryJsonProtocol.scala#L228) in `QueryJsonProtocol`.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/spray/BigQueryRestJsonProtocol.scala
armanbilge: I just added support+tests for the `BYTES` type via Akka's `ByteString`. So that covers about everything, besides the `GEOGRAPHY` type.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/BigQueryJobs.scala
```diff
@@ -0,0 +1,171 @@
...
+      .retryRequestWithOAuth(HttpRequest(POST, uri.withQuery(query)))
+      .flatMap { response =>
+        Unmarshal(response.entity).to[JobCancelResponse]
+      }(system.classicSystem.dispatcher)
+  }
+
+  /**
+   * Loads data into BigQuery via a series of asynchronous load jobs, configurable by [[akka.stream.alpakka.googlecloud.bigquery.LoadJobSettings]].
+   * @note WARNING: Pending the resolution of [[ https://issuetracker.google.com/176002651 BigQuery issue 176002651]] this method may not work as expected.
+   *       As a workaround, you can use the config setting `akka.http.parsing.conflicting-content-type-header-processing-mode = first` with Akka HTTP v2.6.4 or later.
```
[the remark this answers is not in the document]
armanbilge: Oops! Fixed, it was in 5 places.

armanbilge:
> I would like to merge this in the next few days so I can prepare to release it with the 3.0.0-M1 milestone.

That's great, thank you!!

> IMO it would be fine to label the public APIs as `@ApiMayChange`

I agree, I marked the core public APIs as `@ApiMayChange` referencing #2353 (is there an easy way to annotate the whole package?).

armanbilge:
@seglo After adding fatal warnings we seem to be in a CI deadlock.

`a3add92` [fails](https://github.com/akka/alpakka/runs/2081169688#step:6:1021) because `HttpsConnectionContext` is deprecated.

However, the fix in `f52677d` [fails](https://github.com/akka/alpakka/runs/2081726990#step:6:433) because the new method isn't available.

seglo:
> @seglo After adding fatal warnings we seem to be in a CI deadlock.

Add the silencer dependency and use the `@silent` attribute on the warning line.

seglo:
> I agree, I marked the core public APIs as `@ApiMayChange` referencing #2353 (is there an easy way to annotate the whole package?).

I'm not aware of a way. Just put it on the top level of the main APIs with a link to this PR. That should make it clear.

seglo:
@armanbilge I fixed the build in master. If you merge we'll run another build and I'll merge the PR when the big-query tests are successful.

armanbilge:
@seglo Should be good to go.

PR #2551, opened 2021-01-05
GCS: Allow adding custom metadata for resumable uploads

aperkins1310:
Pretty self explanatory, metadata is currently part of StorageObjects but wasn't able to be added on upload.

I wasn't sure which combination of overloaded functions/default args to add, so for now I've just added an additional overloaded function for where chunkSize is already set.

I didn't add it for the `simpleUpload` case as it [seems to be quite messy to upload a single file with metadata](https://cloud.google.com/storage/docs/uploading-objects), so I felt it best to leave until someone had a use case for that

### DIFF
--- a/google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/javadsl/GCStorage.scala
+++ b/google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/javadsl/GCStorage.scala
@@ -250,21 +250,8 @@ object GCStorage {
                       objectName: String,
                       contentType: ContentType,
                       chunkSize: java.lang.Integer,
-                      metadata: java.util.Map[String, String]): Sink[ByteString, CompletionStage[StorageObject]] = {
-    assert(
-      (chunkSize >= (256 * 1024)) && (chunkSize % (256 * 1024) == 0),
-      "Chunk size must be a multiple of 256KB"
-    )
-
-    GCStorageStream
-      .resumableUpload(bucket,
-                       objectName,
-                       contentType.asInstanceOf[ScalaContentType],
-                       chunkSize,
-                       Some(metadata.asScala.toMap))
-      .asJava
-      .mapMaterializedValue(func(_.toJava))
-  }
+                      metadata: java.util.Map[String, String]): Sink[ByteString, CompletionStage[StorageObject]] =
+    resumableUpload(bucket, objectName, contentType, chunkSize, Some(metadata))
 
   /**
    * Uploads object by making multiple requests
@@ -280,14 +267,27 @@ object GCStorage {
   def resumableUpload(bucket: String,
                       objectName: String,
                       contentType: ContentType,
-                      chunkSize: java.lang.Integer): Sink[ByteString, CompletionStage[StorageObject]] = {
+                      chunkSize: java.lang.Integer): Sink[ByteString, CompletionStage[StorageObject]] =
+    resumableUpload(bucket, objectName, contentType, chunkSize, metadata = None)
+
+  private def resumableUpload(
+      bucket: String,
+      objectName: String,
+      contentType: ContentType,
+      chunkSize: java.lang.Integer,
+      metadata: Option[java.util.Map[String, String]]
+  ): Sink[ByteString, CompletionStage[StorageObject]] = {
     assert(
       (chunkSize >= (256 * 1024)) && (chunkSize % (256 * 1024) == 0),
       "Chunk size must be a multiple of 256KB"
     )
 
     GCStorageStream
-      .resumableUpload(bucket, objectName, contentType.asInstanceOf[ScalaContentType], chunkSize)
+      .resumableUpload(bucket,
+                       objectName,
+                       contentType.asInstanceOf[ScalaContentType],
+                       chunkSize,
+                       metadata.map(_.asScala.toMap))
       .asJava
       .mapMaterializedValue(func(_.toJava))
   }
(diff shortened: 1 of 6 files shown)


google-cloud-storage/src/test/scala/akka/stream/alpakka/googlecloud/storage/scaladsl/GCStorageWiremockBase.scala
```diff
@@ -711,16 +712,21 @@ abstract class GCStorageWiremockBase(_system: ActorSystem, _wireMockServer: Wire
...
     )
 
-  def mockLargeFileUpload(firstChunkContent: String, secondChunkContent: String, chunkSize: Int): Unit = {
+  def mockLargeFileUpload(firstChunkContent: String, secondChunkContent: String, chunkSize: Int): Unit =
+    mockLargeFileUpload(firstChunkContent, secondChunkContent, chunkSize, None)
+
+  def mockLargeFileUpload(firstChunkContent: String,
+                          secondChunkContent: String,
+                          chunkSize: Int,
+                          metadata: Option[(String, String)]): Unit = {
```
aperkins1310, line 721: I've only allowed one key-value pair as metadata here, purely to make creating the json for this simple, without duplicating the code itself (IE `metadata.toJson.compactPrint`)
I'm happy to take suggestions here, as I can see how it would be confusing to to only be able to add a key-value for metadata, when in reality it's a map. I could duplicate the main code for this one section, as it is really only 2 spray json calls, or could create the json "manually" for a map instead of just the tuple

google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/javadsl/GCStorage.scala
```diff
@@ -233,6 +234,38 @@ object GCStorage {
...
+   * @param contentType `ContentType`
+   * @param chunkSize the size of the request sent to google cloud storage in bytes, must be a multiple of 256KB
+   * @param metadata custom metadata for the object
+   * @return a `Sink` that accepts `ByteString`'s and materializes to a `Future` of `StorageObject`
+   */
+  def resumableUpload(bucket: String,
+                      objectName: String,
+                      contentType: ContentType,
+                      chunkSize: java.lang.Integer,
+                      metadata: java.util.Map[String, String]): Sink[ByteString, CompletionStage[StorageObject]] = {
```
seglo, line 253: I think the extra overloads are fine. I was considering whether we should mark the other as deprecated, but considering how this has only been addressed now there are probably many users not concerned with metadata.

google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/javadsl/GCStorage.scala
```diff
@@ -233,6 +234,38 @@ object GCStorage {
...
+    assert(
+      (chunkSize >= (256 * 1024)) && (chunkSize % (256 * 1024) == 0),
+      "Chunk size must be a multiple of 256KB"
+    )
```
seglo, line 257: Make the old overloads call this one (with a `Map.empty`) so we can DRY up the pre-conditions.

google-cloud-storage/src/test/scala/akka/stream/alpakka/googlecloud/storage/scaladsl/GCStorageWiremockBase.scala
seglo: I think the mock should support the Map as well.

> I could duplicate the main code for this one section, as it is really only 2 spray json calls, or could create the json "manually" for a map instead of just the tuple?

Either solution sounds fine to me.

seglo (commented):
Thanks for the PR. I left some comments.

> I didn't add it for the `simpleUpload` case as it [seems to be quite messy to upload a single file with metadata](https://cloud.google.com/storage/docs/uploading-objects), so I felt it best to leave until someone had a use case for that

I'm not sure what you mean by messy, but I agree that it's not important to implement it in this PR.

google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/javadsl/GCStorage.scala
aperkins1310: Okay so I just had a look at this, and there's a tiny issue with doing this:
Currently the overload takes a util.Map, and converts that to Some[Map[String, String]]. It's currently the option on which we decide to create a request with metadata or not. Passing it through that Some, means we create a "metadata request", but with no metadata. That leaves us with a few different options:
- Do we make the file upload mock expect this, and send requests with an empty metadata map?
- Do we check both the Option being defined, *and* the metadata being nonEmpty (at which point the Option is a bit redundant, and we could just use an empty map for no metadata)
- Do we simply extract the chunk size preconditions so they aren't repeated

Not sure I have a huge preference, maybe using the empty map case as I did debate the merit of using an option in the first place?

google-cloud-storage/src/test/scala/akka/stream/alpakka/googlecloud/storage/scaladsl/GCStorageWiremockBase.scala
aperkins1310: cool, I'll sort this now

google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/javadsl/GCStorage.scala
seglo: Ah, makes sense. Using the `Option` seems fine.

> Do we simply extract the chunk size preconditions so they aren't repeated, either via another (private?) overload that takes an optional metadata map, or just a private method for the chunk size?

I guess it's not a big deal to DRY things up, but this suggestion is a reasonable solution.

aperkins1310:
The comments should all be addressed now, please have another look 🙂

seglo (approved):
LGTM. Thanks for the contribution!

Issue #1128, opened 2018-08-02
Not all Slick sources stream by default


arnold-jr:
This is an issue of documentation clarity: as per [Slick's documentation](http://slick.lightbend.com/doc/3.2.0/dbio.html#streaming), not all databases (most notably PostgreSQL) require `.withStatementParameters` in order to get results returned asynchronously. So a truly streaming query would look like

```
val typedSelectAllUsers = TableQuery[Users].result
          .withStatementParameters(
            rsType = ResultSetType.ForwardOnly,
            rsConcurrency = ResultSetConcurrency.ReadOnly,
            fetchSize = conf.getInt("batchSize")
          )
```

This might be a useful addition to [the docs](https://github.com/akka/alpakka/blob/57c9718104db1fab7d424a3697d2ceb2cd279a8d/slick/src/test/scala/akka/stream/alpakka/slick/scaladsl/SlickSpec.scala#L60), since DBs that will fit entirely into memory are more the exception than the rule :)

Thanks!

P.S. 
Incidentally, for PostgreSQL, the [fetchSize is not respected](https://github.com/slick/slick/issues/1305)!

2m:
Ahh, good insight! Would you be up for creating a PR for documentation amendment?

WellingR:
Before I have done some tests with this. And I indeed found that the above is _almost_ the full solution. The last thing that is needed (which is also mentioned in https://github.com/slick/slick/issues/1305 ) is to wrap the query in a .transactionally block

So the truly streaming query would look like

```scala
val typedSelectAllUsers = TableQuery[Users].result
          .withStatementParameters(
            rsType = ResultSetType.ForwardOnly,
            rsConcurrency = ResultSetConcurrency.ReadOnly,
            fetchSize = conf.getInt("batchSize")
          )
          .transactionally
```

Sh1ftry:
Shouldn't this be explicitly pointed in the [Slick Alpakka documentation](https://doc.akka.io/docs/alpakka/current/slick.html#typed-queries)? By looking at example that is present in the documentation I can assume that using `Slick.source(TableQuery[Users].result)` is enough to enable streaming, which is not true, because whole result would be first loaded into memory for some databases.

PR #2558, opened 2021-01-17
Slick: Add note about enabling streaming with PostgreSQL (#1128)

markarasev:
Some additional parameters need to be set to stream some databases such as PostgreSQL

I got the surprise of my application loading my whole database in-memory while I was thinking I was streaming it, so I think it's worth highlighting this.

References https://github.com/akka/alpakka/issues/1128


raboof (approved):
wow, that seems surprising indeed!

ennru (approved):
LGTM.

PR #2572, opened 2021-02-05
FTP: fix deprecations for `IOResult.createFailed`

dwickern:
Stop using the failure side of `IOResult` which was [deprecated in akka 2.6.x](https://doc.akka.io/docs/akka/current/project/migration-guide-2.5.x-2.6.x.html#iosources-fileio).
Previously, failures materialized as a _successful_ `Future` with the exception buried in the `IOResult`.

I'm using `IOOperationIncompleteException` as the exception for `Ftp.fromPath`/`Ftp.toPath` to keep the APIs consistent with `FileIO`.

References #2474


seglo:
Thanks. There's more to do to complete #2474, but we don't have to do it all in this PR.

PR #2579, opened 2021-03-01
FTP: Fix SFTP keyfile authentication when no password is provided

Bobo1239:
Previously it would always assume a single specific keyfile format (`OpenSSHKeyFile` a.k.a. the "old" OpenSSH keyfile format) which obviously doesn't always work. But it only did the wrong thing when no password was provided for probably historic reasons. (Downstream I'm currently working around this issue by providing a wrong password.) Now we just use [`authPublickey()`](https://github.com/akka/alpakka/blob/10c268dd3add3f4a0956877df04a7cabd7c2a98c/ftp/src/main/scala/akka/stream/alpakka/ftp/impl/SftpOperations.scala#L213) from the other branch which does the same but correctly and in addition we can remove some duplicate code.


seglo (approved):
LGTM

seglo:
Thanks!

PR #2589, opened 2021-03-06
Update OrientDB to Akka 2.6

mrooding:
- Update OrientDB according to the checklist in #2491 
- Update OrientDB dependencies to their latest version

I've kept `fatalWarnings` set to `false` due to the deprecation warnings from the OrientDB SDK. I looked at it a little bit but it seems like quite a large refactor to update Alpakka to not use deprecated APIs

References #2491


ennru (approved):
LGTM.

Issue #2577, opened 2021-02-21
Generating time-based path for each new file


bondarenko:
### Short description

It's about a possibility to create a file path for each new file depending on timestamp supplied in hdfs messages.

### Details

Currently the only possibility is to have file paths created from current time in milliseconds and a row count. Moreover, each file will be written into the same folder defined at flow-creation time.
So, it would be great to have a possibility to create a file path (including its folder name) for each new file depending on timestamp supplied in hdfs messages. It's the real business case I've stumble upon my current project.

PS. I already have an implementation for this and could proceed with creating PR, just wanted to know what other people thoughts could be on the matter.

seglo:
Sounds reasonable. We would welcome a PR. It might make sense to make the destination path more generic. For example, providing a user defined function to generate a path, given the HDFS message/metadata.

PR #2607, opened 2021-03-10
HDFS: New path per file

bondarenko:
References #2577

Adding by the PR a FilePathGenerator option of creating a new path for each file depending on rotation count and timestamp supplied by hdfs messages.
An assumption is that each hdfs message in addition to source and passThrough contains a timestamp. This timestamp in succession is used to create a file path. The specific file path format as before is defined in a function supplied to FilePathGenerator.
An example of specific path template we can use for path creation with this option on:
`s"/$year/$month/$day/$hour-$rotationCount"`.


hdfs/src/main/scala/akka/stream/alpakka/hdfs/impl/writer/HdfsWriter.scala
```diff
@@ -53,8 +56,8 @@ private[hdfs] trait HdfsWriter[W, I] {
 @InternalApi
 private[writer] object HdfsWriter {
 
-  def createTargetPath(generator: FilePathGenerator, c: Long): Path =
-    generator(c, System.currentTimeMillis / 1000)
+  def createTargetPath(generator: FilePathGenerator, c: Long, timestamp: Long = -1): Path =
+    generator(c, if (generator.newPathForEachFile) timestamp else currentTimeMillis / 1000)
```
seglo, line 60: I suppose we're making the assumption that the `timestamp` is at the scale we want?  The default behaviour scaled timestamp to seconds, for `newPathForEachFile` we use whatever is provided.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/model.scala
```diff
@@ -144,6 +153,21 @@ object FilePathGenerator {
...
+    }
+
+  /**
+   * Scala API: creates [[FilePathGenerator]] to rotate output
+   *
+   * @param f    a function that takes rotation count and timestamp to return path of output
+   * @param pathForEachFile turns on/off a possibility to create a new path for each file
+   *                        which depends on rotation count and timestamp supplied by hdfs messages
+   */
+  def apply(f: (Long, Long) => String, pathForEachFile: Boolean): FilePathGenerator =
```
seglo, line 167: I think we should also support the `temp` directory argument in the other constructor.  Maybe we could create a type that uses the builder pattern and encapsulates the various options.

i.e.

```
class FilePathGeneratorSettings private(val tempDirectory: String, newPathForEachFile: Boolean) {
  def withTempDirectory(tempDir: String): FilePathGeneratorSettings = ...
  def withNewPathForEachFile(newPathForEachFile: Boolean): FilePathGeneratorSettings = ...
  private def copy ...
}
```

hdfs/src/main/scala/akka/stream/alpakka/hdfs/model.scala
```diff
@@ -144,6 +153,21 @@ object FilePathGenerator {
...
+    new FilePathGenerator {
+      val tempDirectory: String = DefaultTempDirectory
+      val newPathForEachFile: Boolean = pathForEachFile
       def apply(rotationCount: Long, timestamp: Long): Path = new Path(f(rotationCount, timestamp))
     }
```
seglo, line 173: Java APIs seem to be missing.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/model.scala
```diff
@@ -144,6 +153,21 @@ object FilePathGenerator {
...
+   * Scala API: creates [[FilePathGenerator]] to rotate output
+   *
+   * @param f    a function that takes rotation count and timestamp to return path of output
+   * @param pathForEachFile turns on/off a possibility to create a new path for each file
+   *                        which depends on rotation count and timestamp supplied by hdfs messages
+   */
+  def apply(f: (Long, Long) => String, pathForEachFile: Boolean): FilePathGenerator =
+    new FilePathGenerator {
+      val tempDirectory: String = DefaultTempDirectory
+      val newPathForEachFile: Boolean = pathForEachFile
```
seglo, line 170: I'm confused about the purpose of `newPathForEachFile`.  If a timestamp extractor perhaps that should imply that the target path is evaluated per record.

hdfs/src/main/scala/akka/stream/alpakka/hdfs/model.scala
seglo: Maybe we could add the optional `timestampExtractor` function to this new type instead of it being separate.

hdfs/src/test/scala/docs/scaladsl/HdfsWriterSpec.scala
```diff
@@ -293,6 +295,94 @@ class HdfsWriterSpec
...
     }
+
+    "kafka-example - store data with passThrough and a new path for each file" in {
+      //#define-kafka-classes
+      case class Book(title: String)
+      case class KafkaOffset(offset: Int)
+      case class KafkaMessage(book: Book, offset: KafkaOffset, timestamp: Long)
+      //#define-kafka-classes
+
+      val TWENTY_FOUR_HOURS = 24 * 60 * 60
```
seglo, line 306:
```suggestion
      val TwentyFourHoursInSeconds = 24 * 60 * 60
```

seglo (commented):
Looking good. Left some suggestions and comments.

patriknw:
@bondarenko Is this something you would be able to complete?

ennru:
Please open a new PR, if you want to get back to this.

Issue #2547, opened 2020-12-18
Retry gcs upload on OutgoingConnectionBlueprint$UnexpectedConnectionClosureException


francisdb:
### Versions used 

Alpakka version: `2.0.2`
Akka: `2.6.8`

### Expected Behavior

The upload is retried when the connection is lost

### Actual Behavior

`OutgoingConnectionBlueprint$UnexpectedConnectionClosureException` is thrown

### Relevant logs

A similar fix on the AWS code was applied in #2535

```scala
    GCStorage
      .resumableUpload(
        googleStorageConfiguration.bucket,
        fileName,
        ContentType.parse(contentType).getOrElse(ContentTypes.`application/octet-stream`)
      )
```

```
akka.http.impl.engine.client.OutgoingConnectionBlueprint$UnexpectedConnectionClosureException: The http server closed the connection unexpectedly before delivering responses for 1 outstanding requests
    at akka.http.impl.engine.client.OutgoingConnectionBlueprint$.$anonfun$apply$6(OutgoingConnectionBlueprint.scala:123)
    at akka.http.impl.engine.client.OutgoingConnectionBlueprint$.$anonfun$apply$6$adapted(OutgoingConnectionBlueprint.scala:123)
    at akka.http.impl.util.One2OneBidiFlow$One2OneBidi$$anon$1$$anon$4.onUpstreamFinish(One2OneBidiFlow.scala:97)
    at akka.stream.impl.fusing.GraphInterpreter.processEvent(GraphInterpreter.scala:523)
    at akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:390)
    at akka.stream.impl.fusing.GraphInterpreterShell.runBatch(ActorGraphInterpreter.scala:625)
    at akka.stream.impl.fusing.ActorGraphInterpreter$SimpleBoundaryEvent.execute(ActorGraphInterpreter.scala:56)
    at akka.stream.impl.fusing.ActorGraphInterpreter$SimpleBoundaryEvent.execute$(ActorGraphInterpreter.scala:52)
    at akka.stream.impl.fusing.ActorGraphInterpreter$BatchingActorInputBoundary$OnComplete.execute(ActorGraphInterpreter.scala:86)
    at akka.stream.impl.fusing.GraphInterpreterShell.processEvent(ActorGraphInterpreter.scala:600)
[...]

Issue #1705, opened 2019-05-17
Alpakka Google Cloud Pub/Sub: Allow HTTP(S) proxy configuration


cokeSchlumpf:
### Short description

The Alpakka Google Cloud Pub/ Sub Connector does not support configuration of a Proxy. In our enterprise environment we have to go through a proxy to reach Google API endpoints. Thus I would suggest to implement the support for Proxy configuration.

### Details

It should be possible to configure the proxy, I see two options to implement it:

a) Check for `http_proxy`, `https_proxy` as well as `no_proxy` environment variables to configure proxy for request as described in [Akka HTTP documentation](https://doc.akka.io/docs/akka-http/current/client-side/client-transport.html). Theoratically this could also be a feature of Akka HTTP.

b) Make Proxy configurable via Alpakka's Google Pub/Sub API. Similar to authentication properties.

I could also imagine that a combination of both might be implemented where as b) wins over a) when configured at both places.

I could support implementing it, but would first like to discuss which option might be the better one. Or maybe there is even another option....

ennru:
Please note that a similar feature was recently added to Alpakka S3 in [#1639](https://github.com/akka/alpakka/pull/1639).

Please feel free to suggest a similar solution for Google Pub/Sub.

gkatzioura:
Hi all! In my recent pull request for [BigQuery](https://github.com/akka/alpakka/pull/2106) I added the proxy feature from Alpakka S3.  Is it ok to proceed on this issue and add the proxy feature on Pub/Sub?

seglo:
@gkatzioura That would be wonderful!  Please go ahead.

armanbilge:
This will be fixed in #2613.

seglo:
Fixed in #2613

PR #2613, opened 2021-03-12
Create google-common module

armanbilge:
References #773 #2548; cc @tg44 

This is an attempt at the beginnings of a long-awaited `google-common` module.

The implementation is mostly taken from the new BigQuery connector, which includes:

* OAuth2 `Credentials` providers w/ Service Account and Compute Engine configurations
  * A lightweight wrapper as a Google `Credentials` class, for interop with gRPC
* A `GoogleHttp` class with auth, retry, and proxy support for requests
* A `PaginatedRequest` source for paginated resources
* A `ResumableUpload` flow (will be shared between GCS and BigQuery, at least)
* `GoogleSettings` and stream attributes classes for managing config

The idea is that this config is generally done centrally (although of course can be overridden) so that multiple connectors can share access tokens from the same `Credentials` instance.

If it's not too ambitious, I'd like to try and migrate all the Google connectors to this module in this PR. I think this is the best path towards determining a relatively stable (internal) API for this module that meets the needs of all of the connectors.

Unfortunately, it seems that the various Google connectors don't share a top level package: they are split between `a.s.a.google` and `a.s.a.googlecloud`, probably too late to change?

Looking forward to your feedback!


tg44:
Nice! I think we don't need to keep it internal only, simply just document that everything can change until we freeze the API. So I don't think the namespaces are problematic.

> I'd like to try and migrate all the Google connectors to this module in this PR.

This will be the interesting part. If you want to I can try to migrate FCM.

armanbilge:
Thanks @tg44! I'm open to making it public API with those caveats.

> This will be the interesting part. If you want to I can try to migrate FCM.

Thanks, that would be a big help!! Before we start migrating code, my big question is if (and how much) we can break the public APIs of the existing connectors? Of course this module is still useful even if only used to replace the internal implementations of the connectors, without changing their public APIs.

But, to achieve truly central configuration (and shared credentials/access tokens) will probably require breaking changes to config and settings classes for each of the connectors. If we are going to do it, Alpakka 3 would probably be our best chance. Hopefully @seglo or @ennru can comment on this?

tg44:
I created a commit; https://github.com/tg44/alpakka/commit/601a46413c28d7b2e583238cc031e0ff59c2f6df
The java API and the docs are missing (also I haven't tested against Google). But the "flow" of the migration seemed ok to me. Also, I don't think I broke any existing API.

(I can't PR it to your fork, I think you need to enable something.)

I removed the `extends AnyVal` and `final` from `GoogleHttp` so we can mock it from now on. I eased up some visibility rules too to write backward compatible conversions.

armanbilge:
Wow, this is awesome @tg44, thanks for this! Love that your commit nearly halved the size of the code. Not sure why you couldn't PR so I started it for you at https://github.com/armanbilge/alpakka/pull/1.

> But the "flow" of the migration seemed ok to me.

I'm glad you had a positive experience, any feedback? Actually, I'm still noodling a bit with the `GoogleHttp` class so that it directly takes unmarshallers and has customizable retrying.

> Also, I don't think I broke any existing API.
I added some comments, but you're right, it looks like you pulled this off without breaking much/any public API! I hope it can be this easy for the other connectors.

seglo:
> If it's not too ambitious, I'd like to try and migrate all the Google connectors to this module in this PR. I think this is the best path towards determining a relatively stable (internal) API for this module that meets the needs of all of the connectors.

To make sure IUC, you're asking if you can update the existing connectors to use this updated module in this PR?  That sounds reasonable to me, and will help to make sure we don't miss any functionality required by one or more of google connectors.

> Unfortunately, it seems that the various Google connectors don't share a top level package: they are split between `a.s.a.google` and `a.s.a.googlecloud`, probably too late to change?

If we did this I would choose a package that most of the connectors use and deprecate all the APIs found in the other packages. I don't think we should make a clean break. I'm not sure this is really worth the effort though.

> Thanks, that would be a big help!! Before we start migrating code, my big question is if (and how much) we can break the public APIs of the existing connectors? Of course this module is still useful even if only used to replace the internal implementations of the connectors, without changing their public APIs.
> 
> But, to achieve truly central configuration (and shared credentials/access tokens) will probably require breaking changes to config and settings classes for each of the connectors. If we are going to do it, Alpakka 3 would probably be our best chance. Hopefully @seglo or @ennru can comment on this?

We can deprecate existing configuration, but it would be unfortunate to have to make breaking changes across the board for Alpakka 3.0.0.  Alpakka major releases don't _have_ to be as rare as they are right now.  If there's a good reason (and maybe this is one of them) then we could release an Alpakka 4.0.0 in a more reasonable interval, maybe after 3 months?

armanbilge:
> make sure we don't miss any functionality required by one or more of google connectors.

Exactly my thinking!

> We can deprecate existing configuration, but it would be unfortunate to have to make breaking changes across the board for Alpakka 3.0.0.

Thanks for your thoughts! This makes sense; I agree. Ok, let's focus on not breaking stuff for now, and once the dust settles see where that leaves us regarding need for 4.0.

armanbilge:
Just pushed my first migration effort for the Google Cloud Storage module. All tests pass, but I did have to break a couple things.

1. In the original config you could specify non-standard endpoints for the Google APIs. I assume that this existed purely for testing purposes (but please correct me if I'm mistaken) and dropped support for non-default values. Users who needed this functionality can use the new `forward-proxy` config.
2. There was a single test for a failing request that broke: previously, it failed silently (returned empty Source) but now it fails with an exception. This behavior seemed inconsistent with the rest of the connector's API (only 404s fail silently like this) so I didn't fix it.

@seglo Are these breaking changes acceptable?

google-cloud-storage/src/main/resources/reference.conf
```diff
@@ -1,6 +1,22 @@
...
+  }
+
+  cloud.storage {
+
+    # Deprecated, use config path alpakka.google.credentials
+    project-id = deprecated
+    client-email = deprecated
+    private-key = deprecated
+
+    # No longer supported, use config path alpakka.google.forward-proxy
```
armanbilge, line 14: These are the configurable endpoints I dropped support for.

google-cloud-storage/src/test/scala/docs/scaladsl/GCStorageSourceSpec.scala
```diff
@@ -221,23 +257,30 @@ class GCStorageSourceSpec
...
+        mockBucketListingFailure()
+      )
 
       val listSource = GCStorage.listBucket(bucketName, None)
 
       listSource.runWith(Sink.seq).failed.futureValue.getMessage shouldBe "[400] Bucket listing failed"
     }
 
-    "return empty source listing bucket with wrong settings" in {
+    // TODO Is this really the desired behavior?
```
armanbilge, line 270: Couldn't get this test to pass (it expects a silent failure) and not sure if it should anyway. Other tests with `mockBucketListingFailure` like the one immediately above it fail loudly.

armanbilge:
@tg44 I just pushed the FCM migration based on your code, thanks for the help! Would be great if you could have a look at it.

tg44:
I think it will be fine!

armanbilge:
Migrated both PubSub implementations. Breaking change:
* The PubSub gRPC previously depended on `google-auth-library-java` which [loads credentials automatically](https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java) from the following places:

> 1. Credentials file pointed to by the `GOOGLE_APPLICATION_CREDENTIALS` environment variable
> 2. Credentials provided by the Google Cloud SDK `gcloud auth application-default login` command
> 3. Google App Engine built-in credentials
> 4. Google Cloud Shell built-in credentials
> 5. Google Compute Engine built-in credentials

The new Alpakka implementation in this PR also [automatically loads credentials](https://github.com/armanbilge/alpakka/blob/wip-google-common/google-common/src/main/resources/reference.conf#L11), but only supports strategies 1, 2, and 5. So for anyone currently running on App Engine or Cloud Shell this could be technically a breaking change.

@seglo is this okay? Should we add support for any more of these strategies?

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/scaladsl/GooglePubSub.scala
```diff
@@ -67,6 +68,7 @@ protected[pubsub] trait GooglePubSub {
   def subscribeFlow(subscription: String, config: PubSubConfig): Flow[Done, ReceivedMessage, Future[NotUsed]] = {
     flow(config)(httpApi.pull(subscription, config.pullReturnImmediately, config.pullMaxMessagesPerInternalBatch))
       .mapConcat(_.receivedMessages.getOrElse(Seq.empty[ReceivedMessage]).toIndexedSeq)
+      .buffer(1, OverflowStrategy.backpressure) // Needed to prevent a deadlock in IntegrationSpec
```
armanbilge, line 71: Anyone know a better fix for this deadlock?

armanbilge:
I think this PR is ready for final reviews: I've finished migrating all of the connectors and also updated the docs for each. All existing tests pass without breaking changes to public APIs (apart from the few exceptions I've indicated above). Thanks in advance for your reviews; would be awesome to see this make the 3.0.0-M1 release!

google-cloud-storage/src/main/resources/reference.conf
seglo: I don't know. Maybe some people with some recent experience using GCS can comment. @francisdb, @DanieleSassoli, @gabrielreid ?

See Arman's note about introducing configuration breaking changes here: https://github.com/akka/alpakka/pull/2613#issuecomment-802101303

google-cloud-storage/src/test/scala/docs/scaladsl/GCStorageSourceSpec.scala
seglo: It seems like an odd way to handle a misconfiguration (if that configuration is really necessary in the first place). I think it's fine to drop this.

seglo:
> Migrated both PubSub connectors. Breaking change:
> 
> * The PubSub gRPC connector previously depended on `google-auth-library-java` which [loads credentials automatically](https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java?rgh-link-date=2021-03-19T17%3A13%3A40Z) from the following places:
> 
> > 1. Credentials file pointed to by the `GOOGLE_APPLICATION_CREDENTIALS` environment variable
> > 2. Credentials provided by the Google Cloud SDK `gcloud auth application-default login` command
> > 3. Google App Engine built-in credentials
> > 4. Google Cloud Shell built-in credentials
> > 5. Google Compute Engine built-in credentials
> 
> The new Alpakka implementation in this PR also [automatically loads credentials](https://github.com/armanbilge/alpakka/blob/wip-google-common/google-common/src/main/resources/reference.conf?rgh-link-date=2021-03-19T17%3A13%3A40Z#L11), but only supports strategies 1, 2, and 5. So for anyone currently running on App Engine or Cloud Shell this could be technically a breaking change.
> 
> @seglo is this okay? Should we add support for any more of these strategies?

IIUC the new implementation is based on what you had for BigQuery. Was this mostly your invention or did you find it referenced somewhere else in Alpakka or otherwise? Without knowing a lot about google auth options it seems to me that it would be more convenient to use a library like `google-auth-library-java` if it saves us from having to manage more boilerplate. What are the tradeoffs?

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/scaladsl/GooglePubSub.scala
seglo: I'm not sure what failure you were experiencing in `IntegrationSpec`, but when I added a line to cancel the test probe the test passed successfully (without the buffer). The other test doesn't have an issue because `Sink.head` cancels the stream when the first element arrives.

seglo:
> I think this PR is ready for final reviews: I've finished migrating all of the connectors and also updated the docs for each. All existing tests pass without breaking changes to public APIs (apart from the few exceptions I've indicated above). Thanks in advance for your reviews; would be awesome to see this make the 3.0.0-M1 release!

Thanks a lot for taking this on. I agree that it would be nice to get this into Alpakka's first 3.0.0 milestone. Though at some point we'll need to stop accepting more contributions (it's hard!)

I'll try to expedite a review this week if I can spare some cycles.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/scaladsl/GooglePubSub.scala
armanbilge: The test `"receive a published message and acknowledge it"` was failing the `assertAllStagesStopped`. Which line did you change?

google-cloud-storage/src/main/resources/reference.conf
francisdb: Just had a look, seems this was added by @jkobejs when he worked further on my pr. We are not using these settings so for my part ok to make unsupported.

armanbilge:
> IIUC the new implementation is based on what you had for BigQuery. Was this mostly your invention or did you find it referenced somewhere else in Alpakka or otherwise?

It was a mix; the `GoogleOAuth2` class is adapted from one of the other connectors (they all have essentially some version of it). My biggest contribution was the Actor-based token cache/refresh.

> Without knowing a lot about google auth options it seems to me that it would be more convenient to use a library like google-auth-library-java if it saves us from having to manage more boilerplate. What are the tradeoffs?

@tg44 has some strong opinions on this https://github.com/akka/alpakka/pull/650#issuecomment-364968964
> The only thing we shouldn't do; is using the official google sdk-s as deps. I walked through some of their code and trust me, you really don't want to depend on that :)

It would be great if @tg44 could clarify exactly what he was referring to, but if I had to guess it's [because of all the locking in their implementation](https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java#L127)? Besides that, it adds several dependencies.

tg44:
I'm not up to date with the current situation, I only can refer the codebase about 2-2.5 years ago, and that time it was a "google-java-sdk" lib.

The dep was huge. It requested tokens before every request. Almost everything was blocking, the things what was not blocking used locks. One of the ugliest codebase I ever saw (global states, static classes with magic initializations, using two accounts from the same app was almost impossible). We tried to use their bigquery, it performed really bad (mostly bcs the lack of streaming and token caching). Trying out and failing with it needed almost the same time as reimplementing the simplest case from the documentation (which was also hard based on the quality of the documentation).

Probably I'm too picky with java libs (I also dislike apache-commons), but when I see strange deps, and hard to test/mock functionality in them I simply avoid them. I hope it get better till then.

Also these libs includes a new http client as dep, which we don't need, we already has akka-http as a dep. I will check the auth lib tomorrow to get an up to date view of it.

armanbilge:
> I think we don't need to keep it internal only, simply just document that everything can change until we freeze the API.

At @tg44's suggestion I added `{scaladsl,javadsl}.Google` which exposes public APIs (`singleRequest`, `paginatedRequest`, `resumableUpload`) to easily interact with any Google API. Actually, IMO they seem straightforward enough that I'd be willing to commit to these APIs.

I also exposed `Credentials.asGoogle` so Alpakka's Google credentials can be provided to Google's own client libraries.

tg44:
I checked the current state of the [google-auth-library-java](https://github.com/googleapis/google-auth-library-java) and its deps, and it is much-much better now (it has multi cred support, and token caching now). I still think that we don't need this functionality. We switched an ~2kLoc java lib (which depends on a ~500kLoc google HTTP client and gson), to a 1kLoc scala code. Also, we (and most of the alpakka users) familiar with akka-http, while we probably never used or debugged google-http-client.

So I still think using our homegrown authenticator gives us more control and confidence, also it is less code, less deps, and known asynchronousity.

armanbilge:
> So I still think using our homegrown authenticator gives us more control and confidence, also it is less code, less deps, and known asynchronousity.

Thanks for looking into this @tg44! I'm inclined to agree.

To add on to this: even if we did decide to use `google-auth-library-java`, we would probably still have to wrap it in our own async caching code because their implementation uses a lock even when retrieving a cached token (which occurs before every HTTP request). So it would ultimately only be a modest reduction of boilerplate.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/scaladsl/GooglePubSub.scala
seglo: I cancelled the probe with `result2.cancel()`.

https://github.com/armanbilge/alpakka/compare/wip-google-common...seglo:seglo/wip-google-common?expand=1

google-cloud-pub-sub/src/test/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApiSpec.scala
```diff
@@ -244,17 +250,17 @@ class PubSubApiSpec extends AnyFlatSpec with BeforeAndAfterAll with ScalaFutures
       WireMock
         .post(
           urlEqualTo(
-            s"/v1/projects/${config.projectId}/subscriptions/sub1:pull"
+            s"/v1/projects/${TestCredentials.projectId}/subscriptions/sub1:pull?prettyPrint=false"
```
seglo, line 253: How come prettyPrint is disabled?

google-cloud-pub-sub/src/test/scala/docs/scaladsl/IntegrationSpec.scala
```diff
@@ -42,7 +42,7 @@ class IntegrationSpec
 
   // The gCloud emulator is selected via environment parameters (in build.sbt)
   // as created in docker-compose.yml
-  private val projectId = "alpakka"
+//  private val projectId = "alpakka"
```
seglo, line 45: delete?

google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/GCStorageExt.scala
```diff
@@ -8,13 +8,16 @@ import akka.actor.{ClassicActorSystemProvider, ExtendedActorSystem, Extension, E
 
 /**
  * Manages one [[GCStorageSettings]] per `ActorSystem`.
+ * Deprecated, please use [[akka.stream.alpakka.google.GoogleSettings]].
```
seglo, line 11: Could use the `@deprecated` scala/javadoc tag here.

google-cloud-storage/src/test/scala/docs/scaladsl/GCStorageSinkSpec.scala
```diff
@@ -41,7 +40,10 @@ class GCStorageSinkSpec
     val secondChunkContent = Random.alphanumeric.take(chunkSize).mkString
     val metadata = Map(Random.alphanumeric.take(5).mkString -> Random.alphanumeric.take(5).mkString)
 
-    mockLargeFileUpload(firstChunkContent, secondChunkContent, chunkSize, Some(metadata))
+    mock.simulate(
+      mockTokenApi,
+      mockLargeFileUpload(firstChunkContent, secondChunkContent, chunkSize, Some(metadata))
+    )
```
seglo, line 46: Curious why this changed.

google-common/src/test/scala/akka/stream/alpakka/google/PaginatedRequestSpec.scala
```diff
@@ -2,52 +2,44 @@
...
     TestKit.shutdownActorSystem(system)
     super.afterAll()
   }
 
-  implicit val settings = BigQuerySettings().copy(credentialsProvider = new CredentialsProvider {
-    override def projectId: String = ???
-    override def getToken()(implicit ec: ExecutionContext, settings: BigQuerySettings): Future[OAuth2BearerToken] =
-      Future.successful(OAuth2BearerToken("yyyy.c.an-access-token"))
-  })
+  implicit val patience = PatienceConfig(timeout = 3.seconds)
```
seglo, line 38: If this is being changed it would be nice to use the Akka test kit `remainingOrDefault` with an `Assert` so the timeout can be scaled with configuration.  Not a big deal since I wouldn't expect this to ever timeout.

https://doc.akka.io/api/akka/2.6/akka/testkit/TestKit.html#remainingOrDefault:scala.concurrent.duration.FiniteDuration

seglo (commented):
Looking good. There's so much churn, but I understand that it's mostly moving things around. I'm relying mostly on the tests to validate this work.

I want to try and get a release out before Friday. I think this is mostly ready to go. I just found some minor things and asked a few questions.

seglo:
> Thanks for looking into this @tg44! I'm inclined to agree.
> 
> To add on to this: even if we did decide to use `google-auth-library-java`, we would probably still have to wrap it in our own async caching code because their implementation uses a lock even when retrieving a cached token (which occurs before every HTTP request). So it would ultimately only be a modest reduction of boilerplate.

Thanks @tg44 @armanbilge. Ok, let's stick with this implementation instead of the lib.

> The new Alpakka implementation in this PR also automatically loads credentials, but only supports strategies 1, 2, and 5. So for anyone currently running on App Engine or Cloud Shell this could be technically a breaking change.

These are probably corner cases anyway. We can implement them later if there's a request.

docs/src/main/paradox/google-cloud-bigquery.md
```diff
@@ -49,11 +49,17 @@ The table below shows direct dependencies of this module and the second tab show
...
-When running in a [Compute Engine](https://cloud.google.com/compute) instance, credentials can be loaded automatically by setting `alpakka.google.bigquery.credentials.provider = compute-engine`.
-If you use a non-standard configuration path or need multiple different configurations, please refer to @ref[the attributes section below](google-cloud-bigquery.md#apply-bigquery-settings-to-a-part-of-the-stream) to see how to apply different configuration to different parts of the stream.
-All of the available configuration settings can be found in the @github[reference.conf](/google-cloud-bigquery/src/main/resources/reference.conf).
+Shared settings for all Google connectors are read by default from the `alpakka.google` configuration section in your `application.conf`.
+Credentials will be loaded automatically:
+
+1. From the file path specified by the `GOOGLE_APPLICATION_CREDENTIALS` environment variable or another [“well-known” location](https://medium.com/google-cloud/use-google-cloud-user-credentials-when-testing-containers-locally-acb57cd4e4da); or
```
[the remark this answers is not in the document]
armanbilge: Much agreed!

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/BigQueryJobs.scala
```diff
@@ -68,21 +65,16 @@ private[scaladsl] trait BigQueryJobs { this: BigQueryRest =>
...
     import SprayJsonSupport._
-    implicit val ec = system.classicSystem.dispatcher
     val uri = BigQueryEndpoints.jobCancel(settings.projectId, jobId)
     val query = ("location" -> location) ?+: Query.Empty
-    BigQueryHttp()
-      .retryRequestWithOAuth(HttpRequest(POST, uri.withQuery(query)))
-      .flatMap { response =>
-        Unmarshal(response.entity).to[JobCancelResponse]
-      }(system.classicSystem.dispatcher)
+    singleRequest[JobCancelResponse](HttpRequest(POST, uri.withQuery(query)))
```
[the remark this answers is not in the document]
armanbilge: Yes, sorry, maybe this became too convoluted.

1. Inside `GoogleHttp` I consolidated `singleRequest[WithOAuth]` with `retryRequest[WithOAuth]` to `single[Authenticated]Request` which retries if a `Retry(ex)` is thrown.
2. In the public API `Google` I expose `singleRequest` which just calls `singleAuthenticatedRequest`.
3. I mixed `Google` into `BigQuery` and preferred calls to public `Google` APIs instead of directly to internal `GoogleHttp`

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/BigQueryTables.scala
```diff
@@ -85,19 +82,17 @@ private[scaladsl] trait BigQueryTables { this: BigQueryRest =>
    * @return a [[scala.concurrent.Future]] containing the [[akka.stream.alpakka.googlecloud.bigquery.model.TableJsonProtocol.Table]]
    */
   def createTable(table: Table)(implicit system: ClassicActorSystemProvider,
-                                settings: BigQuerySettings): Future[Table] = {
+                                settings: GoogleSettings): Future[Table] = {
     import BigQueryException._
     import SprayJsonSupport._
-    implicit val ec = system.classicSystem.dispatcher
+    implicit val ec = ExecutionContexts.parasitic
```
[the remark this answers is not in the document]
armanbilge: Decisively no :) Akka's own `Http().singleRequest` method doesn't even take a execution context as an (implicit) parameter. Thanks for keeping an eye on this!

google-cloud-pub-sub/src/test/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApiSpec.scala
armanbilge: `prettyPrint=false` is the default because otherwise (at least some) Google APIs return huge JSON responses full of whitespace. We could set `prettyPrint=true` for tests?

google-common/src/test/scala/akka/stream/alpakka/google/PaginatedRequestSpec.scala
armanbilge: Actually I explicitly set this because I did have some timeout issues because Hoverfly was slow to spin up I think. Would `remainingOrDefault` still work for this?

armanbilge:
Thanks for the review @seglo, working my way through it.

> These are probably corner cases anyway.

As far as I can tell the `compute-engine` strategy also works on cloud shell since that runs in a compute engine instance anyway. It would be a happy surprise if App Engine could also use this but not sure.

google-cloud-storage/src/test/scala/docs/scaladsl/GCStorageSinkSpec.scala
armanbilge: Why this specific test changed, or why I had to redo all the GCS tests? I couldn't figure out how to get WireMock to work with the forward proxy so I rewrote them with Hoverfly. If someone knows how to configure this I'd love to revert these back to the original WireMock tests.

google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/scaladsl/GooglePubSub.scala
armanbilge: Thanks. I was confused why this test was passing without the probe cancel before this PR.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/ErrorProtoJsonProtocol.scala
```diff
@@ -22,25 +22,25 @@ object ErrorProtoJsonProtocol {
...
-    def withReason(reason: String) =
-      copy(reason = reason)
+    def withReason(reason: util.Optional[String]) =
+      copy(reason = reason.asScala)
     def withLocation(location: util.Optional[String]) =
       copy(location = location.asScala)
-    def withMessage(message: String) =
-      copy(message = message)
+    def withMessage(message: util.Optional[String]) =
+      copy(message = message.asScala)
```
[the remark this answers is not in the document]
armanbilge: Sorry, I'm confused, which accessors are missing?

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/ErrorProtoJsonProtocol.scala
```diff
@@ -52,8 +52,8 @@ object ErrorProtoJsonProtocol {
    * @param message A human-readable description of the error
    * @return an [[ErrorProto]]
    */
-  def createErrorProto(reason: String, location: util.Optional[String], message: String) =
-    ErrorProto(reason, location.asScala, message)
+  def createErrorProto(reason: util.Optional[String], location: util.Optional[String], message: util.Optional[String]) =
```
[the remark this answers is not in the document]
armanbilge: Well usually it would be `ErrorProto.create` but in this case it is `ErrorJsonProtocol.createErrorProto`. Actually, I've been having doubts about keeping all the model classes inside `JsonProtocol` objects. It's ok for Scala but it is much more difficult to use from Java. I can fix this in a quick followup PR?

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/ErrorProtoJsonProtocol.scala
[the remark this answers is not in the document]
seglo: The `with` methods all accept `java.util.Optional`. We should have overloads that take `Option` as well.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/BigQueryJobs.scala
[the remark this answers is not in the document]
seglo: Got it.

google-cloud-pub-sub/src/test/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApiSpec.scala
seglo: It might make debugging easier in the future.

google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/impl/DeprecatedCredentials.scala
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Used purely as a placeholder class to help migrate to common Google auth.
+ */
+@InternalApi
+private[grpc] object DeprecatedCredentials extends CallCredentials {
+  override def applyRequestMetadata(requestInfo: CallCredentials.RequestInfo,
+                                    appExecutor: Executor,
+                                    applier: CallCredentials.MetadataApplier): Unit = ???
+  override def thisUsesUnstableApi(): Unit = ???
```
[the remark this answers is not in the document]
armanbilge: My bad, actually this is totally a breaking change if someone directly uses the credentials from the `PubSubSettings` case class. I can mitigate this but there has to be another slight breaking change :(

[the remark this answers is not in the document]
armanbilge: Of course the other option is that we keep the `google-auth-library-java` dependency.

google-common/src/test/scala/akka/stream/alpakka/google/PaginatedRequestSpec.scala
seglo: `remainingOrDefault` will return a dilated value of the `akka.test.single-expect-default` configuration from the Akka testkit, which is actually 3 seconds by default. Sometimes it's useful to apply a multiple with `akka.test.timefactor` when the test is running in a slow environment (i.e. a build server). It's just a convenient way to increase the timeouts without changing any code.

google-cloud-storage/src/test/scala/docs/scaladsl/GCStorageSinkSpec.scala
seglo: Got it. I don't have any particular attachment to WireMock. As long as the same interactions are recorded/replayed it should be fine.

google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/PubSubSettings.scala
```diff
@@ -17,7 +18,7 @@ final class PubSubSettings private (
     val port: Int,
     val useTls: Boolean,
     val rootCa: Option[String],
-    val callCredentials: Option[CallCredentials]
+    @deprecated("Use akka.stream.alpakka.google.GoogleSettings", "3.0.0") val callCredentials: Option[CallCredentials]
```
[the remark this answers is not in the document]
armanbilge: Added these, let me know if I missed any.

google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/PubSubSettings.scala
```diff
@@ -63,14 +74,21 @@ object PubSubSettings {
   /**
    * Create settings from config instance.
    */
-  def apply(config: Config): PubSubSettings =
+  def apply(config: Config)(implicit system: ClassicActorSystemProvider): PubSubSettings =
```
armanbilge, line 77: @seglo This is a breaking change to address your comment https://github.com/akka/alpakka/pull/2613#discussion_r599848241. We need the system to create credentials.

The non-breaking option is to keep the dependency on Google's auth lib.

google-cloud-storage/src/test/scala/docs/scaladsl/GCStorageSinkSpec.scala
armanbilge: The WireMock and Hoverfly DSLs are very similar so it was a straightforward change.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/ErrorProtoJsonProtocol.scala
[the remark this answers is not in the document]
armanbilge: I see, I thought that `with` methods are Java-only APIs since Scala case classes have `copy` available, is this not the case? If so, I need to fix this in many many places ...

google-common/src/test/scala/akka/stream/alpakka/google/PaginatedRequestSpec.scala
armanbilge: Cool, will be making more use of this!

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/ErrorProtoJsonProtocol.scala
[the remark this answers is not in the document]
seglo: In Akka core as well as other Akka projects, types that have Scala and Java APIs are not usually defined as case classes, because the extra methods they generate will break bin compatibility whenever fields are changed (though perhaps not code compatibility for Scala users who recompile against new deps). It's safer to have a class where the accessors are controlled by us so that we can make additive changes easily and control deprecations more explicitly. We don't allow bin compat breaking changes in patch versions, but we also try to make a best effort across all upgrades to avoid it as well.

A common scenario is when users depend on multiple Alpakka connectors in the same project. We want to have some overlap of compatibility across versions to give users a chance to upgrade and deprecating old APIs. Since there have been no releases that have deprecated the google config across the connectors this common library will replace we have to be careful not to break the upgrade experience since there was no deprecation forewarning.

google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/impl/DeprecatedCredentials.scala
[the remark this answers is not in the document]
seglo: In the spirit of my reply in https://github.com/akka/alpakka/pull/2613#discussion_r600503144 let's try to break anything. At least no code breaking changes. If that means including the dependency `google-auth-library-java` in a vestigial manner for a little while, then I think we should.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/ErrorProtoJsonProtocol.scala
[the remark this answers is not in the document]
seglo: Follow up PR is fine :+1:

google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/PubSubSettings.scala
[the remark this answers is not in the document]
seglo: Looking good, thanks. 

I noticed that the deprecation message was different for the javadocs and the scala annotation in `PubSubSetting` and `Credentials`. Let's sync those up.

google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/PubSubSettings.scala
seglo: See https://github.com/akka/alpakka/pull/2613#discussion_r600505557

In this case, maybe we can create a different factory method to avoid ambiguity with `apply`, and deprecate `apply` (we can replace it later).

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/ErrorProtoJsonProtocol.scala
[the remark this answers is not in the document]
armanbilge: Tracking in #2625

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/ErrorProtoJsonProtocol.scala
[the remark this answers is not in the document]
armanbilge: Thank you for the detailed explanation! This was very helpful and would be a great addition to the contributing guide.

google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/impl/DeprecatedCredentials.scala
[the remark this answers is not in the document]
armanbilge: Makes sense, reverted this and brought back `google-auth-library-oauth2-http` for now.

google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/PubSubSettings.scala
armanbilge: Reverted. The `system` would only have been necessary during the migration period: post-migration, credentials will no longer be handled by `PubSubSettings` anyway.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/ErrorProtoJsonProtocol.scala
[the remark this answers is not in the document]
armanbilge: BTW I added Scala overrides for `with` methods for the `GoogleSettings` class and family.

google-cloud-pub-sub/src/test/scala/akka/stream/alpakka/googlecloud/pubsub/impl/PubSubApiSpec.scala
armanbilge: I'm going to leave this for future PRs to decide on a per-module basis. In my case setting `prettyPrint=false` for BigQuery actually made debugging easier, albeit due to the terrible formatting of the Hoverfly recording/replay scenario:
https://github.com/akka/alpakka/blob/master/google-cloud-bigquery/src/test/resources/BigQueryEndToEndSpec.json#L81

armanbilge:
@seglo did I miss any comments or anything else left to do? Otherwise I think should be in good shape for merge.

google-cloud-storage/src/test/scala/docs/scaladsl/GCStorageSourceSpec.scala
```diff
@@ -289,11 +341,14 @@ class GCStorageSourceSpec
 
     "download file when file exists" in {
       // Ensure file content is above the size limit set by akka.http.client.parsing.max-content-length
-      val fileContent = "Google storage file content" + ("x" * 10000)
+      val fileContent = "Google storage file content" + ("x" * 1000)
```
armanbilge, line 344: This was a mistake (I didn't read the comment properly), need to revert and fix this test.

armanbilge: Fixed.

seglo (approved):
LGTM! I'll wait for the build to go thru and merge this evening or tomorrow.

armanbilge:
Thanks for merging! Besides #773 there's a number of issues this potentially closes: #523, #825, #923, #1385, #1705, #2547.

seglo:
Got it :+1:

PR #2614, opened 2021-03-15
Update Google Cloud BigQuery to Akka 2.6

seglo:
References #2476

### DIFF
--- a/google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/BigQueryAttributes.scala
+++ b/google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/BigQueryAttributes.scala
@@ -4,9 +4,10 @@
 
 package akka.stream.alpakka.googlecloud.bigquery
 
+import akka.actor.ClassicActorSystemProvider
+import akka.stream.Attributes
 import akka.stream.Attributes.Attribute
 import akka.stream.alpakka.googlecloud.bigquery.impl.BigQueryExt
-import akka.stream.{Attributes, Materializer}
 
 /**
  * Akka Stream [[Attributes]] that are used when materializing BigQuery stream blueprints.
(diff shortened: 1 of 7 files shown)


google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/BigQueryAttributes.scala
```diff
@@ -26,12 +27,12 @@ object BigQueryAttributes {
   /**
    * Resolves the most specific [[BigQuerySettings]] for some [[Attributes]]
    */
-  def resolveSettings(attr: Attributes, mat: Materializer): BigQuerySettings =
```
seglo, line 29: @armanbilge Not sure if this was intended to be a public method, if not we can roll this back and make it internal.

armanbilge: It was intended to be public—access to these settings/attributes is public anyway, so I would think only helps to have this utility method.

armanbilge (commented):
I'm a little confused why the use of a `Materializer` in the method signature is anti-2.6? The signature of `resolveSettings` was intended to match `fromMaterializer`, since that's where it's used.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/BigQueryAttributes.scala
seglo: I see its convenience when called from `fromMaterializer` but I'm not sure when this method be useful to an end user.

armanbilge: It's useful whenever a `Future`-returning method is used from within a Stream operator. Here's an example from our code base:
```scala
Sink.fromMaterializer { (attr, mat) =>
  implicit val settings = BigQueryAttributes.resolveSettings(attr, mat)
  Sink.lazyFutureSink { () =>
    for {
      table <- BigQuery.createTable[A](datasetId, tableId) // Uses implicit settings
    } yield BigQuery.insertAll(datasetId, tableId, ...)
  }
}
```

Unless the problem is the use of `Future`-returning methods in the BigQuery API. I've seen some of the other Alpakka APIs prefer the use of `Source` everywhere even if returning a single item.

seglo: I see. In the main stream you settings would be added as an attribute so they can be accessed like this for stages you build separately from each other. 

It wasn't immediately obvious to me that settings would be reused in more than one stage, but it makes sense now.

armanbilge: This is why I was asking about the need for the signature change—I know it's minor but `resolveSettings(attr, mat)` is just a little less annoying than `resolveSettings(attr, mat.system)`.

seglo: Yes, you've convinced me.  I'll close the PR.

armanbilge: Thanks, appreciate that!

Issue #2364, opened 2020-07-24
S3: Consider not deprecating path-style access


mpdn:
While AWS is deprecating path-style access, many other providers of S3-like services are not. Path-style access is usually the easiest way to setup MinIO as it does not require fudging around with the DNS.

So while path-style access for AWS should be deprecated, path-style access for custom endpoints are still relevant.

For the same reason, it would be nice to have a way to disable the warning here: https://github.com/akka/alpakka/blob/4c01bb13fc31cd8900955127356cff6f2601d13b/s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala#L179

I would be willing to put some work into this. The simplest way of achieving this would probably be something like un-deprecating the `withPathStyleAccess` function and remove the warning for custom endpoints.

ennru:
You are right, for compatible storage solutions Alpakka should continue to allow path-style access in one way or the other.

Would it be possible to use the `enpoint-url` setting to achieve something without removing the deprecation?
https://github.com/akka/alpakka/blob/01bb7e4c35f1550dc04b82ab211fa9a9768dae4a/s3/src/main/resources/reference.conf#L87-L91

KGausel:
@ennru I tried that, but maybe I configured it wrong. Do you have any input on how to make it work with MinIO using the `endpoint-url` template method you are suggesting?

I tried `http://myminio.local/{bucket}` and `http://myminio.local/{bucket}/`

ennru:
Alpakka uses Minio with virtual-host style access. The trick is to configure a forward-proxy and a dynamic `enpoint-url`:

https://github.com/akka/alpakka/blob/63b22cb0735215fece5ce91b7d09f0a733b3e0c1/s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3IntegrationSpec.scala#L638-L657

KGausel:
This might not be an option for all users though and seems like an overly complicated way to stop Alpakka from spamming log warnings of path style access being deprecated.

ennru:
@KGausel Good news, @laszlovandenhoek just convinced me that we can should [re-think the path-style access settings](https://github.com/akka/alpakka/pull/2193#issuecomment-686304485

ennru:
With #2392 the deprecation shows only if you don't use an `endpointUrl` thus use AWS' services.
This cleaner solution resolved your request.

PR #2618, opened 2021-03-15
Drop pre-3.0.0 deprecated APIs

seglo:
I dropped all deprecations. Let's discuss in the PR if any should be restored.


s3/src/main/resources/reference.conf
```diff
@@ -74,19 +66,6 @@ alpakka.s3 {
     }
   }
```
seglo, line 68: I may have gone too far with dropping all the `path-style-access`/`access-path` code.

The only reason I can think is to support Minio users, but it was found that `endpoint-url` could be used to satisfy this case https://github.com/akka/alpakka/issues/2364.

ennru (approved):
Nice cleanup.
LGTM.

ennru:
S3 tests are unhappy.

s3/src/main/resources/reference.conf
seglo: Let's keep these deprecated APIs for now. AWS has decided to support path-style-access indefinitely, so until it's dropped, and there are alternatives for issues like #2620, we should have the option available for users.

seglo: > Update (September 23, 2020) – Over the last year, we’ve heard feedback from many customers who have asked us to extend the deprecation date. Based on this feedback we have decided to delay the deprecation of path-style URLs to ensure that customers have the time that they need to transition to virtual hosted-style URLs.
>
> We have also heard feedback from customers that virtual hosted-style URLs should support buckets that have dots in their names for compatibility reasons, so we’re working on developing that support. Once we do, we will provide at least one full year prior to deprecating support for path-style URLs for new buckets.

https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

PR #2619, opened 2021-03-18
Add module for BigQuery Storage API

gkatzioura:
### Purpose
This adds support for the BigQuery Storage API which now seems to be the preferred method of reading data from BigQuery, due to its built in streaming support, and support for parallel reads. It doesn't offer any resource management capabilities however.

This PR references #2106, I'd probably see it as an alternative rather than a replacement, in a similar way to the PubSub HTTP/gRPC modules.

### Background Context
It is quite similar in setup to the PubSub gRPC module, and similarly makes use of akka-gRPC to reduce the complexity in this module. A lot of the auth and configuration is very similar, and could be shared if there was a need/want for a base module.

In order to test this I've created an implementation of the gRPC server, which is not the most thorough, but at least allows local tests with a reasonable representation of the real behaviour, and an exact representation of the API format. I've currently only added support for Avro GenericRecords as a starting point, but Apache Arrow is also supported by the API itself, and I would imagine you could extend this to also support Avro Specific Records for generated classes in the future.


gkatzioura:
This is the continuation of #2180. For start I tackled the comments on the original PR. I will continue with the proposals as well as the [comment](https://github.com/akka/alpakka/pull/2180#issuecomment-796304848) from @armanbilge .

armanbilge:
Great to see this effort continuing, thanks @gkatzioura! You should be aware of #2613 which centralizes common code and configuration for all of the Google modules, would be really excellent if you could target it. You should just be able to grab `credentials` from the `GoogleSettings` case class and use `asGoogle()` to make them compatible with RPC stuff.

gkatzioura:
Hi @armanbilge will try to go for it. I see it's not yet merged so will merge on my branch.

armanbilge:
Yup, not merged yet, we're trying to migrate all the existing modules in that same PR ... let me know if you have any questions/issues with it.

> What I see immediately applicable is the AVRO decoder proposal.

Which proposal is this?

gkatzioura:
This comment here
https://github.com/akka/alpakka/pull/2180#issuecomment-796304848

armanbilge (changes requested):
Thanks for migrating to `google-common`! Added some pointers to help you with this.

I think this is still WIP, but I hope you don't mind that I took a quick look at your `GoogleBigQueryStorage` API and I still think we could represent the entire API with a single generic method e.g.

```scala
def read[A](datasetId: String, tableId: String, readOptions: Option[TableReadOptions], maxNumStreams: Int)(implicit um: FromByteStringUnmarshaller[A]): Source[A, Future[ReadSession.Schema]]
```

Then, we can implement `FromByteStringUnmarshaller`s for various Avro or Arrow implementations or users can bring their own. This makes the API un-opinionated and very flexible with regards to the underlying implementation.

Also notice that I dropped `projectId: String` from the method signature (we can read this from the `GoogleSettings` attached to the stream). And I assume that the `ReadSession.Schema` will be constant, so instead of repeatedly emitting it in the stream we can make it the materialized value?

Thanks for all your efforts on this!

gkatzioura:
Hi!
I did finish the Arrow integration. Somehow I get an [pub/sub grpc](https://github.com/akka/alpakka/pull/2619/checks?check_run_id=2780666593#step:6:570) error on `sbt site`. Wonder if this is a know error or my branch.
@armanbilge may you have a look?

armanbilge:
Wow, thanks for all your work to finish this up! Not sure about the error, looks strange, but I've had lots of strange issues with `sbt site`. Maybe try merging in latest master?

gkatzioura:
> Wow, thanks for all your work to finish this up! Not sure about the error, looks strange, but I've had lots of strange issues with `sbt site`. Maybe try merging in latest master?

Since the error does not ring a bell probably has to do with my changes. I have the branch sync so I will try locally to create a new branch apply the changes and see if it works. 

In any case the code changes will be the same, so any feedback is welcome.

armanbilge:
> probably has to do with my changes.

Could be, but you didn't change the pubsub code right? Maybe it's a weird caching issue.

> so any feedback is welcome.

For sure! I'll make a quick pass for now. Also, do you mind if I ask your thoughts about using the Unmarshalling to provide a generic API from my previous review? https://github.com/akka/alpakka/pull/2619#pullrequestreview-632586796

gkatzioura:
@armanbilge 
My bad missed it, thanks for reminding. Will proceed accordingly.

google-cloud-bigquery-storage/src/main/resources/reference.conf
```diff
@@ -0,0 +1,13 @@
...
+  host = "bigquerystorage.googleapis.com"
+  port = 443
+
+  # Set to "none" to disable TLS
+  # TLS and certificates should be configured via the underlying
+  # SSL-config library
+  # https://lightbend.github.io/ssl-config/WSQuickStart.html#point-the-trust-manager-at-the-pem-file
+  # rootCa = "none"
+
+  credentials.default-scopes = ${?alpakka.google.credentials.default-scopes} ["https://www.googleapis.com/auth/bigquery.readonly"]
```
armanbilge, line 11: This is at the wrong path.
Currently `alpakka.google.cloud.bigquery.grpc.credentials.default-scopes`
Should be `alpakka.google.credentials.default-scopes` (to match the override)

google-cloud-bigquery-storage/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/storage/BigQueryStorageSettings.scala
```diff
@@ -0,0 +1,126 @@
...
+
+import akka.actor.ClassicActorSystemProvider
+import akka.stream.alpakka.google.GoogleSettings
+import com.typesafe.config.Config
+
+final class BigQueryStorageSettings private (
+    val host: String,
+    val port: Int,
+    val rootCa: Option[String] = None,
+    val googleSettings: Option[GoogleSettings] = None
```
armanbilge, line 15: `GoogleSettings` are handled separately so they can be shared across the modules, so this can be removed.

google-cloud-bigquery-storage/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/storage/javadsl/BigQueryStorage.scala
```diff
@@ -0,0 +1,98 @@
...
+ * Copyright (C) 2016-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.googlecloud.bigquery.storage.javadsl
+
+import java.util.concurrent.CompletionStage
+import akka.NotUsed
+import akka.stream.javadsl.Source
+import com.google.cloud.bigquery.storage.v1.ReadSession.TableReadOptions
+import org.apache.avro.generic.GenericRecord
```
armanbilge, line 11: This import will cause `ClassNotFoundExceptions` at runtimes for Arrow users.

google-cloud-bigquery-storage/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/storage/scaladsl/BigQueryArrowStorage.scala
```diff
@@ -0,0 +1,86 @@
...
+      .fromMaterializer { (mat, attr) =>
+        val client = reader(mat.system, attr).client
+        readSession(client, projectId, datasetId, tableId, DataFormat.ARROW, readOptions, maxNumStreams)
+          .map { session =>
+            session.schema match {
+              case ReadSession.Schema.ArrowSchema(_) => ArrowSource.readRecordsMerged(client, session)
+              case other => throw new IllegalArgumentException(s"Only Arrow format is supported, received: $other")
+            }
+          }
+      }
```
armanbilge, line 35: Seems like a lot of duplicated code? Can this be abstracted?

google-cloud-bigquery-storage/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/storage/scaladsl/BigQueryAvroStorage.scala
```diff
@@ -0,0 +1,86 @@
...
+        val client = reader(mat.system, attr).client
+        readSession(client, projectId, datasetId, tableId, DataFormat.AVRO, readOptions, maxNumStreams)
+          .map { session =>
+            session.schema match {
+              case ReadSession.Schema.AvroSchema(_) => AvroSource.readRecordsMerged(client, session)
+              case other => throw new IllegalArgumentException(s"Only Avro format is supported, received: $other")
+            }
+          }
+      }
+      .flatMapConcat(a => a)
```
armanbilge, line 36: Same duplication here

google-cloud-bigquery-storage/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/storage/scaladsl/BigQueryStorage.scala
```diff
@@ -0,0 +1,142 @@
...
+import com.google.cloud.bigquery.storage.v1.avro.AvroRows
+import com.google.cloud.bigquery.storage.v1.storage.{
+  BigQueryReadClient,
+  CreateReadSessionRequest,
+  ReadRowsRequest,
+  ReadRowsResponse
+}
+import com.google.cloud.bigquery.storage.v1.stream.ReadSession.TableReadOptions
+import com.google.cloud.bigquery.storage.v1.stream.{DataFormat, ReadSession}
+import org.apache.avro.generic.GenericRecord
```
armanbilge, line 22: Same `ClassNotFoundException` issue here

armanbilge (changes requested):
See also: #2678 about the new Akka gRPC module, may be worth upgrading? On the other hand, then it will be impossible to use this module with the pub-sub gRPC module until that one is upgraded too. So, not sure, maybe @seglo or @ennru can weigh in on this.

gkatzioura:
Hi @armanbilge , may you please review?
Meanwhile I will create a draft pr to identify and fix the Link Validator issue. If everything is ok will migrate the  work there.

google-cloud-bigquery-storage/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/storage/scaladsl/BigQueryStorage.scala
```diff
@@ -0,0 +1,134 @@
...
+                  } else {
+                    resp.avroRows.get.serializedBinaryRows
+                  }
+                })
+                .map(_.toByteArray)
+                .map(ByteString.apply)
+                .map(a => um.asScala(a))
+                .mapAsync(parallelism) { a =>
+                  a
+                }
```
armanbilge, line 93:
```suggestion
                .mapAsync(1)(resp => {
                  val bytes = if (resp.isArrowRecordBatch)
                    resp.arrowRecordBatch.get.serializedRecordBatch
                  else
                    resp.avroRows.get.serializedBinaryRows
                  um(ByteString(bytes.toByteArray))
                })
```

google-cloud-bigquery-storage/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/storage/scaladsl/BigQueryStorage.scala
```diff
@@ -0,0 +1,134 @@
...
+  private val RequestParamsHeader = "x-goog-request-params"
+
+  def createMergedStreams(
+      projectId: String,
+      datasetId: String,
+      tableId: String,
+      dataFormat: DataFormat,
+      readOptions: Option[TableReadOptions] = None,
+      maxNumStreams: Int = 0
+  ): Source[(ReadSession.Schema, ReadRowsResponse.Rows), Future[NotUsed]] =
```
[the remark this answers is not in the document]
armanbilge: Also, I think the return signature should be `Source[ReadRowsResponse.Rows, Future[ReadSession.Schema]]`.

Although, I'm not entirely sure why we are exposing this method.

google-cloud-bigquery-storage/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/storage/scaladsl/BigQueryStorage.scala
```diff
@@ -0,0 +1,134 @@
...
+import scala.concurrent.{ExecutionContextExecutor, Future}
+
+/**
+ * Google BigQuery Storage Api Akka Stream operator factory.
+ */
+object BigQueryStorage {
+
+  private val RequestParamsHeader = "x-goog-request-params"
+
+  def createMergedStreams(
```
armanbilge, line 30: Can we rename `typed[A]` as `createMergedStreams[A]` and implement an implicit unmarshaller for `ReadRowsResponse.Rows`? So everything can go through this method.

armanbilge (commented):
@gkatzioura sorry, I missed your message, thanks for all your work! Just took a quick pass now.

ennru:
@gkatzioura The Link validator [fails on duplicated gRPC-generated classes](https://github.com/akka/alpakka/runs/3155526016#step:6:533), I believe that is not an effect of your PR.

ennru:
CI fails on code style checks.
e.g. https://app.travis-ci.com/github/akka/alpakka/jobs/536315319#L1246

gkatzioura:
Will look during the weekend.

google-cloud-bigquery-storage/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/storage/scaladsl/BigQueryStorage.scala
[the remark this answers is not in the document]
gkatzioura: Hi @armanbilge This is mainly to allow the user to change the way data are read based on the schema.

google-cloud-bigquery-storage/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/storage/scaladsl/BigQueryStorage.scala
gkatzioura: The main goal for `createMergedStreams` is to make one stream from the multiple ones and give to the user the choice to use an unmarshaller. 
If it is unnecessary can be removed and leave the typed only as is. 
What do you thing @armanbilge ?

google-cloud-bigquery-storage/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/storage/scaladsl/BigQueryStorage.scala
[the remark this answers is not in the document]
armanbilge: I think if we materialize the schema as a `Future[ReadSession.Schema]`, then the user can access it from there? That way it's not repeated with every element in the stream: they can use it once, to set up the rest of the stream.

google-cloud-bigquery-storage/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/storage/scaladsl/BigQueryStorage.scala
armanbilge: > The main goal for createMergedStreams is to make one stream from the multiple ones and give to the user the choice to use an unmarshaller.

I think this makes a lot of sense 👍 If I understand correctly, the only difference between `createMergedStreams` and `typed` is that `typed` uses an unmarshaller, right? So I was thinking, maybe `createMergedStreams` can use an unmarshaller too, and we can provide a default unmarshaller for `ReadRowsResponse.Rows`.

gkatzioura: Ok will apply changes on this, and let you know

gkatzioura:
Hi @armanbilge 
I resolved the cases since I replaced `typed` to `createMergedStreams`. Also added the Java Bindings since there was a consensus. 
There is an issue with duplicate generated GRPC classes which make the pipeline fail and block the merge.
I am ok to apply a solution if there is one, or id someone else more familiar with the issue of Akka-Grpc generation and CI wants to take over.

docs/src/main/paradox/google-cloud-bigquery-storage.md
```diff
@@ -0,0 +1,103 @@
...
+> SQL text filtering statement, similar to a WHERE clause in a query. Currently, only a single predicate that is a comparison between a column and a constant value is supported. Aggregates are not supported
+Scala
+: @@snip (/google-cloud-bigquery-storage/src/test/scala/docs/scaladsl/ExampleReader.scala) { #read-options }
+
+Java
+: @@snip (/google-cloud-bigquery-storage/src/test/java/docs/javadsl/ExampleReader.java) { #read-options }
+
+You can then choose to read and process these streams sequentially, essentially with no parallelism:
+Scala
+: @@snip (/google-cloud-bigquery-storage/src/test/scala/docs/scaladsl/ExampleReader.scala) { #read-sequential }
```
ennru, line 84: The `read-sequential` example code is missing.

docs/src/main/paradox/google-cloud-bigquery-storage.md
```diff
@@ -0,0 +1,103 @@
...
+
+Java
+: @@snip (/google-cloud-bigquery-storage/src/test/java/docs/javadsl/ExampleReader.java) { #read-sequential }
+
+Or in parallel:
+Scala
+: @@snip (/google-cloud-bigquery-storage/src/test/scala/docs/scaladsl/ExampleReader.scala) { #read-parallel }
```
ennru, line 91: The `read-parallel` example code is missing.

ennru (commented):
I added a few Alpakka infra things to push this forward.

ennru:
The example code does not compile https://app.travis-ci.com/github/akka/alpakka/jobs/548074823#L1175

ennru (approved):
Great work.
LGTM.

ennru:
Thank you for your efforts. We'll merge this work from #2780 with a squashed commit history.

PR #2626, opened 2021-03-24
Add missing ClassicActorSystemAdapter overloads to DSLs

seglo:
References #xxxx


cassandra/src/main/scala/akka/stream/alpakka/cassandra/CqlSessionProvider.scala
```diff
@@ -38,6 +38,8 @@ trait CqlSessionProvider {
  */
 class DefaultSessionProvider(system: ActorSystem, config: Config) extends CqlSessionProvider {
 
+  def this(system: ClassicActorSystemProvider, config: Config) = this(system.classicSystem, config)
```
ennru, line 41: IIRC this is problematic for the use side as `ActorSystem` is a `ClassicActorSystemProvider`. The compiler can't choose.

seglo: You're right. I did some testing locally and it wasn't allowed. I'll just replace the default constructor with `ClassicActorSystemProvider`.

s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3.scala
```diff
@@ -691,7 +691,7 @@ object S3 {
    * @param materializer materializer to run with
```
seglo, line 691: I deprecated all the Java DSL that required a `Materializer` and used `ClassicActorSystemProvider` instead of `ActorSystem` for the new overloads.

s3/src/main/scala/akka/stream/alpakka/s3/scaladsl/S3.scala
```diff
@@ -3,15 +3,16 @@
  */
```
seglo, line 3: Breaking change. Replaced `Materializer` implicit params with `ClassicActorSystemProvider`.

ennru (approved):
LGTM.

PR #2628, opened 2021-03-25
Enable fatalwarnings for most Alpakka and internal projects

seglo:
References #2456


seglo:
Re-evaluate later.

seglo:
Took another crack at this because of #2456. All projects have `fatalWarnings := true` except for OrientDB which needs a refactor.

build.sbt
```diff
@@ -107,56 +107,49 @@ TaskKey[Unit]("verifyCodeFmt") := {
...
 
 lazy val ftp = alpakkaProject(
   "ftp",
   "ftp",
   Dependencies.Ftp,
   Test / fork := true,
   // To avoid potential blocking in machines with low entropy (default is `/dev/random`)
-  javaOptions in Test += "-Djava.security.egd=file:/dev/./urandom",
-  fatalWarnings := true
+  javaOptions in Test += "-Djava.security.egd=file:/dev/./urandom"
```
marcospereira, line 152:
```suggestion
  Test / javaOptions += "-Djava.security.egd=file:/dev/./urandom"
```

marcospereira: To be sbt 1.5 friendly.

influxdb/src/main/scala/akka/stream/alpakka/influxdb/impl/AlpakkaResultMapperHelper.scala
```diff
@@ -63,7 +63,7 @@ private[impl] class AlpakkaResultMapperHelper {
         val columnName: String = column.name()
         val fieldType: Class[_] = field.getType()
 
-        if (!field.isAccessible()) {
+        if (!field.canAccess(model)) {
```
marcospereira, line 66: Not sure if this is relevant here, but I don't think `canAccess` exists in Java 8.

seglo: Ah, you're right. I'll silence this warning.

build.sbt
seglo: True, but I think we should tackle this later when we upgrade.

marcospereira (approved):

PR #2637, opened 2021-03-29
Remove the CMSClassUnloadingEnabled jvm flag

PepRoll:
The flag `-XX:+CMSClassUnloadingEnabled` deprecated after the Java 9 release. And it is pointless without the flag `-XX:+UseConcMarkSweepGC` that deprecated too. I would like to remove it from the `.jvmopts`

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,4 +1,3 @@
 -Dfile.encoding=UTF8
 -Xms1g
 -Xss6M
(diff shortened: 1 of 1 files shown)


seglo:
We do use Java 8 in the build, but I take your point that it's not useful without CMS enabled.

seglo (approved):

PR #2639, opened 2021-04-05
Google Common: Fix authorization error

PepRoll:
When using a new version of the BigQuery connector I faced that it starts to return the HTTP response with the status code 401 and the message - "Invalid Credentials" after a while. While I was exploring the code of getting a google access token, I noticed, that the field `expiresAt`  of the `AccessToken` class is initializing with the concatenation of seconds and milliseconds. Because of this, the access token did not refresh.

References #xxxx


armanbilge (approved):
LGTM. Thanks for this fix!

seglo (approved):

PR #2641, opened 2021-04-05
Google: Drop old jwt-core dependencies

armanbilge:
This drops the `jwt-core` dependency from the Google connectors since this is now handled in `google-common`.

Unfortunately, this is currently a potentially fatal bug: the `google-common` module depends on a recent version of `jwt-scala` that has a new Maven organization name but retains the same package name as `jwt-core`, which IIUC creates a classpath conflict that SBT will not automatically resolve via eviction. Due to API changes, if the old `JWT` classes are used instead of the new then authorization requests to Google will fail.


seglo:
I see. Hopefully that's a small subset of users who directly depend on the old `jwt-core` and an Alpakka Google connector. If there is a problem I expect people will find this PR and comment.

seglo (approved):
LGTM

PR #2642, opened 2021-04-05
Create Google Cloud Logging connector

armanbilge:
This is perhaps an unusual addition to Alpakka: a connector that provides logging backends for Google Cloud Logging, specifically:
* an Akka Classic logger
* a Logback appender

As it stands, this connector is primarily intended for logging so only implements a barebones `scaladsl` (used internally) and lacks a `javadsl`. I hope this is okay. The logging backends are supported from both Scala and Java of course. I've also left plenty of room for the API to be fleshed out in a followup PR should anyone want to e.g. implement streaming of logs from Cloud Logging.

CI test coverage is currently very poor: although I added tests for both backends which are straightforward to run against a live account, running a simulation replay is difficult due to the precise nature of logging (changing timestamps, ActorSystem addresses, threadnames, etc.). Furthermore, the Cloud Logging API is dead simple (send logs, get empty response). So essentially all that can be tested effectively in a sandbox environment is that the JSON schema is implemented correctly. I'd appreciate any suggestions for this.

Thanks in advance for your reviews and feedback.


seglo:
Interesting. I'd like to release 3.0.0 final this week so there won't be time to give this a thorough review, but we can add connectors any time.

I think the lack of a java dsl in this case is fine.

How do you feel about the current state of testing?

armanbilge:
Thanks for the heads up about 3.0.0!

> How do you feel about the current state of testing?

I improved it in https://github.com/akka/alpakka/pull/2642/commits/0ed82c226331d5dba3dafd92c2f544d5b3a84349, which deterministically generates some synthetic log entries and plays them against a Hoverfly simulation (ironically, this test will fail against a live account because the log timestamps are no longer within the acceptable range). As I mentioned, the API itself is dead simple, so this is probably sufficient. Ideally if this ever gets fleshed out into a complete API there could be a proper E2E test that writes logs and then re-reads them to check for equality.

patriknw:
At a first glance this looks like a lot of code for a log appender. What is the advantage of using this google specific api compared to using a JsonLayout? Example of JsonLayout can be found in https://github.com/lightbend/kalix-jvm-sdk/blob/main/sdk/java-sdk/src/main/java/kalix/javasdk/logging/LogbackJsonLayout.java and configured with https://github.com/lightbend/kalix-jvm-sdk/blob/main/samples/java-first-service/src/main/resources/logback.xml
More things can be customized in the JsonLayout than is shown in that example. For example additional fields can be added.

Issue #2655, opened 2021-04-14
XmlParser throws exceptions when source is empty


moqichenle:
Hi, I m new to akka streams and I was trying the XML parser: https://doc.akka.io/docs/alpakka/current/data-transformations/xml.html

```
Source.empty[ByteString]
  .via(XmlParsing.parser)
  .toMat(Sink.seq)(Keep.right)
```
is throwing the exception ```IllegalStateException  Stream finished before event was fully parsed```
which is an odd behavior to me. 
I thought an empty source has nothing to emit and it will just complete.
How can I make it work with empty source? thank you

ennru:
Fixed with #2658

PR #2658, opened 2021-04-23
Fixed empty source XML parsing #2655

Seetaramayya:
Added unit test to prove issue exists and afterwords fixed it

### DIFF
--- a/xml/src/main/scala/akka/stream/alpakka/xml/impl/StreamingXmlParser.scala
+++ b/xml/src/main/scala/akka/stream/alpakka/xml/impl/StreamingXmlParser.scala
@@ -6,6 +6,7 @@ package akka.stream.alpakka.xml.impl
 import akka.annotation.InternalApi
 import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
 import akka.stream.alpakka.xml._
+import akka.stream.alpakka.xml.impl.StreamingXmlParser.withStreamingFinishedException
 import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
 import akka.util.ByteString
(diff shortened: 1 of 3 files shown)


xml/src/main/scala/akka/stream/alpakka/xml/impl/StreamingXmlParser.scala
```diff
@@ -55,11 +56,13 @@ import scala.annotation.tailrec
       @tailrec private def advanceParser(): Unit =
         if (parser.hasNext) {
           parser.next() match {
-            case AsyncXMLStreamReader.EVENT_INCOMPLETE =>
-              if (!isClosed(in)) pull(in)
-              else failStage(new IllegalStateException("Stream finished before event was fully parsed."))
+            case AsyncXMLStreamReader.EVENT_INCOMPLETE if isClosed(in) && !started => completeStage()
```
Seetaramayya, line 59: In general, I like the checks on left hand side of `=>` (actions on the right hand side of `=>`). If it is not good then I'll convert it to `if elseif else`

xml/src/main/scala/akka/stream/alpakka/xml/impl/StreamingXmlParser.scala
```diff
@@ -55,11 +56,13 @@ import scala.annotation.tailrec
...
+            case AsyncXMLStreamReader.EVENT_INCOMPLETE if isClosed(in) =>
+              failStage(new IllegalStateException("Stream finished before event was fully parsed."))
+            case AsyncXMLStreamReader.EVENT_INCOMPLETE if !isClosed(in) => pull(in)
```
ennru, line 62:
```suggestion
            case AsyncXMLStreamReader.EVENT_INCOMPLETE => pull(in)
```

ennru (approved):
LGTM.

PR #2660, opened 2021-04-29
UDP: flow with socket options

CptLemming:
I have extended the UDP bind / send flows to allow passing SocketOptions to the Akka UDP manager.

I have a requirement to join channels once the socket is bound and this wasn't possible without the ability to pass options.

This is my first stab at writing any Scala code, so apologies if there are any linting / coding idioms I may have missed.

### DIFF
--- a/udp/src/main/scala/akka/stream/alpakka/udp/impl/UdpBind.scala
+++ b/udp/src/main/scala/akka/stream/alpakka/udp/impl/UdpBind.scala
@@ -9,16 +9,19 @@ import java.net.InetSocketAddress
 import akka.actor.{ActorRef, ActorSystem}
 import akka.annotation.InternalApi
 import akka.io.{IO, Udp}
+import akka.io.Inet.SocketOption
 import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
 import akka.stream.alpakka.udp.Datagram
 import akka.stream.stage._
 
+import scala.collection.immutable.Traversable
 import scala.concurrent.{Future, Promise}
 
 /**
  * Binds to the given local address using UDP manager actor.
  */
 @InternalApi private[udp] final class UdpBindLogic(localAddress: InetSocketAddress,
+                                                   options: Traversable[SocketOption] = Nil,
                                                    boundPromise: Promise[InetSocketAddress])(
     val shape: FlowShape[Datagram, Datagram]
 )(implicit val system: ActorSystem)
@@ -31,7 +34,11 @@ import scala.concurrent.{Future, Promise}
 
   override def preStart(): Unit = {
     implicit val sender = getStageActor(processIncoming).ref
-    IO(Udp) ! Udp.Bind(sender, localAddress)
+    if (options != null) {
+      IO(Udp) ! Udp.Bind(sender, localAddress, options)
+    } else {
+      IO(Udp) ! Udp.Bind(sender, localAddress)
+    }
   }
 
   override def postStop(): Unit =
@@ -77,8 +84,10 @@ import scala.concurrent.{Future, Promise}
   )
 }
 
-@InternalApi private[udp] final class UdpBindFlow(localAddress: InetSocketAddress)(implicit val system: ActorSystem)
-    extends GraphStageWithMaterializedValue[FlowShape[Datagram, Datagram], Future[InetSocketAddress]] {
+@InternalApi private[udp] final class UdpBindFlow(localAddress: InetSocketAddress,
+                                                  options: Traversable[SocketOption] = Nil)(
+    implicit val system: ActorSystem
+) extends GraphStageWithMaterializedValue[FlowShape[Datagram, Datagram], Future[InetSocketAddress]] {
 
   val in: Inlet[Datagram] = Inlet("UdpBindFlow.in")
   val out: Outlet[Datagram] = Outlet("UdpBindFlow.in")
@@ -86,6 +95,6 @@ import scala.concurrent.{Future, Promise}
   val shape: FlowShape[Datagram, Datagram] = FlowShape.of(in, out)
   override def createLogicAndMaterializedValue(inheritedAttributes: Attributes) = {
     val boundPromise = Promise[InetSocketAddress]
-    (new UdpBindLogic(localAddress, boundPromise)(shape), boundPromise.future)
+    (new UdpBindLogic(localAddress, options, boundPromise)(shape), boundPromise.future)
   }
 }
--- a/udp/src/main/scala/akka/stream/alpakka/udp/impl/UdpBind.scala
+++ b/udp/src/main/scala/akka/stream/alpakka/udp/impl/UdpBind.scala
@@ -14,6 +14,7 @@ import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
 import akka.stream.alpakka.udp.Datagram
 import akka.stream.stage._
 
+import scala.Option
 import scala.collection.immutable.Traversable
 import scala.concurrent.{Future, Promise}
 
@@ -21,7 +22,7 @@ import scala.concurrent.{Future, Promise}
  * Binds to the given local address using UDP manager actor.
  */
 @InternalApi private[udp] final class UdpBindLogic(localAddress: InetSocketAddress,
-                                                   options: Traversable[SocketOption] = Nil,
+                                                   options: Option[Traversable[SocketOption]],
                                                    boundPromise: Promise[InetSocketAddress])(
     val shape: FlowShape[Datagram, Datagram]
 )(implicit val system: ActorSystem)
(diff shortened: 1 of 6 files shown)


lightbend-cla-validator:

CptLemming:
Regarding the cla issue above; I had committed with an email not linked to my github account. Now fixed.

udp/src/main/scala/akka/stream/alpakka/udp/impl/UdpBind.scala
```diff
@@ -9,16 +9,19 @@ import java.net.InetSocketAddress
...
 import akka.stream.stage._
 
+import scala.collection.immutable.Traversable
 import scala.concurrent.{Future, Promise}
 
 /**
  * Binds to the given local address using UDP manager actor.
  */
 @InternalApi private[udp] final class UdpBindLogic(localAddress: InetSocketAddress,
+                                                   options: Traversable[SocketOption] = Nil,
```
seglo, line 24: Use an `scala.Option` here and for the `null` check in the stage `preStart`

udp/src/main/scala/akka/stream/alpakka/udp/scaladsl/Udp.scala
```diff
@@ -33,6 +35,24 @@ object Udp {
   def sendFlow(system: ActorSystem): Flow[Datagram, Datagram, NotUsed] =
     Flow.fromGraph(new UdpSendFlow()(system))
 
+  /**
+   * Creates a flow that will send all incoming [UdpMessage] messages to the remote address
+   * contained in the message. All incoming messages are also emitted from the flow for
+   * subsequent processing.
+   */
+  def sendFlow(
+      options: Traversable[SocketOption]
```
seglo, line 44: Interesting that `Traversable` used. I see why you're using it, since eventually you call `Udb.bind` which accepts that type. Generally speaking though we should use `Iterable` instead.

https://www.scala-lang.org/blog/2017/02/28/collections-rework.html

udp/src/main/scala/akka/stream/alpakka/udp/scaladsl/Udp.scala
```diff
@@ -33,6 +35,24 @@ object Udp {
   def sendFlow(system: ActorSystem): Flow[Datagram, Datagram, NotUsed] =
     Flow.fromGraph(new UdpSendFlow()(system))
 
+  /**
+   * Creates a flow that will send all incoming [UdpMessage] messages to the remote address
+   * contained in the message. All incoming messages are also emitted from the flow for
+   * subsequent processing.
```
seglo, line 41: Can you add some docs about the new parameter?

udp/src/main/scala/akka/stream/alpakka/udp/scaladsl/Udp.scala
```diff
@@ -45,6 +65,21 @@ object Udp {
    */
   def sendSink(system: ActorSystem): Sink[Datagram, NotUsed] = sendFlow(system).to(Sink.ignore)
 
+  /**
+   * Creates a sink that will send all incoming [UdpMessage] messages to the remote address
+   * contained in the message.
```
seglo, line 70: Docs for options param.

udp/src/main/scala/akka/stream/alpakka/udp/javadsl/Udp.scala
```diff
@@ -33,6 +36,23 @@ object Udp {
   def sendFlow(system: ClassicActorSystemProvider): Flow[Datagram, Datagram, NotUsed] =
     scaladsl.Udp.sendFlow()(system).asJava
 
+  /**
+   * Creates a flow that will send all incoming [UdpMessage] messages to the remote address
+   * contained in the message. All incoming messages are also emitted from the flow for
+   * subsequent processing.
```
seglo, line 42: Docs for options param.

udp/src/main/scala/akka/stream/alpakka/udp/javadsl/Udp.scala
```diff
@@ -47,6 +67,20 @@ object Udp {
   def sendSink(system: ClassicActorSystemProvider): Sink[Datagram, NotUsed] =
     scaladsl.Udp.sendSink()(system).asJava
 
+  /**
+   * Creates a sink that will send all incoming [UdpMessage] messages to the remote address
+   * contained in the message.
```
seglo, line 72: Docs for options param.

seglo (commented):
Thanks for the PR. I left some comments.

udp/src/test/scala/docs/scaladsl/UdpSpec.scala
```diff
@@ -87,6 +88,49 @@ class UdpSpec
...
+      Source(1 to messagesToSend)
+        .map(i => ByteString(s"Message $i"))
+        .map(Datagram(_, destination))
+        .runWith(Udp.sendSink(List(InetProtocolFamily())))
+      // #send-datagrams
+
+      (1 to messagesToSend).foreach { _ =>
+        sub.requestNext()
+      }
+      sub.cancel()
```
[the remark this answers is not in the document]
CptLemming: I'm struggling with this one: I can't see a way to hook into the underlying implementation to confirm options have been called on the socket. I checked the akka docs/tests hoping for inspiration but came up short.

udp/src/main/scala/akka/stream/alpakka/udp/impl/UdpBind.scala
seglo: On second thought, since this is an `Iterable` with a default argument we can just check to see if it's empty. i.e. with `nonEmpty` which is like a java iterable's `hasNext`. 

```scala
    implicit val sender = getStageActor(processIncoming).ref
    if (options.nonEmpty)
      IO(Udp) ! Udp.Bind(sender, localAddress, options)
    else
      IO(Udp) ! Udp.Bind(sender, localAddress)
  }
  ...
```

There's only one internal caller of UdbBindLogic so it can always take an iterable (no default argument), then move the default param for options to `UdpBindFlow`

```scala
@InternalApi private[udp] final class UdpBindFlow(localAddress: InetSocketAddress,
                                                  options: Iterable[SocketOption] = Nil)(
```

Then for the non-options use case you can call it like this from the DSLs:

```scala
Flow.fromGraph(new UdpBindFlow(localAddress)(system))
```

udp/src/test/scala/docs/scaladsl/UdpSpec.scala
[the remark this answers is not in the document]
seglo: I did some reading on Java [UDP broadcast](https://support.biamp.com/Audia-Nexia/Control/TCP_and_UDP_discovery_methods#:~:text=UDP Broadcast is an automatic,and can be entered manually.) and modified your test to send UDP messages to a broadcast address instead of a specific hostname/ip. I [stole some code from here](https://www.baeldung.com/java-broadcast-multicast) to get the first broadcast address. If the broadcast socket option is not enabled then no messages are received, so I think this might be a valid test, can you confirm?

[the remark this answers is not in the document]
seglo: I put up a PR on your PR branch: https://github.com/CptLemming/alpakka/pull/1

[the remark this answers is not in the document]
CptLemming: Sorry for the delay; I did some testing with the code you put up and did confirm without the socket options the tests failed so it seemed a valid way to check this. Thanks!

ennru:
Please run `sbt javafmtAll` to get the correct code formatting.

udp/src/main/scala/akka/stream/alpakka/udp/impl/UdpBind.scala
```diff
@@ -31,7 +35,10 @@ import scala.concurrent.{Future, Promise}
 
   override def preStart(): Unit = {
     implicit val sender = getStageActor(processIncoming).ref
-    IO(Udp) ! Udp.Bind(sender, localAddress)
+    options match {
+      case Some(socketOptions) => IO(Udp) ! Udp.Bind(sender, localAddress, socketOptions)
+      case None => IO(Udp) ! Udp.Bind(sender, localAddress)
+    }
```
ennru, line 41: I agree with https://github.com/akka/alpakka/pull/2660#discussion_r631845751 -- the wrapping into `Option` shouldn't be necessary. And I assume you may even pass an empty sequence of options to `Udp.Bind` to mean the same as no parameter.

CptLemming: Apologies, I missed that updated comment when I did the last updates. Hopefully I got everything now.

udp/src/main/scala/akka/stream/alpakka/udp/impl/UdpBind.scala
```diff
@@ -31,7 +34,10 @@ import scala.concurrent.{Future, Promise}
 
   override def preStart(): Unit = {
     implicit val sender = getStageActor(processIncoming).ref
-    IO(Udp) ! Udp.Bind(sender, localAddress)
+    if (options.nonEmpty)
+      IO(Udp) ! Udp.Bind(sender, localAddress, options)
+    else
+      IO(Udp) ! Udp.Bind(sender, localAddress)
```
ennru, line 40: You can remove the check, not passing `options` is equal to passing an empty traversable:
https://github.com/akka/akka/blob/d83cda8f56aa97f4cd1d1d8e6c494fc2996818fa/akka-actor/src/main/scala/akka/io/Udp.scala#L104-L107

ennru (approved):
LGTM.

ennru:
Thank you for contribution! and cheers for the first Scala code you wrote!

PR #2661, opened 2021-05-07
Materialize stream instead of actor for Google token requests

armanbilge:
This PR reimplements the Google `OAuth2Credentials` internals to materialize a stream (instead of spawning an actor) to process access token requests. This is necessary because the `system.actorOf` method is not available for typed actor systems (and thus spawning actors is not an option). As a side benefit the stream-based implementation is simpler.


seglo (approved):
Nice work. I think it would have been ok to spawn system actors under the system guardian in Akka 2.6. There's precedent for other Alpakka connectors and Akka extensions to do this. i.e.)

```
classicSystem.asInstanceOf[ExtendedActorSystem].systemActorOf(...)
```

I agree this impl is cleaner though.

PR #2662, opened 2021-05-11
Google Common - Use effective port for GoogleHttp().cachedHostConnectionPool

armanbilge:
A small bug fix. Took a while to catch, because it didn't seem to be a problem when running behind a proxy.


seglo (approved):

seglo:
Thanks.

PR #2663, opened 2021-05-12
BigQuery API improvements

armanbilge:
1. Adds a labels field to `JobConfiguration` and `insertAllAsync`. Not 100% binary compatible to 3.0.0-M1 because of case class stuff, but if important can probably be fixed?
2. Creates a `Schema` singleton object to help materialize implicit `TableSchema` for a class `T`

### DIFF
--- a/google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/JobJsonProtocol.scala
+++ b/google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/JobJsonProtocol.scala
@@ -22,9 +22,9 @@ import scala.compat.java8.OptionConverters._
  * @param jobReference reference describing the unique-per-user name of the job
  * @param status the status of this job
  */
-final case class Job(configuration: Option[JobConfiguration],
-                     jobReference: Option[JobReference],
-                     status: Option[JobStatus]) {
+final case class Job private (configuration: Option[JobConfiguration],
+                              jobReference: Option[JobReference],
+                              status: Option[JobStatus]) {
 
   def getConfiguration = configuration.asJava
   def getJobReference = jobReference.asJava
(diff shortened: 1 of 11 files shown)


google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/JobJsonProtocol.scala
```diff
@@ -70,18 +70,35 @@ object Job {
  * @see [[https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#jobconfiguration BigQuery reference]]
  *
  * @param load configures a load job
+ * @param labels the labels associated with this job
  */
-final case class JobConfiguration(load: Option[JobConfigurationLoad]) {
+final case class JobConfiguration(load: Option[JobConfigurationLoad], labels: Option[Map[String, String]]) {
```
seglo, line 75: We could make this constructor private to avoid breaking change, I guess?  Not a big deal.

seglo (approved):
LGTM. Before we merge can you enable the `MimaPlugin` for big query and google common? You may need to add some exceptions, but that's fine.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/JobJsonProtocol.scala
armanbilge: Not sure if that would help? I think we need to add a constructor with the old signature. I should have marked all of these constructors private to begin with...

seglo: Let's make it private and just add the MimaException. It won't be bin compat but since you added the factory method it should be fine for the Scala DSL at least.

armanbilge: Sounds good. While I'm at it, can I mark all the other case class constructors private?

seglo: Sounds good.

armanbilge: Done.

armanbilge:
I will open a followup PR to enable MiMa for google-common.

PR #2667, opened 2021-05-13
google-common: Improve log messages for application default credentials

armanbilge:
One last fix for 3.0.0 that substantially improves the log messages for application default credentials, especially when they are not found. The previous logging code was badly broken.


seglo (approved):
LGTM. Thanks!

PR #2671, opened 2021-05-15
mqtt-streaming: Accommodate source.queue change in Akka 2.6

seglo:
The `Source.queue` implementation has been modified a few times since Akka 2.6 was released which broke this test.


ennru (approved):
LGTM.

hepin1989:
@ennru When will this be ready?

ennru:
Thank you for the ping. There is a test case that keeps failing on Travis, I don't have time to look into it to understand if it is related to this change.

PR #2678, opened 2021-05-25
Google PubSub gRPC and BigQuery Storage: Akka gRPC 2.1.4

raboof:


ennru:
Thank you, this is an important update. But the breaking changes from ScalaPB are a bit scary to push into a point release, as it upgrades ScalaPB and Akka HTTP.
The only excuse would be that Alpakka 3.0.0 is effectively broken as of #2680.

armanbilge:
In https://github.com/akka/alpakka/pull/2613#issuecomment-800335344 @seglo mentioned that
> Alpakka major releases don't have to be as rare as they are right now. If there's a good reason (and maybe this is one of them) then we could release an Alpakka 4.0.0 in a more reasonable interval, maybe after 3 months?

This update seems like it could be a good reason, along with the ability to make several other breaking changes (including to this module) that would complete the migration of all Google modules to `google-common` introduced in #2613. For example, `google-pub-sub-grpc` also currently has a vestigial dependency to the massive `google-auth-library-java` although it is not used by the internal API (all auth runs through `google-common` now).

raboof:
> Thank you, this is an important update. But the breaking changes from ScalaPB are a bit scary to push into a point release, as it upgrades ScalaPB and Akka HTTP.
>
> The only excuse would be that Alpakka 3.0.0 is effectively broken as of #2680.

Yes, it's not ideal, even if it'd only affect users of `akka-grpc-pub-sub-grpc` that are already on 3.0.0 and use transitive dependencies that are on an older version of Akka gRPC. Without this change there is no version of akka-grpc-pub-sub-grpc that can coexist with Akka gRPC 2.0.0, which is not great either.

ennru:
Thanks for your input @armanbilge.

We decided to postpone this update and will try to schedule another major release later in 2021.

ennru:
Let's bring this PR back to live now!

ennru:
I just merged the upgrade to Akka HTTP 10.2.9 with #2820

ennru:
Rebased on current master.

ennru:
What is breaking this build is the `BigQueryMockServer` which relies on the Akka gRPC generated power APIs for the server-side. Those APIs changed since Akka gRPC 1.0 and need to be updated for the new style.

Furthermore, it looks like `Test / akkaGrpcGeneratedSources := Seq(AkkaGrpc.Server)` doesn't kick in to generate the server code only for testing.

I don't have much time to spend on this, so please help out.

ennru (approved):
LGTM.

seglo:
Thanks! 😁

PR #2679, opened 2021-05-25
[BigQuery] Add extra params to the QueryRequest

peproll:
I want to add extra params to the BigQuery query request. 
- `location` the geographic location where the job should run
- `maximumBytesBilled` limits the number of bytes billed for this query

These params described here https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query#QueryRequest

References #xxxx

### DIFF
--- a/google-cloud-bigquery/src/main/mima-filters/3.0.0.backwards.excludes/2679-query-request
+++ b/google-cloud-bigquery/src/main/mima-filters/3.0.0.backwards.excludes/2679-query-request
@@ -1,0 +1,4 @@
+ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.stream.alpakka.googlecloud.bigquery.model.QueryRequest.unapply")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.googlecloud.bigquery.model.QueryRequest.copy")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.googlecloud.bigquery.model.QueryRequest.this")
+ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.stream.alpakka.googlecloud.bigquery.model.QueryRequest.unapply")
--- a/google-cloud-bigquery/src/main/mima-filters/3.0.0.backwards.excludes/2679-query-request
+++ b/google-cloud-bigquery/src/main/mima-filters/3.0.0.backwards.excludes/2679-query-request
@@ -1,4 +1,3 @@
 ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.stream.alpakka.googlecloud.bigquery.model.QueryRequest.unapply")
 ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.googlecloud.bigquery.model.QueryRequest.copy")
 ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.googlecloud.bigquery.model.QueryRequest.this")
-ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.stream.alpakka.googlecloud.bigquery.model.QueryRequest.unapply")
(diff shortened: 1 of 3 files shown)


google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/QueryJsonProtocol.scala
```diff
@@ -103,15 +121,19 @@ object QueryRequest {
              timeout: util.Optional[Duration],
              dryRun: util.Optional[lang.Boolean],
              useLegacySql: util.Optional[lang.Boolean],
-             requestId: util.Optional[String]) =
+             requestId: util.Optional[String],
+             location: util.Optional[String],
+             maximumBytesBilled: util.OptionalLong) =
```
ennru, line 126: You should not add new parameters here, it is not backwards compatible. 
The `withXyz` methods should be used to set them.

ennru (commented):
Thank you for suggesting this.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/QueryJsonProtocol.scala
```diff
@@ -31,14 +31,18 @@ import scala.concurrent.duration.FiniteDuration
...
 final case class QueryRequest private (query: String,
                                        maxResults: Option[Int],
                                        defaultDataset: Option[DatasetReference],
                                        timeout: Option[FiniteDuration],
                                        dryRun: Option[Boolean],
                                        useLegacySql: Option[Boolean],
-                                       requestId: Option[String]) {
+                                       requestId: Option[String],
+                                       location: Option[String],
+                                       maximumBytesBilled: Option[Long]) {
```
armanbilge, line 45: Just a tiny tiny nitpick, could we keep the order consistent with the [BigQuery reference](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query#QueryRequest)? Here and in `apply`/`create`. Not a huge deal :)

armanbilge (commented):
Thanks for adding these!

While you are making these changes, would you mind adding `labels: Map[String, String]` as well? I think it's another useful param.

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/QueryJsonProtocol.scala
PepRoll: Thanks! No problems, I've changed the order in the private constructor.

PepRoll:
> Thanks for adding these!
> 
> While you are making these changes, would you mind adding `labels: Map[String, String]` as well? I think it's another useful param.

Ok, I think so) I've added it.

google-cloud-bigquery/src/main/mima-filters/3.0.0.backwards.excludes/2679-query-request
```diff
@@ -0,0 +1,4 @@
+ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.stream.alpakka.googlecloud.bigquery.model.QueryRequest.unapply")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.googlecloud.bigquery.model.QueryRequest.copy")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.googlecloud.bigquery.model.QueryRequest.this")
+ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.stream.alpakka.googlecloud.bigquery.model.QueryRequest.unapply")
```
armanbilge, line 4: I think you added it twice?

google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/model/QueryJsonProtocol.scala
```diff
@@ -142,7 +160,8 @@ object QueryRequest {
     "useLegacySql",
     "requestId",
     "location",
-    "maximumBytesBilled"
+    "maximumBytesBilled",
+    "labels"
```
armanbilge: Make sure this order matches the constructor.

armanbilge (changes requested):

google-cloud-bigquery/src/main/mima-filters/3.0.0.backwards.excludes/2679-query-request
PepRoll: You're right.

armanbilge (approved):
LGTM! Thanks for taking my suggestions :)

ennru (approved):
LGTM.

Issue #2683, opened 2021-06-02
Clean up attributes hacks after Akka 2.6.15 is released


armanbilge:
References akka/akka#29973, akka/akka#30142, akka/akka#30161, akka/akka#30171

While working on #2548 and #2613 I had to use various hacks to work around attribute inheritance bugs ([most memorably](https://github.com/akka/alpakka/blob/master/google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/impl/GCStorageStream.scala#L155)). Now that these will be fixed in Akka 2.6.15 these can all be cleaned up.

ennru:
I hope most of these went into #2777.

PR #2688, opened 2021-06-10
Update to Akka v2.6.15

armanbilge:
References #2683

### DIFF
--- a/google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/BigQueryQueries.scala
+++ b/google-cloud-bigquery/src/main/scala/akka/stream/alpakka/googlecloud/bigquery/scaladsl/BigQueryQueries.scala
@@ -100,7 +100,6 @@ private[scaladsl] trait BigQueryQueries { this: BigQueryRest =>
                       else
                         throw BigQueryException("Query job not complete.")
                     }
-                    .addAttributes(attr)
                   val restartSettings = RestartSettings(minBackoff, maxBackoff, randomFactor)
                   RestartSource.onFailuresWithBackoff(restartSettings)(() => pages).map(_.get)
                 } getOrElse Source.empty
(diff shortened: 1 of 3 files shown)


ennru:
Thank you for keeping things up here!
We try to avoid changing the versions of Akka dependencies in patch releases, I suggest we park this until we are ramping up for Alpakka 3.1.0 or 4.0.0

rtyley:
I'd particularly like to see an upgrade to Akka v2.6.**16**, as it has https://github.com/akka/akka/pull/30375 - a fix for the `scala-java8-compat` issue with sbt v1.5's eviction-errors that's been spilling into various places (like https://github.com/playframework/playframework/pull/10966 & https://github.com/scanamo/scanamo/pull/1176).

ennru:
> I'd particularly like to see an upgrade to Akka v2.6.16

As Akka is binary-compatible, you're free to use a later Akka version than what Alpakka ships with. 
See https://doc.akka.io/docs/akka/current/project/downstream-upgrade-strategy.html

rtyley:
> As Akka is binary-compatible, you're free to use a later Akka version than what Alpakka ships with.

That's certainly true, however, even though I can definitely specify Akka v2.6.16 as a dependency in my own projects, if those projects are also _even indirectly_ dependent on an alpakka artifact, sbt 1.5 still regards the varying versions of `scala-java8-compat` in the dependency tree as fatally-incompatible, and won't compile unless those errors are disabled for the entire project with `evictionErrorLevel := Level.Info`.

ennru:
Right, that is very annoying. Akka had to introduce a Scala-version-dependent version for it:

https://github.com/akka/akka/blob/8e5493fb1af29db8f5c98b8897f2dd557aa11456/project/Dependencies.scala#L71-L78

ennru:
@rtyley Does specifying the dependency explicitly in your build solve this?
```sbt
"org.scala-lang.modules" %% "scala-java8-compat" % "1.0.0"
```

ennru:
@rtyley I was told, you'll have to tell sbt to rely on the version you picked:

`ThisBuild / libraryDependencySchemes += "org.scala-lang.modules" %% "scala-java8-compat" % VersionScheme.Always`

rtyley:
> @rtyley I was told, you'll have to tell sbt to rely on the version you picked:
> 
> `ThisBuild / libraryDependencySchemes += "org.scala-lang.modules" %% "scala-java8-compat" % VersionScheme.Always`

Thanks @ennru, you're completely right - many thanks for pointing that out!

ennru:
@armanbilge We're ramping up for Alpakka 4.0 would you like to bring Alpakka to the latest Akka version?
BTW: We intend to drop Scala 2.12 for Alpakka 4.0 #2777

armanbilge:
@ennru thanks for the ping, unfortunately I'm spread too think to be able to work on this. It would be great if whoever takes that on can cherry-pick my changes from here, but they were nothing critical, just removing workarounds for now-fixed bugs.

ennru:
Ok, I made the upgrade to Akka 2.6.18 part of #2777

PR #2693, opened 2021-06-25
elasticsearch: Nop

### LINKED ISSUE #2694: elasticsearch: not possible to skip doing an `Operation` for a given `WriteMessage`
### Versions used 

3.0.1

Akka version: 

2.6.15

### Expected Behavior

Please describe the expected behavior of the issue, starting from the first action.

elasticsearch:
if stream Source is a kafka message and Sink is a kafka commit, some messages may not warrant any elasticsearch index changes. Kafka stream cannot reorder messages. Currently, there is no way to omit doing an elastic operation for a given message. So, need to pass the message through with a "nop".

### Actual Behavior

Please provide a description of what actually happens, working from the same starting point.

elasticsearch requires doing an `Operation` for every `WriteMessage`.

### Reproducible Test Case

Please provide a PR with a failing test.

N/A

If the issue is more complex or requires configuration, please provide a link to a project that reproduces the issue.

N/A

ashendon:
`Operation.Nop` to permit a pass-through `WriteMessage` that does not do any index changes.

For example, if `Source` is a kafka message and `Sink` is a kafka commit, some messages may not warrant any elasticsearch index changes. So pass the message through with a nop.

Fixes #2694

### DIFF
--- a/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/WriteMessage.scala
+++ b/elasticsearch/src/main/scala/akka/stream/alpakka/elasticsearch/WriteMessage.scala
@@ -28,6 +28,7 @@ private[elasticsearch] object Operation {
   object Update extends Operation("update")
   object Upsert extends Operation("update")
   object Delete extends Operation("delete")
+  object Nop extends Operation("nop")
 }
 
 final class WriteMessage[T, PT] private (val operation: Operation,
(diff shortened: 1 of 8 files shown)


ennru (approved):
Cool feature.
LGTM.

ennru:
Thank you for this contribution!

PR #2695, opened 2021-06-25
Drain JMS acks queue when session is closed

jaceksokol:
I found that in some cases when JMS session is closed the acks queue is left non-empty. This is leading to reprocessing uncommitted messages on another session. The test cases that I modified is irregularly failing if you comment out my changes to the session code.


johanandren (approved):
LGTM, thanks @jaceksokol !

Issue #2697, opened 2021-06-29
Archiver.tarReader stuck because of upstream chunk size


johanandren:
When combining Compression.gunzip with Archiver.tarReader with the default gunzip max chunk size is 64kb, in some circumstances (have not verified it is all) where chunk size hits a larger number the stage is not failed but somehow is stuck running but not doing anything.

Sadly cannot share offending tar.gz file, will try to create on that can be shared.

johanandren:
With debug logs enabled, the last entry I see before the stage hanging is this:

```
12:53:54 | DEBUG  akka.stream.alpakka.file.impl.archive.TarReaderStage$$anon$1:101 - emitting completed source for TarArchiveMetadata(filePathPrefix=None,filePathName=something,size=59488,lastModification=2020-04-18T14:10:01Z,linkIndicatorByte=0) 
```

I think the reason could potentially be that we end up in `ReadPastTrailer.onPush`, buffer with that latest element is larger than a trailer (440 bytes) , but subSource is `None` so we move on to trying to `readHeader()` from the remaining bytes after dropping the trailer (416 bytes), but the remaining bytes is only 24, we transition to `CollectHeader` but we have not done a pull to get more bytes from upstream so I think we get stuck there.

PR #2698, opened 2021-07-02
Tar reader stuck with certain files and chunk sizes 

johanandren:
Cannot share repeating tar files (from customer) and did not figure out how to create a repeating file safe to share, but when reading certain tar files the stage would get stuck waiting for more data to be pushed but without having pulled upstream to actually get the next chunk.

References #2697


file/src/main/scala/akka/stream/alpakka/file/impl/archive/TarReaderStage.scala
```diff
@@ -37,12 +36,15 @@ private[file] class TarReaderStage
           extends SubSourceOutlet[ByteString]("fileOut")
           with TarReaderStage.SourceWithTimeout
 
-      readHeader(ByteString.empty)
+      setHandlers(flowIn, flowOut, new CollectHeader(ByteString.empty))
```
johanandren, line 39: We can't do a pull in the constructor, only set handlers, so instead of calling `readHeader` we set collect handlers here directly, we anyway know the buffer is empty.

file/src/main/scala/akka/stream/alpakka/file/impl/archive/TarReaderStage.scala
```diff
@@ -37,12 +36,15 @@ private[file] class TarReaderStage
...
 
       def readHeader(buffer: ByteString): Unit = {
         if (buffer.length >= TarArchiveEntry.headerLength) {
           readFile(buffer)
-        } else setHandlers(flowIn, flowOut, new CollectHeader(buffer))
+        } else {
+          if (!hasBeenPulled(flowIn)) pull(flowIn)
```
johanandren, line 45: For the problematic files/chunk size scenario we'd get here, without `flowIn` being pulled, and then switch to `CollectHeader` waiting to get the next chunk.

ennru (approved):
LGTM.

PR #2702, opened 2021-07-15
Handle failed Future when invoking ES endpoint

jaceksokol:
Invocation of an ES endpoint may not be complete and thus returning failed future. This can happen if the host is not reachable (covered in the test) or in other cases. The one that happened in my project was caused by misconfigured Akka HTTP Client. The server was not responding in the desired time (1 second) and the Future was completed with a failure (`akka.stream.scaladsl.TcpIdleTimeoutException: TCP idle-timeout encountered on connection to [xxx:yyyy], no bytes passed in the last 1 second`). The failure was not handled, and the whole stream was stoping processing without any notification.

In this fix, I added a `recoverWith` and invoke the `failureHandler` within it.


johanandren (approved):
LGTM, thanks!

PR #2703, opened 2021-07-15
AWS S3: use strict http entity for in-memory chunks

phiSgr:
We seem to hit a very peculiar bug where the HTTP pool is stuck in the `WaitingForEndOfRequestEntity` state.
It's sort of a workaround for that as strict entities don't go that state.

Regardless, I believe using a strict entity should be good for performance.

If this is a change that can be merged, I will update the test code.


ennru (commented):
Thank you for digging this up. Looks good to make it aware of strict entities.
Please go ahead with some tests.

ennru (approved):
LGTM.

ennru:
Thank you!

PR #2704, opened 2021-07-15
Fix equals method for MetaHeaders and AES256

mdedetrich:
The `equals` method in `MetaHeaders` incorrectly checks if the other type is a `S3Headers`. This is likely an accident since in both cases both types have a field called `metaHeaders`


ennru (approved):
Nice catch.
LGTM.

mdedetrich:
@ennru No worries, when is it likely for a new release of Alpakka with this fix to be published? Currently using Alpakka in an upcoming project and parts of it don't work due to this broken equals method (mainly testing, wiring up the config).

ennru:
I can't promise, but I believe we have another patch release coming up.
Now that it is merged, you may try the latest [snapshot](https://doc.akka.io/docs/alpakka/current/other-docs/snapshots.html).

ennru:
...and thank you for the fix!

PR #2705, opened 2021-07-16
Only remove ETag quotes conditionally

mpdn:
So we've been trying to use alpakka with IBM Cloud Object Storage (COS). We observed that COS worked with all the standard S3 tooling, but `multipartCopy` did not work with Alpakka and would fail with a "no such part" error. After some digging, it turns out that COS does not quote the ETag in the `CopyPart` response. Alpakka assumes this to be quoted, and simply drops the first and last character. This ends up mangling the ETag, causing it to not be found when completing the copy.

Standard S3 libraries [removes quotes *if present*](https://github.com/aws/aws-sdk-java/blob/f935a3758b771a25f628f1d296cb61044a82b4ac/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/ServiceUtils.java#L122), so it works correctly in this case. Although one might argue that COS should quote the ETag in the `CopyResult`, I also think Alpakka should follow the standard S3 tooling and handle this case gracefully.

This PR removes quotes in the same manner as the AWS tooling does. It basically just trims and then checks if the leading/trailing character is a `"` and removes it if so.

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/Utils.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/Utils.scala
@@ -1,0 +1,17 @@
+/*
+ * Copyright (C) 2016-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.s3
+
+private[s3] object Utils {
+
+  /** Removes S3 ETag quotes in the same way the official AWS tooling does. See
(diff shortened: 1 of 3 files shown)


ennru (approved):
Agreed, simply dropping chars is harsh.
LGTM.

Please fix add the header (try `sbt verifyCodeStyle`).

mpdn:
Should be fixed now - all green. :)

ennru:
Thank you!

PR #2720, opened 2021-08-11
AWS S3: Skip forward-proxy settings on optional overrides

mdedetrich:
The config parser for `S3Settings` uses `c.hasPath("forward-proxy")` to see if any of the values inside of `forward-proxy` in order to determine whether to skip parsing the entire `forward-proxy` section.

Unfortunately `c.hasPath` returns `true` if you provide optional environment overrides even when they are not set, i.e. if you do something like this (which is a supported typesafe config way of overriding specific config)

```
forward-proxy {
  host = ${?HOST}
  port = ${?PORT}
  credentials {
    username = ${?CREDENTIALS_USERNAME}
    password = ${?CREDENTIALS_PASSWORD}
  }
}
```

and you don't actually set any of these environment variables then `S3Settings` will immediately fail on instantiation with 

```
com.typesafe.config.ConfigException$Missing: reference.conf  No configuration setting found for key 'host'
```

Even though none of the variables have actually been overridden. The solution to this is to not only check if `c.hasPath("forward-proxy")` returns `true`, but also the other mandatory fields that are part of the `forward-proxy` configuration (in this case `host` and `port`)

### DIFF
--- a/s3/src/test/scala/akka/stream/alpakka/s3/S3SettingsSpec.scala
+++ b/s3/src/test/scala/akka/stream/alpakka/s3/S3SettingsSpec.scala
@@ -12,6 +12,8 @@ import software.amazon.awssdk.auth.credentials._
 import software.amazon.awssdk.regions.Region
 import software.amazon.awssdk.regions.providers._
 
+import scala.annotation.nowarn
+
 class S3SettingsSpec extends S3WireMockBase with S3ClientIntegrationSpec with OptionValues {
   private def mkSettings(more: String): S3Settings =
     S3Settings(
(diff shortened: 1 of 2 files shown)


s3/src/test/scala/akka/stream/alpakka/s3/S3SettingsSpec.scala
```diff
@@ -250,4 +252,20 @@ class S3SettingsSpec extends S3WireMockBase with S3ClientIntegrationSpec with Op
     settings.forwardProxy.value.credentials.value.username shouldEqual "knock-knock"
     settings.forwardProxy.value.credentials.value.password shouldEqual "whos-there"
   }
+
+  it should "skip parsing forward-proxy when optional environment overrides exist but aren't set" in {
+    @nowarn("msg=possible missing interpolator: detected an interpolated expression")
```
mdedetrich, line 257: This warning occurs because `$` is in the raw string however its never used for actual interpolation (it is the valid syntax for doing an environment variable override).

ennru (approved):
LGTM.

PR #2721, opened 2021-08-19
Google common: ignore forward-proxy on optional environment vars

mdedetrich:
This PR is the equivalent of https://github.com/akka/alpakka/pull/2720 however its done on `GoogleSettings` rather than `S3Settings`.

Note that since there wasn't any existing test for `GoogleSettings` I created a `GoogleSettingsSpec`


ennru (approved):
LGTM.

Issue #2690, opened 2021-06-18
alpakka jms does not support null message properties


BertRobben:
The Alpakka implementation does not support null values for JMS message properties. Looking at the spec (https://javaee.github.io/javaee-spec/javadocs/javax/jms/Message.html), this is a valid practice.

"As noted above, property values may be NULL. "

Trying to use them with alpakka jms however yields an exception.

stack_trace: akka.stream.alpakka.jms.NullMessageProperty: null value was given for Jms property '_wb_tx_uid'. at akka.stream.alpakka.jms.impl.JmsMessageProducer.$anonfun$populateMessageProperties$1(JmsMessageProducer.scala:88) at akka.stream.alpakka.jms.impl.JmsMessageProducer.$anonfun$populateMessageProperties$1$adapted(JmsMessageProducer.scala:77) at scala.collection.immutable.BitmapIndexedMapNode.foreach(HashMap.scala:1076) at scala.collection.immutable.HashMap.foreach(HashMap.scala:1083) at akka.stream.alpakka.jms.impl.JmsMessageProducer.populateMessageProperties(JmsMessageProducer.scala:77) at akka.stream.alpakka.jms.impl.JmsMessageProducer.send(JmsMessageProducer.scala:23) at akka.stream.alpakka.jms.impl.JmsProducerStage$$anon$2.$anonfun$sendWithRetries$1(JmsProducerStage.scala:165) at 

And the code from akka.stream.alpakka.jms.impl.JmsMessageProducer (line 88) clearly specifically throws an exception on null properties.

The expected behavior is that alpakka follows the jms spec and allows null properties.

We're using Alpakka JMS 2.0.2

ennru:
Thanks for reporting. Would you be in a position to suggest the necessary changes?

BertRobben:
Company policy doesn't allow me to contribute source code directly.

But from reading the code in JmsMessageProducer, I would expect

          case null => throw NullMapMessageEntry(key, jmsMessage)

should be replaced by

          case null => message.setObject(key, null)


And that on line 88.

ennru:
Fixed by #2725

PR #2725, opened 2021-09-02
JMS producer should support null values for property #2690

ihostage:
References #2690


ennru (approved):
Thanks @ihostage (and @bertrobben).
LGTM.

PR #2726, opened 2021-09-03
Add GCSSettings to make endpointUrl + basePath configurable

mdedetrich:
This PR adds a `GCSSettings` for `google-cloud-stroage` in order to give the ability to configure the host and the basePath. This is mainly done due to the fact that its currently not possible to test GCS using an implementation such as https://github.com/fsouza/fake-gcs-server since the endpoints are hardcoded at https://github.com/akka/alpakka/blob/f2971ca8a4a71b541cddbd5bf35af3a2a56efe71/google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/impl/GCStorageStream.scala#L33-L34.

Of note is there was a previous `GCStorageSettings` which was deprecated beforehand since there are a lot of settings that are shared between Google services. In order to not create conflicts/confusion, a seperate `GCSSettings` was created and it contains settings that are **only** specific to Google Cloud Storage so from a design PoV there shouldn't be any issues (common settings go into `GoogleSettings`, google service specific settings i.e. endpoints for a specific service go into the projects specific settings).

Tests were also made to make sure that the settings were correctly loaded. Afaik no binary incompatible changes were introduced.

If this PR is accepted/merged it means its also possible to use https://github.com/fsouza/fake-gcs-server to test GCS in alpakka in a similra fashion to how S3 is being tested with https://min.io/ (currently the tests just an in memory mock which is arguably not ideal).

### DIFF
--- a/google-cloud-storage/src/main/resources/reference.conf
+++ b/google-cloud-storage/src/main/resources/reference.conf
@@ -20,3 +20,12 @@ alpakka.google {
     token-scope = deprecated // "https://www.googleapis.com/auth/devstorage.read_write" // default
   }
 }
+
+alpakka.google {
+
+    cloud-storage {
+        endpoint-url = "https://storage.googleapis.com/"
+        base-path = "/storage/v1"
+    }
+
+}
--- a/google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/GCSAttributes.scala
+++ b/google-cloud-storage/src/main/scala/akka/stream/alpakka/googlecloud/storage/GCSAttributes.scala
@@ -1,0 +1,36 @@
+/*
+ * Copyright (C) 2016-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.googlecloud.storage
+
+import akka.stream.Attributes
+import akka.stream.Attributes.Attribute
+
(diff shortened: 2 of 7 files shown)


google-cloud-storage/src/main/resources/reference.conf
```diff
@@ -20,3 +20,8 @@ alpakka.google {
     token-scope = deprecated // "https://www.googleapis.com/auth/devstorage.read_write" // default
   }
 }
+
+alpakka.gcs {
```
mdedetrich, line 24: I used `alpakka.gcs` here. `alpakka.google-cloud-storage` can also work and it doesn't conflict with `alpakka.google.cloud.storage`. It is also possible to do `alpakka.google.cloud.storage` however it overlaps with the previously deprecated settings, due to the way typesafe config works this shouldn't cause any problems it just might be confusing for users.

ennru (commented):
Thank you for keeping up with all that config ceremony.

mdedetrich:
@ennru I have just done as request, the main config is now under `alpakka.google.cloud-storage` and I have fixed the old class reference.

Also adjusted the commit message so its shorter, let me know if anything else is needed.

ennru (approved):
LGTM.

ennru:
Thank you!

PR #2741, opened 2021-10-02
S3: Make eTag naming consistent

mdedetrich:
The intention of this PR is to make `eTag`/`ETag` consistent. This was done with internal private API but also the public API where `MultipartUploadResult` used `etag`/`getEtag`/`getEtag` where as every other public model used `eTag`/`getETag`/`withETag`. Since this is a public API I added `deprecated` methods so that people are warned when using the old inconsistent etag naming and to maintain binary compatibility.

There is one case where (I think) source compatibility is broken but NOT binary compatibility (see comments). Unfortunately I couldn't get mima running in Alpakka to verify this, I am getting the following error when I run `s3 / mimaReportBinaryIssues`

```
[error] (s3 / mimaReportBinaryIssues) java.lang.NoClassDefFoundError: Could not initialize class com.typesafe.tools.mima.core.Config$
```

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/model.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/model.scala
@@ -20,9 +20,13 @@ final class MultipartUploadResult private (
     val location: Uri,
     val bucket: String,
     val key: String,
-    val etag: String,
+    val eTag: String,
     val versionId: Option[String]
 ) {
+
+  /** Scala API */
+  @deprecated("Use eTag", "3.0.3")
+  val etag: String = eTag
 
   /** Java API */
   def getLocation: akka.http.javadsl.model.Uri = akka.http.javadsl.model.Uri.create(location)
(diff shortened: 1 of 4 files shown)


s3/src/main/scala/akka/stream/alpakka/s3/model.scala
```diff
@@ -89,13 +99,13 @@ object MultipartUploadResult {
       location: Uri,
       bucket: String,
       key: String,
-      etag: String,
+      eTag: String,
```
mdedetrich, line 102: AFAIK this can break source compatibility (i.e. if someone calls `apply` with named parameters) but it doesn't break binary compatibility.

s3/src/main/scala/akka/stream/alpakka/s3/model.scala
```diff
@@ -104,13 +114,13 @@ object MultipartUploadResult {
       location: akka.http.javadsl.model.Uri,
       bucket: String,
       key: String,
-      etag: String,
+      eTag: String,
```
mdedetrich, line 117: AFAIK this can break source compatibility (i.e. if someone calls `create` with named parameters) but it doesn't break binary compatibility.

ennru (approved):
Mima on Travis is happy with your changes https://app.travis-ci.com/github/akka/alpakka/jobs/540918111
LGTM.

PR #2742, opened 2021-10-04
S3: Allow bucket creation besides `us-east-1` for AWS

mdedetrich:
There seems to be another inconsistency between real AWS and Minio wrt creating a bucket in the non default region (i.e. `us-east-1`). With AWS S3 you need an addition XML payload (along with the host/access-style which typically contains the region) in order to create a bucket in a non default `us-east-1` region, with Minio this payload is not required. If the XML payload is not in the make bucket request on AWS you will get the following error

```
The unspecified location constraint is incompatible for the region specific endpoint this request was sent to..
```

For more details you can read https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateBucket.html , there is already an existing issue at https://github.com/aws/aws-cli/issues/2603#issuecomment-422271388 which demonstrates the same problem (i.e. even with the aws cli you need to provide an additional `LocationConstraint`).

A test is provided but do note the fixes on this PR are only verifiable when running the test against `AWSS3IntegrationSpec`. With `MinioS3IntegrationSpec` you won't notice any difference.

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/HttpRequests.scala
@@ -199,6 +199,17 @@ import scala.concurrent.{ExecutionContext, Future}
     }
   }
 
+  def createBucketRegionPayload(region: Region)(implicit ec: ExecutionContext): Future[RequestEntity] = {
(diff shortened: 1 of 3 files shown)


ennru (approved):
LGTM.

ennru:
Thank you for all the great additions and fixes!

PR #2747, opened 2021-10-08
Add ListBucketVersions and improve S3.deleteBucketContents

mdedetrich:
This PR adds the listBucketVersions API and also improves the S3.deteBucketContents so that it also deletes all version objects (previously if you had versioned objects then the `S3.deleteBucketContents` wouldn't properly clear the bucket which means that a consequent `S3.deleteBucket` would fail). This also mirrors S3's official documentation on how to clear a bucket, i.e. see https://docs.aws.amazon.com/AmazonS3/latest/userguide/delete-bucket.html.

There were also some minor documentation errors with `listMultipartUploadAndCommonPrefixes` which I fixed.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -217,6 +217,9 @@ services:
       sh -c "
         aws --endpoint-url http://minio:9000 s3api create-bucket --bucket my.test.frankfurt
         aws --endpoint-url http://minio:9000 s3api create-bucket --bucket my-test-us-east-1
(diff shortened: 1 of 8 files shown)


mdedetrich:
Okay so it seems the minio that is used in Travis doesn't really support versioning so I had to disable the test

mdedetrich:
Regarding this issue with not being able to make a versioned bucket in Minio to test properly, should I make an issue for it and annotate it as a comment in the code to fix it later?

PR #2751, opened 2021-10-11
Fixes for list object versions result

mdedetrich:
This PR fixes a few changes that was introduced in my previous PR at https://github.com/akka/alpakka/pull/2747. The easiest way to navigate the PR is to look at the commits which are described below.

### Make Owner fields optional for ListObjectVersionsResult
This commit is mandatory since it appears that in some circumstances you do not get the `Owner` field in the S3 response. Ideally and originally I wanted to make a test to replicate this but it seems like you need a specific AWS account with a certain set of permissions in order to have the missing fields. i.e. on one AWS account I get this as a response

```xml
<ListVersionsResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
<Name>wzlfljq6</Name>
<Prefix/>
<KeyMarker/>
<VersionIdMarker/>
<MaxKeys>1000</MaxKeys>
<IsTruncated>false</IsTruncated>
<Version>
    <Key>1970-01-01T00:00:00.179Z.json</Key>
    <VersionId>null</VersionId>
    <IsLatest>true</IsLatest>
    <LastModified>2021-10-11T18:09:50.000Z</LastModified>
    <ETag>&quot;436e1b9fdef26ffe3f6c9f20524f7cfe-1&quot;</ETag>
    <Size>8845</Size>
    <StorageClass>STANDARD</StorageClass>
</Version>
<Version>
    <Key>1970-01-01T00:00:00.341Z.json</Key>
    <VersionId>null</VersionId>
    <IsLatest>true</IsLatest>
    <LastModified>2021-10-11T18:09:50.000Z</LastModified>
    <ETag>&quot;dc997d6c24774e24e7efb2b027761be4-1&quot;</ETag>
    <Size>3796</Size>
    <StorageClass>STANDARD</StorageClass>
</Version>
</ListVersionsResult>
```

where as with another AWS account I get this as a response

```xml
<ListVersionsResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
    <Name>alpakka-test-bucket-1</Name>
    <Prefix/>
    <KeyMarker/>
    <VersionIdMarker/>
    <MaxKeys>1000</MaxKeys>
    <IsTruncated>false</IsTruncated>
    <Version>
        <Key>deleteBucketContentsTest</Key>
        <VersionId>null</VersionId>
        <IsLatest>true</IsLatest>
        <LastModified>2021-10-11T18:23:19.000Z</LastModified>
        <ETag>&quot;83beb8c4fa4596c8f7b565d390f494e2&quot;</ETag>
        <Size>11</Size>
        <Owner>
            <ID>REDACTED</ID>
            <DisplayName>awstest</DisplayName>
        </Owner>
        <StorageClass>STANDARD</StorageClass>
    </Version>
</ListVersionsResult>
```

I have a suspicion that this may be due to how AWS permissions are set up on the account which means that I can't really write a test for this (changing/creating actual AWS accounts is out of scope for tests).

### Make VersionId optional by detecting null

When you call `S3.listObjectVersions` on a bucket that doesn't have versioning enabled (which is the default) you get `versionId=null` in the XML response (see above XML response snippets as well as https://docs.aws.amazon.com/AmazonS3/latest/userguide/AddingObjectstoVersionSuspendedBuckets.html). This commit makes the `versionId` field an `Option` where `None` is the case where we get `null` from AWS as a quality of life improvement.

### Deleting versions with S3.deleteBucketContents is now optional

Upon the realization that `S3.listObjectVersions` completely fails if you are using Minio I deemed it appropriate to make the behavior of deleting all object versions in `S3.deleteBucketContents`/`S3.deleteObjectsByPrefix` optional with a default of **NOT** deleting object versions.

A test was added to verify this behavior.

This is mainly done as a safety precaution (i.e. the default behavior is now the same at the point in time before https://github.com/akka/alpakka/pull/2747 was merged) and also as a means of allowing Minio users to actually use `S3.deleteBucketContents`/`S3.deleteObjectsByPrefix` (otherwise this method will always fail because Minio doesn't even seem to support the `S3.listObjectVersions` call).

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/javadsl/S3.scala
@@ -141,7 +141,7 @@ object S3 {
    *
    * @param bucket the s3 bucket name
    * @param key the s3 object key
-   * @return A [[akka.stream.javadsl.Source Source]] that will emit [[java.lang.Void]] when operation is completed
+   * @return A [[akka.stream.javadsl.Source Source]] that will emit [[akka.Done]] when operation is completed
    */
   def deleteObject(bucket: String, key: String): Source[Done, NotUsed] =
     deleteObject(bucket, key, Optional.empty(), S3Headers.empty)
@@ -152,7 +152,7 @@ object S3 {
    * @param bucket the s3 bucket name
    * @param key the s3 object key
    * @param versionId optional version id of the object
-   * @return A [[akka.stream.javadsl.Source Source]] that will emit [[java.lang.Void]] when operation is completed
+   * @return A [[akka.stream.javadsl.Source Source]] that will emit [[akka.Done]] when operation is completed
    */
   def deleteObject(bucket: String, key: String, versionId: Optional[String]): Source[Done, NotUsed] =
     deleteObject(bucket, key, versionId, S3Headers.empty)
@@ -164,7 +164,7 @@ object S3 {
    * @param key the s3 object key
    * @param versionId optional version id of the object
    * @param s3Headers any headers you want to add
-   * @return A [[akka.stream.javadsl.Source Source]] that will emit [[java.lang.Void]] when operation is completed
+   * @return A [[akka.stream.javadsl.Source Source]] that will emit [[akka.Done]] when operation is completed
    */
   def deleteObject(bucket: String,
                    key: String,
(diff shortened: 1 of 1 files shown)


johanandren:
Yeah, I guess it can't hurt either, feel free to do as you like (it's not in here now as far as I can see).

mdedetrich:
@johanandren The tests have failed but its due to getting rate limited by docker, i.e.

```
ERROR: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

So its unrelated.

PR #2759, opened 2021-10-20
Pass PubSub gRPC call credentials if not using deprecated API

balagez:
I want set up Pub/Sub gRPC by manually passing service account credentials using the `alpakka.google.credentials.service-account` config. I do not have the `GOOGLE_APPLICATION_CREDENTIALS` environment variable set up since I do not want to package sensitive information such as a service account credentials file in a docker image.

I have this configuration:
```
alpakka.google {
	credentials {
		default-scopes = ${?alpakka.google.credentials.default-scopes} []
		scopes = ${alpakka.google.credentials.default-scopes}

		provider = service-account

		service-account {
			project-id = ${GCP_PROJECT_ID}
			client-email = ${GCP_SERVICE_ACCOUNT_EMAIL}
			private-key = ${GCP_SERVICE_ACCOUNT_PRIVATE_KEY}

			scopes = ${alpakka.google.credentials.scopes}
		}
		...
	}
	...
	cloud.pubsub.grpc {
		host = "pubsub.googleapis.com"
		port = 443
		use-tls = true
		rootCa = "none"
		callCredentials = "none"
	}
}
```

### Problem 1

If I don't set `alpakka.google.cloud.pubsub.grpc.callCredentials` to `none` its default value will make [`PubSubSetting`](https://github.com/akka/alpakka/blob/f2971ca/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/PubSubSettings.scala#L85-L92) try to read the `GOOGLE_APPLICATION_CREDENTIALS` service account credentials file, but the environment variable is not set so it fails with this error:

```
Caused by: java.io.IOException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
	at com.google.auth.oauth2.DefaultCredentialsProvider.getDefaultCredentials(DefaultCredentialsProvider.java:134)
	at com.google.auth.oauth2.GoogleCredentials.getApplicationDefault(GoogleCredentials.java:124)
	at com.google.auth.oauth2.GoogleCredentials.getApplicationDefault(GoogleCredentials.java:96)
	at akka.stream.alpakka.googlecloud.pubsub.grpc.PubSubSettings$.apply(PubSubSettings.scala:87)
	at akka.stream.alpakka.googlecloud.pubsub.grpc.PubSubSettings$.apply(PubSubSettings.scala:104)
	...
```

### Problem 2

If I do set `alpakka.google.cloud.pubsub.grpc.callCredentials` to `none` in order to skip reading `GOOGLE_APPLICATION_CREDENTIALS`, all Pub/Sub gRPC calls will fail with this error:

```
io.grpc.StatusRuntimeException: PERMISSION_DENIED: The request is missing a valid API key.
	at io.grpc.Status.asRuntimeException(Status.java:533)
	at akka.grpc.internal.AkkaNettyGrpcClientGraphStage$$anon$1.onCallClosed(AkkaNettyGrpcClientGraphStage.scala:165)
	at akka.grpc.internal.AkkaNettyGrpcClientGraphStage$$anon$1.$anonfun$callback$1(AkkaNettyGrpcClientGraphStage.scala:71)
	...

```

### Solution

I believe this PR fixes this problem by passing call credentials from `GoogleSettings` to `GrpcClientSettings` exactly the same way as when using the deprecated `alpakka.google.cloud.pubsub.grpc.callCredentials = deprecated` configuration (default behavior).

### DIFF
--- a/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/impl/AkkaGrpcSettings.scala
+++ b/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/impl/AkkaGrpcSettings.scala
@@ -33,7 +33,9 @@ import io.grpc.auth.MoreCallCredentials
     )
 
     (config.callCredentials: @silent("deprecated")) match {
-      case None => settings
+      case None =>
+        val credentials = googleSettings.credentials.asGoogle(sys.dispatcher, googleSettings.requestSettings)
+        settings.withCallCredentials(MoreCallCredentials.from(credentials))
       case Some(DeprecatedCredentials(_)) => // Deprecated credentials were loaded from config so override them
         sys.log.warning(
           "Config path alpakka.google.cloud.pubsub.grpc.callCredentials is deprecated, use alpakka.google.credentials"


raboof:
> I do not want to package sensitive information such as a service account credentials file in a docker image.

That is very sensible. Indeed when using `GOOGLE_APPLICATION_CREDENTIALS` I'd typically recommend providing it 'from the outside' when starting the docker container, not 'embedding' it into the docker container.

> I want set up Pub/Sub gRPC by manually passing service account credentials using the `alpakka.google.credentials.service-account` config.

I think that's reasonable to offer as an option as well - though the same concerns apply as to how to populate that config :)

> I believe this PR fixes this problem by passing call credentials from `GoogleSettings` to `GrpcClientSettings` exactly the same way as when using the deprecated `alpakka.google.cloud.pubsub.grpc.callCredentials = deprecated` configuration (default behavior).

I don't think this is how this was intended to work: setting the `callCredentials` to `"none"` seems like it would specify that you don't use any callCredentials, which is indeed the current implementation.

I'm not sure how falling back to the credentials from `GoogleSettings` would help - wouldn't that still try to read the credentials from `GOOGLE_APPLICATION_CREDENTIALS`?

balagez:
@raboof My understanding is that Pub/Sub API calls do need authentication, no matter what. The `callCredentials` configuration is deprecated because it used to only support reading the default credentials [see this](https://github.com/akka/alpakka/blob/f2971ca/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/PubSubSettings.scala#L85-L92), and it was deprecated in favor of the `alpakka.google.credentials` config. Setting the config to `none` signals that I no longer want to use this config. But regardless, Pub/Sub API calls still need to be authenticated.

So the problem is that if I want to use `alpakka.google.credentials` to set the credentials, and thus I set `callCredentials` to `none` to avoid the deprecation warning and the code I linked above trying to read the credentials file from `GOOGLE_APPLICATION_CREDENTIALS`, Pub/Sub requests will not be authenticated at all because the last `case Some(creds) =>` branch on `AkkaGrpcSettings.scala#45` will not be reached.

Another solution to this problem could be to remove the [deprecated code](https://github.com/akka/alpakka/blob/f2971ca/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/PubSubSettings.scala#L85-L92) but that would possibly be a breaking change for many users.

Issue #2760, opened 2021-10-22
S3: Suggestions on how to modify multipart upload to signal successful chunk upload?


mdedetrich:
For context + general discussion here is the relevant lightbend discord thread https://discuss.lightbend.com/t/committing-a-kafka-cursor-after-chunk-upload-on-multipart-s3-upload/8941.

I currently have a problem where as a client/user I need to know whenever a `S3.multipartUpload` successfully uploads a chunk and to act upon it. In my case specifically I want to upload Kafka commit information when a chunk/part has been uploaded into S3 (as well as when the multipart upload finishes but this is possible to figure out currently).

With the current API this does not appear to be possible since the all of the chunk/part handling is abstracted away, i.e. https://github.com/akka/alpakka/blob/master/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala#L849-L855 only takes `Sink[ByteString, Future[MultipartUploadResult]]`.

One idea I had in mind is to provide an alternate implementation (also mentioned in the lightbend thread) that looks like this

```scala
def multipartUploadWithContext[T](
    s3Location: S3Location,
    contentType: ContentType = ContentTypes.`application/octet-stream`,
    s3Headers: S3Headers,
    chunkSize: Int = MinChunkSize,
    chunkingParallelism: Int = 4,
    chunkUploadSink: UploadPartResponse => Sink[T, NotUsed]
): Sink[(ByteString, T), Future[MultipartUploadResult]]
```

Here the `T` represents some context (in my case specifically its Kafka commit information) and the `chunkUploadSink` is a function that takes `UploadPartResponse` as an input (currently this is private so one solution is to make it public, it contains the result of a chunk/part upload i.e. such as if its successful or failed) and then as an output you have to provide a `Sink[T, NotUsed]` that defines what to do with that `T` (in my case commit the latest applicable Kafka cursor context). Basically the `chunkUploadSink` is like a callback that gets triggered whenever a chunk is uploaded.

It would also possible to just change `multipartUpload` to the above implementation since you can just define `T` as `NotUsed` and provide a a `chunkUploadSink` that doesn't do anything (or just make `chunkUploadSink` optional with `Option`).

Any thoughts?

johanandren:
[...]

PR #2770, opened 2021-11-05
S3: Add multipart upload with context

mdedetrich:
References #2760 

This PR is an implementation of the above referenced ticket. It adds custom versions of `multipartUpload`/`resumeMultipartUpload` that allows you to pass a context with the intention of being able to act on that context whenever a chunk is uploaded to S3, either successfully or unsuccessfully. This is done by providing a `chunkUploadSink: Sink[(UploadPartResponse, immutable.Iterable[C]), NotUsed]` which is a `Sink` that gets executed whenever the said Chunk is uploaded. For a very brief summary of the context of this PR, I want to be able to commit kafka commit cursors after S3 successfully uploads a chunk (with the source being kafka messages).

Notes are listed below

- You may have noticed that the `Sink[(UploadPartResponse, immutable.Iterable[C]), NotUsed]` has an `Iterable[C]` rather than just `C`. This is because for any given chunk you can have multiple emitted elements from the initial source. Initially I wanted to just expose the last context belonging to the last emitted element for the chunk but this behavior is not correct, either in the generic API case or in my context of Kafka (i.e. specifically with Kafka I have to upload the latest given cursor for **every** `GroupTopicPartition` and not just the last commit in the chunk)
  - There is an argument that just storing every single commit context inside an `immutable.Iterable` for an entire chunk in memory can be expensive in terms of memory, an alternative is that the user provides a data structure along with how to update it whenever an element is emitted by `SplitAfterSizeWithContext` (in the context of Kafka this would be your batching mechanism with `CommittableOffsetBatch` by maintaining a `Map[GroupTopicPartition, Long]` and only updating an element in place if its offset is larger than the one currently in the map which is currently implemented in Alpakka's Kafka client with `CommittableOffsetBatch`/`CommittableOffsetBatch.apply`)). I am not against this change (and design wise its considered to be superior) however it does make public API more complex.
    - As a counter argument to the "public API more complex part", one can simply provider the current implementation as a default in the argument (i.e. `immutable.Iterator.empty` as the constructor and `+=` as the append) and then for specific use cases like Kafka you can do `CommittableOffsetBatch.empty`/`CommittableOffsetBatch.apply`). One can also provide a data-structure that wraps this constructor/updater to keep things clean.
- The `UploadPartResponse` in the callback sink was an internal `UploadPartResponse` that was made public. This is because when an S3 chunk is uploaded the user of this public API may want to know details, along with whether or not it was successful or not.
  - In regards to `MultipartUpload`, initially it contained an `S3Location` which is an internal API. Rather than making `S3Location` public as well, I instead just modified `MultipartUpload` to have a `key`/`bucket` to minimize whats being exposed
  - In regards to `UploadPartResponse`/`SuccessfulUploadPart`/`FailedUploadPart` the field `index` was renamed to `partNumber`. Since its now public `partNumber` is more clear (its the official terminology in S3 API for the index of an uploaded part).
  - Regarding the `UploadPartResponse` and the fact that its a `trait` that is inherited, I don't think this should cause any issues with Java compatibility. The `sealed` part of the trait is ignored in Java and a bare Scala `trait` converts directly to a Java `interface` (which has 2 inherited implementations). The only impact here is that since Java has no concept of `sealed`, some Java user of alpakka could create their own implementation of `UploadPartResponse` which bypasses the `sealed` check but I fail to see how this is a real problem.
- Originally I wanted to have on generic `chunkAndRequest` that could be reused for the original `multipartUpload`/`resumeMultipartUpload` (by passing in `Nothing` as the context and having a stub for the `Sink` callback)  however I came across some issues so the current implementation just creates a new `chunkAndRequestWithContext`.
  - Buffering is currently not supported with `chunkAndRequestWithContext`. The original reason for this is that with the currently proposed API changes there is no way to support disk buffering because there is no conversion from the context `C` to a `ByteString` (which we need to have to save on disk). Although its definitely possible to add this as a parameter to the public apis (i.e. a `Sink[C, ByteString, Nothing]`) in my view this would just further complicate the feature
    - Its theoretically possible to still support the in memory buffering (a `Sink[C, ByteString, Nothing]` is not required here since we are already storing the context `C` in memory anyways) however there is still an issue where there is a corner case with `maxChunkSize` that splits a `ByteString` (see https://github.com/akka/alpakka/blob/master/s3/src/main/scala/akka/stream/alpakka/s3/impl/SplitAfterSize.scala#L59). Although we can split a `ByteString`, we have no way of splitting a completely generic context `C`. Theoretically this can be fixed in `SplitAfterSizeWithContext` by pushing the entire next pulled element rather than splitting the current element that is being rolled over but again in the case of simplicity I just removed the buffering aspect (also such an implementation would mean that strictly speaking the maximum size is no longer whats defined by `maxChunkSize` but rather `maxChunkSize` + the remaining bytes from the current `pos` of the currently processed ByteString). If this is important I can try and implement it and document this behavior.
  - There is a `null` sentinel value being used where `atLeastOneByteStringAndEmptyContext` is defined. This is the initial case where initially we just passed an empty `ByteString` however now we also have to pass in an empty context for a generic `C` and since this is all internal I used `null` because I found it cleaner/simpler (the other alternative is to have an `Option[C]` and then having to deal with `Option` everywhere). Due to this sentinel value we have to check that `C` is not null in `MemoryWithContext` before we add it to the final `ListBuffer`. I can also change this to use `Option[C]` if so desired.
  - `requestInfoOrUploadState` is one of the few places where I could easily extract methods to reduce some of the boilerplate
- The returning `Sink` thats exposed in the public API is a `Sink[(ByteString, C)]` (or for java a `Pair[ByteString, C]`). Ideally I would have liked to use a `SinkWithContext` but that doesn't exist in akka-streams yet (maybe I should make a feature request for this in akka-streams as this appears to be a valid use case?) 
  
Note that I still need to write tests but I created this PR in the interest of getting feedback if the API/implementation is appropriate (@ennru / @johanandren let me know your thoughts). I also want to do an end2end test for my personal usecase with Kafka to make sure everything is being implemented correctly. I also want to add docs to https://doc.akka.io/docs/alpakka/current/s3.html to add an example of how to use this API since its usage is non trivial.

I also need to add tests, both unit-based style tests for the new `SplitAfterSizeWithContext`/`MemoryWithContext` and also more generic tests that makes sure the whole process works as intended.

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
@@ -42,28 +42,6 @@ import scala.util.{Failure, Success, Try}
 }
 
 /** Internal Api */
-@InternalApi private[impl] final case class MultipartUpload(s3Location: S3Location, uploadId: String)
-
-/** Internal Api */
-@InternalApi private[impl] sealed trait UploadPartResponse {
-  def multipartUpload: MultipartUpload
-
-  def index: Int
-}
-
-/** Internal Api */
-@InternalApi private[impl] final case class SuccessfulUploadPart(multipartUpload: MultipartUpload,
-                                                                 index: Int,
-                                                                 eTag: String)
-    extends UploadPartResponse
-
-/** Internal Api */
-@InternalApi private[impl] final case class FailedUploadPart(multipartUpload: MultipartUpload,
-                                                             index: Int,
-                                                             exception: Throwable)
-    extends UploadPartResponse
-
-/** Internal Api */
 @InternalApi private[impl] final case class CompleteMultipartUploadResult(location: Uri,
                                                                           bucket: String,
                                                                           key: String,
@@ -857,6 +835,21 @@ import scala.util.{Failure, Success, Try}
       .toMat(completionSink(s3Location, s3Headers.serverSideEncryption))(Keep.right)
 
   /**
+   * Uploads a stream of ByteStrings along with a context to a specified location as a multipart upload. The
+   * chunkUploadSink parameter allows you to act upon the context when a chunk has been uploaded to S3.
+   */
+  def multipartUploadWithContext[C](
+      s3Location: S3Location,
+      chunkUploadSink: Sink[(UploadPartResponse, immutable.Iterable[C]), NotUsed],
+      contentType: ContentType = ContentTypes.`application/octet-stream`,
+      s3Headers: S3Headers,
+      chunkSize: Int = MinChunkSize,
+      chunkingParallelism: Int = 4
+  ): Sink[(ByteString, C), Future[MultipartUploadResult]] =
+    chunkAndRequestWithContext[C](s3Location, contentType, s3Headers, chunkSize, chunkUploadSink)(chunkingParallelism)
+      .toMat(completionSink(s3Location, s3Headers.serverSideEncryption))(Keep.right)
+
+  /**
    * Resumes a previously created a multipart upload by uploading a stream of ByteStrings to a specified location
    * and uploadId
    */
@@ -869,10 +862,35 @@ import scala.util.{Failure, Success, Try}
                             chunkingParallelism: Int = 4): Sink[ByteString, Future[MultipartUploadResult]] = {
     val initialUpload = Some((uploadId, previousParts.size + 1))
     val successfulParts = previousParts.map { part =>
-      SuccessfulUploadPart(MultipartUpload(s3Location, uploadId), part.partNumber, part.eTag)
+      SuccessfulUploadPart(MultipartUpload(s3Location.bucket, s3Location.key, uploadId), part.partNumber, part.eTag)
     }
     chunkAndRequest(s3Location, contentType, s3Headers, chunkSize, initialUpload)(chunkingParallelism)
       .prepend(Source(successfulParts))
+      .toMat(completionSink(s3Location, s3Headers.serverSideEncryption))(Keep.right)
+  }
+
+  /**
+   * Resumes a previously created a multipart upload by uploading a stream of ByteStrings to a specified location
+   * and uploadId. The chunkUploadSink parameter allows you to act upon the context when a chunk has been uploaded to
+   * S3.
+   */
+  def resumeMultipartUploadWithContext[C](
+      s3Location: S3Location,
+      uploadId: String,
+      previousParts: immutable.Iterable[Part],
+      chunkUploadSink: Sink[(UploadPartResponse, immutable.Iterable[C]), NotUsed],
+      contentType: ContentType = ContentTypes.`application/octet-stream`,
+      s3Headers: S3Headers,
+      chunkSize: Int = MinChunkSize,
+      chunkingParallelism: Int = 4
+  ): Sink[(ByteString, C), Future[MultipartUploadResult]] = {
+    val initialUpload = Some((uploadId, previousParts.size + 1))
+    val successfulParts = previousParts.map { part =>
+      SuccessfulUploadPart(MultipartUpload(s3Location.bucket, s3Location.key, uploadId), part.partNumber, part.eTag)
+    }
+    chunkAndRequestWithContext[C](s3Location, contentType, s3Headers, chunkSize, chunkUploadSink, initialUpload)(
+      chunkingParallelism
+    ).prepend(Source(successfulParts))
       .toMat(completionSink(s3Location, s3Headers.serverSideEncryption))(Keep.right)
   }
 
@@ -883,7 +901,7 @@ import scala.util.{Failure, Success, Try}
       s3Headers: S3Headers
   )(implicit mat: Materializer, attr: Attributes): Future[MultipartUploadResult] = {
     val successfulParts = parts.map { part =>
-      SuccessfulUploadPart(MultipartUpload(s3Location, uploadId), part.partNumber, part.eTag)
+      SuccessfulUploadPart(MultipartUpload(s3Location.bucket, s3Location.key, uploadId), part.partNumber, part.eTag)
     }
     Source(successfulParts)
       .toMat(completionSink(s3Location, s3Headers.serverSideEncryption).withAttributes(attr))(Keep.right)
@@ -983,7 +1001,7 @@ import scala.util.{Failure, Success, Try}
 
     Source
       .future(
-        completeMultipartUploadRequest(parts.head.multipartUpload, parts.map(p => p.index -> p.eTag), headers)
+        completeMultipartUploadRequest(parts.head.multipartUpload, parts.map(p => p.partNumber -> p.eTag), headers)
       )
       .flatMapConcat(signAndGetAs[CompleteMultipartUploadResult](_, populateResult(_, _)))
       .runWith(Sink.head)
@@ -1067,21 +1085,6 @@ import scala.util.{Failure, Success, Try}
 
     val chunkBufferSize = chunkSize * 2
 
-    val requestInfoOrInitialUploadState = initialUploadState match {
-      case Some((uploadId, initialIndex)) =>
-        // We are resuming from a previously aborted Multipart upload so rather than creating a new MultipartUpload
-        // resource we just need to set up the initial state
-        Source
-          .single(s3Location)
-          .flatMapConcat(_ => Source.single(MultipartUpload(s3Location, uploadId)))
-          .mapConcat(r => Stream.continually(r))
-          .zip(Source.fromIterator(() => Iterator.from(initialIndex)))
-      case None =>
-        // First step of the multi part upload process is made.
-        //  The response is then used to construct the subsequent individual upload part requests
-        initiateUpload(s3Location, contentType, s3Headers.headersFor(InitiateMultipartUpload))
-    }
-
     val headers = s3Headers.serverSideEncryption.toIndexedSeq.flatMap(_.headersFor(UploadPart))
 
     Flow
@@ -1125,7 +1128,7 @@ import scala.util.{Failure, Success, Try}
           .mergeSubstreamsWithParallelism(parallelism)
           .filter(_.size > 0)
           .via(atLeastOne)
-          .zip(requestInfoOrInitialUploadState)
+          .zip(requestInfoOrUploadState(s3Location, contentType, s3Headers, initialUploadState))
           .groupBy(parallelism, { case (_, (_, chunkIndex)) => chunkIndex % parallelism })
           // Allow requests that fail with transient errors to be retried, using the already buffered chunk.
           .via(RetryFlow.withBackoff(minBackoff, maxBackoff, randomFactor, maxRetries, retriableFlow) {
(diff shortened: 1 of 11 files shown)


s3/src/main/scala/akka/stream/alpakka/s3/impl/SplitAfterSizeWithContext.scala
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Internal Api
+ *
+ * Splits up a byte stream source into sub-flows of a minimum size while maintaining a Context. Does not attempt to
+ * create chunks of an exact size. Unlike `SplitAfterSize` there is no maximum size since this would imply splitting
+ * up a Context which is not possible to do if the Context is generic.
+ *
+ * This also means that `SplitAfterSizeContext` currently doesn't support buffering since that would require a way to
+ * serialize the Context in the case of a disk buffer which is currently unsupported (we would have to add in a
+ * C => ByteString serializer in the public API)
```
johanandren, line 23: That could potentially be deferred to the Akka serialization infrastructure though, and not require any additional parameters in the API, but probably best if it was opt in anyway in that case, and ok to keep in mem like this for now and add once there is a need rather than to make this PR even more complex.

s3/src/main/scala/akka/stream/alpakka/s3/impl/SplitAfterSizeWithContext.scala
```diff
@@ -0,0 +1,65 @@
...
+ *
+ */
+@InternalApi private[impl] object SplitAfterSizeWithContext {
+  // The warning being suppressed is abstract type pattern C is unchecked since it is eliminated by erasure which
+  // happens in the pattern match on collect.
+  @silent
```
johanandren, line 29: can it be `case (bs: ByteString, context: C @unchecked)` instead of silencing the warning?

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -856,6 +834,21 @@ import scala.util.{Failure, Success, Try}
...
+   * chunkUploadSink parameter allows you to act upon the context when a chunk has been uploaded to S3.
+   */
+  def multipartUploadWithContext[C](
+      s3Location: S3Location,
+      chunkUploadSink: Sink[(UploadPartResponse, immutable.Iterable[C]), NotUsed],
+      contentType: ContentType = ContentTypes.`application/octet-stream`,
+      s3Headers: S3Headers,
+      chunkSize: Int = MinChunkSize,
+      chunkingParallelism: Int = 4
+  ): Sink[(ByteString, C), Future[MultipartUploadResult]] =
```
johanandren, line 848: Given that it returns a sink and accepts a sink, and will feed elements through one side to the other, should it perhaps rather return a `Flow[(ByteString, C), (UploadPartResponse, Iterable[C]), Future[MultipartUploadResult]]`, which expresses that more clearly, and is materialized as one stream, or am I missing something that makes it specifically require sink + sink?

s3/src/main/scala/akka/stream/alpakka/s3/impl/SplitAfterSizeWithContext.scala
mdedetrich: Wasn't aware there is an akka serialization infrustructure (I was just thinking of providing a `Flow[C, ByteString, NotUsed]` but yes we are in agreement this can be added later.

s3/src/main/scala/akka/stream/alpakka/s3/impl/SplitAfterSizeWithContext.scala
mdedetrich: Ah so that is what `@unchecked` does, I had no idea, doing this now!

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
mdedetrich: I can see both sides, from one end a `Flow` from a technically makes sense for the reasons you stated however from a usability point of view its slightly odd because you will always use it as a `Sink` and hence you will end up doing `.via(multipartUploadWithContext(...)).to(Sink.ignore)` all of the time rather than just `.to(multipartUploadWithContext(...))`.

I spoke with a colleague about this and he said that as far as he could recall there were already existing examples in Alpakka where a `Flow` is used so I will do some quick research and if its the case then I will change it for the sake of consistency.

johanandren: Yeah, I think that's how the other connectors in general look. A flow can also be run like so `source.via(flow).run()` (short for `.runWith(Sink.ignore)`, so the syntax difference vs `source.runWith(sink)` is not huge. 

I did expect that you want to run this with something consuming those contexts though, isn't that kind of the point?

mdedetrich: True, but in my case (I actually adjusted my code to work with this PR) the "callback" is a `Sink` (actually `Committer.sink(committerSettings)` from Alpakka Kafka) and since the current implementation is keeping the return type of `multipartUploadWithContext` as a `Sink` that part is unchanged (I just had to adjust `contramap` to pass in the `CursorContext`).

So if I changed it to a `Flow[(ByteString, C), (UploadPartResponse, Iterable[C]), Future[MultipartUploadResult]]` I would have do more adjustments (i.e. using `Sink`'s within a flow) but as you said this is conventional for this usecase.

I will do change to `Flow[(ByteString, C), (UploadPartResponse, Iterable[C]), Future[MultipartUploadResult]]` locally and update my PR with those changes to see if there are any problems (there shouldn't be) and I will also do some future testing

mdedetrich: So I tried changing it to `Flow[(ByteString, C), (UploadPartResponse, Iterable[C]), Future[MultipartUploadResult]]` and I realized that this is the wrong type signature for what is being done.

I will explain this the best way I can, but the situation here is different to what is done with other Alpakka connectors. For other Alpakka modules, there are "lower" level building blocks (composed of `Source`/`Flow`/`Sink`) which are then combined by the user of their API. On the other hand specifically with S3's `multipartUpload`/`resumeMultipartUpload` the implementation is the opposite, i.e. this "lower" level building block composition is abstracted away in the API (in `chunkAndRequest`/`chunkAndRequestWithContext`) and for good reason (this is unnecessary complication that the user doesn't need to know).

The other issue with `Flow[(ByteString, C), (UploadPartResponse, Iterable[C]), Future[MultipartUploadResult]]` is that its saying the output of the flow is `[(UploadPartResponse, Iterable[C])` which is not correct, this is actually an input (i.e. the user has to define how to act on `(UploadPartResponse, immutable.Iterable[C])`, this is not something that is implemented in the Alpakka S3 flow.

To simplify it down further, what we are actually doing is interposing one sink (the chunkUploadSink) inside of another sink (the s3UploadSink) with the method of interposing being an implementation detail of s3UploadSink (since the implementation of s3UploadSink itself hidden as an implementation detail).

Ultimately I think we have 3 ways of solving this feature

1. This current implementation, where the user has to provide an additional `Sink[(UploadPartResponse, immutable.Iterable[C]), NotUsed]` as a parameter.
2. What you suggested at https://github.com/akka/alpakka/issues/2760#issuecomment-966064402 which is exposing the internals, this means that a lot of the internal implementation of `S3Stream.scala` would become public so that users are able to combose the multipart upload themselves how they want. In other words users would actually write https://github.com/akka/alpakka/pull/2770/commits/3aec0df850768bf8c015ad8e71e7a19636929c26#diff-cbe9040cf5467cee6ffdbd2896491e2da2acea2aec409e9075c5cb15e3180e59R1227-R1261 themselves (and by association any of the functions in that method would need to be public).
3. Another 3rd alternative which I just came up with is that rather than `multipartUpload`/`resumeMultipartUpload` only returning a `Sink[ByteString, Future[MultipartUploadResult]]`, it would instead return a `(Sink[ByteString, Future[MultipartUploadResult]], Source[String])`. The `Source[String]` is then a reference which you can pass into a second function `chunkResponseSink(uploadId: String): Sink[(UploadPartResponse, immutable.Iterable[C]), NotUsed]` as an `uploadId` which you would call right after calling `multipartUpload`/`resumeMultipartUpload`. The issue with this approach is that in order to not break the public API of `multipartUpload`/`resumeMultipartUpload` we would have to make new versions and secondly we would have to store some additional state within `multipartUpload`/`resumeMultipartUpload` since we have to hold onto an `uploadId` reference.

johanandren: Thanks for investigating that a bit. Since it is anyway a bit of a power-user feature/API I think the current sink-sink API is ok.

mdedetrich: Perfect thanks, I will work on adding tests to finalize the PR. Do you think its realistic to have this out before the next release (i.e. https://github.com/akka/alpakka/issues/2775) if I finish it in the next couple of days or is this pushing it too much?

johanandren: I wouldn't want make it a promise but we can try, wdyt @ennru ?

ennru: We can wait a few more days before kicking off the release.

s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3IntegrationSpec.scala
```diff
@@ -471,7 +473,9 @@ trait S3IntegrationSpec
       result.lift(currentChunk) match {
         case Some(currentString) =>
           val newString = ByteString(s"\n${newAcc.toString()}")
-          if (currentString.length <= S3.MinChunkSize) {
+          //TODO: Performance here can be improved by using the same technique as in
+          //createStringCollectionContextWithMinChunkSizeRec to keep track of the current string size in bytes
+          if (currentString.toArray.length < S3.MinChunkSize) {
```
mdedetrich, line 478: There was a fencepost error which I realized when writing `createStringCollectionContextWithMinChunkSizeRec`. For this specific test it didn't make a difference since we were only checking the entire file and didn't care about chunks however for the `perform a chunked multi-part upload with the correct context` test this is relevant

s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3IntegrationSpec.scala
```diff
@@ -692,6 +696,112 @@ trait S3IntegrationSpec
...
+   * `S3.MinChunkSize` bytes. Unlike `createStringCollectionWithMinChunkSizeRec` this version also adds an index
+   * to each individual `ByteString` which is helpful when dealing with testing for context
+   * @param numberOfChunks The number of chunks to create
+   * @return A List of `List[ByteString]` where each element is at least `S3.MinChunkSize` in size along with an
+   *         incrementing index for each `ByteString`
+   */
+  def createStringCollectionContextWithMinChunkSize(numberOfChunks: Int): List[List[(ByteString, BigInt)]] =
+    createStringCollectionContextWithMinChunkSizeRec(numberOfChunks).toList.map(_.toList)
+
+  it should "perform a chunked multi-part upload with the correct context" in {
```
mdedetrich, line 762: One thing to note about this test is that not only does it test the contexts are correct as a whole, but also that there aren't any fencepost errors for contexts in a single chunk, i.e. the last context for chunk n that is being uploaded will also be the same in the `collectContextSink` callback.

This is why the data structure for `originalContexts`/`resultingContexts` is a `List[List[BigInt]]` where the outer list represents the chunk

mdedetrich:
@ennru @johanandren PR is ready for review. Another thing I was thinking about is that we could also add `@ApiMayChange` to the newly added functions in the public Java/Scala API incase we find another better way of solving this problem? I guess if we find a better solution we can just add new overloaded functions and deprecate the current one's being added (the function signature should be different after erasure so I think this should be fine).

s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3IntegrationSpec.scala
```diff
@@ -692,6 +696,112 @@ trait S3IntegrationSpec
       fullInputData.utf8String shouldEqual fullDownloadedFile.utf8String
     }
   }
+  @tailrec
+  final def createStringCollectionContextWithMinChunkSizeRec(
```
mdedetrich, line 700: We could probably abstract the core algorithm for creating the chunks along with `createStringCollectionWithMinChunkSizeRec`to reduce boilerplate but I wanted to get this PR out fast so we don't delay release.

s3/src/test/scala/akka/stream/alpakka/s3/impl/auth/SplitAfterSizeWithContextSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+      .runWith(Sink.seq)
+      .futureValue should be(
+      Seq(
+        Seq((ByteString(1, 2, 3, 4, 5), 1), (ByteString(6, 7, 8, 9, 10, 11, 12), 2)),
+        Seq((ByteString(13, 14), 3))
+      )
+    )
+  }
+
+  it should "not split large elements (unlike SplitAfterSize)" in assertAllStagesStopped {
```
mdedetrich, line 58: Unlike `SplitAfterSize` we do **NOT** want to split large elements because there is no way to spit the context associated with that large element so this test makes sure this doesn't happen.

s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3IntegrationSpec.scala
```diff
@@ -45,7 +47,7 @@ trait S3IntegrationSpec
   )
   implicit val ec: ExecutionContext = actorSystem.dispatcher
 
-  implicit val defaultPatience: PatienceConfig = PatienceConfig(90.seconds, 100.millis)
+  implicit val defaultPatience: PatienceConfig = PatienceConfig(3.minutes, 100.millis)
```
mdedetrich, line 50: I had to increase this to 3 minutes because depending on laptop spec/internet connection it can be possible that you cannot download/upload the large chunks for the related tests in time.

johanandren (commented):
Looking good as far as I can see, only one comment about a small API doc improvement.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -856,6 +834,21 @@ import scala.util.{Failure, Success, Try}
     chunkAndRequest(s3Location, contentType, s3Headers, chunkSize)(chunkingParallelism)
       .toMat(completionSink(s3Location, s3Headers.serverSideEncryption))(Keep.right)
 
+  /**
+   * Uploads a stream of ByteStrings along with a context to a specified location as a multipart upload. The
+   * chunkUploadSink parameter allows you to act upon the context when a chunk has been uploaded to S3.
+   */
```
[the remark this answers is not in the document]
mdedetrich: No worries, I added the documentation as a separate commit so you can check it out in isolation. If more details are needed then let me know and I can update + rebase that commit

johanandren (approved):
LGTM

ennru:
Thank you @mdedetrich and @johanandren for discussing this advanced feature in detail.

mdedetrich:
@johanandren @ennru Thanks for putting the time into this, greatly appreciate it!

PR #2780, opened 2021-11-23
BigQuery storage module

jrudolph:
This is a version of #2619 consisting of only two commits that should have the same diff as #2619.

Should be merged and not squashed into master.


ennru (approved):
This covers the exact content of #2619.

LGTM.

ennru:
I'll fix the headers after merge to our main branch.

PR #2788, opened 2021-12-02
google-common: Catch and return JWT errors

balagez:
The `generateJwt` method (or rather, the underlying library) might throw an exception which is unhandled and makes the streams that require authentication fail without anything logged and any way to recover.

This can be reproduced by simply changing the private key to something invalid, like adding a few extra characters, in this case a `java.security.InvalidKeyException` is thrown.

This PR aims to catch the exception and return it in the `Future` so that the downstream can properly handle it.


ennru (approved):
LGTM.

ennru:
Thank you!

PR #2792, opened 2021-12-14
Google common: user access credentials

bjornnes:
Our devs normally acquire credentials through `gcloud auth application-default login` on their personal computers. This creates a file in the well-known credentials file location.

We already try to parse the file found in this location into a service account credentials file. However, credentials acquired this way will have a different format and can not be used as service account credentials. In addition, tokens are obtained a bit differently as we have to use a refresh token and a different endpoint.

I have essentially duplicated the functionality for authorizing with service accounts, altered the expected json format, and changed the endpoint for token generation. Let me know if this is something we would be interested to merge into the main repo and I'll gladly make the necessary changes :)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 We follow the standard GitHub [fork & pull](https://help.github.com/articles/using-pull-requests/#fork--pull) approach to pull requests. Just fork the official repo, develop in a branch, and submit a PR!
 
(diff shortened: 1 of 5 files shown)


leozilla:
+1 for supporting this

I was actually surprised that alpakka did not support this. Even the alpakka google-common documentation refers an [article](https://medium.com/google-cloud/use-google-cloud-user-credentials-when-testing-containers-locally-acb57cd4e4da) describing this functionality [here](https://doc.akka.io/docs/alpakka/current/google-common.html#credentials).

ennru (approved):
LGTM.

ennru:
Thank you! Great DX addition.

Issue #2753, opened 2021-10-14
GCS call to GCStorage.rewrite(..).withAttributes(GoogleAttributes.settings(myCustomSettings)) does not take into account customsettings


ramazanyich:
Alpakka version: 3.0.3

We are using custom settings to provide google configuration settings like projectId, privatekey, etc. to CGSStorage API calls.
we make use of withAttributes() api call to different GCStorage API calls: GCStorage.deleteObject, listObject,simpleUpload, etc.
it works for all of them excerpt call to GCStorage.rewrite(..).withAttributes(GoogleAttributes.settings(myCustomSettings)) 
We get a 401 error. This is the code which we use. The   withAttributes(GoogleAttributes.settings(customSettings)) code is used for all our other GCStorage calls and it is the same.
```
  GCStorage
          .rewrite(
            googleStorageConfiguration.bucket,
            getFileName(oldObject.id),
            googleStorageConfiguration.bucket,
            getFileName(newObject.id)
          )
          .withAttributes(GoogleAttributes.settings(customSettings))
```
with alpakka version 2.0.2 it did work.
this is old code which did work:
```
 GCStorage
          .rewrite(
            googleStorageConfiguration.bucket,
            getFileName(oldObject.id),
            googleStorageConfiguration.bucket,
            getFileName(newObject.id)
          )
          .withAttributes(GCStorageAttributes.settings(customSettings))
```

stacktrace
```
[401] {"error":{"code":401,"message":"Invalid Credentials","errors":[{"message":"Invalid Credentials","domain":"global","reason":"authError","locationType":"header","location":"Authorization"}]}}
java.lang.RuntimeException: [401] {"error":{"code":401,"message":"Invalid Credentials","errors":[{"message":"Invalid Credentials","domain":"global","reason":"authError","locationType":"header","location":"Authorization"}]}}
	at akka.stream.alpakka.googlecloud.storage.impl.GCStorageStream$.$anonfun$optionUnmarshaller$5(GCStorageStream.scala:258)
	at scala.concurrent.Future.$anonfun$flatMap$1(Future.scala:256)
	at akka.http.scaladsl.util.FastFuture$.$anonfun$transformWith$1(FastFuture.scala:36)
	at akka.http.scaladsl.util.FastFuture$.strictTransform$1(FastFuture.scala:40)
	at akka.http.scaladsl.util.FastFuture$.transformWith$extension(FastFuture.scala:44)
[...]

PR #2794, opened 2021-12-17
pass correct GoogleAttributes in rewrite flow

brunoballekens:
References #2753 
Th rewrite creates it's own internal streams without setting the correct attribute


ennru (approved):
LGTM.

ennru:
Thank you!

Issue #2816, opened 2022-02-02
FTP: include in CI workflow


ennru:
### Expected Behavior

FTP tests run in the GitHub actions workflow as all other Alpakka modules.

### Actual Behavior

FTP is excluded as paths into the FTP docker instances don't work out in the GH environment

Related to #1810

ennru:
Just as in #1810 the user UID in the FTP container must match the user UID running sbt so that it is allowed to delete those files.

Updating the `ftp` container config for Docker Compose helps:
```
      FTP_USER_UID: 1001
      FTP_USER_GID: 121
```

ennru:
Let's combine https://github.com/actions/checkout/issues/211#issuecomment-680125896 and https://docs.docker.com/compose/environment-variables/ to solve it.

ennru:
Fixed with #2825

PR #2821, opened 2022-02-05
ftp: investigate directories for GH

ennru:
References #2816


ennru:
Conclusion in https://github.com/akka/alpakka/issues/2816#issuecomment-1030807102

PR #2822, opened 2022-02-05
build: check if build needed earlier

ennru:
Avoid steps altogether if no changes are detected.

### DIFF
--- a/scripts/gh-detect-changes.sh
+++ b/scripts/gh-detect-changes.sh
@@ -3,8 +3,6 @@
 set -x
 
 DIR=$1
-PRE_CMD=$2
-CMD=$3
 
 if [ "$GITHUB_EVENT_NAME" == "pull_request" ]
 then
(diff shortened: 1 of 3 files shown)


scripts/gh-detect-changes.sh
```diff
@@ -0,0 +1,33 @@
...
+fi
+
+git diff "$COMPARE_TO" --exit-code --quiet "${CONNECTOR}" build.sbt project/
+# git diff "$COMPARE_TO" --exit-code --quiet "${CONNECTOR}" build.sbt project/ .github/workflows/
+DIFF_EXIT_CODE=$?
+
+if [ "$GITHUB_EVENT_NAME" == "schedule" ]
+then
+  echo "Building everything as it is a scheduled build"
+  echo "execute_build=true" >> $GITHUB_ENV
```
johanandren, line 25: Will this really work?

From the gh actions docs on the $GITHUB_ENV file:
> This file is unique to the current step and changes for each step in a job

So it isn't actually used in the following steps?

johanandren: Isn't `set-output` the tool to use for passing data between steps? https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

ennru: This works as the tiny change in CSV shows (all other connectors builds were short-circuited).

Using outputs is more heavy-weight and allows for passing values across jobs.

ennru: Here are the docs: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

johanandren: Thanks for clarifying, all good then 👍

johanandren (approved):

octonato (approved):
LGTM, seems to work pretty well. 
Any reason for a draft?

jrudolph:
Btw. there's also https://github.com/dorny/paths-filter for GHA but maybe not necessary to add a third-party plugin if we have our own script working.

For whole workflows there's also the built-in possibility to add a path filter to the `on` directive in yaml.

Issue #1810, opened 2019-07-03
FTP: Tests fail locally


MarcinAman:
### Versions used 
Code: latest master from 03.07.2019

Env:
```
java -version
openjdk version "1.8.0_212"
OpenJDK Runtime Environment (build 1.8.0_212-8u212-b03-0ubuntu1.19.04.2-b03)
OpenJDK 64-Bit Server VM (build 25.212-b03, mixed mode)
```
```
uname -a
Linux 5.0.0-20-generic #21-Ubuntu SMP Mon Jun 24 09:32:09 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```
(but the same thing is on fedora 30)

### Issue:
After cloning the repo, setting up docker containers and running sbt majority of ftp tests fail.

Commands used:
```
docker-compose up ftp sftp

<in other tab>
sbt ftp/test
```

### Expected Behavior
Tests should pass

### Relevant logs
Run ends up with:

[logs](https://gist.github.com/MarcinAman/af4382a16876ccd5a0d2802f354d538f) or with [logs](https://gist.github.com/MarcinAman/da07893874e16003781a291c0967724c)

As far as i can tell second issue is permission related and running:
```
ls -lahR ftp/tmp
ftp/tmp:
total 12K
drwxr-xr-x 3 root   root   4,0K lip  3 18:07 .
drwxr-xr-x 5 woolfy woolfy 4,0K lip  3 18:12 ..
drwxr-xr-x 2   2000 root   4,0K lip  3 18:13 home

ftp/tmp/home:
total 24M
drwxr-xr-x 2 2000 root 4,0K lip  3 18:13 .
drwxr-xr-x 3 root root 4,0K lip  3 18:07 ..
-rw-r--r-- 1 2000 2000   25 lip  3 18:13 file.txt
-rw-r--r-- 1 2000 2000   50 lip  3 18:13 file.txt.gz
-rw-r--r-- 1 2000 2000 2,0M lip  3 18:09 sample_bigger_file_120000000
-rw-r--r-- 1 2000 2000 2,0M lip  3 18:12 sample_bigger_file_270000000
-rw-r--r-- 1 2000 2000 2,0M lip  3 18:09 sample_bigger_file_375000000
-rw-r--r-- 1 2000 2000 2,0M lip  3 18:13 sample_bigger_file_454000000
-rw-r--r-- 1 2000 2000 2,0M lip  3 18:12 sample_bigger_file_493000000
[...]

PR #2825, opened 2022-02-07
FTP: user ID for containers from environment

ennru:
Commit 1:
* Upgrade docker compose file version
* Shell script to start FTP containers (didn't get default values to work locally)

Refs #1810 and #2816 

Commit 2:
* Remove remaining Travis references


johanandren (approved):

ennru:
> I wonder if it only works accidentally because UIDs are numerical and the local user might have a standard UID that may by accident also be present in the container?

That was exactly what broke it. It had hardcoded ids that worked on Travis, but not on GitHub CI. In #1810 users reported similar issues with running the FTP tests locally.

Issue #2185, opened 2020-03-10
Failed: HBase all tests failed


seglo:
All the HBase tests timed out on this build.  Attached the log so the stage could be retried and not wipe out build info.

[log.txt](https://github.com/akka/alpakka/files/4313852/log.txt)

seglo:
Happening in master too.

https://travis-ci.com/github/akka/alpakka/jobs/296475571

Build log has an error:

```
2020-03-10 13:52:34,501 DEBUG [pool-1-thread-1] [org.apache.hadoop.util.Shell]  Failed to detect a valid hadoop home directory
java.io.IOException: HADOOP_HOME or hadoop.home.dir are not set.
	at org.apache.hadoop.util.Shell.checkHadoopHome(Shell.java:329)
	at org.apache.hadoop.util.Shell.<clinit>(Shell.java:354)
	at org.apache.hadoop.util.StringUtils.<clinit>(StringUtils.java:80)
	at org.apache.hadoop.conf.Configuration.getBoolean(Configuration.java:1437)
	at org.apache.hadoop.hbase.HBaseConfiguration.checkDefaultsVersion(HBaseConfiguration.java:67)
	at org.apache.hadoop.hbase.HBaseConfiguration.addHbaseResources(HBaseConfiguration.java:81)
	at org.apache.hadoop.hbase.HBaseConfiguration.create(HBaseConfiguration.java:96)
	at docs.scaladsl.HBaseStageSpec.<init>(HBaseStageSpec.scala:102)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.lang.Class.newInstance(Class.java:442)
	at org.scalatest.tools.Framework$ScalaTestTask.execute(Framework.scala:450)
	at sbt.ForkMain$Run.lambda$runTest$1(ForkMain.java:304)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
[...]

PR #2826, opened 2022-02-08
hbase: re-enable

ennru:
References #2185


ennru:
This is not enough to fix the HBase connector.

PR #2829, opened 2022-02-11
Pravega: Renaming builder method

cheleb:
References #xxxx

### DIFF
--- a/pravega/src/main/scala/akka/stream/alpakka/pravega/PravegaSettings.scala
+++ b/pravega/src/main/scala/akka/stream/alpakka/pravega/PravegaSettings.scala
@@ -348,6 +348,9 @@ class TableReaderSettingsBuilder[K, V](
   /**
   Build the settings.
    */
+  @deprecated("Use withKeyExtractor instead", "4.0.0")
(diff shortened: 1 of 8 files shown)


pravega/src/main/scala/akka/stream/alpakka/pravega/PravegaSettings.scala
```diff
@@ -479,7 +482,7 @@ class TableWriterSettingsBuilder[K, V](
   /**
     Build the settings.
    */
-  def withSerializers(
+  def withKeyExtractor(
       extractor: K => TableKey
   ): TableWriterSettingsBuilder[K, V] = copy(tableKeyExtractor = Some(extractor))
```
ennru, line 487: Something is off here. The comment should go on the `build()` method.
`withSerializers` seems to have been the wrong name, right, that's why you just rename it now?

cheleb: Or just be removed !

ennru (approved):
LGTM.

ennru:
Thank you!

PR #2830, opened 2022-02-17
json streaming: test for multiple top level array's

mdedetrich:
This PR adds a test to verify if unwrapping a JSON array from a top level element works if there are multiple top level JSON array's. The context of the test is here https://discuss.lightbend.com/t/working-with-eof-for-a-source-flow/9481

I think adding this test would be a good idea to verify such behaviour


ennru (approved):
LGTM.

ennru:
Thank you!

PR #2820, opened 2022-02-03
Use akka-http 10.2.9

ennru:
Rebased @mrooding's PR #2650 to only use the current Akka HTTP version.
Will need a few more changes to avoid deprecated APIs.

Supersedes #2650

PR #2833, opened 2022-03-01
Use Akka HTTP 10.2.9

jrudolph:
On top of #2820 to check if latest snapshot works (in preparation of 10.2.9).

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -17,7 +17,7 @@ object Dependencies {
   val AwsSpiAkkaHttpVersion = "0.0.11"
   // Sync with plugins.sbt
   val AkkaGrpcBinaryVersion = "1.0"
-  val AkkaHttpVersion = "10.2.7"
+  val AkkaHttpVersion = "10.2.9"
   val AkkaHttpBinaryVersion = "10.2"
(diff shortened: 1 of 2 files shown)


jrudolph:
@ennru this is a fix against your PR to use 10.2.9 which fixes the SSE issues.

jrudolph:
@ennru this is a fix against your PR to use 10.2.9 which fixes the SSE issues.

jrudolph:
Interesting, the docs problem seems to be that the akka-http apidocs are now missing a lot of classes. Probably not critical but I'll look into what's happened there.

jrudolph:
> Interesting, the docs problem seems to be that the akka-http apidocs are now missing a lot of classes. Probably not critical but I'll look into what's happened there.

Ok, I cannot find those apidocs in older versions as well, so maybe we just shouldn't link to internal classes.

jrudolph:
> Ok, I cannot find those apidocs in older versions as well, so maybe we just shouldn't link to internal classes.

I see, 10.2.7 links were ignored in the link checker. Updated for 10.2.9

ennru (approved):
LGTM.

Issue #2809, opened 2022-01-22
Support MQTT 5.0 for the streaming connector


JustinPihony:
Add support for MQTT 5.0 akin to #2622 but for the streaming connector.

ennru:
[MQTT Version 5.0 - What's New?](https://www.cwnp.com/mqtt-version-5/)

PR #2836, opened 2022-03-06
mqtt-streaming: protocol version 5 support (#2809)

tilmann-bartsch:
I'm planning to add MQTT v5 support to the mqtt-streaming subproject in the next weeks.

I currently implemented most of the neccessary encoding/decoding in the file model.scala and updated the corresponding tests.

Encoding/Decoding:

* Updated file model.scala to allow mqttv311 and mqttv3 encoding/decoding
* Tests in MqttCodecSpec.scala updated for both variants
* Encoding/Decoding for Packets PUBREC, PUBREL, PUBCOMP, UNSUBSCRIBE, UNSUBACK not updated yet

ClientSessionFlow
* Small changes to work with updated encoding/decoding

References #2809


ennru:
Hi @tilmann-bartsch 
Thank you for jumping on this. Did you get any further already?

Issue #2841, opened 2022-03-22
Use testcontainers-scala-scalatest for tests requiring docker


mdedetrich:
### Short description

Currently for the integration tests we are using docker to spin up the relevant service. The current approach to do this is to directly start docker using github actions (i.e. https://github.com/akka/alpakka/blob/master/.github/workflows/check-build-test.yml#L80-L127).  While as an initial implementation this is fine, it has the following issues

* It makes it less convenient for users to test alpakka locally since the test won't automatically start docker. This means you have to manually manage starting/stopping the docker container yourself (especially if you want to wipe state for a clean run)
* It makes the github workflow actions build quite complex. Ideally in the future I want to explore using https://github.com/djspiewak/sbt-github-actions to generate a github workflow directly from `build.sbt`. This would greatly reduce maintenance burden of alpakka and the integration of the tool is much easier if it doesn't have to worry about spinning up docker containers. Note that I have already integrated sbt-github-actions into other lightbend/akka projects and also other OS projects without any problems, for example it has solved the `target`/coursier caching issue completely.
* Greatly improve performance of tests since the docker containers will only be running when the specific test is running it rather than just globally for the entire test (i.e. you don't need a docker container running if you are just doing integration tests)
  * In turn this makes it very clear in the code which tests actually need a service running and which tests don't since tests that need a docker service will have a specific trait mixed in.

In short the idea is to remove the spinning up of the docker containers in the github workflow and instead use https://github.com/testcontainers/testcontainers-scala.

### Details

As a preface one thing to note is that [testcontainers-scala](https://github.com/testcontainers/testcontainers-scala) is that its a wrapper over testcontainers which is a very popular and highly maintained java library (see https://github.com/testcontainers/testcontainers-java/ and https://www.testcontainers.org/). In other words there shouldn't be any qualms about maintenance/stability here.

[...]

PR #2842, opened 2022-03-22
AWS S3: Use testcontainer-scala-scalatest to setup Minio docker

mdedetrich:
Replaces manual starting of minio docker containers in github actions/docker-compose with https://github.com/testcontainers/testcontainers-scala. The changes are a bit more involved than originally anticipated mainly as a result of a race condition between Actor initialization of a typesafe `Config` and testcontainers starting (basically the issue is that when you do `ActorSystem(config)` this will evaluate immediately when the test initializes which is an issue since that is before the `beforeAll` of scalatest (and hence before container actually starts via testcontainers-scala-scalatest). Note that the usage of `lazy val` is also due to this, i.e. we only want to evaluate settings when needed (which is when the tests are actually run and not immediately)

The solution was rather than using typesafe config to configure the S3 client in tests, we instead use `S3Settings` which along with the usage of `S3Attributes.settings` allows us to evaluate settings whenever we want. The usage of `S3Settings` directly has also provided other benefits.

* Boilerplate was removed from `MinioS3IntegrationSpec`since everything can now easily be configured via the usage of builder pattern on `S3Settings()`. A lot of methods which needed to be overridden in `MinioS3IntegrationSpec` now no longer need to be done so.
* We can now use `TestKit` to setup `ActorSystem` for testing which we couldn't before since we were passing non-static configuration to `ActorSystem`.

I have also verified that the `AWSIntegrationSpec` still works by adding the settings in `s3/src/test/resources/application.conf` as documented.

References: https://github.com/akka/alpakka/issues/2841

### DIFF
--- a/s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3IntegrationSpec.scala
+++ b/s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3IntegrationSpec.scala
@@ -5,18 +5,17 @@
 package akka.stream.alpakka.s3.scaladsl
 
 import akka.actor.ActorSystem
-import akka.http.scaladsl.model.{ContentTypes, StatusCodes}
 import akka.http.scaladsl.Http
+import akka.http.scaladsl.model.{ContentTypes, StatusCodes}
 import akka.stream.{Attributes, KillSwitches, SharedKillSwitch}
-import akka.stream.alpakka.s3.AccessStyle.PathAccessStyle
+import akka.stream.alpakka.s3.AccessStyle.{PathAccessStyle, VirtualHostAccessStyle}
 import akka.stream.alpakka.s3.BucketAccess.{AccessGranted, NotExists}
 import akka.stream.alpakka.s3._
 import akka.stream.alpakka.testkit.scaladsl.LogCapturing
 import akka.stream.scaladsl.{Keep, Sink, Source}
-import akka.testkit.TestKit
+import akka.testkit.{TestKit, TestKitBase}
 import akka.util.ByteString
 import akka.{Done, NotUsed}
-import com.typesafe.config.ConfigFactory
 import org.scalatest.Inspectors.forEvery
 import org.scalatest._
 import org.scalatest.concurrent.ScalaFutures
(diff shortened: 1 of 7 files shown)


s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3IntegrationSpec.scala
```diff
@@ -1040,75 +1045,59 @@ trait S3IntegrationSpec
...
+        system,
+        attributes
+      )
+    } yield ()
+    Await.result(makeBuckets, 10.seconds)
   }
 
-  it should "properly set the endpointUrl" in {
-    S3Settings().endpointUrl.value shouldEqual endpointUrlVirtualHostStyle
+  it should "properly set the endpointUrl using VirtualHostAccessStyle" in {
```
mdedetrich, line 1096: I renamed the test because it was unclear what it was actually testing

s3/src/test/scala/akka/stream/alpakka/s3/scaladsl/S3IntegrationSpec.scala
```diff
@@ -1040,75 +1045,59 @@ trait S3IntegrationSpec
...
-        )
+  override lazy val defaultS3Settings: S3Settings = s3Settings
+    .withS3RegionProvider(
+      new AwsRegionProvider {
+        val getRegion: Region = Region.US_EAST_1
+      }
     )
+    .withAccessStyle(PathAccessStyle)
+
+  override def beforeAll(): Unit = {
```
mdedetrich, line 1070: This is a replacement for what was previously in the `docker-compose.yml` `minio_prep` section. We can just programatically create the S3 buckets needed for the tests in the standard scalatest `beforeAll`

mdedetrich:
Test failures are false positive, there was a previous run which passed without any issues (see https://github.com/akka/alpakka/actions/runs/2025105650) and the only difference was the formatting of a docker command.

mdedetrich:
I just rebased the PR but the only changes was adding back in the

```
 * Run the tests from inside sbt:
 * s3/testOnly *.MinioS3IntegrationSpec
 ```
 
 documentation at `MinioS3IntegrationSpec` which I accidentally removed.

PR #2855, opened 2022-04-15
Use testcontainers to setup Rabbitmq docker

mdedetrich:
Replaces manual starting of rabbitmq docker containers in github actions/docker-compose with https://github.com/testcontainers/testcontainers-scala.

Ideally I would have liked to used testcontainers in the idiomatic way (i.e. each test file starting its own docker container with a dynamic IP allowing you to have multiple rabbitmq instances on a single machine). Instead we are forced to use a `FixedPortContainer` that binds a docker container to a static IP (which means you can only have global instance of rabbitmq running), for 2 main reasons

* There appears to be unusual behavior in testing for authentication failed, i.e. https://github.com/akka/alpakka/blob/81580cbc4de9a5f97f6541a0096392e80d9c126f/amqp/src/test/scala/akka/stream/alpakka/amqp/scaladsl/AmqpConnectorsSpec.scala#L52-L70 and https://github.com/akka/alpakka/blob/81580cbc4de9a5f97f6541a0096392e80d9c126f/amqp/src/test/java/akka/stream/alpakka/amqp/javadsl/AmqpConnectorsTest.java#L92-L120. Basically from what I can tell, when using a Rabbitmq docker container its impossible to test this functionality if rabbitmq is not running on `localhost:5672` because you get a different exception (i.e. `ConnectionRefused` instead of `AuthenticationException`). See https://groups.google.com/g/rabbitmq-users/c/IrcosevMexs for more details

* We have specific tests usually called `LocalAmqpConnection`* which tests the `AmqpLocalConnectionProvider.getInstance()` method. This method creates an Amqp connection provider bound to localhost:5672 which means the only way to test this functionality is to have rabbitmq running on a fixed port (5672)

This means that even if we managed to solve the first point, in order to test `AmqpLocalConnectionProvider.getInstance()` we need a rabbitmq with a fixed port anyways (either that or we would have to disable the tests). There is an argument though that these tests are over compensating, i.e. if you want to test that `AmqpLocalConnectionProvider.getInstance()` points to `localhost:5672` you can just write a unit test that confirms the connection details.

Regarding not stopping the testcontainers docker-container, you don't need to do this since its already handled by https://github.com/testcontainers/moby-ryuk which will automatically close all created docker containers when the JVM process exits

### DIFF
--- a/.github/workflows/check-build-test.yml
+++ b/.github/workflows/check-build-test.yml
@@ -77,7 +77,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - { connector: amqp,                         pre_cmd: 'docker-compose up -d amqp' }
+          - { connector: amqp }
           - { connector: avroparquet }
           - { connector: awslambda }
           - { connector: aws-event-bridge,             pre_cmd: 'docker-compose up -d amazoneventbridge' }
--- a/amqp/src/test/java/akka/stream/alpakka/amqp/javadsl/AmqpConnectionProvidersTest.java
+++ b/amqp/src/test/java/akka/stream/alpakka/amqp/javadsl/AmqpConnectionProvidersTest.java
@@ -20,8 +20,7 @@ import org.junit.Test;
 
 import java.net.ConnectException;
 
-public class AmqpConnectionProvidersTest {
-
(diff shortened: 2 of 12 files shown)


mdedetrich:
Unrelated tests are failing

PR #2857, opened 2022-04-15
Use testcontainers to setup Mongodb docker

mdedetrich:
Replaces manual starting of mongodb docker containers in github actions/docker-compose with https://github.com/testcontainers/testcontainers-scala.

The only notable thing about the implementation in this PR is that due to the fact that some tests use paradox snippets this will mean some documentation will change, i.e. instead of saying

```scala
private val client = MongoClients.create(s"mongodb://localhost:27017")
```

it will instead say

```scala
private lazy val client = MongoClients.create(connectionString)
```

The `lazy val`s are required because otherwise they will initialize before the container is even started in testcontainers and the `connectionString`/`getConnectionString()` is required because testcontainers doesn't start docker with a global fixed port of 27017.

I am not sure if the change to the docs are a deal breaker but note that if you don't want these changes in the test I will have to use `FixedPortContainer` like in https://github.com/akka/alpakka/pull/2855 which does bring disadvantages, namely the fact that the container will run for the entire lifecycle of the JVM process rather than when only the tests are running (as they are now). This is not going to cause an issue now but as per https://github.com/akka/alpakka/issues/2841#issuecomment-1077380058 ideally once everything is migrated I would like to use a single runner for all of the tests which would only work well if each test starts and shuts down its own docker containers and doing this with both JUnit and ScalaTest tests is problematic.

### DIFF
--- a/docs/src/main/paradox/mongodb.md
+++ b/docs/src/main/paradox/mongodb.md
@@ -38,7 +38,8 @@ The table below shows direct dependencies of this module and the second tab show
 
 In the code examples below we will be using Mongo's support for automatic codec derivation for POJOs.
 For Scala we will be using a case class and a macro based codec derivation.
-For Java a codec for POJO is derived using reflection.
+For Java a codec for POJO is derived using reflection. We assume that you already have an 
+@apidoc[akka.actor.ActorSystem] set up.
 
 Scala
 : @@snip [snip](/mongodb/src/test/scala/docs/scaladsl/MongoSourceSpec.scala) { #pojo }
@@ -62,15 +63,6 @@ Scala
 
 Java
 : @@snip [snip](/mongodb/src/test/java/docs/javadsl/MongoSourceTest.java) { #init-connection }
-
-We will also need an @apidoc[akka.actor.ActorSystem].
-
-Scala
-: @@snip [snip](/mongodb/src/test/scala/docs/scaladsl/MongoSourceSpec.scala) { #init-system }
-
-Java
-: @@snip [snip](/mongodb/src/test/java/docs/javadsl/MongoSourceTest.java) { #init-system }
-
 
 ## Source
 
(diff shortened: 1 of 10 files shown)


mdedetrich:
Test is failing but its unrelated (due to Cassandra).

mongodb/src/test/scala/docs/scaladsl/MongoSourceSpec.scala
```diff
@@ -36,9 +37,6 @@ class MongoSourceSpec
   implicit val system = ActorSystem()
   // #init-system
 
-  override protected def beforeAll(): Unit =
```
mdedetrich, line 39: This is not required anymore since every test run creates a fresh docker instance with empty state.

mdedetrich:
So I just rebased the PR, I just changed it so the documentation generated with the paradox snippets is more appealing, i.e. rather than

```scala
private val client = MongoClients.create(s"mongodb://localhost:27017")
```

there is

```scala
private val client = MongoClients.create(ConnectionString)
```

and for Java

```java
client = MongoClients.create(CONNECTION_STRING);
```

docs/src/main/paradox/mongodb.md
```diff
@@ -38,7 +38,8 @@ The table below shows direct dependencies of this module and the second tab show
 
 In the code examples below we will be using Mongo's support for automatic codec derivation for POJOs.
 For Scala we will be using a case class and a macro based codec derivation.
-For Java a codec for POJO is derived using reflection.
+For Java a codec for POJO is derived using reflection. We assume that you already have an
```
mdedetrich, line 41: Due to us no longer manually doing `val actorSystem = ActorSystem.create()` in test code since we get `System` from `TestKit` I have updated the doc's here to mention you need an `ActorSystem`.

mdedetrich, line 41: Due to us no longer manually doing `val actorSystem = ActorSystem.create()` in test code since we get `System` from `TestKit` I have updated the doc's here to mention you need an `ActorSystem`.

mdedetrich:

```
 ```

PR #2858, opened 2022-04-15
Use testcontainers to setup InfluxDb docker

mdedetrich:
Replaces manual starting of mongodb docker containers in github actions/docker-compose with https://github.com/testcontainers/testcontainers-scala.

The only thing of note is there are some slight changes in the configuration of the paradox snippets, i.e. `setupConnection` was moved around but this shouldn't effect the generated documentation since its still within the `#init-client` snippet, i would however double check the generated documentation is the same.


mdedetrich:

```
 ```

PR #2859, opened 2022-04-16
Use testcontainers to setup OrientDb docker

mdedetrich:
Replaces manual starting of orientdb docker containers in github actions/docker-compose with https://github.com/testcontainers/testcontainers-scala.

Similar to the other PR's the only notable thing here is that the generated paradox documentation changes slightly, i.e. instead of having

```java
private static String url = "remote:127.0.0.1:2424/";
```
In the `#init-settings` its now removed due to the fact that the url is no longer static but retrieved from the newly created docker container. Instead the example code references `ORIENTDB_URL`/`OrientDbUrl` constant which in a self explanatory manner describes the same thing.

### DIFF
--- a/orientdb/src/test/java/docs/javadsl/OrientDbBaseTest.java
+++ b/orientdb/src/test/java/docs/javadsl/OrientDbBaseTest.java
@@ -42,22 +42,13 @@ import java.util.stream.Collectors;
 
 import static org.junit.Assert.assertEquals;
 
-public class OrientDbTest {
+public class OrientDbBaseTest extends OrientDbJUnitTest {
   @Rule public final LogCapturingJunit4 logCapturing = new LogCapturingJunit4();
 
   private static OServerAdmin oServerAdmin;
   private static OPartitionedDatabasePool oDatabase;
   private static ODatabaseDocumentTx client;
   private static ActorSystem system;
-
-  // #init-settings
-
-  private static String url = "remote:127.0.0.1:2424/";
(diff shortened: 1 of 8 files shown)


orientdb/src/test/java/docs/javadsl/OrientDbBaseTest.java
```diff
@@ -176,9 +172,6 @@ public static void tearDown() throws Exception {
     unregister(sink3);
     unregister(sink6);
 
-    if (oServerAdmin.existsDatabase(dbName, "plocal")) {
```
mdedetrich, line 179: No longer required since every test run has a new clean slate from Docker

orientdb/src/test/scala/docs/scaladsl/OrientDbBaseSpec.scala
```diff
@@ -91,19 +93,19 @@ class OrientDbSpec extends AnyWordSpec with Matchers with BeforeAndAfterAll with
...
   }
 
-  override def afterAll() = {
+  override def beforeStop(): Unit = {
     unregister(sourceClass)
     unregister(sink4)
     unregister(sink5)
     unregister(sink7)
 
-    if (oServerAdmin.existsDatabase(dbName, "plocal")) {
```
mdedetrich, line 100: No longer required since every test run has a new clean slate from Docker

mdedetrich:

```
 ```

PR #2873, opened 2022-06-17
google-cloud-pub-sub-grpc: more imports in sample code

raboof:
especially ByteString is useful to include explicitly, not to confuse
it with `akka.util.ByteString`


ennru (approved):
LGTM.

Issue #2866, opened 2022-05-19
Memory leak due to stream leak with S3 paginated calls when recovering from errors


mdedetrich:
### Versions used 

Akka version: 2.6.19
Alpakka version: 3.0.4
Akka-Http version: 10.2.9

### Expected Behavior

When recovering from an Alpakka S3 calls (such as `S3.listMultipartUpload`) either by using `recover`/`recoverWith` on the materialized `Future` or directly within the stream itself by using `recover` that it works as expected.

### Actual Behavior

For starters, a lot of warnings are logged from the `WaitingForResponseEntitySubscription`, i.e.

```
[WARN ] 15:43:22.956 PoolId - [2 (WaitingForResponseEntitySubscription)]Response entity was not subscribed after 1 second. Make sure to read the response `entity` body or call `entity.discardBytes()` on it -- in case you deal with `HttpResponse`, use the shortcut `response.discardEntityBytes()`. GET /2022-02-04T06:38:00Z.json Empty -> 200 OK Default(64185 bytes)
```

Note that even if you increase the polling timeout using `akka.http.host-connection-pool.response-entity-subscription-timeout` from https://github.com/akka/akka-http/issues/1836#issuecomment-370807885 to something like 10 seconds (just to see if there is some buffering of open requests going on) it just delays the leak, not eliminate it.

This can be confirmed by running an application that continuously calls `S3.listMultipartUpload` when recovering from errors and looking  at the memory usage which continuously grows over time.

### Relevant logs

This is just a sample of the logs showing the `WaitingForResponseEntitySubscription` warnings from the project behind the linked issue.

```
[WARN ] 16:04:55.096 PoolId - [2 (WaitingForResponseEntitySubscription)]Response entity was not subscribed after 10 seconds. Make sure to read the response `entity` body or call `entity.discardBytes()` on it -- in case you deal with `HttpResponse`, use the shortcut `response.discardEntityBytes()`. GET /2022-02-04T12:52:00Z.json Empty -> 200 OK Default(66771 bytes)
[...]

PR #2874, opened 2022-06-19
Add S3.getObject and deprecate S3.download

mdedetrich:
PR adds in the `S3.getObject` method which is an alternative to `S3.download` that has a far simpler/idiomatic API (returns a `Source[ByteString, Future[ObjectMetadata]]` instead of a `Source[Option[(Source[ByteString, NotUsed], ObjectMetadata)], NotUsed]` as well as deprecating `S3.download`. Since no `Option` is returned in `S3.getObject` there is no chance of accidentally creating a leak (with `S3.download` if it returned a `Some` but you didn't consume the underlying `ByteString` `Source` a leak would be created, see referenced issue).

Nothing that fancy in the actual implementation of `Se.getObject`, most changes were mechanical. The materialized value `ObjectMetdata` is propagated by using a `Promise` and all of the tests have been updated to use `S3.getObject`. The old `S3.download` remains completely untouched (apart from depreciation annotation).

One thing to note of is that the documentation generated from the paradox code snippers has changed. While the Scala version essentially the same one of the Java code examples has been changed but for the better since the code sample no longer has blocking `.get` code and instead uses `.thenApply` (which is honestly much better since we shouldn't ever be advocating/documenting blocking code).

References #2866

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
@@ -212,12 +212,11 @@ import scala.util.{Failure, Success, Try}
         issueRequest(s3Location, rangeOption = range, versionId = versionId, s3Headers = headers)(mat, attr)
           .map(response => response.withEntity(response.entity.withoutSizeLimit))
           .mapAsync(parallelism = 1)(entityForSuccess)
-          .map {
+          .flatMapConcat {
             case (entity, headers) =>
               objectMetadataMat.success(computeMetaData(headers, entity))
               entity.dataBytes
           }
-          .flatMapConcat(identity)
           .mapMaterializedValue(_ => objectMetadataMat.future)
       }
       .mapMaterializedValue(_.flatMap(identity)(ExecutionContexts.parasitic))
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
@@ -27,8 +27,7 @@ import akka.{Done, NotUsed}
 import software.amazon.awssdk.regions.Region
 
 import scala.collection.immutable
-import scala.collection.immutable.Seq
-import scala.concurrent.Future
+import scala.concurrent.{Future, Promise}
 import scala.util.{Failure, Success, Try}
 
 /** Internal Api */
@@ -197,6 +196,32 @@ import scala.util.{Failure, Success, Try}
           }
       }
   }.mapMaterializedValue(_ => NotUsed)
+
+  def getObject(
+      s3Location: S3Location,
+      range: Option[ByteRange],
+      versionId: Option[String],
+      s3Headers: S3Headers
+  ): Source[ByteString, Future[ObjectMetadata]] = {
+    val headers = s3Headers.headersFor(GetObject)
+    val objectMetadataMat = Promise[ObjectMetadata]()
+
+    Source
+      .fromMaterializer { (mat, attr) =>
+        implicit val materializer: Materializer = mat
+        issueRequest(s3Location, rangeOption = range, versionId = versionId, s3Headers = headers)(mat, attr)
+          .map(response => response.withEntity(response.entity.withoutSizeLimit))
+          .mapAsync(parallelism = 1)(entityForSuccess)
+          .map {
+            case (entity, headers) =>
+              objectMetadataMat.success(computeMetaData(headers, entity))
+              entity.dataBytes
+          }
+          .flatMapConcat(identity)
+          .mapMaterializedValue(_ => objectMetadataMat.future)
+      }
+      .mapMaterializedValue(_.flatMap(identity)(ExecutionContexts.parasitic))
+  }
 
   /**
    * An ADT that represents the current state of pagination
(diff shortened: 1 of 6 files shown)


s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -198,6 +197,32 @@ import scala.util.{Failure, Success, Try}
...
+
+    Source
+      .fromMaterializer { (mat, attr) =>
+        implicit val materializer: Materializer = mat
+        issueRequest(s3Location, rangeOption = range, versionId = versionId, s3Headers = headers)(mat, attr)
+          .map(response => response.withEntity(response.entity.withoutSizeLimit))
+          .mapAsync(parallelism = 1)(entityForSuccess)
+          .map {
+            case (entity, headers) =>
+              objectMetadataMat.success(computeMetaData(headers, entity))
```
mdedetrich, line 217: My only slight concern is when this gets evaluated to complete the `objectMetadataMat` `Promise` due to the lazy nature of akka-streams.

This can be problematic because ideally you want the materialized value to complete ASAP (i.e. as soon as you get the response headers) so that you can fullfill the usecase of reusing the `Source[ByteString,_]` in other stream computations as soon as the `Future[ObjectMetadata]` gets completed (and you **don't** want this to occur when the body has completed streaming)

s3/src/test/scala/docs/scaladsl/S3SourceSpec.scala
```diff
@@ -167,29 +161,28 @@ class S3SourceSpec extends S3WireMockBase with S3ClientIntegrationSpec {
...
     mock404s()
 
     val download = S3
-      .download("nonexisting-bucket", "nonexisting_file.xml")
+      .getObject("nonexisting-bucket", "nonexisting_file.xml")
       .runWith(Sink.head)
-      .futureValue
 
-    download shouldBe None
+    download.failed.futureValue should matchPattern {
```
mdedetrich, line 183: So even though it can be argued that this test is not necessary anymore since we aren't wrapping the 404 case into an `Option`, I rewrote it to check for the same exception just to make sure its behaving the same way.

ennru (approved):
So much simpler!
S3 is was one of the first Alpakka connectors so I think we have learned a lot since than.
LGTM.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -198,6 +197,32 @@ import scala.util.{Failure, Success, Try}
...
+              entity.dataBytes
+          }
```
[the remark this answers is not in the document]
mdedetrich: Done

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -198,6 +197,32 @@ import scala.util.{Failure, Success, Try}
...
+          .flatMapConcat(identity)
+          .mapMaterializedValue(_ => objectMetadataMat.future)
+      }
+      .mapMaterializedValue(_.flatMap(identity)(ExecutionContexts.parasitic))
```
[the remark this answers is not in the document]
mdedetrich: Done, I used `mapError` to retrieve the error and `.failure` to complete the `Promise`. Let me know if this is appropriate.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -198,6 +197,32 @@ import scala.util.{Failure, Success, Try}
...
   }.mapMaterializedValue(_ => NotUsed)
 
+  def getObject(
+      s3Location: S3Location,
+      range: Option[ByteRange],
+      versionId: Option[String],
+      s3Headers: S3Headers
+  ): Source[ByteString, Future[ObjectMetadata]] = {
+    val headers = s3Headers.headersFor(GetObject)
+    val objectMetadataMat = Promise[ObjectMetadata]()
```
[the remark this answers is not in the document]
mdedetrich: Thanks, for the explanation. Done!

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
johanandren: Do you have a more concrete example where you suspect this would happen? It's not obvious to me there is any problem lurking here.

mdedetrich: What I mentioned earlier is just a hypothetical problem that popped in my head so I mentioned it but in reality it can be a complete non concern in which case you can just ignore it.

johanandren: Ok, in that case I'd expect that it is fine because when materializing the stream there will always be a downstream triggering demand, leading to the request coming in here and completing the promise, if a stream for some reason has backpressure but still want that promise to complete early they can `detach` to trigger the request early.

jroper:
While I can understand why this change was done, there are use cases for attempting to download a file if and only if it exists. How can we use Alpakka to do that without using the deprecated `download` method? It looks like I have to manually recover the source and check the exception thrown to see if the error code matches the string `NoSuchKey`. Is there an easier way than that?

PR #2879, opened 2022-06-29
Google Pub/Sub: Configurable host (for ordering)

LukBed:
If ordering is required on Pub/Sub then messages must be sent only to specific region (using regional Pub/Sub endpoint). If two messages with the same orderng key are sent to two different regions, then ordering is not guaranteed.

All quotes from Pub/Sub documentation.
https://cloud.google.com/pubsub/docs/ordering
https://cloud.google.com/pubsub/docs/resource-location-restriction

> If messages have the same ordering key and are in the same region, you can enable message ordering and receive the messages in the order that the Pub/Sub service receives them.

It's possible to restrict region on topic level

> If you publish messages to the global Pub/Sub endpoint, Pub/Sub automatically stores the messages in the nearest Google Cloud region. Pub/Sub's topic message storage policy offers a way to ensure that messages published to a topic are never persisted outside a set of Google Cloud regions you specify, regardless of where the publish requests originate. When multiple regions are allowed by the policy, Pub/Sub chooses the nearest allowed region.

But:
> If you publish a message with an ordering key and the message storage policy excludes the nearest region, the Pub/Sub service returns an error.

So the best solution is specify region when message is published.

PR constains proposition with extending publish flow, by new optional parameter with Pub/Sub host.

Example usage:
`GooglePubSub.publish(topic, config,  Some("europe-west1-pubsub.googleapis.com"));`

### DIFF
--- a/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/javadsl/GooglePubSub.scala
+++ b/google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/javadsl/GooglePubSub.scala
@@ -4,17 +4,16 @@
 
 package akka.stream.alpakka.googlecloud.pubsub.javadsl
 
-import java.util.concurrent.CompletionStage
 import akka.actor.Cancellable
 import akka.stream.alpakka.googlecloud.pubsub.scaladsl.{GooglePubSub => GPubSub}
 import akka.stream.alpakka.googlecloud.pubsub.{AcknowledgeRequest, PubSubConfig, PublishRequest, ReceivedMessage}
 import akka.stream.javadsl.{Flow, FlowWithContext, Sink, Source}
 import akka.{Done, NotUsed}
 
-import scala.jdk.CollectionConverters._
+import java.util.concurrent.CompletionStage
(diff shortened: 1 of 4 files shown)


google-cloud-pub-sub/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/javadsl/GooglePubSub.scala
```diff
@@ -23,15 +23,27 @@ object GooglePubSub {
 
   /**
    * Creates a flow to that publishes messages to a topic and emits the message ids.
+   * @param overrideHost if present publish message will be sent to specific host,
+   *                     can be used to send message to specific regional endpoint,
+   *                     which can be important when ordering is enabled
    */
   def publish(topic: String,
               config: PubSubConfig,
+              overrideHost: java.util.Optional[String],
```
ennru, line 32: Wouldn't `String` be enough? If you don't want to pass a host you'd use the other overload?

LukBed: Right, changed

ennru (approved):
LGTM.

Issue #2881, opened 2022-07-12
Google pubSub Grpc. Impl PublishWithContext and make subscriber and publisher accessible 


bc-tayvs:
## Publish with context

https://github.com/akka/alpakka/blob/d27748d1c1f29ae3b9e93f20d769109bd90bc60a/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/scaladsl/GooglePubSub.scala#L28

publish method is not enough when you need to propagate context. FlowWithContext variant also needed.

## Subscriber and publisher accessibillity 

When you need custom `publish` flow, subscriber and publisher will be very useful

https://github.com/akka/alpakka/blob/d27748d1c1f29ae3b9e93f20d769109bd90bc60a/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/scaladsl/GooglePubSub.scala#L130

https://github.com/akka/alpakka/blob/d27748d1c1f29ae3b9e93f20d769109bd90bc60a/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/scaladsl/GooglePubSub.scala#L136

### Details

I made PR with small changes
1. Introduce publishWithContext method
2. Made subscriber and publisher package private. In case when you need custom approach or just client

Explain your idea in more detail and possibly suggest different ways to achieve it.

PR #2882, opened 2022-07-12
Impl PublishWithContext and make subscriber and publisher accessible

bc-tayvs:
Made subscriber and publisher package private. In case when you need custom approach or just client

[Issue](https://github.com/akka/alpakka/issues/2881)

### DIFF
--- a/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/scaladsl/GooglePubSub.scala
+++ b/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/scaladsl/GooglePubSub.scala
@@ -5,14 +5,15 @@
 package akka.stream.alpakka.googlecloud.pubsub.grpc.scaladsl
 
 import akka.actor.Cancellable
+import akka.annotation.ApiMayChange
 import akka.dispatch.ExecutionContexts
 import akka.stream.{Attributes, Materializer}
-import akka.stream.scaladsl.{Flow, Keep, Sink, Source}
+import akka.stream.scaladsl.{Flow, FlowWithContext, Keep, Sink, Source}
 import akka.{Done, NotUsed}
 import com.google.pubsub.v1.pubsub._
 
 import scala.concurrent.duration._
-import scala.concurrent.{Future, Promise}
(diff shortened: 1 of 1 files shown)


google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/scaladsl/GooglePubSub.scala
```diff
@@ -127,15 +140,28 @@ object GooglePubSub {
       .mapMaterializedValue(_.flatMap(identity)(ExecutionContexts.parasitic))
   }
 
-  private def publisher(mat: Materializer, attr: Attributes) =
+  @ApiMayChange
```
patriknw, line 143: These are not supposed to be public user facing API? Then they should be marked with `@InternalApi` instead.

google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/scaladsl/GooglePubSub.scala
```diff
@@ -26,12 +27,24 @@ object GooglePubSub {
...
+      Flow[PublishRequest].mapAsyncUnordered(parallelism)(publisherClient.publish)
+    }
+
+  /**
+   * Create a flow with context to publish messages to Google Cloud Pub/Sub. The flow emits responses that contain published
+   * message ids and provided context.
+   *
+   * @param parallelism controls how many messages can be in-flight at any given time
+   */
+  def publishWithContext[Ctx](parallelism: Int): FlowWithContext[PublishRequest, Ctx, PublishResponse, Ctx, NotUsed] =
```
patriknw, line 40: no corresponding in javadsl?

patriknw (commented):
looks like a nice convenience

google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/scaladsl/GooglePubSub.scala
johanandren: I think they should be left private, or if I'm missing a use case, please describe it a bit.

PR #2894, opened 2022-08-28
Bump Slick version to 3.4.0

nafg:


nafg:
It doesn't seem like you have Scala Steward or an alternative so this has to be done manually ;)

ennru:
Thank you. Yes, with the myriad of dependencies we opt for manual management.

ennru:
CI is upset as logging is not set up correctly for Slick.
https://github.com/akka/alpakka/runs/8071882089?check_suite_focus=true#step:6:50

MancunianSam:
There are version conflicts with alpakka-slick and the new slick versions. alpakka-slick needs 3.3.3 and everything else depends on 3.4.0. There is a pull request https://github.com/akka/alpakka/pull/2894 open for that library to upgrade the version so it's worth waiting a couple of days to get that merged which should fix these problems.

ennru:
The issue seems to be caused by the `slf4j-api` upgrade from 1.7.30 to 2.0.0 which Slick 3.4.0 brings in.

patriknw:
It will be problematic if we bring in another (incompatible) version of slf4j than the other Akka libraries are using. Let's stay on the older Slick version.

nafg:
I may have seen the issue before, do you know why it's happening?

Also what's keeping you from bumping the version everywhere?

Anyway if you think it makes sense I could downgrade it back to 1.x and release 3.4.1

patriknw:
> what's keeping you from bumping the version everywhere?

We don't want to be the one that forces everyone to update their logging implementations (eg Logback) to newer versions, if that is what is required by slf4j 2.0. We'll be happy to jump on the train when/if it has become mainstream.

nafg:
Also what about just adding an exclusion to the Slick dependency?

patriknw:
> Also what about just adding an exclusion to the Slick dependency?

That could work, but it's likely that someone using this library is also using akka-persistence-jdbc, so good to use [the same version](https://github.com/akka/akka-persistence-jdbc/blob/master/project/Dependencies.scala#L14). We probably don't want to bump that right now.

If someone is in need of Slick 3.4.0 I assume they can just update the dependency in their application build.

patriknw:
Updated to 3.4.1 in https://github.com/akka/alpakka/pull/2909

PR #2895, opened 2022-08-30
Remove doc references to abandoned Azure libraries

leviramsey:
Our docs for Azure Event Hubs and Azure IoT Hub link to libraries which have been archived (no further upstream contributions accepted).  Since these libraries are built on Akka 2.4 or 2.5 (neither of which will be supported at some point in the near future if they're not already unsupported) and only built against Scala versions before 2.13 (2.12 support will likely be removed soon?), I don't think we can in good conscience suggest their use.

If there are active forks (I believe both are MIT licensed) that have received updates, we can instead recommend those (if anyone is aware of one).  Alternatively, if someone wants to implement/maintain an integration, at least a link in the docs could be incorporated into Alpakka.


johanandren (approved):

ennru (approved):
Thanks.

PR #2899, opened 2022-09-05
Add S3.listBuckets

mdedetrich:
This PR adds the AWS S3 list buckets call (see https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListBuckets.html). The call itself is trivial however there are some non conventional differences

* This seems to be one of the few S3 calls that returns a list of entities that is not paginated. Due to this its an open question whether the `S3.listBuckets` method should just return a `Future[Seq[ListBucketsResultContents]]` or the current version which is a `Source[ListBucketsResultContents, NotUsed]`. I opted for the `Source` variant due to consistency with the rest of the S3 API (every single other method in S3 that returns a list of entities is a `Source`) plus there is a chance that AWS may add a paginated `list-buckets` call in the future in which case its trivially easy to use the newly updated version. I don't feel strongly about it either way however, so if you want me to change it to `Future[Seq[ListBucketsResultContents]]` then just let me know
* The base construction of the URI is also very different to any other S3 call, hence why the `listBuckets` call in `HttpRequests.scala` is manually constructing the URI itself, i.e. there isn't any Path/Virtual access style and the authority is hardcoded to `s3.amazonaws.com` (for AWS) since the call is region agnostic. I have tested this both with Minio and a real AWS S3 account so I can confirm its working

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/scaladsl/S3.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/scaladsl/S3.scala
@@ -264,6 +264,25 @@ object S3 {
     S3Stream.getObject(S3Location(bucket, key), range, versionId, s3Headers)
 
   /**
+   * Will return a list containing all of the buckets for the current AWS account
+   *
+   * @see https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListBuckets.html
+   * @return [[akka.stream.scaladsl.Source Source]] of [[ListBucketsResultContents]]
+   */
+  def listBuckets(): Source[ListBucketsResultContents, NotUsed] =
+    listBuckets(S3Headers.empty)
+
+  /**
+   * Will return a list containing all of the buckets for the current AWS account
+   *
+   * @see https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListBuckets.html
+   * @param s3Headers any headers you want to add
+   * @return [[akka.stream.scaladsl.Source Source]] of [[ListBucketsResultContents]]
+   */
+  def listBuckets(s3Headers: S3Headers): Source[ListBucketsResultContents, NotUsed] =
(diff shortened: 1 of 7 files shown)


mdedetrich:
Test failure for `CI / ScalaDoc, Documentation with Paradox (pull_request)` is unrelated (link validator is failing).

s3/src/main/scala/akka/stream/alpakka/s3/scaladsl/S3.scala
```diff
@@ -263,6 +263,25 @@ object S3 {
   ): Source[ByteString, Future[ObjectMetadata]] =
     S3Stream.getObject(S3Location(bucket, key), range, versionId, s3Headers)
 
+  /**
+   * Will return a list containing all of the buckets for the current AWS account
+   *
+   * @see https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListBuckets.html
+   * @return [[akka.stream.scaladsl.Source Source]] of [[ListBucketsResultContents]]
+   */
+  def listBuckets(): Source[ListBucketsResultContents, NotUsed] =
```
patriknw, line 272: > This seems to be one of the few S3 calls that returns a list of entities that is not paginated (i.e. the list of entities is a XML node list in the response). Due to this its an open question whether the S3.listBuckets method should just return a Future[Seq[ListBucketsResultContents]] or the current version which is a Source[ListBucketsResultContents, NotUsed]. I opted for the Source variant due to consistency with the rest of the S3 API (every single other method in S3 that returns a list of entities is a Source) plus there is a chance that AWS may add a paginated list-buckets call in the future in which case its trivially easy to use the newly updated version. I don't feel strongly about it either way however, so if you want me to change it to Future[Seq[ListBucketsResultContents]] then just let me know

I agree, let's use `Source`

ennru (approved):
LGTM. Great work, as always.

PR #2901, opened 2022-09-06
Update release instructions

patriknw:


.github/workflows/publish.yml
```diff
@@ -7,8 +7,6 @@ on:
       - main
       - release-*
     tags: ["v*"]
-  release:
-    types: [published]
```
patriknw, line 11: As far as I can see, this is triggering the release twice. First from the tag and then when publishing the release.
The latter can be seen in action here https://github.com/akka/alpakka/actions/runs/2999580924

docs/release-train-issue-template.md
```diff
@@ -36,8 +36,8 @@ Key links:
 
 ### Cutting the release
 
-- [ ] Wait until the [build finished](https://github.com/akka/alpakka/actions) after merging the release notes
-- [ ] [Fix up the draft release](https://github.com/akka/alpakka/releases) with the next tag version `v$VERSION$`, title and release description linking to announcement and milestone
+- [ ] Wait until the [build finished](https://github.com/akka/alpakka/actions)
+- [ ] [Fix up the draft and Publish release](https://github.com/akka/alpakka/releases) with the next tag version `v$VERSION$`, title and release description linking to announcement and milestone
```
patriknw, line 40: I think it's easies to create the tag from the release page.

johanandren (approved):

ennru (approved):
LGTM.

PR #2777, opened 2021-11-19
bump: Scala 2.13.8 (drop 2.12), Akka 2.6.18

ennru:
Preparing for a major release of Alpakka to only publish Scala 2.13 artefacts.

- Use `@nowarn` instead of Silencer plugin
- Fix matches that were not exhaustive (some ugly fixes)
- Avoid Scala 2.13 warnings

PR #2912, opened 2022-09-20
Add back Scala 2.12

patriknw:
* because it has ripple effect to Alpakka and then Cassandra plugin
* we need to coordinate this decision across all Akka projects, later

This reverts some parts, as little as possible, of https://github.com/akka/alpakka/pull/2777


ennru (approved):
LGTM.
* the link validator failed
* I retriggered Cassandra, it is a flaky test.

PR #2916, opened 2022-09-26
Update Hadoop dependency to 3.3.3

patriknw:
* CVE-2021-37404

PR #2919, opened 2022-09-27
Update hadoop to 3.3.3

patriknw:
* in hdfs and hbase modules

Same version as Avro Parquet that was changed in https://github.com/akka/alpakka/pull/2916


patriknw:
hdfs ✅ https://github.com/akka/alpakka/actions/runs/3135633133/jobs/5091655904

hbase is disabled from CI, but it compiles so 🚢 it, https://github.com/akka/alpakka/issues/2185

johanandren (approved):

PR #2921, opened 2022-09-28
Update Akka to 2.7.0-M2

patriknw:
* Akka to 2.7.0-M2
* Akka HTTP 10.4.0-M1
* logback-classic 1.2.11
* Fix deprecation warnings of TestSource and TestSink


patriknw:
We should also update the Akka gRPC version but that breaks google-cloud-pub-sub-grpc and google-cloud-bigquery-storage somehow, so let's do that in separate PR.

patriknw:
This will have to wait until the version check is fixed.
```
Current version of Akka is [2.7.0-M2], but akka-http requires version [2.7.0-M1]
```

ennru (approved):
LGTM.

PR #2923, opened 2022-10-12
Cassandra: pick current Jackson version

ennru:
Ensure Alpakka Cassandra uses the current version of Jackson.
(java-driver-core 4.14.1 brings in 2.13.2)


johanandren (approved):

project/Dependencies.scala
```diff
@@ -108,7 +108,7 @@ object Dependencies {
           .exclude("org.apache.tinkerpop", "*") //https://github.com/akka/alpakka/issues/2200
           .exclude("com.esri.geometry", "esri-geometry-api"), //https://github.com/akka/alpakka/issues/2225
         "com.typesafe.akka" %% "akka-discovery" % AkkaVersion % Provided
-      )
+      ) ++ JacksonDatabindDependencies
```
patriknw, line 111: a comment would be good, because it's not obvious why we would add Jackson dependency here or that the driver has the Jackson dependency

PR #2928, opened 2022-10-14
ci: current Ubuntu versions

ennru:
Thanks to the brown-out by Github I noticed we had Ubuntu 18.04 still there.


patriknw (approved):
LGTM, probably needed in several other repos

PR #2947, opened 2022-12-21
build: Scala steward conventional commits

johanandren:


ennru:
Note that Scala Steward is currently not enabled for Alpakka.
If we decide to enable it, I'd recommend to set it to run only monthly as it will be overwhelming anyway.

johanandren:
I see, file was already there though. Feel free to close or merge, whichever is best.

ennru (approved):

PR #2951, opened 2023-01-30
bump(google-common): jwt-spray-json 9.0.2 (was 7.1.4)

ennru:
Even bumps
- google-auth-library-credentials 1.15.0 (was 0.24.1)

Even bumps test lib
- Mockito 2.8.1 (was 2.8.0)

PR #2953, opened 2023-02-02
bump: sync Google Auth libs upgrades (1.15.0)

ennru:
I noticed two libraries from the same project could get pulled into Google Pub/Sub gRPC with different versions.

Follow-up to #2951


franciscolopezsancho (approved):
LGTM

Issue #2906, opened 2022-09-16
AWS S3: uploadMultipart* with S3Headers fails


wi101:
### Versions used 

Alpakka version: 4.0.0
Akka version: 2.6.20

### Expected Behavior

`S3.multipartUpload` and `S3.multipartUploadWithHeaders` with S3Headers should succeed as before in v3.0.4

### Actual Behavior

When calling `S3.multipartUpload` or `S3.multipartUploadWithHeaders` with S3Headers there is an error.
```
akka.stream.alpakka.s3.S3Exception: The specified header is not valid in this context
```

When I tried to call `S3.multipartUploadWithHeaders` with an empty header it worked.
An assumption: this change in completionSink might be the reason: https://github.com/akka/alpakka/pull/2844

sikinmettugi:
Hi, I also faced the same issue after updating alpakka-s3 from 3.0.1 to 4.0.0. From my side the current dependency versions are:
akka 2.6.19
alpakka 4.0.0
akka-http 10.2.9 (if this helps)

hagerf:
Hi, just wanted to add that this is a breaking change also for us, we cannot use Alpakka 4.0.0 until this is fixed :(

mvillafuertem:
Hi there, I also faced the same issue. Any workaround?

cptwunderlich:
Alpakka version 5.0 is also affected.

gipeshka:
The issue is apparently caused by sending metadata/custom headers with the multipart upload completion request.

sebastian-alfers:
We published `5.0.0+17-5228417f-SNAPSHOT` in case anyone would like try out that [snapshot](https://doc.akka.io/docs/alpakka/current/other-docs/snapshots.html).

hagerf:
Thanks @sebastian-alfers.
Would be great to test, but since the final version is going to be released in version >5.0.0 which is covered by BSL (I presume), I don't think we'll be able to use it in our company sadly. Or will some patch of 4.0.0 include this?

sebastian-alfers:
I am sorry to hear that, but currently we do not plan to backport this fix.

PR #2844, opened 2022-03-23
AWS S3: Include custom headers in all multipart copy requests

jfromme:
This PR addresses the issue referenced below by ensuring that custom S3 headers passed to multipart copy are included when copying parts and in the final completion request. 

References #2839

PR #2956, opened 2023-02-07
Ensure to only pass sse headers to completeMultipartUpload

sebastian-alfers:
Not very explicit in the [docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html#API_CompleteMultipartUpload_Examples), but other headers beside the sse headers are not allowed for this particular request (`completeMultipartUpload`).

Fixes https://github.com/akka/alpakka/issues/2906 and reverts a small part from https://github.com/akka/alpakka/pull/2844.


patriknw (approved):
LGTM, but I have no knowledge about the details of the s3 api

PR #2957, opened 2023-02-22
Scala3 builds

sebastian-alfers:


johanandren (commented):
Do we need to add something in ci config to compile the supported modules with Scala 3 on PR/nightlies as well?

sebastian-alfers:
> Do we need to add something in ci config to compile the supported modules with Scala 3 on PR/nightlies as well?

I am not sure if we run nightlies, do we? At least I did not see it.

Re scala3 in PRs, I got the understanding that we only want scala3 (same for 2.12) to be used during cros publishing. At least this is how it is right now for the scla3 builds I was working on.

project/Dependencies.scala
```diff
@@ -7,7 +7,9 @@ object Dependencies {
 
   val Scala213 = "2.13.10" // update even in link-validator.conf
   val Scala212 = "2.12.17"
-  val ScalaVersions = Seq(Scala213, Scala212)
+  val Scala3 = "3.1.3"
+  val Scala2Versions = Seq(Scala213, Scala212)
```
patriknw, line 11: > Re scala3 in PRs, I got the understanding that we only want scala3 (same for 2.12) to be used during cros publishing. At least this is how it is right now for the scla3 builds I was working on.

We should at minimum compile with all Scala versions. We typically do that with `+Test/compile`. That should be changed here
https://github.com/akka/alpakka/blob/5228417f38211cb9bfc62bed4f1f7bb2b4aed0cb/.github/workflows/check-build-test.yml#L45

It's probably obsolete to use 2.13 there, since 2.13 is the default version.

Also, currently I think we run tests with the default 2.13, which is good, but it means that we currently don't build with 2.12 at all (unless I missed something, didn't look so close). Anyway, that is fixed by compiling with `+Test/compile`.

sebastian-alfers: Makes sense, will review.

sebastian-alfers:
Something is off with CI, it does not succeed and I can not cancel it...

johanandren:
Looks like that worked nicely (I see 3.1 in the `Compile, Code Style, Binary Compatibility` check), looks like the label/name of the job/step is still wrong though.

mongodb/src/test/scala/docs/scaladsl/MongoSinkSpec.scala
```diff
@@ -42,7 +43,7 @@ class MongoSinkSpec
   implicit val system = ActorSystem()
 
   override protected def beforeAll(): Unit =
-    Source.fromPublisher(db.drop()).runWith(Sink.headOption).futureValue
+    Source.fromPublisher(db.drop()).runWith(Sink.headOption).futureValue(defaultPatience, Position.here)
```
sebastian-alfers, line 46: Not sure why, but this failed 5 times in a row in CI, so...

johanandren: Hmm, isn't defaultPatience already what futureValue picks up implicitly?

csv/src/test/scala/docs/scaladsl/CsvToMapSpec.scala
```diff
@@ -460,5 +460,5 @@ class CsvToMapSpec extends CsvSpec {
       )
     )
     // #header-line
-  }
+  }(mat)
```
johanandren, line 463: Where does this come from and where does it go?

mongodb/src/test/scala/docs/scaladsl/MongoSinkSpec.scala
sebastian-alfers: It is, and I dont see why my changes could impact this. However, that place as consistently failing for a timeout until I added this: https://github.com/akka/alpakka/actions/runs/4253962480/jobs/7400425654.

Should I remove this explicit params again, and see what happens? This was also only showing up in CI.

csv/src/test/scala/docs/scaladsl/CsvToMapSpec.scala
sebastian-alfers: And it is required here: https://github.com/akka/akka/blob/5e9e11b5f6df03ed89c5ca37bdbb295ff5c19ec4/akka-stream-testkit/src/main/scala/akka/stream/testkit/scaladsl/StreamTestKit.scala#L28

johanandren: Hmm, but why isn't that available through implicit just for this last test case, it is for all the other test cases (via the actor system materializer) right?

mongodb/src/test/scala/docs/scaladsl/MongoSinkSpec.scala
johanandren: Maybe instead increase the default patience for this test up top and keep the implicit timeout pickup as it was?

Something like
```
override implicit val patienceConfig: PatienceConfig = PatienceConfig(10.seconds, 100.millis)
```

csv/src/test/scala/docs/scaladsl/CsvToMapSpec.scala
sebastian-alfers: It was failing at some point, but now it seems to work when removing the explicit param again. Thanks for being nit-picky here!

mongodb/src/test/scala/docs/scaladsl/MongoSinkSpec.scala
johanandren: Sorry, looks like the name to override here is `defaultPatience`

sebastian-alfers: 👍

johanandren (approved):
LGTM!

I tried out a +publishLocal to double check that package/docs work fine as well and that looks good.

Issue #2731, opened 2021-09-16
mqtt-streaming ControlPacketFlags.QoS* has incorrect meaning when subscribing


ohze:
The following commands:
+ `Command(Subscribe(Seq("topic" -> QoSAtLeastOnceDelivery))`  will subscribe with qos 2
+ `Command(Subscribe(Seq("topic" -> QoSExactlyOnceDelivery))`  will be encoded to an invalid mqtt package (so, mqtt server will disconnect the subscribing client)

### Versions used 

akka-stream-alpakka-mqtt-streaming 3.0.3  (akka 2.6.14)

### Expected Behavior
correct meaning

### Actual Behavior
incorrect meaning

### Relevant logs
```
Test for qos ControlPacketFlags(0)
SubAck(PacketId(1),Vector(ControlPacketFlags(0)))
Done 0

Test for qos ControlPacketFlags(2)
SubAck(PacketId(1),Vector(ControlPacketFlags(2)))
Done 2

Test for qos ControlPacketFlags(4)
java.lang.AssertionError: assertion failed
  scala.Predef$.assert(Predef.scala:264)
  ammonite.$file.s$.$anonfun$res_6$1(s.sc:26)
  ammonite.$file.s$.$anonfun$res_6$1$adapted(s.sc:9)
  scala.collection.immutable.List.foreach(List.scala:333)
  ammonite.$file.s$.<clinit>(s.sc:9)
```
mosquitto log:
```
1631793666: mosquitto version 2.0.12 starting
1631793666: Config loaded from /mosquitto/config/mosquitto.conf.
1631793666: Opening ipv4 listen socket on port 1883.
1631793666: Opening ipv6 listen socket on port 1883.
1631793666: mosquitto version 2.0.12 running
1631793674: New connection from 172.17.0.1:59778 on port 1883.
1631793674: New client connected from 172.17.0.1:59778 as 0 (p2, c1, k60).
1631793674: New connection from 172.17.0.1:59782 on port 1883.
1631793674: Client 0 closed its connection.
1631793674: New client connected from 172.17.0.1:59782 as 2 (p2, c1, k60).
1631793674: Client 2 closed its connection.
1631793674: New connection from 172.17.0.1:59786 on port 1883.
1631793674: New client connected from 172.17.0.1:59786 as 4 (p2, c1, k60).
1631793674: Invalid QoS in subscription command from 172.17.0.1, disconnecting.
1631793674: Client 4 disconnected due to malformed packet.
```

[...]

PR #2969, opened 2023-04-22
fix(mqtt-streaming): Separate MQTT QoS by message type (#2969

ennru:
As @ohze pointed out a long while ago, when subscribing Alpakka MQTT streaming sent the Quality of Service uncorrectly.

With the details @sbmpost added it became apparent that the code made the wrong assumption the QoS was encoded the same for Subscribe and Publish, but it is not.

The first commit contains the test case suggested by @ohze.
The second commit changes the API slightly to separate QoS encoding for the two message types.
The third commit re-enables a test case that was ignored due to regular failure. Which seems related.

Fixes
- #2731 
- #1549


johanandren (approved):
LGTM when CI is happy

ennru:
The link validator is upset as the [Knowldus blog at https://blog.knoldus.com](https://blog.knoldus.com/) is unavailable.

PR #2970, opened 2023-04-24
fix: logger with proper class ref

sebastian-alfers:
Popped up [here](https://github.com/akka/akka-management/pull/1134#discussion_r1172488640).


ennru (approved):
LGTM.

Issue #2963, opened 2023-04-07
Wrong decoding of UnsubAck


sbmpost:
@ennru 

### Versions used 

akka-stream-alpakka-mqtt-streaming 4.0.0

Akka version: 

2.6.20

### Expected Behavior

UnsubAck should be decoded correctly

### Actual Behavior

UnsubAck isn't decoded correctly. As such an Unsubscribe will lead to UnsubscribeFailed.

### Relevant logs

The error in the code is here:
https://github.com/akka/alpakka/blob/3d8b41ffe8c5904b8d2d197d642939d26a84d4ee/mqtt-streaming/src/main/scala/akka/stream/alpakka/mqtt/streaming/model.scala#L808

Value "ReservedGeneral" should be used instead of ReservedUnsubAck. See also these specs:
http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718022

Related: #2731

I would suggest fixing these two issues in one go as they are both one-liners.

ennru:
Thank you. Fixing it with
- #2971

ennru:
Fixed with #2971.

PR #2971, opened 2023-04-24
fix(mqtt-streaming): correct UNSUBACK flags

ennru:
The flags used with `UnsubAck` should just be 0.

References 
- #2963


johanandren (approved):

ennru:
The link validator is upset as the [Knoldus blog at https://blog.knoldus.com](https://blog.knoldus.com/) is unavailable.

PR #2976, opened 2023-05-04
bump: ScalaTest 2.3.15 (was 2.3.11) + drop ScalaTest+ Mockito

ennru:
Dropping the not too useful ScalaTest+ Mockito dependency which is not published for Scala 2.12 for this ScalaTest version.

Fixed some implicit resolution issue in Mongo tests.


johanandren (approved):
LGTM

PR #2980, opened 2023-05-23
Add zip compression level

### LINKED ISSUE #2766: Support Zip Archive Compression Level
### Short description
There is no way to set the compression level for Zip Archive

### Details
When creating a zip file, we need a way to create an uncompressed zip, since the target file may already be compressed.

gael-ft:
Fixes #2766


ennru (approved):
Thank you for taking this on!
LGTM.

PR #2982, opened 2023-05-23
AWS S3: Add getObjectByRanges to S3 API

### LINKED ISSUE #2981: Add supports for Byte-Range Fetches when downloading from S3
### Short description

Add supports for [Byte-Range Fetches](https://docs.aws.amazon.com/whitepapers/latest/s3-optimizing-performance-best-practices/use-byte-range-fetches.html) when downloading an object from AWS S3.

It means doing like AWS "crt" based clients / TransferManager.

### Details

Given:
- bucket
- S3 key
- rangeLength (should default to 8MB)
- parallelism

We should be able to download _parallelism_ ranges in parallel.
Then pushing downstream them in order, requesting requesting a new range when first is complete.

Have my own implementation, I'll submit a PR

gael-ft:
Fixes #2981 

Note I created a `MergeOrderedN` flow graph, because I didn't find a way to do the same using existing ones.
Happy to remove it if there's a way.

### DIFF
--- a/s3/src/main/scala/akka/stream/alpakka/s3/impl/MergeOrderedN.scala
+++ b/s3/src/main/scala/akka/stream/alpakka/s3/impl/MergeOrderedN.scala
@@ -1,0 +1,128 @@
+package akka.stream.alpakka.s3.impl
+
+import akka.annotation.InternalApi
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
+import akka.stream.{Attributes, Inlet, Outlet, UniformFanInShape}
+
+import scala.collection.{immutable, mutable}
+
+@InternalApi private[impl] object MergeOrderedN {
+  /** @see [[MergeOrderedN]] */
+  def apply[T](inputPorts: Int, breadth: Int) =
+    new MergeOrderedN[T](inputPorts, breadth)
+}
+
+/**
+  * Takes multiple streams (in ascending order of input ports) whose elements will be pushed only if all elements from the
+  * previous stream(s) are already pushed downstream.
+  *
+  * The `breadth` controls how many upstream are pulled in parallel.
+  * That means elements might be received in any order, but will be buffered (if necessary) until their time comes.
+  *
+  * '''Emits when''' the next element from upstream (in ascending order of input ports) is available
+  *
+  * '''Backpressures when''' downstream backpressures
+  *
+  * '''Completes when''' all upstreams complete and there are no more buffered elements
+  *
+  * '''Cancels when''' downstream cancels
+  */
+@InternalApi private[impl] final class MergeOrderedN[T](val inputPorts: Int, val breadth: Int) extends GraphStage[UniformFanInShape[T, T]] {
+  require(inputPorts > 1, "input ports must be > 1")
+  require(breadth > 0, "breadth must be > 0")
+
+  val in: immutable.IndexedSeq[Inlet[T]] = Vector.tabulate(inputPorts)(i => Inlet[T]("MergeOrderedN.in" + i))
+  val out: Outlet[T] = Outlet[T]("MergeOrderedN.out")
+  override val shape: UniformFanInShape[T, T] = UniformFanInShape(out, in: _*)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) with OutHandler {
+    private val bufferByInPort = mutable.Map.empty[Int, mutable.Queue[T]] // Queue must not be empty, if so entry should be removed
+    private var currentHeadInPortIdx = 0
+    private var currentLastInPortIdx = 0
+    private val overallLastInPortIdx = inputPorts - 1
+
+    setHandler(out, this)
+
+    in.zipWithIndex.foreach { case (inPort, idx) =>
+      setHandler(inPort, new InHandler {
+        override def onPush(): Unit = {
+          val elem = grab(inPort)
+          if (currentHeadInPortIdx != idx || !isAvailable(out)) {
+            bufferByInPort.updateWith(idx) {
+              case Some(inPortBuffer) =>
+                Some(inPortBuffer.enqueue(elem))
+              case None =>
+                val inPortBuffer = mutable.Queue.empty[T]
+                inPortBuffer.enqueue(elem)
+                Some(inPortBuffer)
+            }
+          } else {
+            pushUsingQueue(Some(elem))
+          }
+          tryPull(inPort)
+        }
+
+        override def onUpstreamFinish(): Unit = {
+          if (canCompleteStage)
+            completeStage()
+          else if (canSlideFrame)
+            slideFrame()
+        }
+      })
+    }
+
+    override def onPull(): Unit = pushUsingQueue()
+
+    private def pushUsingQueue(next: Option[T] = None): Unit = {
+      val maybeBuffer = bufferByInPort.get(currentHeadInPortIdx)
+      if (maybeBuffer.forall(_.isEmpty) && next.nonEmpty) {
+        push(out, next.get)
+      } else if (maybeBuffer.exists(_.nonEmpty) && next.nonEmpty) {
+        maybeBuffer.get.enqueue(next.get)
+        push(out, maybeBuffer.get.dequeue())
+      } else if (maybeBuffer.exists(_.nonEmpty) && next.isEmpty) {
+        push(out, maybeBuffer.get.dequeue())
+      } else {
+        // Both empty
+      }
+
+      if (maybeBuffer.exists(_.isEmpty))
+        bufferByInPort.remove(currentHeadInPortIdx)
+
+      if (canCompleteStage)
+        completeStage()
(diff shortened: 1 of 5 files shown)


s3/src/main/scala/akka/stream/alpakka/s3/impl/MergeOrderedN.scala
```diff
@@ -0,0 +1,128 @@
...
+  *
+  * '''Emits when''' the next element from upstream (in ascending order of input ports) is available
+  *
+  * '''Backpressures when''' downstream backpressures
+  *
+  * '''Completes when''' all upstreams complete and there are no more buffered elements
+  *
+  * '''Cancels when''' downstream cancels
+  */
+@InternalApi private[impl] final class MergeOrderedN[T](val inputPorts: Int, val breadth: Int) extends GraphStage[UniformFanInShape[T, T]] {
```
johanandren, line 30: `ConcatPreFetch` or something like that would be more correct , merge in Akka streams generally mean emit in any order.

johanandren (commented):
Note that trying to control if the upstreams pre-fetch like you try to do is tricky since the sources could always contain their own buffers and start right away. 

There's also a risk that the request is triggered right away and the response body stream times out with a subscription timeout from being queued based on the stream consumption speed. 

A more safe solution, If I understand correctly what we are aiming for here, might be a `ranges.mapAsync(parallelism)(range => fetchEntireRangeIntoFutureByteString)`

gael-ft:
Hi @johanandren 

Thanks for the review !

> Note that trying to control if the upstreams pre-fetch like you try to do is tricky since the sources could always contain their own buffers and start right away.

> There's also a risk that the request is triggered right away and the response body stream times out with a subscription timeout from being queued based on the stream consumption speed.

I assume this is the pupose of `GraphStages.withDetachedInputs(...)` right ?

--- 

> A more safe solution, If I understand correctly what we are aiming for here, might be a ranges.mapAsync(parallelism)(range => fetchEntireRangeIntoFutureByteString)

Anyway I'll try this as it would be much shorter (no need `MergeOrderedN`) and looks like it does the same in the end

gael-ft:
Just want to be sure to fully understand the issue to fix it correctly (and to know what I should take care of in the future 😉):

> Note that trying to control if the upstreams pre-fetch like you try to do is tricky since the sources could always contain their own buffers and start right away.

If they are, elements will come from the source buffer, not directly from the remote (S3 here), ot sure what is the problem here ?

> There's also a risk that the request is triggered right away and the response body stream times out with a subscription timeout from being queued based on the stream consumption speed.

In the `MergeOrderedN` I am not pulling all inputs ports at the same time.
I only pull `breadth` inputs ports.
As well, I either push them downstream directly if I can (and should) or I put them in a queue and continue to process new elements.
There is no "wait until something happen to continue processing"

Example:

sources: S1, S2, S3
parallelism: 2

- (preStart) -> pulling S1 & S2 only
- S1 push & `out` available -> push downstream and pull it again
- S2 push -> buffer and pull it again
- S1 push & `out` not available -> buffer and pull it again
- S1 complete (S1 has buffered elements, so we do nothing)
- S2 push -> buffer and pull it again
- `out` available and pull -> dequeue buffer of S1 -> S1 completed and buffer empty -> start to pull S3
- `out` available and pull -> dequeue elements from buffer of S2
- ...

I can't figure how I could trigger subscription timeout here ?
If source has its own buffer then problem is not there
If source has not we won't pull it until its time comes so no requests body pending subscription

johanandren:
All the range requests are done immediately on materialization, then your scheme is to only pull from the first N of the requests and backpressure the rest. The are two potential problems with it: 

Any buffer introduced explicitly, or implicitly, for example by an async boundary, will detach the response stream and start consuming it right away for all requests.

If there is no boundary and the backpressure works: if you don't subscribe to those within a timeout (default 1 second, `response-entity-subscription-timeout`) those request streams will fail.  

The only way to make sure the requests are only done N at a time would be to not trigger each request at all until it should run and start fetch data.

gael-ft:
> All the range requests are done immediately on materialization

Ok thanks clear enough to me now.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -279,6 +272,29 @@ import scala.util.{Failure, Success, Try}
...
+      getObject(s3Location, None, versionId, s3Headers)
+    } else {
+      Source(byteRanges)
+        .zipWithIndex
+        .flatMapMerge(parallelism, brToIdx => {
+          val (br, idx) = brToIdx
+          val endMarker = Source.single(ByteString("$END$"))
+          getObject(s3Location, Some(br), versionId, s3Headers).concat(endMarker).map(_ -> idx)
+        })
+        .statefulMapConcat(RangeMapConcat)
```
johanandren, line 294: Composing each range-source with a buffer to allow parallel fetching, and then concatenating the resulting streams to get the resulting bytes out in the right order seems like it would achieve the same but much simpler.

Am I missing something clever that this does?

gael-ft: If I understood correctly, you are thinking about `conflate` or something similar to buffer range sources.
Something like:

```scala
getObject(s3Location, Some(br), versionId, s3Headers).conflate(_ ++ _).concat(endMarker).map(_ -> idx)
```

As `flatMapMerge` may emit in any order, I still need the range `idx` to order (possibly buffered) bytes.
So output item of `flatMapMerge` will look like `(ByteString, Long)` and can be in any order.

How can I order them back, without `statefulMapConcat` ? Range2 could emit before range 1 and range2 could be complete before range1.

johanandren: I was thinking something like

```
Source(byteRanges)
  .mapAsync(parallelism)(range => 
    getObjectByRanges(...).buffer(size, Backpressure)
  ).flatMapConcat(identity)
```

johanandren: But ofc that may not be good enough with buffer sized in chunks instead of bytes, we don't have a buffer with weighted size calculation though, maybe batchWeighted could do, not sure.

gael-ft: Hmm can't make it work:

Tried with:

```scala
Source(byteRanges)
  .mapAsync(parallelism)(br => Future.successful(
    getObject(s3Location, Some(br), versionId, s3Headers).batchWeighted(rangeSize, _.size, identity)(_ ++ _)
  ))
  .flatMapConcat(identity)
```

and

```scala
Source(byteRanges)
  .mapAsync(parallelism)(br => Future.successful(
    Source.fromMaterializer { case (mat, _) =>
      getObject(s3Location, Some(br), versionId, s3Headers)
        .preMaterialize()(mat)
        ._2
        .batchWeighted(rangeSize, _.size, identity)(_ ++ _)
      }
    ))
    .flatMapConcat(identity)
```

But in both situations, ranges are fetched one by one and download perf looks like `getObject`

johanandren: Ah, ofc, they aren't materialized so they can start consume bytes until flatMapConcat:ed, didn't think of that. Pre-materialization creates a running source but the downstream is not materialized until it is used, so you would need to put the batching before preMaterialize.

johanandren: For the record I created an upstream issue with an idea that could make this kind of thing easier: https://github.com/akka/akka/issues/31958 (continue with the current solution here though)

He-Pin:
@gael-ft I have submited a PR for `flatmapConcat` with inflight materialization, would you like to do some review, thanks.

s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Stream.scala
```diff
@@ -232,6 +240,128 @@ import scala.util.{Failure, Success, Try}
...
+          case Some(s3Meta) if s3Meta.contentLength == 0 =>
+            objectMetadataMat.success(s3Meta)
+            Source.empty[ByteString]
+          case Some(s3Meta) =>
+            objectMetadataMat.success(s3Meta)
+            doGetByRanges(s3Location, versionId, s3Headers, s3Meta.contentLength, rangeSize, parallelism)
+          case None =>
+            val exc = new NoSuchElementException(s"Object does not exist at location [${s3Location.mkString}]")
+            objectMetadataMat.failure(exc)
+            Source.failed(exc)
```
He-Pin, line 263: This will close the downstream ASAP, do you want to defer it?

gael-ft: Hmmm not sure of what that implies ?
The idea is that no `ObjectMetadata` means no S3 object, so I think the source should fail as well as the materialized Future

PR #2986, opened 2023-06-12
docs: fix multiple examples for commit after write to elasticsearch

efgpinto:
Noticed that [here](https://doc.akka.io/docs/alpakka/current/elasticsearch.html#passing-data-through-elasticsearchflow), in Scala sample, multiple code snippets are being included while the java sample only has one. I think this was due to copy/paste of the tags when creating new tests. Thus, removing the tags so only one example is shown as with the Java counterpart.


ennru (approved):
LGTM.

PR #2993, opened 2023-07-28
FTP: Add for FTPS the ability to set KeyManager and TrustManager

TheDeadOne:
This fix adds a couple of methods to FtpsSettings to set specific KeyManager and TrustManager.

I understand that the old tag is taken as a basis, but it would be nice to release, for example, version 2.1 with this improvement.

### DIFF
--- a/ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
+++ b/ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
@@ -6,6 +6,8 @@ package akka.stream.alpakka.ftp
 
 import java.net.InetAddress
 import java.net.Proxy
+import javax.net.ssl.KeyManager
+import javax.net.ssl.TrustManager
 import java.nio.file.attribute.PosixFilePermission
 
 import akka.annotation.{DoNotInherit, InternalApi}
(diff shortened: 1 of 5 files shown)


ftp/src/main/scala/akka/stream/alpakka/ftp/model.scala
```diff
@@ -221,7 +231,9 @@ final class FtpsSettings private (
     s"binary=$binary," +
     s"passiveMode=$passiveMode," +
     s"configureConnection=$configureConnection," +
-    s"proxy=$proxy)"
+    s"proxy=$proxy" +
+    s"keyManager=$keyManager" +
```
ennru, line 235:
```suggestion
    s"proxy=$proxy," +
    s"keyManager=$keyManager," +
```

ennru (commented):
Thank you for suggesting, this addition could go into Alpakka 6.0.4.
Please rebase it.

TheDeadOne:
@ennru I hope everything is done right.

ennru:
One more thing...
```
[error] akka-stream-alpakka-ftp: Failed binary compatibility check against com.lightbend.akka:akka-stream-alpakka-ftp_2.13:6.0.2! Found 1 potential problems
[error]  * method this(java.net.InetAddress,Int,akka.stream.alpakka.ftp.FtpCredentials,Boolean,Boolean,scala.Function1,scala.Option)Unit in class akka.stream.alpakka.ftp.FtpsSettings does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.alpakka.ftp.FtpsSettings.this")
```
It is fine to add an exclusion as the constructor is actually private but Mima doesn't see that.

The filter should go into a file in `ftp/src/main/mima-filters/6.0.3.backwards.excludes`.

ennru (approved):
LGTM.
Thank you!

Issue #2994, opened 2023-08-07
6.0.1 release not Java 8 compatible


jphelp32:
### Versions used 
6.0.0, 6.0.1

Appears to be a regression of issue #2680 

Seems to affect only the Enum classes. I've found these so far:
```
akka.stream.alpakka.csv.javadsl.CsvQuotingStyle
akka.stream.alpakka.xml.ParseEventMarker
```

e.g.
```
% javap -v akka.stream.alpakka.xml.ParseEventMarker | grep "major version"
  major version: 55
```

patriknw:
Thanks for reporting. Fix in https://github.com/akka/alpakka/pull/2995
I'll try to release tomorrow.

patriknw:
@jphelp32 6.0.2 has been released

PR #2995, opened 2023-08-08
fix: Release with JDK 8, #2994

patriknw:
--release 8 was not defined in javacOptions 
However, --release 8 flag prevents access to com.sun.nio.file.SensitivityWatchEventModifier so instead we have to use JDK 8 in the release CI job.

References #2994


aludwiko (approved):
LGTM

PR #3003, opened 2023-08-25
Handle XML exceptions

### LINKED ISSUE #2532: [XML] XML parsing exceptions (e.g. WFCException) should cleanly fail the XmlParsing stage rather than go unhandled
In the XmlParsing stage, if any parsing error occurs, the exception from the underlying aalto parser goes unhandled and fails the stage. As a result, the exception is always logged at the ERROR log level by akka's logger. I think the better approach would be to catch all parsing exceptions and fail the stage cleanly via `failStage(ex)`, so that the exception is not automatically logged, and consumers of the XmlParsing stage can determine how to react wrt to logging.

JustinPihony:
This resolves #2532


xml/src/main/scala/akka/stream/alpakka/xml/impl/StreamingXmlParser.scala
```diff
@@ -58,67 +62,72 @@ private[xml] object StreamingXmlParser {
         else if (isAvailable(out)) advanceParser()
       }
 
-      @tailrec private def advanceParser(): Unit =
+       private def advanceParser(): Unit = {
```
JustinPihony, line 65: Scala `tailrec` does not like exceptions. I think the exception handling overrides the annotation but am open to other opinions. This could also be converted to a `while` if really wanted

johanandren: Pull out the try/catch to the graph stage API methods invoking it instead and leave this as it was?

johanandren (approved):
LGTM as is, but left a suggestion as well, you decide if it simplifies things.

xml/src/main/scala/akka/stream/alpakka/xml/impl/StreamingXmlParser.scala
JustinPihony: Yah - I guess I was trying to centralize and got caught up in that. I just pushed a change with that suggestion

ennru (approved):
LGTM.

PR #3018, opened 2023-10-03
bump: protobuf 3.21.12 in aws modules

patriknw:
* some aws dependencies have protobuf-java 3.19.2, which has CVE-2022-3171 CVE-2022-3509 CVE-2022-3510
* bumped those to protobuf 3.21.12, which is the same as in akka-grpc
* also bumped to latest patch of awssdk 2.17.295


ennru (approved):
So this was only needed for AWS Kinesis?

LGTM.

PR #3026, opened 2023-10-20
bump: protobuf-java 3.24.0

patriknw:
* to align with the bump in Akka gRPC


project/Dependencies.scala
```diff
@@ -340,6 +339,7 @@ object Dependencies {
       ) ++ JacksonDatabindDependencies
   )
 
+  val KinesisProtobufJavaVersion = "3.24.0" // sync with Akka gRPC
```
patriknw, line 342: only used by Kinesis, so moved here

ennru (approved):
LGTM.

PR #3030, opened 2023-10-24
Replace Spray JWT dependency

ennru:
Remove the dependency on `com.github.jwt-scala` `jwt-spray-json` which was discontinued.

Add the relevant sources from that library into google-common which was the only Alpakka part relying in this.

References
- https://github.com/jwt-scala/jwt-scala/pull/318

### DIFF
--- a/google-common/src/main/scala/akka/stream/alpakka/google/auth/JwtSprayJsonParser.scala
+++ b/google-common/src/main/scala/akka/stream/alpakka/google/auth/JwtSprayJsonParser.scala
@@ -1,0 +1,77 @@
+/*
+ * Copyright (C) since 2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+/*
+ * This software is licensed under the Apache 2 license.
+ * Copyright 2021 JWT-Scala Contributors.
+ */
+
+package akka.stream.alpakka.google.auth
+
+import pdi.jwt.{JwtAlgorithm, JwtClaim, JwtHeader, JwtJsonCommon}
+
+import java.time.Clock
+import pdi.jwt.exceptions.JwtNonStringException
+import spray.json._
+
+/**
+ * Implementation of `JwtCore` using `JsObject` from spray-json.
+ *
+ * This class originally came from jwt-spray-json,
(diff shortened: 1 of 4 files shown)


google-common/src/main/scala/akka/stream/alpakka/google/auth/JwtSprayJsonParser.scala
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (C) since 2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+/*
+ * This software is licensed under the Apache 2 license.
+ * Copyright 2021 JWT-Scala Contributors.
```
patriknw, line 7: can the copyright be placed this at the top header?

```
/*
 * Copyright 2021 JWT-Scala Contributors.
 * Copyright (C) since 2016 Lightbend Inc. <https://www.lightbend.com>
 */
```

(and it's not really since 2016)

google-common/src/main/scala/akka/stream/alpakka/google/auth/JwtSprayJsonParser.scala
```diff
@@ -0,0 +1,77 @@
...
+import pdi.jwt.exceptions.JwtNonStringException
+import spray.json._
+
+/**
+ * Implementation of `JwtCore` using `JsObject` from spray-json.
+ *
+ * This class originally came from jwt-spray-json,
+ * but was removed in https://github.com/jwt-scala/jwt-scala/commit/bf1131ce02480103c0b953b97da001105a3ee038
+ */
+trait JwtSprayJsonParser[H, C] extends JwtJsonCommon[JsObject, H, C] {
```
patriknw, line 24: make them `private[akka]` and internal api markers?

google-common/src/main/scala/akka/stream/alpakka/google/auth/JwtSprayJsonParser.scala
ennru: That would require excluding this file from the sbt-header plugin.

ennru: But not too hard.

Issue #3038, opened 2023-11-20
Alpakka JMS Consumer memory leak


arcavire:
### Versions used 

Akka version: 2.6.18
Alpakka version: 5.0.0
Scala version: 2.13
IBM MQ AllClient version: 9.3.4.0

### Expected Behavior

`akka.stream.alpakka.jms.impl.JmsConsumerSession` not accumulating in memory

### Actual Behavior

`akka.stream.alpakka.jms.impl.JmsConsumerSession` are accumulating in memory because they are still referenced from jmsListener

### Observations

At the begging the consumer sessions are added to a list (`jmsSessions`) in `akka.stream.alpakka.jms.impl.JmsConnector#onSession`. Additionally, it is referenced by anonymous jms.MessageListener created in  `akka.stream.alpakka.jms.impl.JmsTxSourceStageLogic`.
![image](https://github.com/akka/alpakka/assets/10477756/395bfcde-93e3-4746-aaba-65fce7a01313)

Then I restart the MQ and `akka.stream.alpakka.jms.impl.JmsConnector#handleRetriableException` is called. That method cleans the list of the sessions, but they are still referenced from the listeners:
![image](https://github.com/akka/alpakka/assets/10477756/49912e17-68d5-492b-bae4-4d8039cde37f)

Then the MQ is back and new sessions are added to the list, but the old ones are still kept in memory and will never be garbage collected. The number of the sessions kept in memory grows with each connection lost with MQ.

### Reproducible Test Case

1. Start an application with JmsConsumer.txSource that reads from MQ.
2. Restart the MQ.

ennru:
Thank you for your analysis of this problem.

ennru:
Fixed with
- #3041

arcavire:
Would it be possible to backport the fix to 4.0.0? This is the last version with Apache 2.0 license.

PR #3041, opened 2023-11-28
jms: close sessions on connection loss

leviramsey:
References #3038.

In case of connection loss, we clear the list of current connections without closing them.  In the consumer case, this results in the message processing callback holding onto the session because it closes over it.  Closing the session will close the consumers and free the callback.

### DIFF
--- a/jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
@@ -135,7 +135,8 @@ private[jms] trait JmsConnector[S <: JmsSession] extends TimerGraphStageLogic wi
   }
 
   private def handleRetriableException(ex: Throwable): Unit = {
-    jmsSessions = Seq.empty
+    closeSessions()
+
     connectionState match {
       case JmsConnectorInitializing(_, attempt, backoffMaxed, _) =>
         maybeReconnect(ex, attempt, backoffMaxed)
@@ -270,16 +271,17 @@ private[jms] trait JmsConnector[S <: JmsSession] extends TimerGraphStageLogic wi
     eventualConnection.map(closeConnection).map(_ => Done)
 
   protected def closeSessions(): Unit = {
-    jmsSessions.foreach(s => closeSession(s))
+    jmsSessions.foreach(closeSession)
     jmsSessions = Seq.empty
   }
 
   protected def closeSessionsAsync(): Future[Unit] = {
     val closing = Future
       .sequence {
-        jmsSessions.map(s => Future(closeSession(s)))
+        jmsSessions.map(s => Future { closeSession(s) })
       }
-      .map(_ => ())
+      .flatMap(_ => Future.unit)
+
     jmsSessions = Seq.empty
     closing
   }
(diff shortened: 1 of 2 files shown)


jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
```diff
@@ -270,16 +271,17 @@ private[jms] trait JmsConnector[S <: JmsSession] extends TimerGraphStageLogic wi
...
-    jmsSessions.foreach(s => closeSession(s))
+    jmsSessions.foreach(closeSession)
     jmsSessions = Seq.empty
   }
 
   protected def closeSessionsAsync(): Future[Unit] = {
     val closing = Future
       .sequence {
-        jmsSessions.map(s => Future(closeSession(s)))
+        jmsSessions.map(s => Future { closeSession(s) })
```
leviramsey, line 281: I have a strong preference for curly braces around `Future.apply` calls to make them visually stand out.

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
```diff
@@ -270,16 +271,17 @@ private[jms] trait JmsConnector[S <: JmsSession] extends TimerGraphStageLogic wi
...
       }
-      .map(_ => ())
+      .flatMap(_ => Future.unit)
```
leviramsey, line 283: Not sure if this was written before `Future.unit` existed, but since we're optimizing memory...

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsTxSourceStage.scala
```diff
@@ -42,48 +42,39 @@ private[jms] final class JmsTxSourceStage(settings: JmsConsumerSettings, destina
...
-                      case _: TimeoutException =>
-                        val exception = new JmsTxAckTimeout(settings.ackTimeout)
-                        session.session.rollback()
-                        if (settings.failStreamOnAckTimeout) {
-                          handleError.invoke(exception)
-                        } else {
-                          log.warning(exception.getMessage)
-                        }
+    override protected def onSessionOpened(consumerSession: JmsConsumerSession): Unit =
+      consumerSession
```
leviramsey, line 46: clearer code through trusting the Scala compiler...

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsConnector.scala
```diff
@@ -270,16 +271,17 @@ private[jms] trait JmsConnector[S <: JmsSession] extends TimerGraphStageLogic wi
...
+
     jmsSessions = Seq.empty
     closing
   }
```
patriknw, line 287: since you are anyway cleaning up some stuff, can you also change all `case e: Throwable` to `case NonFatal(e)`

patriknw (approved):
LGTM

jms/src/main/scala/akka/stream/alpakka/jms/impl/JmsTxSourceStage.scala
```diff
@@ -42,48 +42,39 @@ private[jms] final class JmsTxSourceStage(settings: JmsConsumerSettings, destina
...
+                }
+              } catch {
+                case e: IllegalArgumentException => handleError.invoke(e) // Invalid envelope, fail the stage
+                case e: jms.JMSException => handleError.invoke(e)
+              }
             }
-            .onComplete(sessionOpenedCB.invoke)
-
-        case _ =>
-          throw new IllegalArgumentException(
```
leviramsey, line 82: Was wondering why the compiler wasn't warning this was unreachable... turns out that a `null` would be the only value which hits this... the new code would NPE on `createConsumer` which seems to be preferable (you go straight to "why was the argument null").

Scala 3 should complain about this case being unreachable (with the warning saying that it's only reachable in the null case, so recommend explicitly matching `null`).

ennru (approved):
LGTM.

Issue #3019, opened 2023-10-05
Replaced base uri path when using ElasticsearchSink


pramodShehan5:
### Versions used 

Akka version:   "com.lightbend.akka" %% "akka-stream-alpakka-elasticsearch" % "6.0.2"

` val sinkSettings =  ElasticsearchConnectionSettings("https://test.com:9200/elasticsearch")
      .withCredentials("", "")
  val sinkSettings =
  ElasticsearchWriteSettings(connectionSettings)
    .withBufferSize(10)
    .withRetryLogic(RetryAtFixedRate(maxRetries = 5, retryInterval = 1.second))
    .withApiVersion(ApiVersion.V5)
    .withRetryLogic(RetryAtFixedRate(maxRetries = 5, retryInterval = 1.second))
  val elasticsearchSink = ElasticsearchSink.create[DocumentStoreMetadata1](
    constructElasticsearchParams("indexname", "test", ApiVersion.V5),
    settings = sinkSettings
  )`

### Expected Behavior

it should be called POST https://test.com:9200/elasticsearch/_bulk

### Actual Behavior

my elasticsearch base uri is https://test.com:9200/elasticsearch. it shouldn't be removed "/elasticsearch" path.

val uri = baseUri.withPath(Path(endpoint))
"_bulk" has been used in place of "elasticsearch" in this.

### Relevant logs

![image](https://github.com/akka/alpakka/assets/9468378/6d8c442c-f283-451f-a733-2496059a62ca)

ennru:
Thank you for creating a PR to fix this. It must have been an oversight.

PR #3042, opened 2023-11-30
Elasticsearch: keep path from base URI

ennru:
The current implementation drops any path that is present in the baseUri.

References
- #3019


patriknw (approved):
LGTM

PR #3047, opened 2023-12-27
kinesis: use stage materializer with IODispatcher instead of injected EC

jtjeferreira:
The `KinesisSchedulerSourceStage` needs to run the `software.amazon.kinesis.coordinator.Scheduler` in a thread. The scheduler runs a loop and uses `Thread.sleep`, which needs careful use because it is a blocking operation.

Currently, it uses `Future(scheduler.run())`, which will use the implicit `ExecutionContext` from the constructor. By default, this EC will be the akka default dispatcher which is not suitable for blocking. Also, note the use of `ActorAttributes.IODispatcher` in `initialAttributes` which is not having any effect as we are not using the stage materializer.

This PR removes the EC from the constructor, and uses the stage logic materializer EC (which will be the IODispatcher) which is aligned with other custom stages in this project.

### DIFF
--- a/kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
+++ b/kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
@@ -103,7 +103,9 @@ private[kinesis] class KinesisSchedulerSourceStage(
         failStage(SchedulerUnexpectedShutdown(e))
     }
     override def postStop(): Unit =
-      schedulerOpt.foreach(scheduler => Future(if (!scheduler.shutdownComplete()) scheduler.shutdown())(materializer.executionContext))
+      schedulerOpt.foreach(
+        scheduler => Future(if (!scheduler.shutdownComplete()) scheduler.shutdown())(materializer.executionContext)
+      )
 
     protected def executionContext(attributes: Attributes): ExecutionContext = {
       val dispatcherId = (attributes.get[ActorAttributes.Dispatcher](ActorAttributes.IODispatcher) match {
@@ -111,7 +113,7 @@ private[kinesis] class KinesisSchedulerSourceStage(
           ActorAttributes.IODispatcher
         case d => d
       }) match {
-        case d@ActorAttributes.IODispatcher =>
+        case d @ ActorAttributes.IODispatcher =>
           // this one is not a dispatcher id, but is a config path pointing to the dispatcher id
           materializer.system.settings.config.getString(d.dispatcher)
         case d => d.dispatcher
--- a/kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
+++ b/kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
@@ -104,7 +104,7 @@ private[kinesis] class KinesisSchedulerSourceStage(
     }
     override def postStop(): Unit =
       schedulerOpt.foreach(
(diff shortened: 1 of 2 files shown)


jtjeferreira:
@ennru any chance this could be reviewed?

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
```diff
@@ -103,6 +103,23 @@ private[kinesis] class KinesisSchedulerSourceStage(
...
+        case d => d
+      }) match {
+        case d @ ActorAttributes.IODispatcher =>
+          // this one is not a dispatcher id, but is a config path pointing to the dispatcher id
+          materializer.system.settings.config.getString(d.dispatcher)
+        case d => d.dispatcher
+      }
+
+      materializer.system.dispatchers.lookup(dispatcherId)
+    }
```
johanandren, line 123: This is identical to what is done for the JMS connector. Sadly we don't have anything more straight forward out of the box.

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
```diff
@@ -103,6 +103,23 @@ private[kinesis] class KinesisSchedulerSourceStage(
         failStage(SchedulerUnexpectedShutdown(e))
     }
     override def postStop(): Unit =
-      schedulerOpt.foreach(scheduler => Future(if (!scheduler.shutdownComplete()) scheduler.shutdown()))
+      schedulerOpt.foreach(
+        scheduler => Future(if (!scheduler.shutdownComplete()) scheduler.shutdown())(materializer.executionContext)
+      )
```
johanandren, line 108: Contains a lock and possible other blocking stuff, better do it right away inside the stage which is running on the IO dispatcher or make sure this also runs on the `ec`.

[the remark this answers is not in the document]
johanandren: Yes, that will block the stream shutdown until that scheduler shutdown returns (the original code just fired it off and didn't care when it completes) but maybe that is fine?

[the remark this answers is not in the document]
jtjeferreira: I think that's a good change. There is also [startGracefulShutdown](https://github.com/awslabs/amazon-kinesis-client/blob/master/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java#L726) but that is a bigger change...

kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/KinesisSchedulerSourceStage.scala
```diff
@@ -103,6 +103,23 @@ private[kinesis] class KinesisSchedulerSourceStage(
         failStage(SchedulerUnexpectedShutdown(e))
     }
     override def postStop(): Unit =
-      schedulerOpt.foreach(scheduler => Future(if (!scheduler.shutdownComplete()) scheduler.shutdown()))
+      schedulerOpt.foreach(
+        scheduler => if (!scheduler.shutdownComplete()) scheduler.shutdown())
```
jtjeferreira, line 107:
```suggestion
        scheduler => if (!scheduler.shutdownComplete()) scheduler.shutdown()
```

johanandren (approved):
LGTM, Thanks!

PR #3128, opened 2024-02-26
bump: Rollback sbt to 1.9.6

johanandren:
Strange issues in main on 1.9.9 where javax.jms cannot be found. We did not manage to reproduce that locally, but consistently happens in ghactions.

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.9.8
+sbt.version=1.9.7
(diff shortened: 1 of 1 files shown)


ennru (approved):
Refs
- #3123 

LGTM.

johanandren:
Ok, so something with 1.9.7 is the problem. But that's where the recent security fix is and we'd want that. Not obvious from the release notes what could cause it but there is a coursier bump in there: https://github.com/sbt/sbt/releases/tag/v1.9.7

ennru (approved):
Interesting. Will require some digging.
Test failure is unrelated.
LGTM.

Issue #2865, opened 2022-05-16
ackSource fails when acknowledging to IBM MQ


haroldpirum:
I'm trying to connect to an ibm mq source using ack mode and am seeing the following error when attempting to batch up the messages.  Is this a known issue?

```
[ERROR] [05/16/2022 11:36:42.681] [AckSourceApp-akka.actor.default-dispatcher-4] [JmsAckSourceStage$JmsAckSourceStageLogic(akka://AckSourceApp)] Error closing jms session
com.ibm.msg.client.jms.DetailedIllegalStateException: JMSCC0033: A synchronous method call is not permitted when a session is being used asynchronously: 'acknowledge'.
The JMS specification does not permit the use of a session for synchronous methods when asynchronous message delivery is running.
Create a separate session if you wish to use both synchronous methods and asynchronous delivery simultaneously.
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:78)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:480)
	at com.ibm.msg.client.commonservices.j2se.NLSServices.createException(NLSServices.java:319)
	at com.ibm.msg.client.commonservices.nls.NLSServices.createException(NLSServices.java:226)
	at com.ibm.msg.client.jms.internal.JmsErrorUtils.createException(JmsErrorUtils.java:126)
	at com.ibm.msg.client.jms.internal.JmsSessionImpl.checkSynchronousUsage(JmsSessionImpl.java:2923)
	at com.ibm.msg.client.jms.internal.JmsMessageImpl.acknowledge(JmsMessageImpl.java:2170)
	at com.ibm.jms.JMSMessage.acknowledge(JMSMessage.java:478)
	at akka.stream.alpakka.jms.impl.JmsAckSession.$anonfun$ack$1(Sessions.scala:92)
	at akka.stream.alpakka.jms.impl.JmsAckSession.drainAcks(Sessions.scala:123)
[...]

PR #3132, opened 2024-02-29
Handle session thread

JustinPihony:
Addresses
- #2865

### DIFF
--- a/jms/src/main/scala/akka/stream/alpakka/jms/impl/Sessions.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/impl/Sessions.scala
@@ -97,7 +97,7 @@ private[jms] final class JmsAckSession(override val connection: jms.Connection,
 
   private def stopMessageListenerAndCloseSession(): Unit = {
     try {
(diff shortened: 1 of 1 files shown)


jms/src/main/scala/akka/stream/alpakka/jms/impl/Sessions.scala
```diff
@@ -97,7 +97,8 @@ private[jms] final class JmsAckSession(override val connection: jms.Connection,
 
   private def stopMessageListenerAndCloseSession(): Unit = {
     try {
-      drainAcks()
+      if(session.getAcknowledgeMode() == jms.Session.CLIENT_ACKNOWLEDGE)
```
laglangyue, line 100: code style. Need a black, `if (..)`

ennru, line 100:
```suggestion
      if (session.getAcknowledgeMode() == jms.Session.CLIENT_ACKNOWLEDGE)
```

ennru (approved):
LGTM.

PR #3163, opened 2024-03-29
revert: logback may not pull slf4j 2.x

ennru:
Revert
- #3051 

and review all places to stay on Slf4j V1. Some places pulled in V2 in hideous ways.


ennru:
The test failures are real and need attention in more PRs.

johanandren (approved):

PR #3167, opened 2024-04-10
JMS: Swallow IBM MQ exception

JustinPihony:
Better fix for #2865

### DIFF
--- a/jms/src/main/scala/akka/stream/alpakka/jms/impl/Sessions.scala
+++ b/jms/src/main/scala/akka/stream/alpakka/jms/impl/Sessions.scala
@@ -100,7 +100,7 @@ private[jms] final class JmsAckSession(override val connection: jms.Connection,
       if (session.getAcknowledgeMode() == jms.Session.CLIENT_ACKNOWLEDGE)
         drainAcks()
     } catch {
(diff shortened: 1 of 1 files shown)


jms/src/main/scala/akka/stream/alpakka/jms/impl/Sessions.scala
```diff
@@ -99,6 +99,8 @@ private[jms] final class JmsAckSession(override val connection: jms.Connection,
     try {
       if (session.getAcknowledgeMode() == jms.Session.CLIENT_ACKNOWLEDGE)
         drainAcks()
+    } catch {
+      case ex: JMSException if ex.getMessage.contains("JMSCC0033") => // Swallow as ack isn't necessary if this is being thrown
```
ennru, line 103: Can we be sure `getMessage` isn't `null`?

JustinPihony: `JMSException` does not have an empty constructor, but there is no check that the provided value is not `null`. I do not believe it is used in the capacity where `null` is passed in, but I added the check to be safe.

ennru (approved):
LGTM.

PR #3199, opened 2024-05-13
chore: Fix ftps test certs

johanandren:


johanandren:
Hmmm, where does that cert come from then, it's note one of the test certs and when I try the `atmoz/sftp` image locally, it doesn't show any outdated cert....

johanandren:
Aha, found it, its in `test/resources/ftpd/pure-ftpd.pem`

PR #3204, opened 2024-05-14
build: Fix release notes config yaml

johanandren:
Not sure, but looking at the examples they seem to have the labels in a list even when it is just a single one, so let's try that.


ennru (approved):
LGTM.

PR #3220, opened 2024-06-19
bump (Hbase, Hadoop): Hadoop 3.4, Hbase 2.6) fixes CVE-2021-37137

sebastian-alfers:
Note: hbase is currently not tested due to mentioned issue.


ennru:
The failures are unrelated.
I suggest keeping this PR around until the next major or minor bump of Alpakka as libraries shouldn't change versions surprisingly in patch releases.

ennru (approved):
LGTM.

PR #3269, opened 2024-09-26
docs: revise URL structure (libraries)

ennru:
The Link validator is expected to fail.

Keeping these API docs to be produced with JDK 11 as the code also references `java.sql` and `javax.xml` which live in other modules nowadays.

References
- https://github.com/akka/akka/pull/32518


sebastian-alfers (approved):
LGTM

PR #3319, opened 2024-11-18
docs: show template's Akka logo

ennru:
The site currently uses the original Alpakka logo.

PR #3320, opened 2024-11-18
Delete docs/src/main/paradox/_template/logo.st

rasummer:
Delete the Alpakka logo so that it uses the new Akka logo

@ennru can we this repo just use the Akka logo, as the Alpakka logo uses the old Akka brand?


ennru:
I had just opened #3319 to do the same (and other link fixes).

PR #3323, opened 2024-12-11
chore: link validator in separate workflow

ennru:
- Run the link validator workflow on PRs and merges instead of the "Documentation" part in the main CI workflow
- Cancel the Publish workflow in case another merge happens while it is running


sebastian-alfers (approved):
LGTM

Issue #3395, opened 2025-05-08
Update Couchbase Integration


chedim:
### Short description

The Alpakka Couchbase integration has not been updated in quite a while and is stated to reach the EOL. 
It would be nice if someone stepped up to update it instead :)

### Details

I am a Couchbase employee and have been working on the update. I still need to add some more of new tests, but wanted to create this issue for submitting a draft PR for potential suggestions and recommendations by devs.

Thank you!

PR #3396, opened 2025-05-08
Couchbase: Updates the integration to support collections

chedim:
References #3395

This PR introduces new `CouchbaseCollectionSession` class that can be accessed from `CouchbaseSession` via `CouchbaseSession::collection` method. It also updates sinks, sources and flow constructors and their signatures to work with collections. 

All of these are breaking changes. 

Please note that, although fascinated by the framework and some language features, I am not very familiar with either alpakka or scala and more of a Java dev. So, any suggestions on how to improve my changes are greatly appreciated.

Cheers!

### DIFF
--- a/couchbase/src/test/scala/docs/scaladsl/CouchbaseFlowSpec.scala
+++ b/couchbase/src/test/scala/docs/scaladsl/CouchbaseFlowSpec.scala
@@ -7,7 +7,7 @@ package docs.scaladsl
 import akka.Done
 import akka.stream.alpakka.couchbase.scaladsl.CouchbaseFlow
 import akka.stream.alpakka.couchbase.testing.CouchbaseSupport
-import akka.stream.alpakka.couchbase.{CouchbaseDeleteFailure, CouchbaseDeleteResult, CouchbaseDocument}
+import akka.stream.alpakka.couchbase.{CouchbaseDeleteFailure, CouchbaseDocument}
 import akka.stream.alpakka.testkit.scaladsl.LogCapturing
 import akka.stream.scaladsl.{Sink, Source}
 import akka.stream.testkit.scaladsl.StreamTestKit._
@@ -170,8 +170,7 @@ class CouchbaseFlowSpec
 
       // #fromId
       import akka.stream.alpakka.couchbase.scaladsl.CouchbaseFlow
-      import akka.stream.scaladsl.Sink
-      import akka.stream.scaladsl.Source
+      import akka.stream.scaladsl.{Sink, Source}
       import com.couchbase.client.java.json.JsonObject
 
       val ids = immutable.Seq("First", "Second", "Third", "Fourth")
--- a/couchbase/src/test/scala/docs/scaladsl/CouchbaseSessionExamplesSpec.scala
+++ b/couchbase/src/test/scala/docs/scaladsl/CouchbaseSessionExamplesSpec.scala
@@ -4,12 +4,11 @@
 
 package docs.scaladsl
 
-import akka.stream.alpakka.couchbase.scaladsl.CouchbaseSession
 import akka.stream.alpakka.couchbase.testing.CouchbaseSupport
 import akka.stream.alpakka.testkit.scaladsl.LogCapturing
 import com.couchbase.client.core.error.DocumentNotFoundException
+import com.couchbase.client.java.ClusterOptions
 import com.couchbase.client.java.env.ClusterEnvironment
-import com.couchbase.client.java.{Cluster, ClusterOptions}
 import org.scalatest.BeforeAndAfterAll
 import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.matchers.should.Matchers
--- a/couchbase/src/test/scala/docs/scaladsl/CouchbaseSourceSpec.scala
+++ b/couchbase/src/test/scala/docs/scaladsl/CouchbaseSourceSpec.scala
@@ -4,18 +4,14 @@
 
 package docs.scaladsl
 
-import akka.stream.alpakka.couchbase.scaladsl.CouchbaseSource
 import akka.stream.alpakka.couchbase.testing.CouchbaseSupport
 import akka.stream.alpakka.testkit.scaladsl.LogCapturing
-import akka.stream.scaladsl.Sink
 import akka.stream.testkit.scaladsl.StreamTestKit._
-import com.couchbase.client.java.json.JsonValue
 import org.scalatest.BeforeAndAfterAll
 import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpec
 
-import scala.concurrent.Future
 import scala.concurrent.duration._
 
 // #statement


lightbend-cla-validator:
Hi @chedim,

Before we proceed with reviewing this pull request, please sign the Akka Contributors License Agreement:

[https://www.lightbend.com/contribute/cla](https://www.lightbend.com/contribute/cla)

couchbase/src/main/scala/akka/stream/alpakka/couchbase/impl/CouchbaseCollectionSessionImpl.scala
```diff
@@ -0,0 +1,152 @@
...
+    underlying.get(id)
+      .orTimeout(timeout.toMillis, TimeUnit.MILLISECONDS)
+      .thenApply(gr => (id, gr.contentAsBytes()))
+      .asScala
+
+  override def upsert[T](document: (String, T)): Future[(String, T)] = {
+    if (document._2.isInstanceOf[Array[Byte]]) {
+      underlying.upsert(document._1, document._2, UpsertOptions.upsertOptions().transcoder(RawBinaryTranscoder.INSTANCE))
+        .thenApply(_ => document)
+        .asScala
```
chedim, line 87: todo: implement `isInstanceOf[String]` case

lightbend-cla-validator:
Hi @chedim,

Before we proceed with reviewing this pull request, please sign the Akka Contributors License Agreement:

[https://www.lightbend.com/contribute/cla](https://www.lightbend.com/contribute/cla)

couchbase/src/main/scala/akka/stream/alpakka/couchbase/impl/CouchbaseCollectionSessionImpl.scala
chedim: Will update, thank you the suggestion!

couchbase/src/main/scala/akka/stream/alpakka/couchbase/CouchbaseDocument.scala
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (C) since 2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.couchbase
+
+class CouchbaseDocument[T](id: String, document: T) {
```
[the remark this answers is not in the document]
patriknw: perhaps `val id` and `val document`, which is more idiomatic for Scala. Can keep the getters for Java.

couchbase/src/main/scala/akka/stream/alpakka/couchbase/scaladsl/CouchbaseCollectionSession.scala
```diff
@@ -0,0 +1,210 @@
...
+   * Reads a document into an object of target class using appropriate Transcoders:
+   *  * If T === Array[Byte] then RawBinaryTranscoder is used
+   *  * If T === String then RawStringTranscoder is used
+   *  * The default json to class transcoder is used otherwise
+   * @param id id of the document
+   * @param target target class
+   * @tparam T type of the object to return
+   * @return a future that completes with created object or errors if the operation failed
+   */
+  def get[T](id: String, target: Class[T]): Future[CouchbaseDocument[T]]
```
patriknw: In Scala it would be more idiomatic to use:
```
def get[T: ClassTag](id: String): Future[CouchbaseDocument[T]]
```

If you need the actual class of that ClassTag you can use:
```
val clazz = implicitly[ClassTag[T]].runtimeClass
```

chedim: I went ahead and applied this suggestion everywhere I was using generics, please lmk if that's not what you meant.

chedim:
@patriknw , updated the PR according to the comments, had couple of questions if I understood the comments correctly. 
Thank you for the review!

chedim:
Should be fixed now.

couchbase/src/main/scala/akka/stream/alpakka/couchbase/impl/CouchbaseCollectionSessionImpl.scala
```diff
@@ -0,0 +1,206 @@
...
+      .insert(id, document, insertOptions)
+      .asScala
+      .map(_ => Done)(ExecutionContext.parasitic)
+  }
+
+  override def get[T: ClassTag](id: String): Future[CouchbaseDocument[T]] = {
+    val target: Class[T] = implicitly[ClassTag[T]].runtimeClass.asInstanceOf[Class[T]]
+    underlying
+      .get(id, GetOptions.getOptions.transcoder(chooseTranscoder[T]()))
+      .thenApply(gr => new CouchbaseDocument[T](id, gr.contentAs(target)))
```
johanandren: This will likely run on the common FJP if not already completed when `thenApply` is invoked (in which case it executes on the calling thread), better do it as a `map` on the Scala Future where execution context is explicit, to keep it on the same thread pool as the other stuff here.

Same for the other `thenApply`s in this class.

couchbase/src/main/scala/akka/stream/alpakka/couchbase/impl/CouchbaseSessionImpl.scala
```diff
@@ -4,179 +4,79 @@
...
-      )
-    )
+  override def toString: String = s"CouchbaseSession(${underlying.name()})"
 
+  private def getSingleResult(gr: QueryResult): Option[JsonObject] = {
+    val rows = gr.rowsAsObject()
+    if (rows.isEmpty) {
+      return Option.empty
+    }
+    Option.apply(rows.iterator.next)
```
johanandren: Should it not also check that there is only one object and fail if there are more? Or is it perhaps fine that `singleResponseQuery` also can be used as a `firstResponseQuery`?

couchbase/src/main/scala/akka/stream/alpakka/couchbase/scaladsl/CouchbaseSession.scala
```diff
@@ -91,192 +93,32 @@ trait CouchbaseSession {
...
-   *
-   * @param id What counter document id
-   * @param delta Value to increase the counter with if it does exist
-   * @param initial Value to start from if the counter does not exist
-   * @return The value of the counter after applying the delta
-   */
-  def counter(id: String, delta: Long, initial: Long, writeSettings: CouchbaseWriteSettings): Future[Long]
+  private val collectionSessions = new AtomicReference(
+    mutable.WeakHashMap.empty[(String, String), CouchbaseCollectionSession]
+  )
```
johanandren: The general pattern for doing this kind of compare and swap is that the collection inside would be immutable.

I see this is safe since it is probably safe since cloned before adding an entry on each try, but better to use Java concurrent hash map in this case.

WeakHashMap and weak references are problematic for GC in general so I'd recommend not to try to use that as a cache eviction strategy. 

If a pool is really needed it'd be better be done as an akka extension, but I think we should take a step back and consider if this is really worth it. And if it is worth a pool of clients, let's do that a s a separate PR.

```

File: couchbase/src/main/scala/akka/stream/alpakka/couchbase/CouchbaseDocument.scala [hunk 7]
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (C) since 2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.couchbase
+
+import scala.reflect.ClassTag
+
+class CouchbaseDocument[T: ClassTag](val id: String, val document: T) {
```
johanandren: Is the constructor part of the public API also for Java? (That could make sense for example for users constructing `CouchBaseDocument` instances in their tests) In that case `T : ClassTag` is problematic and there should be an explicit constructor for Java accepting `Class<T>`.

Also, make it `final` if not intended for user subclassing as public API.

couchbase/src/main/scala/akka/stream/alpakka/couchbase/impl/CouchbaseSessionImpl.scala
chedim: I don't think it should check it here as that would happen only if the original query returned multiple results and developer explicitly showed the desire to receive only the first result (by choosing methods that use this function). I've documented that behavior. The performance in cases when users don't use `LIMIT 1` and query returns more than one row can be improved by wrapping the passed query as a sub-query inside a query with `LIMIT 1`, but I think that optimization is such a fringe case that it can be implemented separately.

couchbase/src/main/scala/akka/stream/alpakka/couchbase/scaladsl/CouchbaseFlow.scala
```diff
@@ -15,61 +21,138 @@ object CouchbaseFlow {
...
+        implicit val ec = materializer.executionContext
+        val session = CouchbaseSessionRegistry(materializer.system).sessionFor(sessionSettings, bucketName)
+        Flow[String]
+          .mapAsync(sessionSettings.parallelism)(
+            id =>
+              session
+                .map(_.collection(scopeName, collectionName))
+                .flatMap(_.getBytes(id))
+          )
+          .withAttributes(ActorAttributes.supervisionStrategy(_ => Resume))
```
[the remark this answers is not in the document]
johanandren: Surprising! Yeah, that's weird but let's keep the existing behavior then.

couchbase/src/main/scala/akka/stream/alpakka/couchbase/impl/CouchbaseSessionImpl.scala
```diff
@@ -4,179 +4,79 @@
...
-                                                 writeSettings.persistTo,
-                                                 writeSettings.replicateTo,
-                                                 writeSettings.timeout.toMillis,
-                                                 TimeUnit.MILLISECONDS),
-                             id)
-      .map(_.content(): Long)(ExecutionContext.parasitic)
+  override def underlying: AsyncBucket = cluster.bucket(bucketName)
+
+  override def streamedQuery(query: String): Source[JsonObject, NotUsed] = {
+    Source.fromIterator(
```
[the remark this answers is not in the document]
johanandren: Ok, then it doesn't really matter.

johanandren (commented):
Almost good to go now, good work! I left a comment about doc code snippets, please take a look at that.

chedim:
Need help!
I'm getting this error with sbt builds after pulling recent changes in the branch:
```
[warn] 	Note: Some unresolved dependencies have extra attributes.  Check that these dependencies exist with the requested attributes.
[warn] 		com.lightbend.akka.grpc:sbt-akka-grpc:2.5.8 (sbtVersion=1.0, scalaVersion=2.12)
[warn] 		io.akka:sbt-paradox-akka:25.10.2 (sbtVersion=1.0, scalaVersion=2.12)
```
(and then it fails because it can't find these). `sbt clean` did not help, updating my security token in akka.sbt did not help.
not sure what happened...

sebastian-alfers:
@chedim Can you confirm the directory you are in? You should run all commands from the root directory (not any sub directory). Also, the commands you share do not work for me. I am using:

```
➜  alpakka git:(3395-update-couchbase) ✗ cat akka-token-resolver.sbt
resolvers in ThisBuild += "akka-secure-mvn" at "https://repo.akka.io/.../secure"
resolvers in ThisBuild += Resolver.url("akka-secure-ivy", url("https://repo.akka.io/.../secure"))(Resolver.ivyStylePatterns)
➜  alpakka git:(3395-update-couchbase) ✗ pwd
/Users/sebastian/dev/chedim/alpakka
➜  alpakka git:(3395-update-couchbase) ✗ sbt "docs/makeSite"

...

[success] Total time: 59 s, completed Nov 12, 2025, 5:30:45 PM
```

PR #3453, opened 2026-03-19
fix(kinesis): Fix shard end coordination race condition

huajiang-tubi:
## Summary
- Fix race condition in `ShardProcessor` where premature shard checkpoint occurs when KCL delivers an empty final batch
- Replace single semaphore coordination with per-batch last-record tracking
- Add `ShardProcessorSpec` with comprehensive test coverage for the coordination scenarios

## Problem
The original implementation used a single `Semaphore(1)` that was acquired when `isAtShardEnd` was true. This created a race condition:

1. When KCL delivers a batch with records where `isAtShardEnd=false`, the records are processed normally
2. KCL then delivers an empty final batch with `isAtShardEnd=true` (this is valid KCL behavior)
3. The `shardEnded` method is called, but since no records were in the final batch, no semaphore coordination happened
4. `shardEnded` immediately checkpoints the shard with `SHARD_END`
5. In-flight records from the previous batch fail to checkpoint because the shard is already marked as ended

## Solution
Replace the single semaphore approach with a per-batch tracking mechanism:

- Create a new `Semaphore(0)` for the **last record of each batch** (empty batches create no semaphore)
- Store it in `Option[Semaphore]` (volatile) so it persists across batches
- When `shardEnded` is called, wait on the most recent semaphore (which correctly points to the previous batch's last record if the final batch is empty)
- If no records were ever processed, proceed without waiting

## Test Coverage
Added `ShardProcessorSpec` with three test scenarios:
1. **Normal case**: Block `shardEnded` until the last record is checkpointed
2. **Empty final batch**: Block `shardEnded` until the previous batch's last record is checkpointed (the critical bug fix)
3. **No records**: Complete `shardEnded` immediately when no records were ever processed

## Related
This bug was discovered and fixed in https://github.com/adRise/hyades/pull/3668 and has been ported to alpakka.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/ShardProcessor.scala
+++ b/kinesis/src/main/scala/akka/stream/alpakka/kinesis/impl/ShardProcessor.scala
@@ -21,23 +21,16 @@ private[kinesis] class ShardProcessor(
     processRecord: CommittableRecord => Unit
 ) extends ShardRecordProcessor {
 
-  // We need extra coordination in the event of a Shard End (for example, when we double
-  // the number of shards, the old shards are ended and the new shards take their place).
-  //
-  // When a shard needs to be ended, the "shardEnded" method is invoked. We block that
-  // invocation until all records (actually we only care about the latest record)
-  // are checkpointed using the asynchronous stream mechanism.
-  //
-  // When all records are checkpointed, the "shardEnded" method can safely acknowledge
-  // that the shard can be ended.
-  //
-  // If we were not to do this extra work, the shard would be ended before we made sure
-  // that all the records have been successfully consumed by the stream.
-  //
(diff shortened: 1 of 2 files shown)


johanandren (approved):
LGTM

I pushed a tiny improvement to the test on top of the changes.

huayangdu:
Hi Akka @johanandren @sebastian-alfers , I’m a colleague of the PR author. May I ask if there’s a release planned for this fix ? seems it should be in 10.0.5 but no update for long. We recently had another incident during a Kinesis scale-out, where this issue prevented the consumer from checkpointing correctly, causing child leases to be blocked waiting on their parent shards. Thanks!

johanandren:
I'll fire off a release right away to get this out. Sorry for the long wait.

PR #3454, opened 2026-03-23
feat: List blobs/files in azure storage

johanandren:
I've manually run the test suite against actual azure storage.


octonato (commented):
A few comments from myself 

and from a friend...

**No test for multi-page pagination** 
The core value of this PR (automatic pagination via `unfoldAsync`) has zero test coverage for the multi-page case. All WireMock tests return a single page with empty `<NextMarker/>`. A test that returns page 1 with a non-empty marker, then page 2 with an empty marker, would validate the loop actually works.

octonato (approved):
LGTM

PR #3459, opened 2026-04-02
chore: clean up azure-storage

johanandren:
A whole bunch of fixes:

  - Fix duplicate `Content-Length` on requests by setting `renderInRequests = false` on `CustomContentLengthHeader` (Akka HTTP synthesizes it from the entity); `Signer` still reads the header
  value for SharedKey signing.
  - Fix `PutAppendBlock` using `PageBlobHeader` instead of `AppendBlobHeader`.
  - Untangle `StorageHeaders` slot/method naming so fields, helpers, and `copy` parameters line up (`rangeWriteTypeHeader` for `x-ms-write`, `fileTypeHeader` for `x-ms-type`).
  - Clean up `StorageException` parsing: early-return on null/blank bodies, plain `try`/`catch` for XML parse errors.
  - Drop redundant `CustomContentTypeHeader` from `computeMetaData` (the typed `Content-Type` is already added and is the only one `ObjectMetadata.contentType` matches).
  - Add `SharedKeyLite` as a recognized authorization type.
  - Decode the account key defensively in `AzureNameKeyCredential.apply(Config)` so placeholder keys for anon/SAS don't fail base64 decoding.
  - Re-enable previously `@Ignore`d put/update/clear tests (Java + Scala) and `AzuriteIntegrationSpec` now that the Content-Length issue is fixed.
  - Tighten `mockClearRange` to assert `Content-Type` is absent on the request.
  - Fix a bug where 7 digit day-of-months would be dismissed by actual azure (but accepted by azurite)
  - Add `azure-storage` to `check-build-test.yml` so the connector runs in CI.

I've manually run the test suite against actual azure storage.


octonato (approved):
LGTM as far as I can tell. 

I will ask a second opinion to a friend. Stay tuned!

octonato (approved):
LBTM

PR #3464, opened 2026-04-14
feat: Azure identity for azure storage auth

johanandren:
Ammend the azure-storage auth options with azure-identity, enabling OAuth2 bearer token authentication.

 * Opt in dependency on azure-identity to not bring in the entire dependency graph by default, users opt in with an explicit dependency to actually use it
 * When authorization-type is set to BearerToken in config, DefaultAzureCredential is used automatically, supporting managed identity, workload identity, environment credentials, etc.
 * Users can also provide any custom TokenCredential programmatically via StorageSettings.withTokenCredential(credential), e.g. for User Assigned Managed Identity
  with a specific client ID
 * Token acquisition is async (bridges the Azure SDK Mono<AccessToken> into the Akka Stream via Source.future), token caching/refresh handled by the Azure SDK internally

### DIFF
--- a/azure-storage/src/main/scala/akka/stream/alpakka/azure/storage/Settings.scala
+++ b/azure-storage/src/main/scala/akka/stream/alpakka/azure/storage/Settings.scala
@@ -47,7 +47,10 @@ final class StorageSettings(val apiVersion: String,
   /** Java API */
   def getAlgorithm: String = algorithm
 
-  def witApiVersion(apiVersion: String): StorageSettings = copy(apiVersion = apiVersion)
+  def withApiVersion(apiVersion: String): StorageSettings = copy(apiVersion = apiVersion)
+
+  @deprecated("Use withApiVersion instead", "azure-storage 10.0.0")
+  def witApiVersion(apiVersion: String): StorageSettings = withApiVersion(apiVersion)
 
   def withAuthorizationType(authorizationType: String): StorageSettings = copy(authorizationType = authorizationType)
 
@@ -97,6 +100,7 @@ final class StorageSettings(val apiVersion: String,
   override def hashCode(): Int =
     Objects.hash(apiVersion,
                  authorizationType,
+                 endPointUrl,
                  azureNameKeyCredential,
                  sasToken,
                  retrySettings,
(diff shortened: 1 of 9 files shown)


azure-storage/src/main/scala/akka/stream/alpakka/azure/storage/Settings.scala
```diff
@@ -86,13 +88,20 @@ final class StorageSettings(val apiVersion: String,
...
 
   override def hashCode(): Int =
-    Objects.hash(apiVersion, authorizationType, azureNameKeyCredential, sasToken, retrySettings, algorithm)
+    Objects.hash(apiVersion,
+                 authorizationType,
+                 azureNameKeyCredential,
+                 sasToken,
+                 retrySettings,
+                 algorithm,
+                 tokenCredential)
```
octonato, line 108: The equals impl includes the endPointUrl, but this is missing from the hashCode. 

It was already like that, but while we are here it's worth to fix it now.

azure-storage/src/main/scala/akka/stream/alpakka/azure/storage/Settings.scala
```diff
@@ -45,37 +47,37 @@ final class StorageSettings(val apiVersion: String,
   /** Java API */
   def getAlgorithm: String = algorithm
 
-  /** Java API */
   def witApiVersion(apiVersion: String): StorageSettings = copy(apiVersion = apiVersion)
```
octonato:
```suggestion
  def withApiVersion(apiVersion: String): StorageSettings = copy(apiVersion = apiVersion)
```

octonato (approved):
Solid PR. Certified by deep expert. 

I left two minor suggestions, but worth doing before merging.

azure-storage/src/main/scala/akka/stream/alpakka/azure/storage/Settings.scala
octonato: Oh, I realise that this is an existing API. It would be a breaking change.

johanandren:
I'm trying to add an integration test against azurite as well, before merging this.

azure-storage/src/main/scala/akka/stream/alpakka/azure/storage/Settings.scala
johanandren: deprecated and added new correctly spelled method

PR #3500, opened 2026-08-19
chore: Harden GCP object storage access

patriknw:
Diagnosing a failure to reach the GCE metadata server took far longer than it should have: every layer reported a symptom rather than the failure, and the credential path degraded silently, so the trail pointed at IAM instead of the network.

**Credentials**
- `Credentials.apply` logged three warnings that never said what the fallback does. Now one warning stating the consequence (falls back to the `none` provider, so requests are rejected with 401/403), with cause chains and a pointer at the metadata server as the likely cause of a timeout.
- `GoogleComputeMetadata` and `UserAccessMetadata` ignored the response status and unmarshalled errors as tokens, turning an empty-bodied 5xx into `Unexpected end-of-input at input index 0`. They now fail with the status, URL and body. `getProjectId` was worse than a bad message — it accepted an error page (or `""`) as the project ID.
- `compute-engine.timeout` 1s → 3s. A false negative is permanent for the ActorSystem's lifetime (settings are cached, `NoCredentials` never retries), so degrading costs more than waiting.

**Uploads**
- `ResumableUpload.initiateSession` threw `No Location header` for any response lacking one, discarding the entity first. The header is absent from every failed response, so that described the least interesting part; it now reports the status and keeps the API's own error message.

**Docs**
- `google-common.md` presented credential loading as automatic and infallible. Added the `provider` options and a warning covering the silent fallback and the blocking metadata read. `reference.conf` was missing `user-access` from its options list and said nothing about the fallback.

Behaviour is unchanged apart from the timeout default, and errors that previously surfaced as parse failures — or, for `getProjectId`, not at all. Tests added for the metadata status check and failed upload initiation.

### DIFF
--- a/google-common/src/main/scala/akka/stream/alpakka/google/auth/Credentials.scala
+++ b/google-common/src/main/scala/akka/stream/alpakka/google/auth/Credentials.scala
@@ -13,6 +13,7 @@ import com.google.auth.{Credentials => GoogleCredentials}
 import com.typesafe.config.Config
 
 import java.util.concurrent.Executor
+import scala.annotation.tailrec
 import scala.collection.immutable.ListMap
 import scala.concurrent.{Await, ExecutionContext, Future}
 import scala.jdk.CollectionConverters._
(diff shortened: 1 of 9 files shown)


google-common/src/main/scala/akka/stream/alpakka/google/auth/GoogleOAuth2Exception.scala
```diff
@@ -45,4 +54,35 @@ private[google] object GoogleOAuth2Exception {
...
+    else
+      Unmarshal(response.entity)
+        .to[String]
+        .recover { case _ => "<unavailable>" }
+        .flatMap { body =>
+          Future.failed(
+            GoogleOAuth2Exception(
+              ErrorInfo(
+                s"Unexpected ${response.status.value} response from [$uri]",
+                s"Response body: [${body.take(maxErrorBodyLength)}]"
```
johanandren: Append `...` when `body.length > maxErrorBodyLength` so it is visible that it was truncated?

google-common/src/main/scala/akka/stream/alpakka/google/auth/GoogleOAuth2Exception.scala
```diff
@@ -45,4 +54,35 @@ private[google] object GoogleOAuth2Exception {
             ex
       }
       .withDefaultRetry
+
+  private val maxErrorBodyLength = 512
```
johanandren: Constant

```suggestion
  private val MaxErrorBodyLength = 512
```

google-common/src/main/scala/akka/stream/alpakka/google/auth/Credentials.scala
```diff
@@ -68,6 +80,17 @@ object Credentials {
 
   private def parseNone(c: Config) = NoCredentials(c.getConfig("none"))
 
+  /**
+   * Renders the message of `ex` together with the messages of its causes, since the cause chain often carries the
+   * actual reason (e.g. a connection timeout) while the outermost message is generic.
+   */
+  private def describe(ex: Throwable): String = {
+    @tailrec def loop(ex: Throwable, depth: Int, acc: List[String]): List[String] =
+      if ((ex eq null) || depth == 0) acc
```
johanandren: Worth adding a "..." or "[truncated]" to make it clear it was cut if there are more causes?

google-common/src/main/scala/akka/stream/alpakka/google/auth/Credentials.scala
johanandren: I think this still means that the timeout exception can hide the underlying failure, which is sort of what this PR is addressing. Possibly worth adding warning logs for failure on the future returned from apply() or something like that?

johanandren (approved):
LGTM, some suggested improvements

johanandren (approved):

